PR #10, opened 2025-02-08
solana-account: add AccountSharedData::data_clone method

jstarry:
In order to create a lightweight view struct (https://github.com/anza-xyz/agave/pull/4834) over vote account data without creating another copy of the data, I need to copy the internal `Arc` which is private.

### DIFF
--- a/account/src/lib.rs
+++ b/account/src/lib.rs
@@ -597,6 +597,10 @@ impl AccountSharedData {
         self.data.capacity()
     }
 
+    pub fn data_clone(&self) -> Arc<Vec<u8>> {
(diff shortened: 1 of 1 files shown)


joncinque (approved):
I looked for other patterns of cloning an internal `Arc` in a type to see if there's a standard way to name these kind of functions, but couldn't find anything in libraries like tokio / clap / std.

My one concern was that `data_clone` might make it sound like the underlying data is being cloned, but the return type should make that totally clear.

All in all, looks good!

Maliksb11:
Happy 😁

PR #12, opened 2025-02-09
Make ProgramError compatible with pinocchio

kevinheavey:
**Problem**: solana-program-error and pinocchio define ProgramError enums that are identical except for the BorshIoError variant, which looks like `BorshIoError(String)` in solana-program-error and is fieldless (just `BorshIoError`) in pinocchio.

It is going to be a significant source of grief for users to have two almost-identical ProgramError types in their dependency tree. People will accidentally import the wrong one a lot

**Solution**: make solana_program_error::ProgramError the same as the pinocchio one so pinocchio can replace its definition with a re-export.

**Breaking changes**: 
- the BorshIoError variants of ProgramError and InstructionError become fieldless. This will require a major version bump for solana-program-error, solana-instruction-error, solana-program and solana-sdk. I haven't included the version bump in the PR as it breaks compilation because of the solana-system-interface dep. This may or may not be curable with patching but is ultimately not a real problem.
- add a num-traits feature to solana-program-error so that all functionality requiring num-traits becomes optional (pinocchio isn't using this functionality)
- make std optional in solana-program-error as pinocchio is no-std. The only thing this affects is `impl std::error::Error for ProgramError {}`

**Other changes**:
- Extract solana-pubkey-error and solana-instruction-error crates and use them in solana-program-error. This is so pinocchio's deps can be as light as possible
- Make num-traits optional


kevinheavey:
CI error is as expected given the breaking change

program-error/Cargo.toml
```diff
@@ -11,19 +11,24 @@ edition = { workspace = true }
...
 serde = { workspace = true, optional = true }
 serde_derive = { workspace = true, optional = true }
-solana-decode-error = { workspace = true }
-solana-instruction = { workspace = true, default-features = false, features = [
-    "std",
-] }
-solana-msg = { workspace = true }
-solana-pubkey = { workspace = true, default-features = false }
+solana-decode-error = { workspace = true, optional = true }
+solana-instruction-error = { workspace = true }
```
febo: Can we make this one optional?

program-error/Cargo.toml
```diff
@@ -11,19 +11,24 @@ edition = { workspace = true }
...
+solana-msg = { workspace = true, optional = true }
+solana-pubkey-error = { workspace = true, default-features = false }
```
febo: Can we make this one optional?

program-error/src/lib.rs
```diff
@@ -2,15 +2,16 @@
...
 use borsh::io::Error as BorshIoError;
 #[cfg(feature = "serde")]
 use serde_derive::{Deserialize, Serialize};
 use {
-    core::fmt,
-    num_traits::FromPrimitive,
-    solana_decode_error::DecodeError,
-    solana_instruction::error::{
+    core::{convert::TryFrom, fmt},
+    solana_instruction_error::{
```
febo: If we make `solana-instruction-error` optional, we should add a feature here.

program-error/src/lib.rs
```diff
@@ -21,12 +22,12 @@ use {
         MISSING_REQUIRED_SIGNATURES, NOT_ENOUGH_ACCOUNT_KEYS, UNINITIALIZED_ACCOUNT,
         UNSUPPORTED_SYSVAR,
     },
-    solana_msg::msg,
-    solana_pubkey::PubkeyError,
-    std::convert::TryFrom,
+    solana_pubkey_error::PubkeyError,
```
febo: Same here, If we make `solana-pubkey-error` optional, we should add a feature here.

program-error/src/lib.rs
```diff
@@ -122,12 +124,14 @@ impl fmt::Display for ProgramError {
     }
 }
 
+#[cfg(feature = "num-traits")]
 pub trait PrintProgramError {
     fn print<E>(&self)
     where
         E: 'static + std::error::Error + DecodeError<E> + PrintProgramError + FromPrimitive;
```
febo: Is there a way to relax the requirements for `E`? E.g., drop `std::error::Error` and `FromPrimitive`. This way we could use the trait in pinocchio-based errors.

kevinheavey: We can remove the DecodeError and Error constraints, which is technically a breaking change but is trivial and we're already making a breaking change here, so I've committed that.

We can replace the FromPrimitive constraint with TryFrom<u32> but this is a more significant break because users would have to write their own impl or switch to num_enum to derive it https://docs.rs/num_enum/latest/num_enum/derive.TryFromPrimitive.html. For now I've committed this change too

program-error/Cargo.toml
kevinheavey: are you sure? I think this is at the point where the cognitive overhead of the extra feature outweighs any impact on compile times or API surface area

febo: Yeah, the amount of features to enable/disable depending of whether you are using pinocchio or the SDK concerns me - hopefully is not going to be annoying.

Ideally I would not have `solana-instruction-error` and `solana-pubkey-error` exported here – those should be added by the code that wants to use them. So the other option is to remove the export from here, instead of adding another feature.

kevinheavey: pinocchio would want the error codes defined in solana-instruction-error e.g. `pub const INVALID_ARGUMENT: u64 = to_builtin!(2);`. Currently it's redefining these itself. Are you saying you'd want these consts to be in a standalone crate without the InstructionError enum?

febo: I was thinking that we can have the error codes defined in `solana-program-error`, like in pinocchio. So `solana-instruction-error` is the one that depends on `solana-program-error`, not the other way around. The error codes cover the "program errors" variants and `InstructionError` is a super set of `ProgramError`.

kevinheavey: ok i've moved some stuff to program-error

kevinheavey:
any outstanding issues with this?

instruction-error/src/lib.rs
```diff
@@ -202,15 +173,7 @@ pub enum InstructionError {
...
-    ///
-    /// Warning: This error should never be emitted by the runtime.
-    ///
-    /// This error includes strings from the underlying 3rd party Borsh crate
-    /// which can be dangerous because the error strings could change across
-    /// Borsh versions. Only programs can use this error because they are
-    /// consistent across Solana software versions.
-    ///
-    BorshIoError(String),
+    BorshIoError,
```
joncinque, line 176: Confirming that this change is good from the runtime side too -- the error text is never used or stored anywhere

instruction-error/src/lib.rs
```diff
@@ -245,7 +208,6 @@ pub enum InstructionError {
 #[cfg(feature = "std")]
 impl std::error::Error for InstructionError {}
```
joncinque, line 209: Is there a reason to avoid using `core::error::Error` when deriving these? We should be able to essentially remove `std` with that, and only enable it for frozen-abi builds

instruction-error/src/lib.rs
```diff
@@ -385,7 +347,7 @@ impl fmt::Display for InstructionError {
     }
 }
 
-#[cfg(feature = "std")]
+#[cfg(feature = "num-traits")]
```
joncinque, line 350: I looked around to see if there was a standard feature name for num-traits. The highest downloaded crate that has it optional uses `num-traits` https://github.com/starkat99/half-rs, and the second just uses `num` https://github.com/jhpratt/deranged.

All that to say, I'm good with `num-traits`

pubkey-error/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+    #[inline]
+    fn from_u64(n: u64) -> Option<Self> {
+        Self::from_i64(n as i64)
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for PubkeyError {}
```
joncinque: Same here, can we use core::error::Error? Then we can completely remove all notion of `std` I think

program-error/src/lib.rs
```diff
@@ -63,6 +92,7 @@ pub enum ProgramError {
     IncorrectAuthority,
 }
 
+#[cfg(feature = "std")]
 impl std::error::Error for ProgramError {}
```
joncinque, line 92: Same here, can we use core::error::Error?

program-error/Cargo.toml
```diff
@@ -11,19 +11,17 @@ edition = { workspace = true }
...
-    "std",
-] }
-solana-msg = { workspace = true }
-solana-pubkey = { workspace = true, default-features = false }
+solana-msg = { workspace = true, optional = true }
+solana-pubkey-error = { workspace = true, default-features = false, optional = true }
 
 [features]
 borsh = ["dep:borsh"]
+pubkey-error = ["dep:solana-pubkey-error"]
```
joncinque: I don't want to cause too much whiplash, but I'm not sure this feature adds too much value, especially considering solana-pubkey-error is such a light crate.

My vote would be to move the `From<PubkeyError> for ProgramError` implementation to `solana-pubkey-error` instead.

@febo thoughts?

instruction-error/src/lib.rs
```diff
@@ -1,60 +1,31 @@
...
+use solana_program_error::ProgramError;
+pub use solana_program_error::{
+    ACCOUNT_ALREADY_INITIALIZED, ACCOUNT_BORROW_FAILED, ACCOUNT_DATA_TOO_SMALL,
+    ACCOUNT_NOT_RENT_EXEMPT, ARITHMETIC_OVERFLOW, BORSH_IO_ERROR,
+    BUILTIN_PROGRAMS_MUST_CONSUME_COMPUTE_UNITS, CUSTOM_ZERO, ILLEGAL_OWNER, IMMUTABLE,
+    INCORRECT_AUTHORITY, INCORRECT_PROGRAM_ID, INSUFFICIENT_FUNDS, INVALID_ACCOUNT_DATA,
+    INVALID_ACCOUNT_DATA_REALLOC, INVALID_ACCOUNT_OWNER, INVALID_ARGUMENT,
+    INVALID_INSTRUCTION_DATA, INVALID_SEEDS, MAX_ACCOUNTS_DATA_ALLOCATIONS_EXCEEDED,
+    MAX_INSTRUCTION_TRACE_LENGTH_EXCEEDED, MAX_SEED_LENGTH_EXCEEDED, MISSING_REQUIRED_SIGNATURES,
+    NOT_ENOUGH_ACCOUNT_KEYS, UNINITIALIZED_ACCOUNT, UNSUPPORTED_SYSVAR,
```
joncinque, line 19: I do like this change a lot, it always bothered me that the program error codes were defined in instruction error

program-error/src/lib.rs
```diff
@@ -125,17 +155,18 @@ impl fmt::Display for ProgramError {
 pub trait PrintProgramError {
     fn print<E>(&self)
     where
-        E: 'static + std::error::Error + DecodeError<E> + PrintProgramError + FromPrimitive;
+        E: 'static + PrintProgramError + TryFrom<u32>;
 }
 
+#[cfg(feature = "solana-msg")]
```
joncinque: We can figure this out in follow-up work, but since this isn't actually doing any string formatting, we can probably get by with using the syscall directly and completely avoid pulling in `solana-msg` and `format!`, which is typically the source of bloat.

And that way, we don't need the `solana-msg` feature.

What do you think @febo ?

joncinque (commented):
Really great work! I'm a big fan of the change overall, mostly small things and questions to consider

instruction-error/src/lib.rs
kevinheavey: you'd be be bumping the msrv to 1.81, which is higher than solana-program's msrv

joncinque: Ah right, we can do that in the future then, no need to do it here

program-error/Cargo.toml
febo: Makes sense to me. I think it is simpler to import `solana-pubkey-error` directly if you want to use it.

program-error/src/lib.rs
febo: Agree – another option is to use a trait that just converts a `ProgramError` to a `&str`, like [here](https://github.com/anza-xyz/pinocchio/blob/main/sdk/pinocchio/src/program_error.rs#L236-L285).

program-error/Cargo.toml
kevinheavey: that would add an indirect solana-program-error dep to solana-pubkey unless we add a program-error feature to solana-pubkey-error

joncinque: Sorry, I think I'm missing something about the dependency graph -- is there an issue with having solana-pubkey depend on solana-program-error? Or is it the problem of adding any dependencies to solana-pubkey?

kevinheavey: It's more bloaty than having program-error depend on pubkey-error, since if you have to choose a superfluous dep, pubkey-error is the smaller one. I would just have program-error depend on pubkey-error without a feature flag if @febo is ok with Pinocchio depending unnecessarily on pubkey-error

febo: Hmmm, I think it will be a bit strange to have it that way. In a way `ProgramError` is the "root" error type generated by programs, so the direction of the dependency looks more natural if `PubkeyError` depends on `ProgramError` for cases where it will be wrapped as a `ProgramError`. This is what usually will happen to any other custom error type defined that can be wrapped as a `ProgramError`.

kevinheavey: ok i'm moving PubkeyError back to solana-pubkey

al12alex:
how to run project  in git hub

program-error/src/lib.rs
kevinheavey: If @febo is fine with it I suggest we just remove the feature and make solana-msg a regular dependency. It causes no measurable increase in build time and I doubt the solana-msg crate is going to have any major version bumps

joncinque (approved):
This looks good to me, but let's see what @febo thinks before merging

kevinheavey:
Does anyone use PrintProgramError outside of SPL? We could move it somewhere else.

What's the big deal with having unused std code in a dependency? `#![no_std]` still works even if a dependency uses `std` somewhere

febo:
> Does anyone use PrintProgramError outside of SPL? We could move it somewhere else.
> 
> What's the big deal with having unused std code in a dependency? `#![no_std]` still works even if a dependency uses `std` somewhere

Messes around with the panic handler implementation when `std` is indirectly available to the linker: https://github.com/anza-xyz/pinocchio/pull/84

kevinheavey:
I'm not following - isn't std always available to the linker if the platform supports it? And `#[no_std]` just takes it out of scope? Putting `#![no_std]` in a crate doesn't affect that crate's dependencies, and vice versa

febo:
> I'm not following - isn't std always available to the linker if the platform supports it? And `#[no_std]` just takes it out of scope? Putting `#![no_std]` in a crate doesn't affect that crate's dependencies, and vice versa

It might be a specific behaviour of platform-tools toolchain, but this is the difference (full details [here](https://github.com/anza-xyz/pinocchio/pull/84#issuecomment-2709964773)):

std:
when a panic happens, rust runtime will call
```
rust_begin_unwind (std generted)
    std::sys::pal::sbf::panic (sbf compiler generated)
        custom_panic (pinocchio)
        abort syscall (sbf compiler generated)
```

no_std:
```
rust_begin_unwind (pinocchio)
```

This changes which panic handler to include in your program, since for `std` it is more like a "hook", but for `no_std` it is a "full" handler. This is not controlled by whether your program is `no_std` or `std` – having an indirect dependency to the `std` changes the behaviour.

kevinheavey:
If that's true that's a bug in platform-tools rust right? In normal Rust adding a dependency on a crate that uses std doesn't change the behaviour of other crates.

Anyway I've now made it so the std feature is enabled if the solana-msg feature is enabled

febo:
> If that's true that's a bug in platform-tools rust right? In normal Rust adding a dependency on a crate that uses std doesn't change the behaviour of other crates.
> 
> Anyway I've now made it so the std feature is enabled if the solana-msg feature is enabled

I think this is the "normal" behaviour. When the `std` is available, either because your crate depends on it or any of your crate's dependencies, it automatically sets a panic handler. When all crates are `no_std`, there is no panic handler attached and you need to set your one.

You can quickly test this in Rust. Create a new crate with type `"cdylib"`; add `#![no_std]` to the `lib.rs` and try to compile (standard `cargo build`). You will get an error that there is no panic handler set.
```
error: `#[panic_handler]` function required, but not found

error: unwinding panics are not supported without std
```
Now add `solana-msg` as a dependency and add a `solana_msg::msg!("I am std")` to a function. If you try to compile now, it works.

All that to say that it is better to "mark" crates as `no_std` when they do not dependent on `std`.

kevinheavey:
Interesting, that's really annoying! Not to go on a tangent but is there a downside to removing the root `#![no_std]` from pinocchio so that you get predictable panic behaviour? All the other modules in pinocchio could still have `#[no_std]`.

AFAICT Rust tooling provides no good ways of ensuring that all a project's dependencies are no-std, so this is going to be a footgun for users

febo:
> Interesting, that's really annoying! Not to go on a tangent but is there a downside to removing the root `#![no_std]` from pinocchio so that you get predictable panic behaviour? All the other modules in pinocchio could still have `#[no_std]`.
> 
> AFAICT Rust tooling provides no good ways of ensuring that all a project's dependencies are no-std, so this is going to be a footgun for users

Not a lot of downside, but it would prevent someone to completely replace the panic handler with a custom implementation if they wanted to. I imagine most users won't see this issue, since you need to be explicitly writing a `no_std` program  and not use the "standard" `entrypoint!` macro – at that point, you probably will be able to handle the conflict.

I think this will eventually go away when we bump to Rust `1.84` – more [here](https://discord.com/channels/428295358100013066/476811830145318912/1354455905822838904) (Discord).

program-error/src/lib.rs
```diff
@@ -125,17 +153,18 @@ impl fmt::Display for ProgramError {
 pub trait PrintProgramError {
```
febo: Wonder if we should deprecate this trait and/or put it behind the `solana-msg` feature?

cc: @joncinque

joncinque: This is a breaking change anyway, so we can just remove it, but let's do it in a different PR, this one is ready to go on my side.

febo (dismissed):
Looks great and works great as well! Here it is how this is integrated in pinocchio: https://github.com/anza-xyz/pinocchio/pull/112

joncinque:
I'm thinking we should land the deprecation of `PrintProgramError` first with #102, publish a patch release of this crate, then we can land this. How does that sound?

joncinque (commented):
Can you rebase this? After that we can take one more look then land it.

I made some changes around all this, so the rebase might be a bit gross, so let me know if you want me to do it instead.

kevinheavey:
yeah feel free to take this one

joncinque:
@febo can you take a look and make sure it's usable for pinocchio?

febo:
> @febo can you take a look and make sure it's usable for pinocchio?

All good, I updated the PR on pinocchio and everything works.

febo (approved):

apfitzge:
I was just about to make the `BorshIoError` change to get rid of that `String`.
Thanks @kevinheavey!
It takes `TransactionError` size from 32 to 12 bytes ❤️

PR #15, opened 2025-02-10
make is_nonoverlapping private and remove num-traits dep

kevinheavey:
This function is only used here and in solana-bpf-loader program. When used here, it doesn't need num-traits. I have opened a PR in the Agave repo to copy the num-traits version into solana-bpf-loader program https://github.com/anza-xyz/agave/pull/4888

Since the function is hidden this is not a breaking change.

### DIFF
--- a/program-memory/src/lib.rs
+++ b/program-memory/src/lib.rs
@@ -11,19 +11,14 @@ pub mod syscalls {
 }
 
 /// Check that two regions do not overlap.
-///
-/// Hidden to share with bpf_loader without being part of the API surface.
-#[doc(hidden)]
-pub fn is_nonoverlapping<N>(src: N, src_len: N, dst: N, dst_len: N) -> bool
-where
-    N: Ord + num_traits::SaturatingSub,
+fn is_nonoverlapping(src: usize, src_len: usize, dst: usize, dst_len: usize) -> bool
 {
     // If the absolute distance between the ptrs is at least as big as the size of the other,
     // they do not overlap.
     if src > dst {
-        src.saturating_sub(&dst) >= dst_len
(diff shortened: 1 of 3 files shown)


febo:
We could also add the missing `unsafe` markers – that way we can replicate the implementation from pinocchio here.

kevinheavey:
> We could also add the missing `unsafe` markers

Where?

kevinheavey:
Oh guessing you mean these ones:
```
__This function is incorrectly missing an `unsafe` declaration.__
```

Let's leave those for another PR

program-memory/src/lib.rs
```diff
@@ -229,7 +223,7 @@ mod tests {
         for dst in 13..20 {
             assert!(is_nonoverlapping(10, 3, dst, 3));
         }
-        assert!(is_nonoverlapping::<u8>(255, 3, 254, 1));
-        assert!(!is_nonoverlapping::<u8>(255, 2, 254, 3));
+        assert!(is_nonoverlapping(255, 3, 254, 1));
+        assert!(!is_nonoverlapping(255, 2, 254, 3));
```
joncinque: Since this test is meant to look at the limits, can it be updated to work with `usize` limits?
```suggestion
        assert!(is_nonoverlapping(usize::MAX, 3, usize::MAX - 1, 1));
        assert!(!is_nonoverlapping(usize::MAX, 2, usize::MAX - 1, 3));
```

joncinque (commented):
This makes sense to me, thanks for the contribution! Just a suggestion for the test.

And since this is a breaking change anyway, we can mark those functions unsafe as febo recommended in a separate PR, then release a new major version.

joncinque (approved):
Thanks!

PR #24, opened 2025-02-11
fix: typo corrections

rebustron:
Hi! This PR fixes minor typos in comments:

- **`abi_example.rs`**: corrected "kowning" → "knowing".
- **`lib.rs`**: corrected "activiation" → "activation".

No logic changes, just small text fixes.


joncinque (approved):
Thanks for your contribution. The change to the comment on transaction error could be problematic if we used `thiserror` to derive the various traits required by `Error`, but since we implement those by hand, the change is safe.

PR #26, opened 2025-02-11
Bump ed25519-dalek to 2.1.1, ed25519-dalek-bip32 to 0.3.0

kevinheavey:
Ported from https://github.com/anza-xyz/agave/pull/3088

#### Summary of Changes
- Bump the two crate versions as in title and fix what breaks
- This involves one breaking change: `Keypair::generate` now relies on traits from a newer version of the rand crate - it no longer works with rand 0.7.

### DIFF
--- a/keypair/src/lib.rs
+++ b/keypair/src/lib.rs
@@ -83,13 +83,12 @@ impl TryFrom<&[u8]> for Keypair {
     type Error = SignatureError;
 
     fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
-        let Some(keypair_slice) = bytes.get(..ed25519_dalek::KEYPAIR_LENGTH) else {
-            return Err(SignatureError::from_source(String::from(
-                "candidate keypair byte array is too short",
-            )));
-        };
-        // compiler elides the second len check so try_into().unwrap() is as fast as unsafe code
-        let keypair_bytes: &[u8; ed25519_dalek::KEYPAIR_LENGTH] = keypair_slice.try_into().unwrap();
+        let keypair_bytes: &[u8; ed25519_dalek::KEYPAIR_LENGTH] =
+            bytes.try_into().map_err(|_| {
+                SignatureError::from_source(String::from(
+                    "candidate keypair byte array is the wrong length",
+                ))
+            })?;
         ed25519_dalek::SigningKey::from_keypair_bytes(keypair_bytes)
             .map_err(|_| {
                 SignatureError::from_source(String::from(


seanyoung:
How does this affect performance (the bench should tell us that, I think).

kevinheavey:
Looks like the new version is faster. Found this in the CI bench output https://github.com/anza-xyz/solana-sdk/actions/runs/13304474459/job/37152326064?pr=26:

```
test bench_ed25519_len_032 ... bench:      48,795.52 ns/iter (+/- 238.09)
test bench_ed25519_len_128 ... bench:      49,085.67 ns/iter (+/- 368.34)
test bench_ed25519_len_32k ... bench:     110,201.12 ns/iter (+/- 1,078.12)
test bench_ed25519_len_max ... bench:     171,307.76 ns/iter (+/- 1,212.43)
```

Meanwhile in another unrelated PR: https://github.com/anza-xyz/solana-sdk/actions/runs/13344518453/job/37273344074?pr=40
```
test bench_ed25519_len_032 ... bench:      62,532.14 ns/iter (+/- 1,077.31)
test bench_ed25519_len_128 ... bench:      62,774.78 ns/iter (+/- 597.11)
test bench_ed25519_len_32k ... bench:     123,763.83 ns/iter (+/- 1,565.07)
test bench_ed25519_len_max ... bench:     184,640.96 ns/iter (+/- 1,562.99)
```

keypair/src/lib.rs
```diff
@@ -76,12 +71,12 @@ impl Keypair {
 
     /// Returns this `Keypair` as a base58-encoded string
     pub fn to_base58_string(&self) -> String {
-        bs58::encode(&self.0.to_bytes()).into_string()
+        bs58::encode(&self.to_bytes()).into_string()
     }
 
     /// Gets this `Keypair`'s SecretKey
     pub fn secret(&self) -> &ed25519_dalek::SecretKey {
```
juchiast: I think this should just return `[u8; 32]` directly, skip the type alias of ed25519-dalek.

keypair/src/lib.rs
```diff
@@ -271,11 +262,9 @@ pub fn keypair_from_seed(seed: &[u8]) -> Result<Keypair, Box<dyn error::Error>>
     if seed.len() < ed25519_dalek::SECRET_KEY_LENGTH {
         return Err("Seed is too short".into());
     }
-    let secret = ed25519_dalek::SecretKey::from_bytes(&seed[..ed25519_dalek::SECRET_KEY_LENGTH])
+    let secret_key = ed25519_dalek::SecretKey::try_from(&seed[..ed25519_dalek::SECRET_KEY_LENGTH])
```
juchiast: this function never fail, should call unwrap here

keypair/src/lib.rs
kevinheavey: That would have security implications but I'm not the right person to comment on it

joncinque:
We're ready to integrate this now! Do you mind rebasing your branch? Otherwise, I can take care of it. Just let me know

keypair/src/lib.rs
```diff
@@ -78,19 +75,19 @@ impl Keypair {
     /// Returns this `Keypair` as a base58-encoded string
     pub fn to_base58_string(&self) -> String {
         let mut out = [0u8; five8::BASE58_ENCODED_64_MAX_LEN];
-        let len = five8::encode_64(&self.0.to_bytes(), &mut out);
+        let len = five8::encode_64(&self.to_bytes(), &mut out);
```
joncinque, line 61: nit: to avoid the copy, this should be able to use `as_bytes()`
```suggestion
        let len = five8::encode_64(self.as_bytes(), &mut out);
```

keypair/src/lib.rs
joncinque: This function has been deprecated and will be removed anyway

Cargo.toml
```diff
@@ -158,8 +158,8 @@ curve25519-dalek = { version = "4.1.3", features = ["digest", "rand_core"] }
 dashmap = { version = "5.5.3", features = ["serde"] }
 derivation-path = { version = "0.2.0", default-features = false }
 digest = "0.10.7"
-ed25519-dalek = "=1.0.1"
-ed25519-dalek-bip32 = "0.2.0"
+ed25519-dalek = "=2.1.1"
```
joncinque, line 160: note to self for later: 2.2.0 just got released, so we can bump to that after this lands

keypair/src/lib.rs
```diff
@@ -101,32 +98,32 @@ impl Keypair {
...
-        }
-        let secret =
-            ed25519_dalek::SecretKey::from_bytes(&bytes[..ed25519_dalek::SECRET_KEY_LENGTH])
-                .map_err(SignatureError::from_source)?;
-        let public =
-            ed25519_dalek::PublicKey::from_bytes(&bytes[ed25519_dalek::SECRET_KEY_LENGTH..])
-                .map_err(SignatureError::from_source)?;
-        let expected_public = ed25519_dalek::PublicKey::from(&secret);
+        };
+        // compiler elides the second len check so try_into().unwrap() is as fast as unsafe code
```
joncinque: cool! Just for my own knowledge, how did you learn / measure that?

keypair/src/lib.rs
```diff
@@ -101,32 +98,32 @@ impl Keypair {
...
+            &keypair_bytes[ed25519_dalek::SECRET_KEY_LENGTH..],
+        )
+        .map_err(SignatureError::from_source)?;
+        let expected_public = signing_key.verifying_key();
         (public == expected_public)
-            .then_some(Self(ed25519_dalek::Keypair { secret, public }))
+            .then_some(Self(signing_key))
             .ok_or(SignatureError::from_source(String::from(
                 "keypair bytes do not specify same pubkey as derived from their secret key",
             )))
```
joncinque: This looks like it's doing pretty much the same exact thing as [`from_keypair_bytes`](https://docs.rs/ed25519-dalek/latest/ed25519_dalek/struct.SigningKey.html#method.from_keypair_bytes)

How about just calling that instead?

keypair/src/lib.rs
kevinheavey: compared safe and unsafe versions with godbolt, and saw they were the same: https://godbolt.org/z/qf4cnaWax

keypair/src/lib.rs
kevinheavey: `self.0.as_bytes()` would give just the secret 32 bytes, not the full 64 bytes

joncinque: Ah you're right, sorry, I was looking at methods on `SigningKey`

joncinque (commented):
Looks good to me!

keypair/src/lib.rs
```diff
@@ -101,35 +98,28 @@ impl Keypair {
...
-                .map_err(SignatureError::from_source)?;
-        let expected_public = ed25519_dalek::PublicKey::from(&secret);
-        (public == expected_public)
-            .then_some(Self(ed25519_dalek::Keypair { secret, public }))
-            .ok_or(SignatureError::from_source(String::from(
-                "keypair bytes do not specify same pubkey as derived from their secret key",
-            )))
+        };
+        // compiler elides the second len check so try_into().unwrap() is as fast as unsafe code
+        let keypair_bytes: &[u8; ed25519_dalek::KEYPAIR_LENGTH] = keypair_slice.try_into().unwrap();
```
samkim-crypto: What do you think about merging the length check with this line? It seems like the code will be slightly more concise and we can remove the unwrap as well.

```
let keypair_bytes: &[u8; ed25519_dalek::KEYPAIR_LENGTH] = bytes
    .try_into()
    .map_err(|_| SignatureError::from_source(String::from(
        "candidate keypair byte array is the wrong length",
    )))?;
```

keypair/src/lib.rs
```diff
@@ -306,11 +287,10 @@ pub fn keypair_from_seed(seed: &[u8]) -> Result<Keypair, Box<dyn error::Error>>
...
         return Err("Seed is too short".into());
     }
-    let secret = ed25519_dalek::SecretKey::from_bytes(&seed[..ed25519_dalek::SECRET_KEY_LENGTH])
-        .map_err(|e| e.to_string())?;
-    let public = ed25519_dalek::PublicKey::from(&secret);
-    let dalek_keypair = ed25519_dalek::Keypair { secret, public };
-    Ok(Keypair(dalek_keypair))
+    // this won't fail as we've already checked the length
+    let secret_key =
+        ed25519_dalek::SecretKey::try_from(&seed[..ed25519_dalek::SECRET_KEY_LENGTH]).unwrap();
```
samkim-crypto: Nit: an `unwrap` should be safe here, but why not just use `?`?
```suggestion
ed25519_dalek::SecretKey::try_from(&seed[..ed25519_dalek::SECRET_KEY_LENGTH])?;
```
I can't imagine this changing, but this can future-proof any api changes that introduce a new way of erroring.

samkim-crypto (commented):
Some minor comments below, but looks good to me!

keypair/src/lib.rs
kevinheavey: this would change the API to reject slices longer than 64 bytes, but we're in API breaking season so I'll leave the decision up to @joncinque

joncinque: This is a good question... are there any use-cases for keypairs that are longer than 64 bytes? To me it doesn't make any sense.

Maybe if this were taking random seed bytes, then someone could generate more than 32 random bytes and pass them all in. Since this is a keypair though, the first 32 bytes and second 32 bytes must be related, so we probably shouldn't support longer inputs.

What do you think @samkim-crypto ?

samkim-crypto: Yeah if we can allow breaking change, then I would vote for a more strict behavior for safety. I don't see a clear use case for a keypair that is represented by more than 64 bytes. The current implementation silently truncates the input, which can hide bugs.

joncinque (approved):
Looks great, thanks for the work!

PR #30, opened 2025-02-12
add Signature::as_array

fanatid:
It would be nice to have access to the inner array in `Signature` (same as `Pubkey::as_array`) to use custom encoder such as five8

### DIFF
--- a/signature/src/lib.rs
+++ b/signature/src/lib.rs
@@ -287,7 +287,10 @@ mod tests {
         let bytes = [1u8; 64];
         let signature = Signature::from(bytes);
         assert_eq!(signature.as_array(), &bytes);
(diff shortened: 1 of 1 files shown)


joncinque (commented):
Makes sense to me, thanks for the contribution! Can you simply add a sanity test similar to the one for `Pubkey`?

https://github.com/anza-xyz/solana-sdk/blob/6afe65a78cbf0ed52f74cc7455ce7210e4198698/pubkey/src/lib.rs#L1410-L1418

fanatid:
@joncinque updated

joncinque (approved):
Looks great, thanks again for the contribution! Sorry for the delay on re-reviewing

PR #31, opened 2025-02-12
workspace: Bump all crates for publish

joncinque:
#### Problem

As mentioned at #27, the sdk crates were pinning their versions internally, which made them more difficult to use outside of the repo. Now that the version has been relaxed, we need to publish these crates, but they're almost all still on v2.2.0.

#### Summary of changes

Normally crate publishing will be done one-by-one through the GitHub Actions publish workflow, but with ~100 crates, we need to go faster.

Bump all of the v2.2.0 crates to v2.2.1, bump address-lookup-table-interface to v2.2.2, and bump loader-v3-interface to v3.0.0 since it contains a breaking change.

Once this lands, we can publish all of the crates in the repo.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1548,7 +1548,7 @@ checksum = "e19b23d53f35ce9f56aebc7d1bb4e6ac1e9c0db7ac85c8d1760c04379edced37"
 dependencies = [
  "hermit-abi 0.4.0",
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -2536,7 +2536,7 @@ dependencies = [
 
 [[package]]
 name = "solana-account"
-version = "2.2.0"
+version = "2.2.1"
 dependencies = [
  "bincode",
  "qualifier_attr",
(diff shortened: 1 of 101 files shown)


joncinque:
CI error will be fixed with #32

Cargo.lock
```diff
@@ -1548,7 +1548,7 @@ checksum = "e19b23d53f35ce9f56aebc7d1bb4e6ac1e9c0db7ac85c8d1760c04379edced37"
 dependencies = [
  "hermit-abi 0.4.0",
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
```
willhickey: Is this change expected? There's a similar one at line 4651

willhickey (approved):
Aside from the `windows-sys` dependency version changes everything is how I would expect it.

Cargo.lock
joncinque: I was a bit surprised by those changes too.

I reverted them manually, and it doesn't look like cargo tries to downgrade it again, so committed up 97e3f6ef -- thanks for noticing!

willhickey (approved):

Cargo.lock
willhickey: Yeah cargo does something weird. We never did figure out the borsh dependency version change that was happening in the monorepo. Chido eventually added a step to the version bump that just manually reverts it 🙈

joncinque: Yikes, that's rough :sweat_smile:

PR #33, opened 2025-02-12
use five8 in pubkey, signature, hash and keypair crates

kevinheavey:
five8 is faster than bs58.

Offchain benchmarks: https://github.com/kevinheavey/five8/tree/main/crates/five8
Onchain CU estimation (saves about 4500 CUs when using `msg!` on a Pubkey): https://github.com/kevinheavey/five8-sbpf-cu


fanatid:
Yes, please! I have created https://github.com/anza-xyz/solana-sdk/pull/30 to use five8 for Signature in projects, but this would be much better!

juchiast:
I'm adding encode_64_to_str and encode_32_to_str to the five8 crate https://github.com/kevinheavey/five8/pull/9

keypair/src/lib.rs
```diff
@@ -70,13 +70,15 @@ impl Keypair {
...
+        five8::decode_64(s, &mut buf).unwrap();
         Self::from_bytes(&buf).unwrap()
     }
 
     /// Returns this `Keypair` as a base58-encoded string
     pub fn to_base58_string(&self) -> String {
-        bs58::encode(&self.0.to_bytes()).into_string()
+        let mut out = [0u8; five8::BASE58_ENCODED_64_MAX_LEN];
+        let len = five8::encode_64(&self.0.to_bytes(), &mut out);
+        out[..len as usize].iter().map(|c| *c as char).collect()
```
juchiast: Shouldn't this use `from_uft8_unchecked` too?

joncinque (commented):
Sorry for the lateness on this, but the change looks great to me!

Once it's rebased, I'll approve and land.

joncinque (approved):
Thanks for the quick turnaround!

PR #36, opened 2025-02-13
run cargo diet on all crates

kevinheavey:
Some crates are including useless extras in what gets uploaded to crates.io, such as benchmarks and test directories. `cargo diet` is a tool to remove this stuff by specifying a minimal `include` field in Cargo.toml.

I ran `find . -type f -name "Cargo.toml" -execdir cargo diet \;` and then removed the change for `solana-logger` because it made no difference to what I saw when I ran `cargo package -l`


joncinque (approved):
Makes sense, thanks! I confirm the same output when I ran it locally

PR #40, opened 2025-02-15
use fully qualified path in to_instruction macro

kevinheavey:
This macro currently assumes that core::mem::size_of is in scope


kevinheavey:
I would guess this other failure is from removing references to feature IDs that no longer exist in code:
```
 thread 'test_dump_mollusk' panicked at harness/tests/dump_fixture.rs:56:5:
assertion `left == right` failed
  left: 225
 right: 226
```

buffalojoec (approved):

kevinheavey:
I can't merge in this repo

joncinque:
Catching up on the motivation for this change -- do you need this crate released? Or is it possible to use a higher version of rust for the downstream crates?

PR #42, opened 2025-02-17
solana-rent-collector: remove calculation of slots_per_year

LStan:
When `RentCollector::default()` is called (which can actually happen twice per transaction in SVM [here](https://github.com/anza-xyz/agave/blob/0cfcd4788ff943e4d5e6eee5475ae6e04d102838/svm/src/transaction_processor.rs#L415) and [here](https://github.com/anza-xyz/agave/blob/0cfcd4788ff943e4d5e6eee5475ae6e04d102838/svm/src/transaction_processor.rs#L429)) `GenesisConfig::default()` is called which does a lot of computations, including calling `SystemTime::now()`, but only one constant from it is needed.

This PR removes that call and replaces it with a pre-calculated constant. This constant is checked during testing.

Also removed unnecessary  `solana-logger` dependency from `genesis-config` `Cargo.toml`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2963,7 +2963,6 @@ dependencies = [
  "solana-hash",
  "solana-inflation",
  "solana-keypair",
- "solana-logger",
  "solana-native-token",
  "solana-poh-config",
  "solana-pubkey",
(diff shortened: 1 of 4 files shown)


joncinque:
Thanks for your contribution!

This crate is at end-of-life and will be entirely deprecated or removed soon. See https://github.com/anza-xyz/solana-sdk/pull/226 and https://github.com/anza-xyz/agave/pull/6782

While I appreciate the time you took to find this issue and improve it, we won't need the change.

LStan:
This also removed unnecessary `solana-logger` dependency from `genesis-config`. Is it okay if I do it in another PR?

joncinque:
Sure, please!

PR #59, opened 2025-02-26
Remove unused `code.rs` file

LucasSte:
The `codes.rs` file is a remnant from an old version of SIMD-0178. We don't need this file anymore and can be removed completely.


joncinque (approved):
Confirming this change is good and won't cause any breakage because the module wasn't used anywhere or even declared in `lib.rs` -- thanks!

PR #60, opened 2025-02-27
feat: add `From<&Pubkey> for Pubkey`

OliverNChalk:
`From<&Pubkey>` is required for the `EntryRef::insert` [trait bounds](https://docs.rs/hashbrown/latest/hashbrown/hash_map/enum.EntryRef.html#method.insert) in the hashbrown crate.

### DIFF
--- a/pubkey/src/lib.rs
+++ b/pubkey/src/lib.rs
@@ -248,6 +248,13 @@ impl FromStr for Pubkey {
     }
 }
 
+impl From<&Pubkey> for Pubkey {
+    #[inline]
+    fn from(value: &Pubkey) -> Self {
+        *value
+    }
+}
+
 impl From<[u8; 32]> for Pubkey {
     #[inline]
     fn from(from: [u8; 32]) -> Self {


joncinque:
Thanks for your contribution! I think this changes makes sense, but I want to understand the motivation a bit more first.

According to this comment https://github.com/rust-lang/hashbrown/issues/595#issuecomment-2526394752, they say `entry_ref` is used in situations where a key needs to go from borrowed to owned when actually adding the entry to the map.

In the case `Pubkey`, we've made the mistake in the past of abusing its `Copy` derivation, and should prefer using refs as much as possible. So even though `entry` works, we should privilege the `entry_ref` API to avoid unnecessary clones, which means we need to add this `From<&Self>` impl.

Does that cover it?

OliverNChalk:
Yep, the assumption here is that due to the pubkey size we want to prefer entry_ref over entry to avoid a copy if the key is already in the map. Though I admit I have not benchmarked whether this is actually better in terms of performance.

joncinque:
Sorry, can you just rebase your branch? That'll clear up the audit failure, then we can get this in

OliverNChalk:
Done

PR #62, opened 2025-03-02
Add Keypair::try_from_base58_string

juchiast:
Add a conversion method that doesn't panic on errors.

### DIFF
--- a/keypair/src/lib.rs
+++ b/keypair/src/lib.rs
@@ -68,10 +68,15 @@ impl Keypair {
     }
 
     /// Recovers a `Keypair` from a base58-encoded string
-    pub fn from_base58_string(s: &str) -> Self {
+    pub fn try_from_base58_string(s: &str) -> Result<Self, Box<dyn error::Error>> {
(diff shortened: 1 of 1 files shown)


joncinque (commented):
Thanks for your contribution! While I think this makes sense, the `Box<dyn Error>` for the error type really shows the problem with errors in this crate.

We need to add a specific error type for Keypair and use that for this new function. It could probably look very similar to the base error returned by `ed25519_dalek`, which wraps a `Box<dyn Error>` if an `std` feature is enabled: https://docs.rs/signature/2.2.0/signature/struct.Error.html

Then in the future, we can remove `ed25519_dalek` types from the public interface of this crate, so we can finally upgrade the `ed25519_dalek` dependency.

All that to say: let's separately add the proper error type and std feature to this crate, then this PR can reuse that. What do you think?

juchiast:
I think it make more sense to add a `ParseKeypairError` for these functions, rather than an opaque Error.

PR #63, opened 2025-03-03
Make `revise_turbine_epoch_stakes` feature public 

mjain-jump:
`solfuzz-agave` uses the feature set defined here to create a list of supported features for fuzzing. Without this change, we cannot add this feature to our supported features list. This is also the only non-public feature that was made private when the feature was cleaned up from the codebase.


kbhargava-jump (approved):

buffalojoec (approved):
Seems reasonable to me to make the only private feature public. Many of these have been cleaned up without making the IDs private.

PR #64, opened 2025-03-03
Refactor redirect_stderr_to_file to solana-logger

lijunwangs:
redirect_stderr_to_file can be used multiple applications. I found the need to use it in vortexor. Moving this from validator to solana-logger.

Refactored from https://github.com/anza-xyz/agave/blob/master/validator/src/lib.rs#L39 which will be removed once this change made to the SDK

### DIFF
--- a/logger/Cargo.toml
+++ b/logger/Cargo.toml
@@ -12,8 +12,10 @@ edition = { workspace = true }
 [dependencies]
 env_logger = { workspace = true }
 lazy_static = { workspace = true }
-libc = { workspace = true }
 log = { workspace = true }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+libc = { workspace = true }
 signal-hook = { workspace = true }
 
 [lib]
--- a/logger/Cargo.toml
+++ b/logger/Cargo.toml
@@ -2,7 +2,7 @@
 name = "solana-logger"
 description = "Solana Logger"
 documentation = "https://docs.rs/solana-logger"
-version = "2.2.2"
+version = "2.2.1"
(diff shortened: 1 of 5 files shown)


alexpyattaev:
Looks good, this code was stuck in validator crate but can be useful to every CLI tool that uses solana_logger crate.

logger/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "solana-logger"
 description = "Solana Logger"
 documentation = "https://docs.rs/solana-logger"
-version = "2.2.1"
+version = "2.2.2"
```
joncinque: You shouldn't do the version bump here, the release job will take care of that automatically as mentioned at https://github.com/anza-xyz/solana-sdk?tab=readme-ov-file#publishing-a-crate-from-this-repository

logger/src/lib.rs
```diff
@@ -75,3 +79,68 @@ pub fn setup_file_with_default(logfile: &str, filter: &str) {
         .build();
     replace_logger(logger);
 }
+
+#[cfg(unix)]
+#[cfg(not(target_arch = "wasm32"))]
```
joncinque: nit: these can get condensed into one
```suggestion
#[cfg(all(unix, not(target_arch = "wasm32")))]
```

joncinque (commented):
Since solana-logger already transitively depends on libc, the new signal-hook dependency adds very little to the build time, around 0.1s on my machine, so it looks good to me!

Just some tiny comments.

joncinque (approved):
Looks great to me!

PR #66, opened 2025-03-05
Fix a windows compilation issue

lijunwangs:
Ran into a comp issue on Windows platform:

140 |                 setup_file_with_default(&logfile, solana_logger::DEFAULT_FILTER);
       |                                                   ^^^^^^^^^^^^^ use of undeclared crate or module `solana_logger`

Maybe we should add windows comp/clippy checks?

### DIFF
--- a/logger/src/lib.rs
+++ b/logger/src/lib.rs
@@ -137,7 +137,7 @@ pub fn redirect_stderr_to_file(logfile: Option<String>) -> Option<JoinHandle<()>
             #[cfg(not(unix))]
             {
                 println!("logrotate is not supported on this platform");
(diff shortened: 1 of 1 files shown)


joncinque:
That's very strange, we run clippy on windows, here's the run for the breaking PR: https://github.com/anza-xyz/solana-sdk/actions/runs/13663660460/job/38200289091

It looks like it's not doing anything, so I put in https://github.com/anza-xyz/solana-sdk/pull/68 to fix it

joncinque (approved):
Looks good! I'll rebase https://github.com/anza-xyz/solana-sdk/pull/68 on this once it lands

PR #67, opened 2025-03-05
enable optimizer in tests and procmacros

alexpyattaev:
#### Problem
Same as https://github.com/anza-xyz/agave/pull/5155 but for sdk
* Unittests can be faster: optimization is not enabled for them at all, so we wait long time for CI to complete
* Builds are slow: optimizations for procmacros are not enabled either

#### Summary of Changes
* Enable optimizer for unittests
* Enable optimizer for procmacros

If this fails to build there are bugs in tests=)

### DIFF
--- a/program-entrypoint/Cargo.toml
+++ b/program-entrypoint/Cargo.toml
@@ -2,7 +2,7 @@
 name = "solana-program-entrypoint"
 description = "The Solana BPF program entrypoint supported by the latest BPF loader."
 documentation = "https://docs.rs/solana-program-entrypoint"
-version = "2.2.1"
+version = "2.2.2"
 authors = { workspace = true }
 repository = { workspace = true }
 homepage = { workspace = true }
--- a/program-entrypoint/Cargo.toml
+++ b/program-entrypoint/Cargo.toml
@@ -2,7 +2,7 @@
 name = "solana-program-entrypoint"
 description = "The Solana BPF program entrypoint supported by the latest BPF loader."
 documentation = "https://docs.rs/solana-program-entrypoint"
-version = "2.2.2"
+version = "2.2.1"
 authors = { workspace = true }
 repository = { workspace = true }
 homepage = { workspace = true }
--- a/program-entrypoint/src/lib.rs
+++ b/program-entrypoint/src/lib.rs
@@ -301,7 +301,12 @@ impl BumpAllocator {
     /// # Safety
     /// As long as BumpAllocator or any of its allocations are alive,
     /// writing into or deallocating the arena will cause UB.
+    ///
+    /// Integer arithmetic in this global allocator implementation is safe when
+    /// operating on the prescribed `HEAP_START_ADDRESS` and `HEAP_LENGTH`. Any
+    /// other use may overflow and is thus unsupported and at one's own risk.
     #[inline]
+    #[allow(clippy::arithmetic_side_effects)]
     pub unsafe fn new(arena: &mut [u8]) -> Self {
         debug_assert!(
             arena.len() > size_of::<usize>(),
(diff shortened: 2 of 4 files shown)


program-entrypoint/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "solana-program-entrypoint"
 description = "The Solana BPF program entrypoint supported by the latest BPF loader."
 documentation = "https://docs.rs/solana-program-entrypoint"
-version = "2.2.1"
+version = "2.2.2"
```
joncinque: The publish job takes care of bumping the version, so there's no need to do it here

program-entrypoint/src/lib.rs
```diff
@@ -281,18 +281,61 @@ macro_rules! custom_panic_default {
...
+        debug_assert!(
+            arena.len() > size_of::<usize>(),
+            "Arena should be larger than usize"
+        );
+
+        // create a pointer to the start of the arena
+        // that will hold an address of the byte following free space
+        let pos_ptr = arena.as_mut_ptr() as *mut usize;
+        // initialize the data there
+        *pos_ptr = pos_ptr as usize + arena.len();
```
joncinque, line 320: The change looks good, but for my own understanding, how does this avoid undefined behavior exactly?

Is the idea that the `new` call forces the optimizer to keep `heap` around in the tests because some part of it is written to? Would there still be UB if this code here were removed?

joncinque (commented):
The change looks great overall, just a little point and a question

program-entrypoint/Cargo.toml
alexpyattaev: ok should I revert it?

program-entrypoint/src/lib.rs
alexpyattaev: fn new() correctly initializes the arena content where the allocated amount is, which was not done before, as the code assumed all arenas come zeroed. If you were to instantiate the BumpAllocator over a non-zeroed slice, it would cause UB the moment you call alloc().

The reason bug UB was triggered before this change was due to `heap` variable not marked as mut, so compiler assumed that it does not need to reinitialize it (since noone would write into it).

joncinque: Ah very interesting, thanks for the explanation!

program-entrypoint/Cargo.toml
joncinque: Yes please

alexpyattaev: reverted

alexpyattaev: reverted

joncinque (approved):
Looks great!

program-entrypoint/src/lib.rs
illia-bobyr, line 320: I think there is a potential alignment issue here.
`arena: &mut [u8]` is not guaranteed to have proper alignment for it to be cast to a pointer to a `usize`.
A `[u8]` can have any alignment, while a pointer to `usize` needs to have an alignment of 4 or higher.

You can use [`write_unaligned()`](https://doc.rust-lang.org/std/ptr/fn.write_unaligned.html) to avoid the undefined behavior.
[`write_unaligned()` example](https://doc.rust-lang.org/std/ptr/fn.write_unaligned.html#examples) is almost literary the code you are writing here.

alexpyattaev: Write unaligned could work, yes, but I think to properly address this, the entire BumpAllocator should be rewritten to use some form of interior mutability in the struct itself rather than storing the position pointer inline with the allocated data. It would make code a lot cleaner too. I will make a PR for that.

PR #68, opened 2025-03-05
CI: Run clippy script in bash

joncinque:
#### Problem

The windows CI job doesn't seem to be doing anything, likely because the clippy run step doesn't specify to run it under bash. See a no-op run at https://github.com/anza-xyz/solana-sdk/actions/runs/13663660460/job/38200289091

#### Summary of changes

Run the script under bash rather than the default shell.

NOTE: This should fail CI until #66 lands


joncinque:
Rebased on #66, so now it should work, but this did fail previously: https://github.com/anza-xyz/solana-sdk/actions/runs/13685004202/job/38266276543?pr=68

lijunwangs (dismissed):

joncinque:
@lijunwangs it needed one more little fix, and now it passes CI -- can I get another approval?

lijunwangs (approved):

PR #69, opened 2025-03-06
feature-set: Gate ahash's rng usage for non-solana builds

joncinque:
#### Problem

The secp256r1-program crate has code to support being used with on-chain programs at
https://github.com/anza-xyz/solana-sdk/blob/f58afe952551601f3c7fcf60f371fbb9f38003d8/secp256r1-program/src/lib.rs#L739-L750

Unfortunately, `solana-feature-set` uses `ahash`, which uses `getrandom` by default, so the build fails.

#### Summary of changes

The ahash crate has a `no-rng` feature to disable usage of `getrandom` at
https://github.com/tkaitchuck/aHash/blob/7d5c661a74b12d5bc5448b0b83fdb429190db1a3/Cargo.toml#L40-L42, so use that feature for `solana` builds, and properly gate all hasher usage outside of `solana` builds.

Side-note: let me know if there's someone better to tag for review, but since you did the ahash update, I thought you'd be a good reviewer.

### DIFF
--- a/feature-set/Cargo.toml
+++ b/feature-set/Cargo.toml
@@ -10,7 +10,6 @@ license = { workspace = true }
 edition = { workspace = true }
 
 [dependencies]
-ahash = { workspace = true }
 lazy_static = { workspace = true }
 solana-epoch-schedule = { workspace = true }
 solana-frozen-abi = { workspace = true, optional = true, features = [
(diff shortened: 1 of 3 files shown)


feature-set/Cargo.toml
```diff
@@ -19,8 +18,14 @@ solana-frozen-abi = { workspace = true, optional = true, features = [
...
-solana-hash = { workspace = true }
 solana-pubkey = { workspace = true }
+
+[target.'cfg(target_os = "solana")'.dependencies]
+ahash = { workspace = true, default-features = false, features = ["std", "no-rng"] }
+
+[target.'cfg(not(target_os = "solana"))'.dependencies]
+ahash = { workspace = true, default-features = true }
+solana-hash = { workspace = true }
 solana-sha256-hasher = { workspace = true }
```
apfitzge, line 29: can you explain why solana-hash and sha256-hasher are moved under the not(solana) scope?

apfitzge: nvm i see now below - the generated ID only needed for validator.

joncinque:
The refactor for the feature-set has made this no longer needed, since we'll be removing the `verify` functions from all precompiles in the sdk

PR #71, opened 2025-03-06
CI: Update commit message for cargo-release

joncinque:
#### Problem

Right now, the publish job uses a default message for the commit and tag message of `chore: Release`, which isn't very descriptive, especially when there are many possible crates to release.

#### Summary of changes

Similar to what was done at SPL, configure `cargo-release` to use a better tag and commit message.

Here's an example commit:
https://github.com/solana-labs/solana-program-library/commit/a85d36141d639da42fb5f9163ca6aadfc60d8326


yihau (approved):
🪖

Issue #46, opened 2025-02-18
solana-transaction-error missed a major version bump


kevinheavey:
This PR was a breaking change and should have had a major version bump https://github.com/anza-xyz/agave/pull/4150

> It is a breaking change to add a new enum variant if the enum does not use the [#[non_exhaustive]](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) attribute.

https://doc.rust-lang.org/cargo/reference/semver.html#enum-variant-new

joncinque:
Bah, darn. Maybe we should mark it as non-exhaustive and release a new major version to avoid this annoyance in the future?

joncinque:
Fixed with the 3.0.0 release: https://github.com/anza-xyz/solana-sdk/releases/tag/transaction-error%40v3.0.0

PR #74, opened 2025-03-06
Add the responsible program's account index and inner instruction index to each `InstructionError`

steveluscher:
# Problem

Consider a transaction error that originates from a cross-program invocation (ie. an inner instruction). Currently `TransactionError` returns you the index of the outer instruction, which in no way helps you to correlate the content of the error message with the actual program from whence it came.

# Summary of changes

Added the transaction-level account index of the erroring program to the `TransactionError::InstructionError` variant, which will help consumers correlate instruction errors with their _actual_ source.

The account index of the program is designed to be from the perspective of the transaction and not the perspective of the instruction to make this change safe from [SIMD-163](https://github.com/solana-foundation/solana-improvement-documents/pull/163).

Addresses https://github.com/anza-xyz/agave/issues/5152 and https://github.com/anza-xyz/kit/issues/149.
Blocks https://github.com/anza-xyz/agave/pull/6083.

### DIFF
--- a/transaction-error/src/lib.rs
+++ b/transaction-error/src/lib.rs
@@ -4,7 +4,10 @@
 use serde_derive::{Deserialize, Serialize};
 #[cfg(feature = "frozen-abi")]
 use solana_frozen_abi_macro::{AbiEnumVisitor, AbiExample};
-use {core::fmt, solana_instruction::error::InstructionError, solana_sanitize::SanitizeError};
+use {
+    core::fmt, solana_instruction::error::InstructionError, solana_pubkey::Pubkey,
+    solana_sanitize::SanitizeError,
+};
 
 pub type TransactionResult<T> = Result<T, TransactionError>;
 
@@ -42,9 +45,23 @@ pub enum TransactionError {
     /// the `recent_blockhash` has been discarded.
     BlockhashNotFound,
 
-    /// An error occurred while processing an instruction. The first element of the tuple
-    /// indicates the instruction index in which the error occurred.
-    InstructionError(u8, InstructionError),
+    /// An error occurred while processing an instruction.
+    InstructionError {
+        err: InstructionError,
+        /// The index of the inner instruction in which the error was thrown, starting from zero.
+        /// This value will be `None` when the error was thrown from the outer instruction's
+        /// program, and also for all errors stored using a version of this enum variant prior to
+        /// `solana-transaction-error` 3.0.0.
+        inner_instruction_index: Option<u8>,
+        /// The index of the outer instruction in which the error was thrown, starting from zero. Do
+        /// not infer the responsible program from the instruction at this index; the error might
+        /// have been thrown from one of its inner instructions. See `responsible_program_address`.
+        outer_instruction_index: u8,
+        /// The address of the program that threw the error. Use this to decode the error (eg. to
+        /// look up a custom error code in the program's IDL). This value will be `None` for errors
+        /// stored using a version of this enum variant prior to `solana-transaction-error` 3.0.0.
+        responsible_program_address: Option<Pubkey>,
+    },
 
     /// Loader call chain is too deep
     CallChainTooDeep,
@@ -163,7 +180,17 @@ impl fmt::Display for TransactionError {
              => f.write_str("This transaction has already been processed"),
             Self::BlockhashNotFound
              => f.write_str("Blockhash not found"),
-            Self::InstructionError(idx, err) =>  write!(f, "Error processing Instruction {idx}: {err}"),
+            Self::InstructionError {
+                err,
+                outer_instruction_index,
+                ..
+            }
+             // NOTE: We intentionally do not augment the error message in the event that the error
+             // carries the address of the responsible program or the index of the inner
+             // instruction. While it would add value to the log, to do so at this point would also
+             // break any log parser that presumes a stable log format
+             // (eg. https://tinyurl.com/3uuczr68).
+             =>  write!(f, "Error processing Instruction {outer_instruction_index}: {err}"),
             Self::CallChainTooDeep
              => f.write_str("Loader call chain is too deep"),
             Self::MissingSignatureForFee
@@ -415,3 +442,78 @@ impl TransportError {
 
 #[cfg(not(target_os = "solana"))]
 pub type TransportResult<T> = std::result::Result<T, TransportError>;
+
+#[cfg(feature = "serde")]
+#[cfg(test)]
+mod tests {
+    use {
+        crate::TransactionError,
+        serde_json::{from_value, json, to_value},
+        solana_instruction::error::InstructionError,
+        solana_pubkey::Pubkey,
+        test_case::test_case,
+    };
+
+    #[cfg(feature = "serde")]
+    #[test_case(InstructionError::Custom(42), 1, Some(Pubkey::new_unique()), None; "From top-level instruction")]
+    #[test_case(InstructionError::Custom(42), 1, Some(Pubkey::new_unique()), Some(41); "From inner instruction")]
+    #[test_case(InstructionError::Custom(42), 1, None, None; "Legacy instruction without inner/program")]
+    fn test_serialize_instruction_error(
+        err: InstructionError,
+        outer_instruction_index: u8,
+        responsible_program_address: Option<Pubkey>,
+        inner_instruction_index: Option<u8>,
+    ) {
+        let json = to_value(TransactionError::InstructionError {
+            err: err.clone(),
+            inner_instruction_index,
+            outer_instruction_index,
+            responsible_program_address,
+        })
+        .unwrap();
+
+        assert_eq!(
+            json,
+            json!({
+                "InstructionError": {
+                    "err": err,
+                    "inner_instruction_index": inner_instruction_index,
+                    "outer_instruction_index": outer_instruction_index,
+                    "responsible_program_address": responsible_program_address,
+                },
+            }),
+        )
+    }
+
+    #[cfg(feature = "serde")]
+    #[test_case(InstructionError::Custom(42), 1, Some(Pubkey::new_unique()), None; "From top-level instruction")]
+    #[test_case(InstructionError::Custom(42), 1, Some(Pubkey::new_unique()), Some(41); "From inner instruction")]
+    #[test_case(InstructionError::Custom(42), 1, None, None; "Legacy instruction without inner/program")]
+    #[test_case(InstructionError::Custom(42), 1, None, Some(41); "Mixed instruction with inner index but no program address")]
+    fn test_deserialize_instruction_error(
+        err: InstructionError,
+        outer_instruction_index: u8,
+        responsible_program_address: Option<Pubkey>,
+        inner_instruction_index: Option<u8>,
+    ) {
+        let decoded_err: TransactionError = from_value(json!({
+            "InstructionError": {
+                "err": err,
+                "inner_instruction_index": inner_instruction_index,
+                "outer_instruction_index": outer_instruction_index,
+                "responsible_program_address": responsible_program_address,
+            },
+        }))
+        .unwrap();
+
+        assert_eq!(
+            decoded_err,
+            TransactionError::InstructionError {
+                err,
+                inner_instruction_index,
+                outer_instruction_index,
+                responsible_program_address,
+            },
+        )
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4062,10 +4062,13 @@ version = "2.2.1"
 dependencies = [
  "serde",
  "serde_derive",
+ "serde_json",
  "solana-frozen-abi",
  "solana-frozen-abi-macro",
  "solana-instruction",
+ "solana-pubkey",
  "solana-sanitize",
+ "test-case",
 ]
 
 [[package]]
--- a/transaction-error/Cargo.toml
+++ b/transaction-error/Cargo.toml
@@ -17,7 +17,12 @@ solana-frozen-abi-macro = { workspace = true, optional = true }
 solana-instruction = { workspace = true, default-features = false, features = [
     "std",
 ] }
+solana-pubkey = { workspace = true }
 solana-sanitize = { workspace = true }
+
+[dev-dependencies]
+serde_json = { workspace = true }
+test-case = { workspace = true }
 
 [features]
 frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro"]
--- a/transaction/src/sanitized.rs
+++ b/transaction/src/sanitized.rs
@@ -298,11 +298,11 @@ impl SanitizedTransaction {
                 self.message().instructions(),
                 feature_set,
             )
-            .map_err(|err| {
-                TransactionError::InstructionError(
-                    index as u8,
-                    solana_instruction::error::InstructionError::Custom(err as u32),
-                )
+            .map_err(|err| TransactionError::InstructionError {
+                err: solana_instruction::error::InstructionError::Custom(err as u32),
+                inner_instruction_index: None,
+                outer_instruction_index: index as u8,
+                responsible_program_address: Some(*program_id),
             })?;
         }
         Ok(())


steveluscher:
This is a bit of a spitball pr. Things I'm thinking about:

1. This will require a major version bump, because Rust.
2. I imagine that this will increase the storage cost of transactions by the 32 bytes of the pubkey. Maybe this is fine, but perhaps there is an alternative, like storing the _index_ of the program in the _accounts array_ and then reconstructing the program address later (ie. in the RPC method that loads and vends the `InstructionError`)?

transaction-error/src/lib.rs
```diff
@@ -43,8 +46,11 @@ pub enum TransactionError {
     BlockhashNotFound,
 
     /// An error occurred while processing an instruction. The first element of the tuple
-    /// indicates the instruction index in which the error occurred.
-    InstructionError(u8, InstructionError),
+    /// indicates the instruction index in which the error occurred. For avoidance of doubt, the
+    /// third element indicates the address of the program that raised the error, if applicable; the
+    /// error could after all have been raised during a cross-program invocation (ie. in an inner
+    /// instruction).
+    InstructionError(u8, InstructionError, Option<Pubkey>),
```
steveluscher: We won't have a program address in cases where the error is `InstructionError::UnsupportedProgramId`, so this is made an `Option`.

kevinheavey:
btw this is a breaking change, but we're already due a bump to 3.0 because of https://github.com/anza-xyz/solana-sdk/issues/46

transaction-error/src/lib.rs
apfitzge: adding `Option<Pubkey>` here means `TransactionError` is now 64 bytes instead of 32 bytes. I'm not sure how often we move/clone these on the agave side, but this could have some small perf implications. I'm aware of at least a few places we clone these values.

apfitzge: Can you elaborate what cases `UnsupportedProgramId` is for?

apfitzge: ^ this is not a comment on why we shouldn't do this - just a note that we should be aware of this size change, and be more aware of where we're cloning these in our processing pipeline so we can avoid performance hits

steveluscher: Yeah, for sure. It's thrown, for instance, here:

https://github.com/anza-xyz/agave/blob/1bcf743f445270407bccf55681f74e06ef8b9a48/program-runtime/src/invoke_context.rs#L250-L252

ie. if you tried to issue an instruction [without providing the address of a program](https://github.com/anza-xyz/agave/blob/1bcf743f445270407bccf55681f74e06ef8b9a48/program-runtime/src/invoke_context.rs#L250-L252).

The larger point here is that there exist `InstructionErrors` that are more like ‘there was a structural problem with this instruction’ rather than ‘a program died and here's why.’ The latter has a program address associated with it while the former may not.

steveluscher: Feels bad. I should probably endeavour to store _just_ the index of the program account in the accounts array, and then make the downstream thing that _really_ wants to know what the program was look it up.

I'll rifle through the Agave code and see if that's tractable (ie. the instruction will need to be available wherever something needs know what program address was involved).

steveluscher:
So, here's what I'm thinking. There are three ways to approach this.

# Option 1 – Add program address to `TransactionError::InstructionError`

```diff
- TransactionError::InstructionError(u8, InstructionError)
+ TransactionError::InstructionError(u8, InstructionError, Option<Pubkey>)
```

* the address (or address index) will have to be `Option`, because not all `InstructionErrors` are related to a program
* breaks people who construct `TransactionError::InstructionError` (ie. the SVM)
* does not break people who construct `InstructionError`

# Option 2 – Add program address to every `InstructionError` that is related to a program

```diff
  ArithmeticOverflow,  // Leave as is; not related to a program
- InstructionError::CallDepth,
+ InstructionError::CallDepth(Pubkey)
```

* does not break people who construct `TransactionError::InstructionError` (ie. the SVM)
* breaks people who construct `InstructionError`

> [!NOTE]
> There still might exist some `InstructionErrors` that _sometimes_ have programs related to them and other times do not. These would still, unfortunately for complexity, be `Option<Pubkey>`

# Option 3 – Create a `TransactionError::InstructionErrorCausedByProgram` type

```diff
  InstructionError(u8, InstructionError),
+
+ InstructionErrorCausedByProgram(u8, InstructionError, Pubkey)
```

* no need for making the address (or address index) `Option`; just return the relevant `TransactionError::InstructionError`
* would need to make a separate enum to represent the union of both, or use `Either` (ie. functions would have to have return types like `Result<(), Either<InstructionError, InstructionErrorCausedByProgram>)`

joncinque:
So good news, I went through all of the `InstructionError` variants, and they pretty much all pertain to a certain program, or preparing the execution of a certain program. [1]

To go with that, the `TransactionError::InstructionError` variant uses a `u8` for the index of the top-level instruction in the transaction, which means that there's already some notion of "this error requires some greater context to truly understand".

I'm down to go with the option 1, but using a `u8` for the index of the account in the transaction, as mentioned earlier. Looking at the storage protos for transaction errors, I think we should just be able to add a new field for the originating account: https://github.com/anza-xyz/agave/blob/bc09ffa335d9773fd6c4b354e61c44b8fc36724a/storage-proto/proto/transaction_by_addr.proto#L21

It'll be a bit of a slog to get through all the changes, but it should be mostly straightforward. Happy to hear other ideas though!

[1] if I'm incorrect, then we should fix the usage of that particular error variant rather than torpedo the design

steveluscher:
> …pretty much all pertain to a certain program, or preparing the execution of a certain program

I started then scrapped a PR that did option 1 and I seem to remember it being really hard to obtain the program address in places. Here's `cargo check` after adding a `u8` to `TransactionError::InstructionError` (gist).

I can try again, but a few on spec:

* `UnsupportedProgramId` doesn't, by its very nature. https://github.com/anza-xyz/agave/blob/ccbf3f25f332cf4bfa4f1a9bf27db8d3333b3064/program-runtime/src/invoke_context.rs#L539
* Do `MissingAccount` and `NotEnoughAccountKeys` ever happen before we know what the program address is? https://github.com/anza-xyz/agave/blob/ccbf3f25f332cf4bfa4f1a9bf27db8d3333b3064/program-runtime/src/invoke_context.rs#L349, https://github.com/anza-xyz/agave/blob/ccbf3f25f332cf4bfa4f1a9bf27db8d3333b3064/program-runtime/src/invoke_context.rs#L361

The big problem is here, because it takes in a dynamic error after _long_ having forgotten what program is responsible for the error: https://github.com/anza-xyz/agave/blob/ccbf3f25f332cf4bfa4f1a9bf27db8d3333b3064/svm/src/message_processor.rs#L110

So basically you have to follow `process_precompile` and `process_instruction` all the way down, and make sure that they throw the responsible program address up through the `Result`, which means now you're not throwing an `InstructionError` there, you're throwing either an `InstructionErrorWithResponsibleProgram` or a `(InstructionError, u8)`. I found this to get really hairy. https://github.com/anza-xyz/agave/blob/ccbf3f25f332cf4bfa4f1a9bf27db8d3333b3064/svm/src/message_processor.rs#L71-L91

joncinque:
> UnsupportedProgramId doesn't, by its very nature. ([link](https://github.com/anza-xyz/agave/blob/ccbf3f25f332cf4bfa4f1a9bf27db8d3333b3064/program-runtime/src/invoke_context.rs#L539))

That one looks at the owner of the first program account in the transaction context, so we could probably use that, right? https://github.com/anza-xyz/agave/blob/ccbf3f25f332cf4bfa4f1a9bf27db8d3333b3064/program-runtime/src/invoke_context.rs#L523

> Do MissingAccount and NotEnoughAccountKeys ever happen before we know what the program address is? ([link](https://github.com/anza-xyz/agave/blob/ccbf3f25f332cf4bfa4f1a9bf27db8d3333b3064/program-runtime/src/invoke_context.rs#L349), [link](https://github.com/anza-xyz/agave/blob/ccbf3f25f332cf4bfa4f1a9bf27db8d3333b3064/program-runtime/src/invoke_context.rs#L361))

If they do, then that's an issue with their usage, which should be fixable.

> The big problem is here, because it takes in a dynamic error after long having forgotten what program is responsible for the error: ([link](https://github.com/anza-xyz/agave/blob/ccbf3f25f332cf4bfa4f1a9bf27db8d3333b3064/svm/src/message_processor.rs#L110))

That plumbing does look a bit involved, but should end up similar to any big refactor.

steveluscher:
> Rather than plumbing callee IDs all the way from SVM, what about just enabling inner instructions recording on Bank all of the time, and walking that payload to grab the program ID?

Oh, I like this. Also, we wouldn't have to enable CPI logging all the time; I think we could lazily walk the `TransactionContext` to figure out what the last instruction in the trace was before it died. I _think_ all that's involved is to see where the logs _ended_, to know in what program the error originated. I'll give this a shot.

transaction-error/src/lib.rs
steveluscher: Updated the PR! This is now implemented as a u8 that points to the transaction-level account index of the program responsible for the error.

transaction-error/src/lib.rs
```diff
@@ -42,9 +42,12 @@ pub enum TransactionError {
     /// the `recent_blockhash` has been discarded.
     BlockhashNotFound,
 
-    /// An error occurred while processing an instruction. The first element of the tuple
-    /// indicates the instruction index in which the error occurred.
-    InstructionError(u8, InstructionError),
+    /// An error occurred while processing an instruction. The first element of the tuple indicates
+    /// the index of the outer instruction in which the error occurred, and the third the account
+    /// index of the program responsible for the error (ie. the error may have originated from an
+    /// inner instruction). The account index of the responsible program may be `None` for
```
jstarry: I'm not sure this approach actually clears up much ambiguity. There could be multiple inner instructions that invoke the same program. In order to pinpoint the failing inner instruction you would still need to rely on the inner ix metadata that we already provide.

steveluscher: Good catch! Though the [original goal](https://github.com/anza-xyz/kit/issues/149) was to be able to correlate what program `Custom(###)` pertains to so that you can properly decode and handle `###`, we could _also_ add the path to the actual inner instruction.

jstarry: Sorry I don't really follow your response. Is the "original goal" still the current goal? When you say "we could also add the path to the actual inner instruction" is that rhetorical? If this is a "good catch," does this warrant any change with your current approach?

jstarry: Is it correct to say that you're just trying to map a custom program error code to the actual program that returned it?

jstarry: If we're just mapping custom error codes, do we need to add the tx account index for the failing program to all instruction error variants?

steveluscher: > If we're just mapping custom error codes…

We're not. That was the original impetus for this change, but the exploration led to the insight that _most_ `InstructionErrors` would benefit from knowing what program caused them.

steveluscher: > If this is a "good catch," does this warrant any change with your current approach?

Yep! It warrants me adding the inner instruction index so that you can – for instance – tell that a `InsufficientFundsForRent` came from instruction 1.4 as opposed to 1.2, even though 1.2 and 1.4 have the same program address.

steveluscher: I added the index of the inner instruction here in the type, and [here](https://github.com/anza-xyz/agave/pull/6083) in the implementation, @jstarry!

transaction-error/src/lib.rs
```diff
@@ -42,9 +42,14 @@ pub enum TransactionError {
...
-    /// indicates the instruction index in which the error occurred.
-    InstructionError(u8, InstructionError),
+    /// An error occurred while processing an instruction. The first element of the tuple indicates
+    /// the index of the outer instruction in which the error occurred, the third the index of the
+    /// inner instruction if the program responsible for the error was called by cross-program
+    /// invocation (CPI), and the fourth the account index of the program responsible for the error
+    /// (ie. the error may have originated from an inner instruction). The inner instruction index
+    /// may be `None` if the error originated from the top-level program call. The account index of
+    /// the  responsible program may be `None` for transactions created before it was introduced.
+    InstructionError(u8, InstructionError, Option<u8>, Option<u8>),
```
jstarry: If we're going to make a backwards incompatible change, might as well start naming these fields. Pretty easy to misuse this tuple as proposed

transaction-error/src/lib.rs
```diff
@@ -42,9 +42,14 @@ pub enum TransactionError {
     /// the `recent_blockhash` has been discarded.
     BlockhashNotFound,
 
-    /// An error occurred while processing an instruction. The first element of the tuple
-    /// indicates the instruction index in which the error occurred.
-    InstructionError(u8, InstructionError),
+    /// An error occurred while processing an instruction. The first element of the tuple indicates
+    /// the index of the outer instruction in which the error occurred, the third the index of the
+    /// inner instruction if the program responsible for the error was called by cross-program
+    /// invocation (CPI), and the fourth the account index of the program responsible for the error
```
jstarry: The program invoked by the inner instruction might be loaded from an ALT so downstream users may still need to fetch metadata to get the offending program id fyi

jstarry:
I think I'm a bit more up to speed on the evolution of the problem we're trying to solve here. I think it's a very good idea to add an indication that a given instruction error was actually caused by an inner instruction if that was the case. I'm not super excited about making a breaking change to the `InstructionError` variant to solve that problem.

Also, I think we're now trying to solve more than just the problem of knowing whether an error came from an inner instruction. We're trying to know which inner program produced the error and also where in the call tree the error was produced (partly my fault we went down that path due to this thread: https://github.com/anza-xyz/solana-sdk/pull/74#discussion_r2074642754)

I think that it's ok to force users to debug errors by first fetching tx metadata for now. They need metadata to resolve ALT's, to have full context of the call tree leading to the failure, as well as any pertinent logs. I think later we could separately add a better way to get rich error context from the SVM which includes the full call stack in errors so that local development doesn't require metadata fetching.

For differentiating top level vs inner instruction errors, I think I'm most in favor of option 3 where we add a new transaction error variant. As discussed in the PR, we can just use the program account index instead of including the full pubkey, so something like:
```rust
enum TransactionError {
  ..
  InnerInstructionError(InstructionError),
}
```
And then have the user figure out where in the call tree this happened from the tx metadata. I acknowledge this might be too minimalist, adding top level tx index and the account index of the program invoked in the inner instruction could be nice too.

It might also be worth exploring whether adding a new `InstructionError` variant would be less disruptive to users. We wouldn't want crazy recursive nesting but something like this could work:

```rust
enum InstructionError {
  ..
  InnerInstructionFailure(InnerInstructionError)
}

enum InnerInstructionError {
  // subset of InstructionError variants?
}
```

transaction-error/src/lib.rs
joncinque: Agreed, I had to look back at this PR a few times while reviewing the other one to remember which field was which

transaction-error/src/lib.rs
```diff
@@ -163,7 +168,18 @@ impl fmt::Display for TransactionError {
...
+                err,
+                _inner_instruction_index,
+                _responsible_program_account_index,
+            )
+             // NOTE: We intentionally do not augment the error message in the event that the error
+             // carries the index of the inner instruction or the account index of the responsible
+             // program. While it would add value to the log, to do so now would also break any log
+             // parser that presumed the log format to be immutable for all time (eg.
+             // https://tinyurl.com/3uuczr68).
+             =>  write!(f, "Error processing Instruction {index}: {err}"),
```
joncinque: I think that's fine for now, people can also use the `Debug` formatting to get every piece of data.

We could consider creating a new enum variant, but then the UMI parser would be totally useless, which is even worse in my opinion.

joncinque (commented):
Just a couple of comments, looks good otherwise

transaction-error/src/lib.rs
steveluscher: Ugh. This really sucks. I'll have to think about this for a hot second. Ideally we'd just store the program address, but that would increase the size of stored errors quite a bit, which we should try to avoid.

steveluscher:
Thanks for the detailed thoughts, @jstarry.

> I think that it's ok to force users to debug errors by first fetching tx metadata for now.

tx metadata is, unfortunately, insufficient.

Consider these three transactions.

* -> Program A
  * -> Program B
    * -> Program C
    * <- Program C (**ERROR**)
  * <- Program B
* <- Program A

---

* -> Program A
  * -> Program B
    * -> Program C
    * <- Program C
  * <- Program B (**ERROR**)
* <- Program A

---

* -> Program A
  * -> Program B
    * -> Program C
    * <- Program C
  * <- Program B
* <- Program A (**ERROR**)

The inner instruction trace will look the same in all of these cases, making it impossible to discern where the error actually came from.

> They need metadata to resolve ALT's

Guh, this sucks. In _some_ cases you'll have the ALT information locally (eg. you used `@solana/kit` to _create_ the transaction, and the error happened in simulation) but not in all cases for sure. I'll have to consult with @steviez to see if we can stand another 32 bytes of context data in stored errors or not, because the ideal thing would be to just bake the address itself into the error, rather than the index of the program account.

> I think later we could separately add a better way to get rich error context from the SVM…

This is underway in https://github.com/anza-xyz/agave/pull/6083/commits

jstarry:
> The inner instruction trace will look the same in all of these cases, making it impossible to discern where the error actually came from.

Ah yeah thanks for pointing that out, definitely insufficient to just look at metadata as it is right now, you would need to know the stack height of the last instruction which was running. You could probably parse the logs for this but that's not a great solution.

I guess my suggestion with going with option 3 that you listed earlier would be contingent on adding the stack height to the new error variant (something like `InnerInstructionError { stack_height: u8, err: InstructionError}`). If we had that, then we could actually be ok with having users fetch metadata for now.

transaction-error/src/lib.rs
steveluscher: I did not end up doing this, because we'd have to write a custom `serde` serializer to serialize the new struct variant with named fields in the old format the RPC requires.

What I did do, however, is to make type aliases, which gives you a bit of guidance when you're _constructing_ one of these things.

image

transaction-error/src/lib.rs
```diff
@@ -42,9 +46,20 @@ pub enum TransactionError {
...
+    /// outer instruction from which it was spawned via CPI. The account index of the responsible
+    /// program and the inner instruction index may be `None` for transactions created before they
+    /// were introduced. For all other transactions, the inner instruction index may be `None` when
+    /// the error originated from the outer instruction itself.
+    InstructionError(
+        OuterInstructionIndex,
+        InstructionError,
+        Option<ResponsibleProgramAccountIndex>,
+        Option<InnerInstructionIndex>,
+    ),
```
buffalojoec: The aliasing helps read _this file_, but anywhere you access these values in the runtime it's still going to be a tuple-index access, which means maintainers would have to go back to this file to double-check which index is what.

Personally I would rather see a struct here.

transaction-error/src/lib.rs
```diff
@@ -163,7 +178,17 @@ impl fmt::Display for TransactionError {
...
+             idx,
+             err,
+             _responsible_program_account_index,
+             _inner_instruction_index,
+            )
+             // NOTE: We intentionally do not augment the error message in the event that the error
+             // carries the account index of the responsible program. While it would add value to
+             // the log, but to do so at this point would also break any log parser that presumes a
+             // stable log format (eg. https://tinyurl.com/3uuczr68).
+             =>  write!(f, "Error processing Instruction {idx}: {err}"),
```
buffalojoec: Makes sense! I'm pretty sure if we did change this, it would break consensus.

transaction-error/src/lib.rs
buffalojoec: > I did not end up doing this, because we'd have to write a custom `serde` serializer to serialize the new struct variant with named fields in the old format the RPC requires.

this shouldn't be too bad to do, just a bit verbose.

transaction-error/src/lib.rs
buffalojoec: I thought we were just dealing in indices all over the place? Why do we need the ALT if we know the index? Are you guys talking about the eventual end-user on the client side having to resolve a program ID from the index?

jstarry: Yeah when I said downstream users I meant the the client side end user experience

steveluscher:
OK, this puppy is ready to land, and is the prerequisite/companion to https://github.com/anza-xyz/agave/pull/6083.

* We're going with the strategy of breaking the enum type and bumping the major version of `solana-transaction-error`
* Backward compatibility with old data and old clients has been achieved by _airgapping_ the `TransactionError::InstructionError` type from all of the places in Agave where the old serialization and storage formats must be maintained. See the introduction of `StoredTransactionError` and `UiTransactionError` in https://github.com/anza-xyz/agave/pull/6083, the custom serializer/deserializer implemented on each, and the extensive tests.

transaction-error/src/lib.rs
```diff
@@ -42,9 +45,23 @@ pub enum TransactionError {
...
-    /// indicates the instruction index in which the error occurred.
-    InstructionError(u8, InstructionError),
+    /// An error occurred while processing an instruction.
+    InstructionError {
+        err: InstructionError,
+        /// The index of the inner instruction in which the error was thrown, starting from zero.
+        /// This value will be `None` when the error was thrown from the outer instruction's
+        /// program, and also for all errors stored using a version of this enum variant prior to
+        /// `solana-transaction-error` 3.0.0.
+        inner_instruction_index: Option<u8>,
```
jstarry, line 55: 1. I think this would have to be a `u16` because it's possible that an instruction makes more than 256 CPI's.
2. I prefer we don't record this index because it requires fetching the CPI trace metadata to understand and if we require fetching the CPI trace, I think it's better to record the CPI stack height instead.

transaction-error/src/lib.rs
```diff
@@ -42,9 +45,23 @@ pub enum TransactionError {
...
+        /// The index of the outer instruction in which the error was thrown, starting from zero. Do
+        /// not infer the responsible program from the instruction at this index; the error might
+        /// have been thrown from one of its inner instructions. See `responsible_program_address`.
+        outer_instruction_index: u8,
```
jstarry, line 59: While we're making a breaking change, let's change this to a `u16` because I think we have a bug where we incorrectly wrap around the ix index when a tx has more than 256 instructions.

transaction-error/src/lib.rs
```diff
@@ -42,9 +45,23 @@ pub enum TransactionError {
...
+        /// The address of the program that threw the error. Use this to decode the error (eg. to
+        /// look up a custom error code in the program's IDL). This value will be `None` for errors
+        /// stored using a version of this enum variant prior to `solana-transaction-error` 3.0.0.
+        responsible_program_address: Option<Pubkey>,
```
jstarry, line 63: We shouldn't store the program address here. It can be looked up via tx metadata. Let's record the stack height of the failure and the tx account index of the responsible program instead. If the responsible program was resolved from an ALT, the end user should resolve that themselves.

steveluscher: A central principle of this change is that the error alone should be scrutible; all questions you might have about an error should be encapsulated in the error itself.

In any case, you might obtain this error from `simulateTransaction` where you definitely can't fetch tx metadata, and you may not possess the message that was actually simulated from which to decode tx account indexes (eg. it might have been modified by your wallet).

transaction-error/src/lib.rs
steveluscher: The idea behind this value was not to correlate it with the trace, but rather to give clients something they can use for display. For instance, explorers typically think of the first inner instruction of the second top-level instruction as being ‘Instruction 2.1.’ The indexes for such an instruction would be `{ outer_instruction_index: 1, inner_instruction_index: Some(0) }`.

Adding this value also doesn't add any weight in blockstore, since I was able to compress it into the existing uint32 that currently stores the top-level (outer) instruction index.

transaction-error/src/lib.rs
jstarry: > A central principle of this change is that the error alone should be scrutible; all questions you might have about an error should be encapsulated in the error itself.

I think you're trying to maximize error context and I'm trying to minimize it. I was under the assumption that the main problem this change is trying to fix is providing a way to correlate the error with the program that returned it. The other problems you're trying to solve can be solved by dev infra tools that provide users with full instruction traces and stuff.

transaction-error/src/lib.rs
jstarry: The explorer feature you mentioned is fully reliant on the CPI trace. I'm not concerned with extra weight in the blockstore. I just want this to provide the minimal context needed to correlate the error.

steveluscher: Unfortunately, I can't convert this to a `u16`.

In blockstore, we don't store structured errors, we just [dump bytes into a field](https://github.com/anza-xyz/agave/blob/ee9af5b672637fa9bd06b6c5de83076e0cbdfe9a/storage-proto/proto/confirmed_block.proto#L73).

This results in errors appearing like this in long-term storage:

```
// TransactionError::InstructionError(42, InstructionError::Custom(0xdeadbeef))
vec![
    8, 0, 0, 0, /* Eighth enum variant - `InstructionError` */
    42, /* Outer instruction index */
    25, 0, 0, 0, 239, 190, 173, 222, /* InstructionError::Custom(0xdeadbeef) */
];
```

I can't start writing a `u16` into the part of that serialization destined for `outer_instruction_index`. Old versions of the validator won't be able to read that, and new versions of the validator will stop understanding old stored data.

apfitzge: I don't think we need a `u16`, since we only execute up to 64 instructions before we hit our instruction limit (which I think is a separate error). At that point it wouldn't be an instruction error...so while it's technically (now) possible to have an instruction index > 255...I don't think it's actually possible to get an index>255 here?

jstarry: Cool, yeah not possible to get index > 255 right now so can skip the `u16`, sorry for the noise!

PR #76, opened 2025-03-09
chore(pubkey): (may be) proper serde json support (encode as base58 string, not as array) and json schema

dzmitry-lahoda:
- wasm, display, debug are base58 encoded. so serde json to be too
- json schema for serde
- some hopefully proper alloc/std shuffling (as forced by serde impl)
- because breaking, bumped +1 major version

not sure: is anybody uses serde for not json? like if anybody needs serde to array instead of base58 string? anyway, bumped major. also assuming binary can impl serde manually, while json expect convenience (especially in public apis)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3528,7 +3528,7 @@ dependencies = [
 
 [[package]]
 name = "solana-pubkey"
-version = "2.3.0"
+version = "3.0"
 dependencies = [
  "anyhow",
  "arbitrary",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -905,6 +905,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


juchiast:
> because breaking, bumped +1 version

wouldn't this make it v3

kevinheavey:
> not sure: is anybody uses serde for not json? like if anybody needs serde to array instead of base58 string?

yeah there are many places where bincode (which uses serde) is being used to serialize a pubkey

joncinque:
I appreciate the PR, but to be honest, I'm not too keen on the change. This will very likely break things in unexpected ways downstream, without really giving any benefit.

We can leave the PR open to see if any important use-cases emerge, but without a strong motivation, I would not simply make this breaking change.

PR #80, opened 2025-03-18
bn254: restrict loop to ele_len

0x0ece:
Make the loop a bit stricter, analogous to the [original loop](https://github.com/anza-xyz/agave/pull/3341/files#diff-270570bde4c47c131010de66cfe79639a12ce5d8b7c9412a9688c1402c947f8fL250).

### DIFF
--- a/bn254/src/lib.rs
+++ b/bn254/src/lib.rs
@@ -806,4 +806,15 @@ mod tests {
             assert_eq!(result.unwrap(), expected);
         });
     }
+
+    #[test]
+    fn alt_bn128_pairing_invalid_length() {
+        use ark_ff::{BigInteger, BigInteger256};
+
+        let input = [0; 193];
+        let result = alt_bn128_pairing(&input);
+        assert!(result.is_ok());
+        let expected = BigInteger256::from(1u64).to_bytes_be();
+        assert_eq!(result.unwrap(), expected);
+    }
 }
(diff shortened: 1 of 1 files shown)


samkim-crypto (commented):
Looks good to me! Can we maybe just add a simple test along the lines of 
```
    #[test]
    fn alt_bn128_invalid_lengths() {
        let input = [0; 193];
        alt_bn128_pairing(&input).unwrap();
    }
```
just for sanity. This test should fail before this change.

samkim-crypto (approved):

LStan:
Why doesn't this break the consensus?

samkim-crypto:
> Why doesn't this break the consensus?

This was a fix on a change that I made in https://github.com/anza-xyz/agave/pull/3341/files#diff-270570bde4c47c131010de66cfe79639a12ce5d8b7c9412a9688c1402c947f8fR321 that accidentally introduced a breaking change. The `chunks` logic that I added was assuming that https://github.com/anza-xyz/solana-sdk/blob/master/bn254/src/lib.rs#L321-L327 would properly reject invalid length inputs.

LStan:
Then why didn't your change break the consensus?

samkim-crypto:
The breaking change was introduced in the initial version of v2.2, but this PR fixed it for the subsequent versions of v2.2. The validators upgraded to the fixed version directly from v2.1.

PR #82, opened 2025-03-21
add solana-feature-set-interface crate

t-nelson:
step one of monorepo reclaiming the feature-set

this takes a copy of all generic symbols from solana-feature-set. we will wrap it in a new crate with the agave/cluster-specific symbols declared

possible that we don't want this at all and should just take a full copy back to monorepo

### DIFF
--- a/feature-set-interface/Cargo.toml
+++ b/feature-set-interface/Cargo.toml
@@ -1,0 +1,32 @@
+[package]
+name = "solana-feature-set-interface"
+description = "Interface for Solana runtime features."
+documentation = "https://docs.rs/solana-feature-set-interface"
+version = "2.2.4"
+authors = { workspace = true }
+repository = { workspace = true }
+homepage = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+
+[dependencies]
+ahash = { workspace = true }
+solana-epoch-schedule = { workspace = true }
+solana-frozen-abi = { workspace = true, optional = true, features = [
+    "frozen-abi",
+] }
+solana-frozen-abi-macro = { workspace = true, optional = true, features = [
+    "frozen-abi",
+] }
+solana-hash = { workspace = true }
+solana-pubkey = { workspace = true }
+solana-sha256-hasher = { workspace = true }
+
+[features]
+frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro"]
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[lints]
+workspace = true
--- a/feature-set-interface/Cargo.toml
+++ b/feature-set-interface/Cargo.toml
@@ -11,16 +11,13 @@ edition = { workspace = true }
 
 [dependencies]
 ahash = { workspace = true }
-solana-epoch-schedule = { workspace = true }
 solana-frozen-abi = { workspace = true, optional = true, features = [
     "frozen-abi",
 ] }
 solana-frozen-abi-macro = { workspace = true, optional = true, features = [
     "frozen-abi",
 ] }
-solana-hash = { workspace = true }
 solana-pubkey = { workspace = true }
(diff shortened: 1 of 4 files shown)


feature-set-interface/src/lib.rs
```diff
@@ -0,0 +1,75 @@
+//! Collection of all runtime features.
```
jstarry: This doc comment isn't relevant to this interface crate

jstarry (commented):
I think it's useful to have this, just a minor comment

feature-set-interface/src/lib.rs
t-nelson: removed the whole block

jstarry (dismissed):

willhickey:
Is the plan to backport this to v2.2, but not v2.1 (since feature-set/src/lib.rs is still in v2.1)?

feature-set-interface/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+version = "2.2.4"
+authors = { workspace = true }
+repository = { workspace = true }
+homepage = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+
+[dependencies]
+ahash = { workspace = true }
+solana-epoch-schedule = { workspace = true }
```
joncinque: This doesn't look like it's used

feature-set-interface/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+solana-frozen-abi = { workspace = true, optional = true, features = [
+    "frozen-abi",
+] }
+solana-frozen-abi-macro = { workspace = true, optional = true, features = [
+    "frozen-abi",
+] }
+solana-hash = { workspace = true }
+solana-pubkey = { workspace = true }
+solana-sha256-hasher = { workspace = true }
```
joncinque: This isn't used either

feature-set-interface/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+[dependencies]
+ahash = { workspace = true }
+solana-epoch-schedule = { workspace = true }
+solana-frozen-abi = { workspace = true, optional = true, features = [
+    "frozen-abi",
+] }
+solana-frozen-abi-macro = { workspace = true, optional = true, features = [
+    "frozen-abi",
+] }
+solana-hash = { workspace = true }
```
joncinque: This isn't used either

feature-set-interface/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+    solana_pubkey::Pubkey,
+};
+
+/// `FeatureSet` holds the set of currently active/inactive runtime features
+#[cfg_attr(feature = "frozen-abi", derive(solana_frozen_abi_macro::AbiExample))]
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct FeatureSet {
+    pub active: AHashMap<Pubkey, u64>,
+    pub inactive: AHashSet<Pubkey>,
+}
```
joncinque, line 14: I was going to suggest re-exporting this in `solana-feature-set` to avoid breakage, but that isn't possible without other kinds of breakage. But maybe that's ok, and we can have a breaking change on `solana-feature-set`.

What do you think?

joncinque (commented):
Just a couple of questions, makes sense to me!

feature-set-interface/src/lib.rs
t-nelson: `solana-feature-set` will be deprecated in the last pr of this set

feature-set-interface/Cargo.toml
t-nelson: derp, thought i cleaned these up. sec

jstarry (approved):

feature-set-interface/src/lib.rs
joncinque: Gotcha, works for me!

joncinque (approved):

PR #85, opened 2025-03-25
Add PrecompilesFeatureSet

jstarry:
#### Problem

The SDK shouldn't need to be coupled to specific feature keys to enable feature gated functionality in the SDK.

#### Summary
Breaking change to precompile interface to enable feature gates via flags so that the SDK doesn't need to be aware of or coupled to feature keys. This requires a major version bump of the precompiles and transaction crate according to semvar rules

- Update the `Precompile` struct to have an enable function. Really this should have all be based on traits :/


jstarry:
This can wait until after the agave feature set migration is finished if y'all want but I think this will be a good change regardless because we shouldn't have a dependency on agave crates from the sdk if we can help it.

precompiles/src/lib.rs
```diff
@@ -1,46 +1,45 @@
...
     solana_precompile_error::PrecompileError, solana_pubkey::Pubkey,
 };
 
 /// All precompiled programs must implement the `Verify` function
-pub type Verify = fn(&[u8], &[&[u8]], &FeatureSet) -> std::result::Result<(), PrecompileError>;
+pub type Verify =
+    fn(&[u8], &[&[u8]], &PrecompileFeatureSet) -> std::result::Result<(), PrecompileError>;
+
+/// All precompiled programs must implement the `Verify` function
+pub type Enabled = fn(&PrecompileFeatureSet) -> bool;
```
illia-bobyr:
```suggestion
/// All precompiled programs must implement the `Enabled` function
pub type Enabled = fn(&PrecompileFeatureSet) -> bool;
```

jstarry:
> The big drawback with the other approach is redefining the feature ids, which isn't an amazing solution.

Yeah if it weren't for the feature ids I would definitely prefer just switching the feature set interface. Coupling sdk releases to feature id's in agave is far from ideal and likely to cause a lot of headaches and potentially serious bugs.

I think there are other solutions besides this approach. @t-nelson had some ideas as well to have some sort of generic `LocalFeatures` struct but I wasn't super clear on that.

jstarry:
@t-nelson suggested a better approach.. just migrate agave from using `solana-precompiles` to `agave-precompiles` and deprecate any usage of `solana-precompiles` in the sdk. Precompile errors stay for now tho

jstarry:
I'll likely re-introduce this change in agave, it helps make ix iteration faster by doing less feature set lookups

PR #86, opened 2025-03-25
feature-set-interface: Use it in the sdk

joncinque:
#### Problem

solana-feature-set-interface was added to decouple the feature ids from the sdk, but it still isn't used in any of the sdk crates.

#### Summary of changes

Use `solana-feature-set-interface` everywhere. Note that this requires some redefinitions of feature ids from `solana-feature-set`.


ed25519-program/src/lib.rs
```diff
@@ -18,6 +18,11 @@ pub const SIGNATURE_OFFSETS_SERIALIZED_SIZE: usize = 14;
 pub const SIGNATURE_OFFSETS_START: usize = 2;
 pub const DATA_START: usize = SIGNATURE_OFFSETS_SERIALIZED_SIZE + SIGNATURE_OFFSETS_START;
 
+/// Copy from agave-feature-set
```
jstarry, line 21: Yeah this is exactly what I was trying to avoid in https://github.com/anza-xyz/solana-sdk/pull/85. By doing this we are coupling sdk crate versions to runtime releases in a way that will be really tricky to track, especially for rekeys.

joncinque: Yep that's totally fair!

joncinque:
Closing in favor of https://github.com/anza-xyz/agave/pull/5509

PR #87, opened 2025-03-26
move less important methods from Sysvar to SysvarSerialize

kevinheavey:
Problem: the Sysvar trait relies on bincode, but most of the time people just want `Sysvar::get` which doesn't rely on bincode

~Solution: make a `SysvarGet` trait that has one method: `get`. Make the `Sysvar` trait use `SysvarGet::get`. Annoyingly this is still technically a breaking change because we are adding a constraint to `Sysvar`, but for the vast majority of users nothing should break~

Solution: move the other methods out of `Sysvar` and into a new `SysvarSerialize` trait

### DIFF
--- a/sysvar/src/lib.rs
+++ b/sysvar/src/lib.rs
@@ -83,9 +83,9 @@ pub mod __private {
     pub use solana_define_syscall::definitions;
     pub use {solana_program_entrypoint::SUCCESS, solana_program_error::ProgramError};
 }
-use {solana_pubkey::Pubkey, solana_program_error::ProgramError};
 #[cfg(feature = "bincode")]
 use {solana_account_info::AccountInfo, solana_sysvar_id::SysvarId};
+use {solana_program_error::ProgramError, solana_pubkey::Pubkey};
 
 pub mod clock;
 pub mod epoch_rewards;
--- a/sysvar/src/lib.rs
+++ b/sysvar/src/lib.rs
@@ -128,9 +128,7 @@ pub trait SysvarGet: Default + Sized {
 
 #[cfg(feature = "bincode")]
 /// A type that holds sysvar data.
-pub trait Sysvar:
-    SysvarGet + SysvarId + serde::Serialize + serde::de::DeserializeOwned
-{
+pub trait Sysvar: SysvarGet + SysvarId + serde::Serialize + serde::de::DeserializeOwned {
     /// The size in bytes of the sysvar as serialized account data.
     fn size_of() -> usize {
         bincode::serialized_size(&Self::default()).unwrap() as usize
--- a/sysvar/src/lib.rs
+++ b/sysvar/src/lib.rs
@@ -31,7 +31,7 @@
 //!
 //! Since Solana sysvars are accounts, if the `AccountInfo` is provided to the
 //! program, then the program can deserialize the sysvar with
-//! [`Sysvar::from_account_info`] to access its data, as in this example that
+//! [`SysvarSerialize::from_account_info`] to access its data, as in this example that
 //! again logs the [`clock`] sysvar.
 //!
 //! ```
@@ -158,7 +158,7 @@ pub trait SysvarSerialize:
         bincode::serialize_into(&mut account_info.data.borrow_mut()[..], self).ok()
     }
 
-    /// Calls [`SysvarGet::get`].
+    /// Calls [`Sysvar::get`].
     fn get() -> Result<Self, ProgramError> {
         <Self as Sysvar>::get()
     }
--- a/sysvar/src/lib.rs
+++ b/sysvar/src/lib.rs
@@ -128,7 +128,9 @@ pub trait Sysvar: Default + Sized {
 
 #[cfg(feature = "bincode")]
 /// A type that holds sysvar data.
-pub trait SysvarSerialize: Sysvar + SysvarId + serde::Serialize + serde::de::DeserializeOwned {
+pub trait SysvarSerialize:
+    Sysvar + SysvarId + serde::Serialize + serde::de::DeserializeOwned
+{
     /// The size in bytes of the sysvar as serialized account data.
     fn size_of() -> usize {
         bincode::serialized_size(&Self::default()).unwrap() as usize
--- a/sysvar/src/lib.rs
+++ b/sysvar/src/lib.rs
@@ -37,7 +37,7 @@
 //! ```
 //! use solana_account_info::{AccountInfo, next_account_info};
 //! use solana_msg::msg;
-//! use solana_sysvar::Sysvar;
+//! use solana_sysvar::{Sysvar, SysvarSerialize};
 //! use solana_program_error::ProgramResult;
 //! use solana_pubkey::Pubkey;
 //!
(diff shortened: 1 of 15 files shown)


sysvar/src/lib.rs
```diff
@@ -145,10 +142,25 @@ pub fn is_sysvar_id(id: &Pubkey) -> bool {
     ALL_IDS.iter().any(|key| key == id)
 }
 
+/// Interface for loading a sysvar.
+pub trait SysvarGet: Sized {
```
febo: I think this is missing a `Default` constraint since the `get` macro implementation is doing:
```rust
let mut var = Self::default();
```

sysvar/src/lib.rs
```diff
@@ -145,10 +142,25 @@ pub fn is_sysvar_id(id: &Pubkey) -> bool {
...
+    fn get() -> Result<Self, ProgramError> {
+        Err(ProgramError::UnsupportedSysvar)
+    }
+}
+
 #[cfg(feature = "bincode")]
 /// A type that holds sysvar data.
 pub trait Sysvar:
-    SysvarId + Default + Sized + serde::Serialize + serde::de::DeserializeOwned
+    SysvarGet + SysvarId + Default + serde::Serialize + serde::de::DeserializeOwned
```
febo: Once the `Default` constraint is on `SysvarGet`, it can be removed from here.

febo (commented):
This change is great! I wonder if we could take the opportunity of this breaking change to do a bigger refactoring. Ideally we should move the implementation of each sysvar to their own individual crate, instead of having all of them as a dependency on this one. 

So `solana-sysvar` has the trait definitions and macros only. What do you think?

Another alternative is to move the trait definitions and macros to a new crate (`solana-define-sysvar`?), move the implementation to each individual sysvar crate and keep this one re-exporting everything to "minimize" the breaking change.

kevinheavey:
Sounds like another PR to me

febo:
> Sounds like another PR to me

Yeah, definitely. I was mentioning to get an idea if this is an approach that makes sense. I am inclined to go for the `solana-define-sysvar` option, which in a way mimic the syscalls approach.

kevinheavey:
The `solana-define-sysvar` approach is a bit awkward because of the orphan rule. We'd have to move the trait impls to the existing crates like solana-clock. But maybe that's pretty cheap once we have this lightweight SysvarGet trait

febo:
> The `solana-define-sysvar` approach is a bit awkward because of the orphan rule. We'd have to move the trait impls to the existing crates like solana-clock. But maybe that's pretty cheap once we have this lightweight SysvarGet trait

Yeah, the new trait is lightweight and the impl could be behind a feature on the individual crates.

sysvar/src/lib.rs
```diff
@@ -145,11 +142,24 @@ pub fn is_sysvar_id(id: &Pubkey) -> bool {
     ALL_IDS.iter().any(|key| key == id)
 }
 
+/// Interface for loading a sysvar.
+pub trait SysvarGet: Default + Sized {
```
febo: We removed the requirement for the `Default` trait in [pinocchio](https://github.com/anza-xyz/pinocchio/pull/189), so I wonder if we should do the same here as well. Individual sysvars could still implement `Default` when applicable.

febo: Maybe we should do this in a separate PR since we would need to modify the `impl_sysvar_get` macro implementation.

kevinheavey:
I don't think the other methods on `Sysvar` belong in the same trait as `get`:

- 99% of use cases just need `get`
- even if bincode is replaced (which will take a while), there will still be a significant amount of code involved in implementing those methods

febo:
> I don't think the other methods on `Sysvar` belong in the same trait as `get`:
> 
> * 99% of use cases just need `get`
> * even if bincode is replaced (which will take a while), there will still be a significant amount of code involved in implementing those methods

Agree. I should have been more specific when I said "refactor" `Sysvar`. I was thinking extracting the `bincode` related methods to a new trait and leave the `get` on the `Sysvar`. This will make it easier to eventually deprecate `bincode` without having to deprecate the `Sysvar` trait, which is a good trait name.

In other words, inverting the changes of the PR: the new `SysvarGet` stays as `Sysvar` and current `Sysvar` is renamed to `SysvarSerialize`/`SysvarDeserialize` (maybe a better name) with the `bincode` functionality.

kevinheavey:
oh ok yeah makes sense

sysvar/src/lib.rs
```diff
@@ -156,10 +153,25 @@ pub fn is_sysvar_id(id: &Pubkey) -> bool {
     ALL_IDS.iter().any(|key| key == id)
 }
 
+/// Interface for loading a sysvar.
+pub trait Sysvar: Default + Sized {
```
joncinque, line 115: We should eventually support all of the sysvars through this trait, including `SlotHashes` and `StakeHistory`, which are not really `Sized` (technically they are, but not for our purposes).

However, I think we can simply relax this trait and add the bounds to `SysvarSerialize` when we're ready for that work.

sysvar/src/lib.rs
```diff
@@ -188,16 +200,9 @@ pub trait Sysvar:
...
 
-    /// Load the sysvar directly from the runtime.
-    ///
-    /// This is the preferred way to load a sysvar. Calling this method does not
-    /// incur any deserialization overhead, and does not require the sysvar
-    /// account to be passed to the program.
-    ///
-    /// Not all sysvars support this method. If not, it returns
-    /// [`ProgramError::UnsupportedSysvar`].
+    /// Calls [`Sysvar::get`].
```
febo: More out of curiosity, is this `get` still needed on the `SysvarSerialize` trait? Since the implementation will have to provide the `get` due to the `Sysvar` bound, seems like it is redundant and potentially error prone (?) – e.g., technically it can be overridden and implemented differently than `Sysvar::get`.

sysvar/src/lib.rs
febo: Yeah, we also removed the `Default` bound in pinocchio, so we could considerer doing that as well in the future.

febo (commented):
Looks good – just have one more comment.

sysvar/src/lib.rs
kevinheavey: Not sure, maybe @joncinque knows

joncinque: It doesn't seem needed, but I assumed it was included to make upgrading to v3 easier. Users can change their import from `Sysvar` to `SysvarSerialize` and recover exactly what they had before.

On the other hand, it isn't too much worse to tell people "use `SysvarSerialize` with `Sysvar`" instead of "use `SysvarSerialize` instead of `Sysvar`", so I'm good with removing, unless there's something I'm missing.

febo: Yeah, I can't see the need for it. I would expect that in most cases it will still be "use `SysvarSerialize` instead of `Sysvar`" to load a sysvar from an account info as it might be unlikely to be using both at the same time (`from_account_info` and `get`). For users using only `get`, there is no change.

joncinque (approved):
Thanks for all the work!

PR #88, opened 2025-03-26
solana-ed25519-program: make offset fields public

jstarry:
Users of `solana-ed25519-program` need to manually create the offsets to support verifying multiple signatures so the fields need to be public.


jstarry:
The offset fields for the other precompile programs are already all public btw

joncinque (approved):
Thanks for adding these! I can't believe we got this far without them :thinking:

PR #89, opened 2025-03-26
solana-secp256r1-program: make consts for ix generation public

jstarry:
- I'm looking into pulling the verify functions into agave and would like to depend on consts from the sdk if possible
- These two consts are useful for ix generation so any users who want to roll their own ixs should be able to import them.


joncinque (approved):
Looks good to me

PR #91, opened 2025-03-26
CI: Add rebase step during publish

joncinque:
#### Problem

The publish job failed recently because a commit landed *after* the job was started at
https://github.com/anza-xyz/solana-sdk/actions/runs/14089935529/job/39463977139

This happened because the job has an out-of-date view of the repo.

#### Summary of changes

Similar to how this was done in SPL, rebase before publish.

### DIFF
--- a/.github/workflows/publish-rust.yml
+++ b/.github/workflows/publish-rust.yml
@@ -130,6 +130,9 @@ jobs:
           git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
           git config --global user.name "github-actions[bot]"
 
+      - name: Rebase (in case any changes landed after)
+        run: git pull --rebase origin
+
       - name: Publish Crate
         id: publish
         env:


.github/workflows/publish-rust.yml
```diff
@@ -130,6 +130,9 @@ jobs:
           git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
           git config --global user.name "github-actions[bot]"
 
+      - name: Rebase (in case any changes landed after)
```
jstarry, line 133: I think we need to do a release build first because that's what took so long

joncinque: Sorry, I don't understand what you're suggesting.

Do you mean to run the publish verification before rebasing? ie. first run `cargo publish --dry-run`, then rebase, then use `--no-verify` in the cargo release step?

Edit: my concern with that is that we're not verifying what we're uploading, which isn't great

jstarry: I just meant that the publish takes a long time due to the release build it does.. so it doesn't really help to rebase before starting to publish. So if we can run a release build first to cache the artifacts, we can then rebase and publish with a lower chance of something landing in master

joncinque: I'm not sure if that's possible unfortunately.

`cargo publish` runs a dev build against the newest possible dependencies from crates.io, and I can't find where it caches that build, if it caches it at all. If you keep running `cargo publish --dry-run` it'll redo the build each time from scratch

jstarry: Oh dang well then maybe we do a rebase right before pushing the commit?

jstarry: I guess that doesn't work either because we want the commit to be next to avoid confusing what was actually published?

joncinque: Yeah exactly, `cargo release` does bump + commit + publish + tag + push, and we should do that against the newest commit

jstarry (approved):
This won't catch everything but is definitely better than before

PR #94, opened 2025-03-27
solana-sdk: deprecate modules migrated to agave

jstarry:
The following `solana_sdk` modules should no longer be consumed by anyone and are now marked as deprecated
- `precompiles`
- `feature_set`
- `reserved_account_keys`

Both the `feature_set` and `reserved_account_keys` were attempted to be marked as deprecated in the past but adding deprecation annotations to a re-export of a crate does not work. This change puts those re-exports inside a deprecated module so that consumers actually receive the deprecation notice. The other broken deprecation annotations are left as an exercise to anyone who cares to fix it ;)

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -74,6 +74,7 @@ pub mod hash;
 pub mod log;
 pub mod native_loader;
 pub mod net;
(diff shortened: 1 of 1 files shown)


joncinque (approved):
This is great, thanks!

PR #95, opened 2025-03-27
Add back the `solana_sdk::transaction_context` module

jstarry:
By removing the `transaction_context` module we limit ourselves to only being able to release `solana-sdk` as a major version bump. I think it could be nice to relieve ourselves of that limitation by keeping the module for now and removing it later.


joncinque (commented):
This seems to be missing the actual transaction-context crate. Can you add that back in?

joncinque (approved):
Oh sorry, never mind, you're using the version from crates.io, I misunderstood. Looks great!

PR #96, opened 2025-03-27
Add null hasher for Pubkey

alexpyattaev:
Pubkey struct already contains random bytes, and is very commonly a key in a `HashMap`. Hashing pubkeys for hashmap purposes is not useful since they are already effectively random, so just picking a slice will do the trick.

Idea was proposed by @alessandrod to make some of the hashmap-intensive code a little bit faster. Hashing is not *the* bottleneck by any means, but there is truly no point doing useless work.

Implemenation notes:

- the custom Hash implementation prevents the hasher from consuming the length (which is always 32 so no point hashing it anyway)
- The custom hasher is not used by default, one needs to manually build HashMap with it to activate, so it will not break existing code
- The changes to new_unique() are needed to prevent the pubkeys that are full of zeros from creating pathalogical patterns in benchmarks. 
- new_unique uses SipHash from std to fill the Pubkey with random bytes when std is available, and simply pads with last byte of the counter in no_std contexts. This way benches give either correct (for std) or somewhat worse (for no-std) results.

### DIFF
--- a/pubkey/src/lib.rs
+++ b/pubkey/src/lib.rs
@@ -177,7 +177,10 @@ impl Hash for Pubkey {
 mod hasher {
     use {
         crate::PUBKEY_BYTES,
-        core::cell::Cell,
+        core::{
+            cell::Cell,
+            hash::{BuildHasher, Hasher},
+        },
         rand::{thread_rng, Rng},
     };
 
@@ -194,7 +197,7 @@ mod hasher {
         state: u64,
     }
 
-    impl std::hash::Hasher for PubkeyHasher {
+    impl Hasher for PubkeyHasher {
         #[inline]
         fn finish(&self) -> u64 {
             self.state
@@ -252,7 +255,7 @@ mod hasher {
         }
     }
 
-    impl std::hash::BuildHasher for PubkeyHasherBuilder {
+    impl BuildHasher for PubkeyHasherBuilder {
         type Hasher = PubkeyHasher;
         #[inline]
         fn build_hasher(&self) -> Self::Hasher {
@@ -312,7 +315,7 @@ mod hasher {
     }
 }
 #[cfg(all(feature = "rand", not(target_os = "solana")))]
-pub use hasher::PubkeyHasherBuilder;
+pub use hasher::{PubkeyHasher, PubkeyHasherBuilder};
 
 impl solana_sanitize::Sanitize for Pubkey {}
 
--- a/pubkey/src/lib.rs
+++ b/pubkey/src/lib.rs
@@ -23,7 +23,9 @@ use {
     core::{
         array,
         convert::{Infallible, TryFrom},
-        fmt, mem,
+        fmt,
+        hash::Hash,
+        mem,
         str::{from_utf8, FromStr},
     },
     num_traits::{FromPrimitive, ToPrimitive},
@@ -158,9 +160,18 @@ impl From<u64> for PubkeyError {
 #[cfg_attr(all(feature = "borsh", feature = "std"), derive(BorshSchema))]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[cfg_attr(feature = "bytemuck", derive(Pod, Zeroable))]
-#[derive(Clone, Copy, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[derive(Clone, Copy, Default, Eq, Ord, PartialEq, PartialOrd)]
 #[cfg_attr(feature = "dev-context-only-utils", derive(Arbitrary))]
 pub struct Pubkey(pub(crate) [u8; 32]);
+
+/// Custom impl of Hash for Pubkey
+/// allows us to skip hashing the length of the pubkey
+/// which is always the same anyway
+impl Hash for Pubkey {
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+        state.write(self.as_array());
+    }
+}
 
 #[cfg(all(feature = "rand", not(target_os = "solana")))]
 mod hasher {
@@ -197,10 +208,20 @@ mod hasher {
 
     impl Default for PubkeyHasherBuilder {
         fn default() -> Self {
-            let mut rng = rand::prelude::thread_rng();
-            PubkeyHasherBuilder {
-                offset: rng.gen_range(0..32 - 8),
-            }
+            std::thread_local!(static OFFSET: core::cell::Cell<usize>  = {
+                let mut rng = rand::prelude::thread_rng();
+                core::cell::Cell::new(rng.gen_range(0..32 - 8))
+            });
+
+            let offset = OFFSET.with(|offset| {
+                let mut next_offset = offset.get() + 1;
+                if next_offset > 24 {
+                    next_offset = 0;
+                }
+                offset.set(next_offset);
+                next_offset
+            });
+            PubkeyHasherBuilder { offset }
         }
     }
 
@@ -228,8 +249,8 @@ mod hasher {
             let builder = PubkeyHasherBuilder::default();
             let mut hasher1 = builder.build_hasher();
             let mut hasher2 = builder.build_hasher();
-            hasher1.write(&key.0);
-            hasher2.write(&key.0);
+            hasher1.write(key.as_array());
+            hasher2.write(key.as_array());
             assert_eq!(
                 hasher1.finish(),
                 hasher2.finish(),
@@ -240,7 +261,7 @@ mod hasher {
             let builder2 = PubkeyHasherBuilder::default();
             for _ in 0..64 {
                 let mut hasher3 = builder2.build_hasher();
-                hasher3.write(&key.0);
+                hasher3.write(key.as_array());
                 std::dbg!(hasher1.finish());
                 std::dbg!(hasher3.finish());
                 if hasher1.finish() != hasher3.finish() {
@@ -257,8 +278,8 @@ mod hasher {
             let builder = PubkeyHasherBuilder::default();
             let mut hasher1 = builder.build_hasher();
             let mut hasher2 = builder.build_hasher();
-            hasher1.write(&key1.0);
-            hasher2.write(&key2.0);
+            hasher1.write(key1.as_array());
+            hasher2.write(key2.as_array());
             assert_ne!(hasher1.finish(), hasher2.finish());
         }
     }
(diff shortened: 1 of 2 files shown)


pubkey/src/lib.rs
```diff
@@ -162,6 +162,110 @@ impl From<u64> for PubkeyError {
...
+    #[derive(Clone)]
+    pub struct PubkeyHasherBuilder {
+        offset: usize,
+    }
+
+    impl Default for PubkeyHasherBuilder {
+        fn default() -> Self {
+            let mut rng = rand::prelude::thread_rng();
+            PubkeyHasherBuilder {
+                offset: rng.gen_range(0..32 - 8),
```
alessandrod: This is probably too expensive to do for each hashmap. We should do what RandomState does, which is to generate a TLS offset once.

alexpyattaev: Great idea, done in [333a0ed](https://github.com/anza-xyz/solana-sdk/pull/96/commits/333a0ed3c1398b19e44136b130646ad52f0dbb66)

pubkey/src/lib.rs
```diff
@@ -158,10 +160,133 @@ impl From<u64> for PubkeyError {
...
+    }
+}
+
+#[cfg(all(feature = "rand", not(target_os = "solana")))]
+mod hasher {
+    use rand::Rng;
+
+    /// A special Hasher specifically for Pubkeys. Uses random 8 bytes as
+    /// input. Since pubkeys are effectively random, hashing them is pure overhead.
+    /// Use with care, this may be abusable.
```
alessandrod: The documentation contradicts itself. Hashing is pure overhead, but not hashing can be abused?

I'd say something like 

```
A faster, but less collision resistant hasher for pubkeys.

Specialized hasher that uses a random 8 bytes subslice of the pubkey as the hash value. Should not be used when collisions might be used to mount DOS attacks.
```

pubkey/src/lib.rs
```diff
@@ -158,10 +160,133 @@ impl From<u64> for PubkeyError {
...
+    }
+
+    impl std::hash::Hasher for PubkeyHasher {
+        #[inline]
+        fn finish(&self) -> u64 {
+            self.state
+        }
+        #[inline]
+        fn write(&mut self, bytes: &[u8]) {
+            let chunk: &[u8; 8] = bytes[self.offset..self.offset + 8].try_into().unwrap();
```
alessandrod: nit: add a comment that explains why unwrap will never panic

alexpyattaev:
Rough perf numbers for this:
for 2000 entries map of Pubkey->u64
```
Build map with default hasher 154.583µs
Build map with custom hasher 138.683µs
Build frozen map with custom hasher 274.451µs
Build frozen map with default hasher 223.845µs
```
10000 lookups:
```
lookup frozen map custom hasher 73.941µs
lookup frozen map default hasher 126.24µs
lookup normal map 268.1µs
lookup normal map custom hasher 35.628µs
```

pubkey/src/lib.rs
```diff
@@ -158,10 +160,133 @@ impl From<u64> for PubkeyError {
...
+            self.state = u64::from_ne_bytes(*chunk);
+        }
+    }
+
+    /// A special HasherBuilder specifically for Pubkeys. Uses random 8 byte slice as
+    /// input. Since pubkeys are effectively random, hashing them is pure overhead.
+    /// Use with care, this is not a secure hasher!
```
alessandrod: Same here

pubkey/src/lib.rs
```diff
@@ -158,10 +160,133 @@ impl From<u64> for PubkeyError {
...
+    #[derive(Clone)]
+    pub struct PubkeyHasherBuilder {
+        offset: usize,
+    }
+
+    impl Default for PubkeyHasherBuilder {
+        fn default() -> Self {
+            std::thread_local!(static OFFSET: core::cell::Cell<usize>  = {
```
alessandrod: nit: use items don't use full paths (core::cell rand::prelude::thread_rng etc)

pubkey/src/lib.rs
```diff
@@ -158,10 +160,133 @@ impl From<u64> for PubkeyError {
...
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[cfg_attr(feature = "bytemuck", derive(Pod, Zeroable))]
-#[derive(Clone, Copy, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[derive(Clone, Copy, Default, Eq, Ord, PartialEq, PartialOrd)]
 #[cfg_attr(feature = "dev-context-only-utils", derive(Arbitrary))]
 pub struct Pubkey(pub(crate) [u8; 32]);
 
+/// Custom impl of Hash for Pubkey
+/// allows us to skip hashing the length of the pubkey
+/// which is always the same anyway
```
joncinque, line 169: I was so surprised about this! But confirming: https://doc.rust-lang.org/std/hash/trait.Hash.html#impl-Hash-for-%5BT;+N%5D

pubkey/src/lib.rs
```diff
@@ -158,10 +160,133 @@ impl From<u64> for PubkeyError {
...
+                    next_offset = 0;
+                }
+                offset.set(next_offset);
+                next_offset
+            });
+            PubkeyHasherBuilder { offset }
+        }
+    }
+
+    impl std::hash::BuildHasher for PubkeyHasherBuilder {
```
joncinque: nit: I always have a hard time seeing when types became available in core, so maybe this won't be possible without a MSRV bump, but can this use `core::hash::BuildHasher` instead? It doesn't make a big difference since this is gated on `rand` anyway.

pubkey/src/lib.rs
alexpyattaev: addressed in https://github.com/anza-xyz/solana-sdk/pull/96/commits/241d97a2bea2a2fa07efcaf23ccd555a8dc5f06c

pubkey/src/lib.rs
alexpyattaev: done in https://github.com/anza-xyz/solana-sdk/pull/96/commits/241d97a2bea2a2fa07efcaf23ccd555a8dc5f06c

pubkey/src/lib.rs
alexpyattaev: done in https://github.com/anza-xyz/solana-sdk/pull/96/commits/241d97a2bea2a2fa07efcaf23ccd555a8dc5f06c

pubkey/src/lib.rs
alexpyattaev: applied in https://github.com/anza-xyz/solana-sdk/pull/96/commits/241d97a2bea2a2fa07efcaf23ccd555a8dc5f06c

pubkey/src/lib.rs
```diff
@@ -158,10 +160,133 @@ impl From<u64> for PubkeyError {
...
+    /// A special Hasher specifically for Pubkeys. Uses random 8 bytes as
+    /// input. Since pubkeys are effectively random, hashing them is pure overhead.
+    /// Use with care, this may be abusable.
+    #[derive(Default)]
+    pub struct PubkeyHasher {
+        offset: usize,
+        state: u64,
+    }
+
+    impl std::hash::Hasher for PubkeyHasher {
```
joncinque: Same as the other one, can this use `core::hash::Hasher` instead? Unless that entails a MSRV bump

pubkey/src/lib.rs
```diff
@@ -158,10 +160,133 @@ impl From<u64> for PubkeyError {
...
+            hasher2.write(key.as_array());
+            assert_eq!(
+                hasher1.finish(),
+                hasher2.finish(),
+                "Hashers made with same builder should be identical"
+            );
+            // Make sure that when we make new builders we get different slices
+            // chosen for hashing
+            let builder2 = PubkeyHasherBuilder::default();
+            for _ in 0..64 {
```
joncinque, line 292: Sorry, I don't understand why this loop is needed. The builder is created outside of the loop, so every hasher from `build_hasher` will use the same offset, and hash the same key.

Was the intention to make the builder inside the loop to use a different offset and check 23 iterations before we get to the original offset from `builder`?

pubkey/src/lib.rs
```diff
@@ -158,10 +160,133 @@ impl From<u64> for PubkeyError {
...
+            let mut hasher1 = builder.build_hasher();
+            let mut hasher2 = builder.build_hasher();
+            hasher1.write(key1.as_array());
+            hasher2.write(key2.as_array());
+            assert_ne!(hasher1.finish(), hasher2.finish());
+        }
+    }
+}
+#[cfg(all(feature = "rand", not(target_os = "solana")))]
+pub use hasher::PubkeyHasherBuilder;
```
joncinque: Sorry if this is a silly question, but shouldn't this also need to make `PubkeyHasher` available?

joncinque (commented):
Looks great overall! Just a few questions

alessandrod (dismissed):
approving but please wait for one of the maintainers of this code to approve as well

pubkey/src/lib.rs
alexpyattaev: I do not want the test to rely on the builder using stateful logic to pick an offset. Original impl was just using random number every time instead of thread_locals. This test should pass either way.

pubkey/src/lib.rs
alexpyattaev: Actually we do not have to, but there is no harm in making it visible.

pubkey/src/lib.rs
alexpyattaev: done in https://github.com/anza-xyz/solana-sdk/pull/96/commits/32db3275692a9ea63e5e3020902ff94a0f701635

pubkey/src/lib.rs
alexpyattaev: done in https://github.com/anza-xyz/solana-sdk/pull/96/commits/32db3275692a9ea63e5e3020902ff94a0f701635

pubkey/src/lib.rs
alexpyattaev: Yes they do it so two slices of zeros with different lengths do not hash to the same value.

joncinque: Ah very interesting, thanks for the background

pubkey/src/lib.rs
joncinque: Ah gotcha, thanks for the explanation!

pubkey/src/lib.rs
joncinque: Weird, how does that work then? I thought creating the hasher would give a "pub function returns a private type" error

joncinque (approved):
Looks great!

kevinheavey:
Does this require a major version bump?

alexpyattaev:
this should be mostly backwards compatible, unless aomeone was relying on the contents of pubkeys made by new_unique

PR #97, opened 2025-03-27
pin bytemuck_derive <1.9.0

jacobcreech:
Temporary update to pin bytemuck_derive <1.9.0 until we have platform tools for 2.2 using rust >=rustc 1.84


kevinheavey:
This would be a breaking change for anyone using these crates on a non-SBPF target, since in those situations these crates are currently working fine.

Think we need to do the cfg(target_os = "solana") thing

jacobcreech:
We chatted again with the bytemuck_derive dev and they actually [fixed this in bytemuck_derive 1.93](https://github.com/Lokathor/bytemuck/issues/306#issuecomment-2759480874). Closing for now.

PR #100, opened 2025-03-28
message: don't use nonces from alt's

jstarry:
Before clusters adop https://github.com/solana-foundation/solana-improvement-documents/pull/242, we should ensure that no transactions are inadvertently loading nonces from ALT's

### DIFF
--- a/message/src/compiled_keys.rs
+++ b/message/src/compiled_keys.rs
@@ -51,13 +51,9 @@ struct CompiledKeyMeta {
 }
 
 // inlined to avoid solana_nonce dep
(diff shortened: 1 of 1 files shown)


apfitzge:
Does this make our testing more difficult, since now we cannot construct via the normal methods?

jstarry:
> 
Does this make our testing more difficult, since now we cannot construct via the normal methods?

Nope the tests in https://github.com/anza-xyz/agave/pull/5555 construct the lookups manually, they don't use `CompiledKeys`

joncinque (approved):
Looks great!

PR #102, opened 2025-03-28
program-error: Deprecate PrintProgramError

joncinque:
#### Problem

As part of the breaking change to unite pinocchio's program error with the sdk's program error in #12, we are introducing a new `ToStr` trait that doesn't rely on any particular logging.

To go with that, we want to deprecate `PrintProgramError`.

#### Summary of changes

Before landing the breaking change, let's deprecate `PrintProgramError` properly and add `ToStr` as an alternative for downstream users.

Once this lands, we'll do the patch release with the deprecation, then we can finally land #12.

cc @kevinheavey

### DIFF
--- a/program-error/src/lib.rs
+++ b/program-error/src/lib.rs
@@ -122,12 +122,17 @@ impl fmt::Display for ProgramError {
     }
 }
 
+#[deprecated(
+    since = "2.2.2",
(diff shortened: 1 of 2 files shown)


febo (approved):
Looks great!

PR #103, opened 2025-03-28
deps: Bump down intra-repo dependencies

joncinque:
#### Problem

The sdk repo respects semver, but still bumps intra-repo versions any time one of them gets published. This requires downstream users to use all of the newest versions at once, which may not be desirable.

#### Summary of changes

Bump down all of the intra-repo dependency versions to v2.2.1, which is the first version that relaxed all dependencies too.

NOTE: solana-address-lookup-table-interface was a special case, and needs to stay on v2.2.2 to benefit from relaxed dependencies.

At the same time, use the `--dependent-version fix` flag in cargo release, which avoids bumping the intra-repo dependency during a publish. Thankfully, cargo-release does the right thing and *will* bump the intra-repo dependency if it's a major release.


jstarry (approved):

febo (approved):
Nice!

PR #104, opened 2025-03-29
remove solana-decode-error

kevinheavey:
The only part of DecodeError that is not a trivial wrapper over `num_traits::FromPrimitive` is the `type_of` method. AFAICT this method is not used anywhere. We should delete the decode-error dir and stop publishing the crate. For now we can just deprecate the trait and avoid using it anywhere in this repo.

This is a breaking change for solana-precompile-error, solana-program-error, solana-pubkey and solana-vote-interface


febo (dismissed):
Looks great!

kevinheavey:
@febo can you merge this? I don't have permissions

febo:
> @febo can you merge this? I don't have permissions

I can't either. 😅 It might need a rebase so all checks pass.

joncinque:
I'm still going through and marking everything deprecated that we want to remove in the next breaking release. After that, I'll publish new patch releases with those deprecation warnings. And after that, we can start merging breaking changes like this one. I plan to have it ready by the end of this week / early next week.

therealdannzor:
This PR would probably resolve https://github.com/solana-program/libraries/issues/73

Discovered this while working on a MRE to demo the problem

Merging this would help when upgrading to Solana `2.x`

cc @joncinque

joncinque (commented):
Can you rebase this? Since we're landing breaking changes now, and we previously deprecated the trait, you can completely remove `solana-decode-error` and the reference in `scripts/patch-crates-functions.sh` while you're at it

precompile-error/Cargo.toml
```diff
@@ -9,9 +9,8 @@ homepage = { workspace = true }
...
 
-[package.metadata.docs.rs]
-targets = ["x86_64-unknown-linux-gnu"]
-
 [dependencies]
 num-traits = { workspace = true }
-solana-decode-error = { workspace = true }
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
```
joncinque: Looks like CI is still complaining about this

joncinque (commented):
One last bit, then we can land this! Sorry for the holdup

jstarry:
This breaks `spl-pod` in addition to the other crates mentioned in this PR, is someone looking into fixing up these: https://github.com/search?q=repo%3Asolana-program%2Flibraries%20DecodeError&type=code

jstarry:
Oh nvm, I see https://github.com/solana-program/libraries/pull/79 now, we just need to release new versions of those spl crates

joncinque:
This change is unreleased, so most likely there's some other issue. What are you seeing?

jstarry:
I was having problems locally linking the sdk crates to agave due to unreleased breaking changes, but should have been using the maintenance branch instead. Sorry for the noise!

PR #105, opened 2025-03-30
msg!: make std optional

kevinheavey:
In a follow-up PR I'd like to move `sol_log_data` here as I noticed it gets used in a few places

Note: this is a breaking change

### DIFF
--- a/msg/Cargo.toml
+++ b/msg/Cargo.toml
@@ -9,12 +9,12 @@ homepage = { workspace = true }
 license = { workspace = true }
 edition = { workspace = true }
 
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
 [features]
 default = ["std"]
 std = []
-
-[package.metadata.docs.rs]
-targets = ["x86_64-unknown-linux-gnu"]
 
 [target.'cfg(target_os = "solana")'.dependencies]
 solana-define-syscall = { workspace = true }
--- a/msg/Cargo.toml
+++ b/msg/Cargo.toml
@@ -9,6 +9,10 @@ homepage = { workspace = true }
 license = { workspace = true }
 edition = { workspace = true }
 
+[features]
+default = ["std"]
+std = []
+
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
 
--- a/msg/src/lib.rs
+++ b/msg/src/lib.rs
@@ -42,12 +42,15 @@ pub mod syscalls;
 
 /// Print a string to the log.
 #[inline]
-pub fn sol_log(_message: &str) {
+pub fn sol_log(message: &str) {
     #[cfg(target_os = "solana")]
     unsafe {
-        syscalls::sol_log_(_message.as_ptr(), _message.len() as u64);
+        syscalls::sol_log_(message.as_ptr(), message.len() as u64);
     }
 
     #[cfg(all(not(target_os = "solana"), feature = "std"))]
-    std::println!("{_message}");
+    std::println!("{message}");
+
+    #[cfg(all(not(target_os = "solana"), not(feature = "std")))]
+    core::hint::black_box(message);
 }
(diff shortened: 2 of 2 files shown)


msg/src/lib.rs
```diff
@@ -1,3 +1,6 @@
+#![no_std]
+#[cfg(feature = "std")]
+extern crate std;
```
febo, line 3: Wonder if it would be worth to include a `no_std` variation for `msg!` – mostly just cosmetic, but it would allow to use a similar pattern to log static strings.
```rust
/// Print a message to the log.
///
/// Supports simple strings of type `&str` in `no_std` environments. The expression
/// will be passed directly to [`sol_log`].
///
/// This is typically used for logging static strings.
///
/// # Examples
///
/// ```
/// use solana_msg::msg;
///
/// msg!("verifying multisig");
/// ```
#[macro_export]
#[cfg(not(feature = "std"))]
macro_rules! msg {
    ( $msg:expr ) => {
        $crate::sol_log($msg)
    };
}
```

msg/Cargo.toml
```diff
@@ -9,6 +9,10 @@ homepage = { workspace = true }
 license = { workspace = true }
 edition = { workspace = true }
 
+[features]
+default = ["std"]
```
febo, line 16: This is a general query: few other crates (e.g., `solana-program-error`) have a `std` feature, but it is not enabled by default. Do you think it would be useful to have a common pattern, like `std` is always enabled by default or the other way around?

kevinheavey: Here since the crate is already published I want to break less stuff. This PR is still a breaking change because some people could already be depending on it with `default-features = false`, but it's at least a smaller change

febo: But since it is already "breaking", should we update to how ideally we want to go forward with the `std` feature? I wonder if it will become a bit confusing to have some crates with `std` as default and others without. Might be clearer to have a common pattern.

msg/src/lib.rs
kevinheavey: this would give confusing errors to people who don't realise they need the std feature.

If msg doesn't exist the error message is helpful:
```
error[E0432]: unresolved import `solana_msg::msg`
  --> src/lib.rs:1:5
   |
1  | use solana_msg::msg;
   |     ^^^^^^^^^^^^^^^ no `msg` in the root
   |
note: found an item that was configured out
  --> /home/user/solana-sdk/msg/src/lib.rs:33:14
   |
33 | macro_rules! msg {
   |              ^^^
note: the item is gated behind the `std` feature
  --> /home/user/solana-sdk/msg/src/lib.rs:31:7
   |
31 | #[cfg(feature = "std")]
```

If msg exists but doesn't support format args the error gives no hints about feature gating:
```
error: no rules expected `,`
  --> src/lib.rs:3:26
   |
3  |     msg!("Adding {} + {}", left, right);
   |                          ^ no rules expected this token in macro call
   |
note: while trying to match meta-variable `$msg:expr`
  --> /home/user/solana-sdk/msg/src/lib.rs:43:6
   |
43 |     ($msg:expr) => {
```

So I think in the no-std case people should just use sol_log directly instead of a pass-through macro wrapper

febo: Good point!

msg/Cargo.toml
kevinheavey: I looked through the repo crates and I think right now the pattern is "enable std by default if it's required for the core functionality of the crate"

febo: Sounds good, so in this case the core functionality is the `msg!` macro.

febo (dismissed):
Looks good!

joncinque:
Please rebase and then we can land this, thanks for your patience!

joncinque:
There was a breaking change to `cargo-sort`, so you'll need to install the newest version and then run `./cargo nightly sort --workspace`

joncinque (approved):
Thanks!

PR #107, opened 2025-03-31
keypair / signature / ed25519: Deprecate dalek types from public API

joncinque:
#### Problem

The keypair, signature, and ed25519-program crates all expose `ed25519_dalek` types in their public APIs, either as inputs or outputs.

This makes it difficult to land PRs like #26, since we'll just end up causing breaking changes.

#### Summary of changes

Rather than constantly break APIs whenever the ed25519_dalek crates publishes a new major version, deprecate all functions that use ed25519_dalek types and add non-dalek versions.

* 128d5e45043ac25517b52d968d3a4b9ed42216a0: ed25519-program changes
* 87f76f5b53aab128e65a0637341ab4e52d5fa7e3: new Signature error type
* c5be15e499c8841315852b3d2802b6b8f8da53ed: keypair changes

As part of this work, I copied over the Signature error from RustCrypto to use as a generic error. That way, we can choose when to make breaking changes, independent of dependency bumps. I wasn't able to do the same thing since we're not ready to start using `core::error::Error` everywhere.

cc @kevinheavey

### DIFF
--- a/ed25519-program/src/lib.rs
+++ b/ed25519-program/src/lib.rs
@@ -37,7 +37,7 @@ pub struct Ed25519SignatureOffsets {
 /// buffer.
 ///
 /// Note: If the signer for these messages are the same, it is cheaper to concatenate the messages
-/// and have the signer sign the single buffer and use [`new_ed25519_instruction`].
+/// and have the signer sign the single buffer and use [`new_ed25519_instruction_with_signature`].
 pub fn offsets_to_ed25519_instruction(offsets: &[Ed25519SignatureOffsets]) -> Instruction {
     let mut instruction_data = Vec::with_capacity(
         SIGNATURE_OFFSETS_START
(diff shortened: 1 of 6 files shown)


signature/Cargo.toml
```diff
@@ -36,15 +36,16 @@ solana-short-vec = { workspace = true }
 solana-signature = { path = ".", features = ["serde"] }
 
 [features]
-default = ["std"]
+default = ["std", "alloc"]
```
febo, line 39: Since "std" implies "alloc", is it needed to specify both here?

keypair/Cargo.toml
```diff
@@ -18,7 +18,7 @@ solana-derivation-path = { workspace = true, optional = true }
 solana-pubkey = { workspace = true }
 solana-seed-derivable = { workspace = true, optional = true }
 solana-seed-phrase = { workspace = true }
-solana-signature = { workspace = true, features = ["verify"] }
+solana-signature = { workspace = true, features = ["std", "verify"] }
```
febo, line 21: The "std" feature is enabled by default, just checking whether that was expected or not since it is being explicitly enabled here.

ed25519-program/src/lib.rs
```diff
@@ -58,13 +58,18 @@ pub fn offsets_to_ed25519_instruction(offsets: &[Ed25519SignatureOffsets]) -> In
...
+    new_ed25519_instruction_with_signature(message, &signature, &pubkey)
+}
 
-    assert_eq!(pubkey.len(), PUBKEY_SERIALIZED_SIZE);
-    assert_eq!(signature.len(), SIGNATURE_SERIALIZED_SIZE);
-
+pub fn new_ed25519_instruction_with_signature(
+    message: &[u8],
+    signature: &[u8; SIGNATURE_SERIALIZED_SIZE],
+    pubkey: &[u8; PUBKEY_SERIALIZED_SIZE],
```
febo, line 71: These changes make a lot of sense.

febo (commented):
Looks great! I left a couple of tiny queries.

signature/Cargo.toml
joncinque: I was thinking to eventually remove the `std` feature in favor of `alloc`, and by having it already set as a default it'll be simpler (and maybe less breaking?) to make that change in the future.

keypair/Cargo.toml
joncinque: We have `default-features` set to `false` at the workspace level, so we do need to enable ~it~ the std feature somewhere

ed25519-program/src/lib.rs
joncinque: ok great! I tried to follow other verification method signatures like https://docs.rs/ed25519-dalek/latest/ed25519_dalek/fn.verify_batch.html, which go message, then signature, then verifying key

signature/Cargo.toml
febo: Got it, makes sense.

PR #108, opened 2025-03-31
pubkey: Don't always mark variable as mut

joncinque:
#### Problem

If the `std` feature is disabled on `solana-pubkey`, there's currently a warning in the build due to a variable that doesn't need to be marked as `mut`.

#### Summary of changes

Gate marking the variable as `mut` depending on the feature / target. Also, we already pull in the std crate earlier if the `std` feature is enabled, so we don't need to do it twice.

### DIFF
--- a/pubkey/src/lib.rs
+++ b/pubkey/src/lib.rs
@@ -480,7 +480,10 @@ impl Pubkey {
         type T = u32;
         const COUNTER_BYTES: usize = size_of::<T>();
         let mut b = [0u8; PUBKEY_BYTES];
+        #[cfg(any(feature = "std", target_arch = "wasm32"))]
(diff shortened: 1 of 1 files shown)


alexpyattaev:
Do we need a no-std clippy script of some sort for sdk to catch these?

joncinque:
> Do we need a no-std clippy script of some sort for sdk to catch these?

We have a few options. We could update the `cargo hack check` to instead run `cargo hack clippy -- --deny=warnings`, unless there's a way for `cargo check` to return an error on warnings.

Slightly separately, we could also start testing the feature powerset with cargo hack: https://github.com/taiki-e/cargo-hack?tab=readme-ov-file#--feature-powerset

I'll test around with these and enable what seems best.

PR #109, opened 2025-03-31
CI: Add step to check minimal-versions

joncinque:
#### Problem

All of the dependencies are now relaxed and even bumped down to their lowest possible number, but we don't check to make sure that all of the packages can even build with the lowest declared dependencies.

#### Summary of changes

Use the [`cargo-minimal-versions`](https://github.com/taiki-e/cargo-minimal-versions) tool to do exactly that. It goes through each package in the workspace and tries to check it against its lowest declared dependencies.

The tool discovered a lot of instances where we declared a dependency that was too low, so I bumped those up.

And finally, add the check to CI.

IMPORTANT NOTE: This currently doesn't work for intra-repo dependencies. For example if `solana-message` adds a new function, and `solana-transaction` starts using it, then this check will still pass, since the tool doesn't temporarily detach the crate from the workspace when running the check.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -226,6 +226,29 @@ jobs:
       - name: Run hack check
         run: ./scripts/check-hack.sh
 
+  minimal-versions:
+    name: Check minimal-versions
+    runs-on: ubuntu-latest
+    needs: [sanity]
+    steps:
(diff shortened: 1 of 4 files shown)


yihau (approved):
🫡

PR #112, opened 2025-04-01
CI: Add powerset clippy check

joncinque:
#### Problem

As pointed out in #108, we don't explicitly check for all combinations of features in the sdk, and more specifically, we miss warnings.

#### Summary of changes

Add a CI step to run `cargo clippy -- --deny=warnings` on the powerset of features in the sdk. Currently, this means 888 different crate builds, but most of them are quite short.

Let's see how long it takes to run in CI to see if it's worth adding.

Also, fix some little issues that came up while running this.

### DIFF
--- a/message/Cargo.toml
+++ b/message/Cargo.toml
@@ -61,6 +61,7 @@ frozen-abi = [
     "dep:solana-logger",
     "solana-hash/frozen-abi",
     "solana-pubkey/frozen-abi",
+    "serde",
 ]
 serde = [
     "dep:serde",
--- a/signature/src/error.rs
+++ b/signature/src/error.rs
@@ -1,7 +1,7 @@
 //! Signature error copied directly from RustCrypto's opaque signature error at
 //! https://github.com/RustCrypto/traits/tree/master/signature
 
-#[cfg(feature = "alloc")]
+#[cfg(all(feature = "std", feature = "alloc"))]
 use alloc::boxed::Box;
 use core::fmt::{self, Debug, Display};
 
@@ -22,7 +22,7 @@ use core::fmt::{self, Debug, Display};
 #[non_exhaustive]
 pub struct Error {
     /// Source of the error (if applicable).
-    #[cfg(feature = "std")]
+    #[cfg(all(feature = "std", feature = "alloc"))]
     source: Option<Box<dyn std::error::Error + Send + Sync + 'static>>,
 }
 
@@ -33,7 +33,7 @@ impl Error {
     /// errors e.g. signature parsing or verification errors. The intended use
     /// cases are for propagating errors related to external signers, e.g.
     /// communication/authentication errors with HSMs, KMS, etc.
-    #[cfg(feature = "std")]
+    #[cfg(all(feature = "std", feature = "alloc"))]
     pub fn from_source(
         source: impl Into<Box<dyn std::error::Error + Send + Sync + 'static>>,
     ) -> Self {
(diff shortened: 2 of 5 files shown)


joncinque:
The check took 8 minutes, which is a little long, but I think it's worth having.

jstarry:
Yeah seems worth it to me too. Your fixes aren't very obvious, can you annotate them with why they were made?

message/Cargo.toml
```diff
@@ -61,6 +61,7 @@ frozen-abi = [
     "dep:solana-logger",
     "solana-hash/frozen-abi",
     "solana-pubkey/frozen-abi",
+    "serde",
```
joncinque, line 64: The frozen-abi traits require that the types implement `serde::Serialize`, ie https://github.com/anza-xyz/solana-sdk/blob/5cd7a9c9e7acdd02a6fccc3c7970f796d5d6577d/frozen-abi/src/abi_example.rs#L26

Other crates do this, ie https://github.com/anza-xyz/solana-sdk/blob/5cd7a9c9e7acdd02a6fccc3c7970f796d5d6577d/vote-interface/Cargo.toml#L63, and this one was missed

sysvar/src/slot_hashes.rs
```diff
@@ -52,9 +52,12 @@ use bytemuck_derive::{Pod, Zeroable};
 use {crate::Sysvar, solana_account_info::AccountInfo};
 use {solana_clock::Slot, solana_hash::Hash};
 
-#[cfg(all(feature = "bincode", feature = "bytemuck"))]
+#[cfg(feature = "bytemuck")]
 const U64_SIZE: usize = std::mem::size_of::<u64>();
```
joncinque, line 56: `U64_SIZE` is only used in `PodSlotHashes::fetch`, which is only gated on the `bytemuck` feature, so this fixes that gating to work properly

sysvar/src/slot_hashes.rs
```diff
@@ -52,9 +52,12 @@ use bytemuck_derive::{Pod, Zeroable};
...
 
+#[cfg(any(feature = "bytemuck", feature = "bincode"))]
+const SYSVAR_LEN: usize = 20_488; // golden, update if MAX_ENTRIES changes
```
joncinque, line 59: `SlotHashes::size_of()` is required for the `Sysvar` trait, but that trait requites the `bincode` feature, and `PodSlotHashes::fetch()` only requires the `bytemuck` feature, so I factored this out into a separate const, which is reused in both `SlotHashes::size_of()` and `PodSlotHashes::fetch`

sysvar/src/slot_hashes.rs
```diff
@@ -179,7 +182,6 @@ impl PodSlotHashes {
 #[deprecated(since = "2.1.0", note = "Please use `PodSlotHashes` instead")]
 pub struct SlotHashesSysvar;
 
-#[cfg(feature = "bincode")]
```
joncinque, line 182: This used to require bincode due to the `SlotHashes::size_of()` usage in `get_pod_slot_hashes()`, but with the refactoring, this feature-gating isn't required anymore

signature/src/error.rs
joncinque, line 1: We essentially want to drop the `std` part for this once we move towards `core::error::Error` instead of `std::error::Error`, so this fixes the `alloc` feature to essentially be both the `std` and `alloc` features, so that in a future breaking change, we can just drop the requirement on the `std` feature.

joncinque:
Sorry, there was another error due to #107 that landed after the initial PR was put in, so I rebased and fixed errors arising from that PR.

jstarry (approved):

apfitzge (approved):
This is a fantastic check to have for an sdk

PR #113, opened 2025-04-01
CI: Remove nightly check, add more to hack check

joncinque:
#### Problem

We've been adding a lot of new CI steps to the SDK, but some of them might be redundant, causing us to use more CI resources than necessary.

For example, we run `./cargo nightly check` and `./cargo nightly hack check`. Are both of those really necessary?

#### Summary of changes

Remove the nightly check CI job and script, and add some of those features and flags to the hack check.

NOTE: I'm not 100% sure about this, so let me know what you think!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -185,7 +185,7 @@ jobs:
       - name: Run semver checks
         run: ./scripts/check-semver.sh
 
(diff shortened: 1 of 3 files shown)


yihau (approved):
yeah, I think it's a good merge!

PR #114, opened 2025-04-01
secp256k1: Deprecate libsecp256k1 from public interface

joncinque:
#### Problem

The libsecp256k1 crate used by the secp256k1-program is unmaintained, but its types appear in the public API of the crate, so we can't upgrade or change the library without breaking users.

#### Summary of changes

Abstract away the usage of libsecp256k1 by introducing new generic functions that simply take in byte arrays.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3708,6 +3708,7 @@ dependencies = [
  "solana-sdk",
  "solana-sdk-ids",
  "solana-secp256k1-program",
+ "solana-signature",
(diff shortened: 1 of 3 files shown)


samkim-crypto (approved):
Sorry for the delay on this one. The API changes make a lot of sense to me. I double checked the pubkey serialization/indexing. I went through the doc changes and did not find anything. Everything looks fine to me!

PR #115, opened 2025-04-02
Bump stable rust version to 1.85.1

joncinque:
#### Problem

Rust v1.85.1 is now available, but the sdk repo is still on 1.84.1

#### Summary of changes

Bump the version declared in rust-toolchain.toml


yihau (approved):
🫡

PR #116, opened 2025-04-02
Replace `std::error::Error` with `core::error::Error`

### LINKED ISSUE #81: MSRV: Use core::error::Error instead of std::error::Error
#### Problem

A lot of places in the SDK use the `std` feature to enable an implementation of `std::error::Error`, but this isn't needed because the `Error` trait was moved down to `core.

As mentioned at https://github.com/anza-xyz/solana-sdk/pull/12#discussion_r2002912485 though, using `core::error::Error` instead of `std::error::Error` requires an MSRV bump to 1.81.

#### Proposed Solution

Whenever we want to make an MSRV bump, let's remove the `std::error::Error` usage everywhere in favor of `core::error::Error`. Maybe we can include that whenever we bump the edition to 2024.

divyaranjan1905:
This PR simply replaces instances of `std::error::Error` with `core::error::Error`. It fixes #81, but as mentioned in the issue, it should probably be merged after #115.

Regards,

Divya.

### DIFF
--- a/bls-signatures/src/keypair.rs
+++ b/bls-signatures/src/keypair.rs
@@ -120,7 +120,7 @@ impl Keypair {
     pub fn write_json_file<F: AsRef<Path>>(
         &self,
         outfile: F,
-    ) -> Result<String, Box<dyn std::error::Error>> {
+    ) -> Result<String, Box<dyn core::error::Error>> {
         let outfile = outfile.as_ref();
 
         if let Some(outdir) = outfile.parent() {
--- a/hash/src/lib.rs
+++ b/hash/src/lib.rs
@@ -91,8 +91,7 @@ pub enum ParseHashError {
     Invalid,
 }
 
-#[cfg(feature = "std")]
-impl std::error::Error for ParseHashError {}
+impl core::error::Error for ParseHashError {}
 
 impl fmt::Display for ParseHashError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
--- a/pubkey/src/lib.rs
+++ b/pubkey/src/lib.rs
@@ -103,8 +103,7 @@ impl FromPrimitive for PubkeyError {
     }
 }
 
-#[cfg(feature = "std")]
-impl std::error::Error for PubkeyError {}
+impl core::error::Error for PubkeyError {}
 
 impl fmt::Display for PubkeyError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
@@ -368,8 +367,7 @@ impl FromPrimitive for ParsePubkeyError {
     }
 }
 
-#[cfg(feature = "std")]
-impl std::error::Error for ParsePubkeyError {}
+impl core::error::Error for ParsePubkeyError {}
 
 impl fmt::Display for ParsePubkeyError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
--- a/quic-definitions/src/lib.rs
+++ b/quic-definitions/src/lib.rs
@@ -56,5 +56,5 @@ pub const QUIC_MIN_STAKED_RECEIVE_WINDOW_RATIO: u64 = 128;
 pub const QUIC_MAX_STAKED_RECEIVE_WINDOW_RATIO: u64 = 512;
 
 pub trait NotifyKeyUpdate {
-    fn update_key(&self, key: &Keypair) -> Result<(), Box<dyn std::error::Error>>;
+    fn update_key(&self, key: &Keypair) -> Result<(), Box<dyn core::error::Error>>;
 }
--- a/signature/src/error.rs
+++ b/signature/src/error.rs
@@ -1,7 +1,7 @@
 //! Signature error copied directly from
 //! [RustCrypto's opaque signature error](https://github.com/RustCrypto/traits/tree/master/signature)
 
-#[cfg(all(feature = "std", feature = "alloc"))]
+#[cfg(feature = "alloc")]
 use alloc::boxed::Box;
 use core::fmt::{self, Debug, Display};
 
@@ -12,10 +12,10 @@ use core::fmt::{self, Debug, Display};
 /// (e.g. [BB'06]).
 ///
 /// When the `std` feature is enabled, it impls
-/// [`std::error::Error`](https://doc.rust-lang.org/std/error/trait.Error.html).
+/// [`core::error::Error`](https://doc.rust-lang.org/core/error/trait.Error.html).
 ///
 /// When the `alloc` feature is enabled, it supports an optional
-/// [`std::error::Error::source`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source),
+/// [`core::error::Error::source`](https://doc.rust-lang.org/core/error/trait.Error.html#method.source),
 /// which can be used by things like remote signers (e.g. HSM, KMS) to report
 /// I/O or auth errors.
 ///
@@ -24,8 +24,8 @@ use core::fmt::{self, Debug, Display};
 #[non_exhaustive]
 pub struct Error {
     /// Source of the error (if applicable).
-    #[cfg(all(feature = "std", feature = "alloc"))]
-    source: Option<Box<dyn std::error::Error + Send + Sync + 'static>>,
+    #[cfg(feature = "alloc")]
+    source: Option<Box<dyn core::error::Error + Send + Sync + 'static>>,
 }
 
 impl Error {
@@ -35,9 +35,9 @@ impl Error {
     /// errors e.g. signature parsing or verification errors. The intended use
     /// cases are for propagating errors related to external signers, e.g.
     /// communication/authentication errors with HSMs, KMS, etc.
-    #[cfg(all(feature = "std", feature = "alloc"))]
+    #[cfg(feature = "alloc")]
     pub fn from_source(
-        source: impl Into<Box<dyn std::error::Error + Send + Sync + 'static>>,
+        source: impl Into<Box<dyn core::error::Error + Send + Sync + 'static>>,
     ) -> Self {
         Self {
             source: Some(source.into()),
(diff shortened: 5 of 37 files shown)


hash/src/lib.rs
```diff
@@ -95,7 +95,7 @@ pub enum ParseHashError {
 }
 
 #[cfg(feature = "std")]
```
sonicfromnewyoke: i guess it can be dropped

pubkey/src/lib.rs
```diff
@@ -361,7 +361,7 @@ impl FromPrimitive for ParsePubkeyError {
 }
 
 #[cfg(feature = "std")]
```
sonicfromnewyoke: the same here

pubkey/src/lib.rs
```diff
@@ -103,7 +103,7 @@ impl FromPrimitive for PubkeyError {
...
```
sonicfromnewyoke: and here

instruction/src/error.rs
```diff
@@ -442,7 +442,7 @@ pub enum LamportsError {
 }
 
 #[cfg(feature = "std")]
```
sonicfromnewyoke: and this one also

instruction/src/error.rs
```diff
@@ -243,7 +243,7 @@ pub enum InstructionError {
...
```
sonicfromnewyoke: one more place

instruction/src/error.rs
```diff
@@ -243,7 +243,7 @@ pub enum InstructionError {
...
-impl std::error::Error for InstructionError {}
+impl core::error::Error for InstructionError {}
 
 #[cfg(feature = "std")]
```
sonicfromnewyoke: this is also can be omitted, because this is a `core::fmt::Dispalay` trait

divyaranjan1905: Okay, I originally considered dropping these attributes, but I wasn't sure. Thank you for confirming.

divyaranjan1905:
Rebased the last few commits and removed the feature attributes for std.

joncinque:
There have been quite a few changes since the PR was first opened, but we should be ready to integrate it -- can you rebase and update your branch? Let me know if you don't have time, and I'll be happy to take this over

divyaranjan1905:
> There have been quite a few changes since the PR was first opened, but we should be ready to integrate it -- can you rebase and update your branch? Let me know if you don't have time, and I'll be happy to take this over

NO worries, will rebase and push by today.

kevinheavey:
Could you please add `rust-version = "1.81.0"` to the Cargo.toml for every crate using `core::error`? 🙏

Otherwise the error messages for people on old Rust versions will be confusing

.github/workflows/main.yml
```diff
@@ -83,24 +83,6 @@ jobs:
       - name: Check formatting
         run: ./scripts/check-fmt.sh
 
-  check:
```
apfitzge: why is this PR changing CI?

joncinque:
@divyaranjan1905 do you still have time for this? If not I can take over the PR

divyaranjan1905:
> @divyaranjan1905 do you still have time for this? If not I can take over the PR

I tried to rebase, but it didn't resolve nicely. Anything should I be specifically paying attention to? I'll try once again.

joncinque:
There have been a lot of changes since the PR was created, so it would probably be easier to redo it from scratch. It looks like a few merge commits snuck in too, which probably isn't helping

divyaranjan1905:
@joncinque Would this be required anymore? I was told to remove the `std` attribute from other places, but I am not sure about this.

https://github.com/anza-xyz/solana-sdk/blob/ec56d76d688fcdba3fa59e9cf6cb77f2cf83cace/signature/src/error.rs#L27

bls-signatures/src/keypair.rs
```diff
@@ -120,7 +120,7 @@ impl Keypair {
     pub fn write_json_file<F: AsRef<Path>>(
         &self,
         outfile: F,
-    ) -> Result<String, Box<dyn std::error::Error>> {
+    ) -> Result<String, Box<dyn core::error::Error>> {
```
joncinque, line 123: Can you add the rust-version to `bls-signatures/Cargo.toml`?

quic-definitions/src/lib.rs
joncinque, line 1: Can you add rust-version to `quic-definitions/Cargo.toml`?

signature/src/error.rs
```diff
@@ -12,10 +12,10 @@ use core::fmt::{self, Debug, Display};
 /// (e.g. [BB'06]).
 ///
 /// When the `std` feature is enabled, it impls
-/// [`std::error::Error`](https://doc.rust-lang.org/std/error/trait.Error.html).
+/// [`core::error::Error`](https://doc.rust-lang.org/nightly/core/error/trait.Error.html).
```
joncinque: nit: No reason to reach for nightly docs here
```suggestion
/// [`core::error::Error`](https://doc.rust-lang.org/core/error/trait.Error.html).
```

signature/src/error.rs
```diff
@@ -12,10 +12,10 @@ use core::fmt::{self, Debug, Display};
...
 ///
 /// When the `alloc` feature is enabled, it supports an optional
-/// [`std::error::Error::source`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source),
+/// [`core::error::Error::source`](https://doc.rust-lang.org/nightly/core/error/trait.Error.html#method.source),
```
joncinque: nit: same here, can just use the stable core docs
```suggestion
/// [`core::error::Error::source`](https://doc.rust-lang.org/core/error/trait.Error.html#method.source),
```

signature/src/lib.rs
```diff
@@ -139,7 +139,6 @@ pub enum ParseSignatureError {
     Invalid,
 }
 
-#[cfg(feature = "std")]
```
joncinque, line 142: Be sure to also import `Error` from `core` to go with this change

signature/src/error.rs
```diff
@@ -25,7 +25,7 @@ use core::fmt::{self, Debug, Display};
 pub struct Error {
     /// Source of the error (if applicable).
     #[cfg(all(feature = "std", feature = "alloc"))]
```
joncinque: Can you remove all references to the `std` feature in this file? Now we only need the `alloc` feature

signature/src/lib.rs
```diff
@@ -13,7 +13,8 @@ extern crate alloc;
 #[cfg(feature = "std")]
 extern crate std;
 #[cfg(feature = "std")]
-use std::{error::Error, vec::Vec};
+use core::error::Error;
+use std::vec::Vec;
```
joncinque: This isn't correct -- the whole point is to only bring in `vec` if the `std` feature is enabled:
```suggestion
#[cfg(feature = "std")]
use std::vec::Vec;
use core::error::Error;
```

signature/src/error.rs
```diff
@@ -1,7 +1,7 @@
 //! Signature error copied directly from
 //! [RustCrypto's opaque signature error](https://github.com/RustCrypto/traits/tree/master/signature)
 
-#[cfg(all(feature = "std", feature = "alloc"))]
+#[cfg(all(feature = "alloc"))]
```
joncinque: nit: for all of these, we don't need `all` anymore
```suggestion
#[cfg(feature = "alloc")]
```

divyaranjan1905:
That should be it :)

Sorry for the tiny mistakes.

joncinque:
Looks like there are some CI failures:

* be sure to run `./cargo nightly fmt --all` to format the code
* check the test errors by running `./scripts/build-sbf.sh` or `./scripts/check-hack.sh`

divyaranjan1905:
Looks like it's still failing. Interesting:
https://github.com/anza-xyz/solana-sdk/actions/runs/16509156545/job/46687480671?pr=116#step:6:659

I have not touched it in my diff though.

divyaranjan1905:
Changing it to `self.source()`, I get the following complaint from the compiler:

```
error[E0599]: the method `as_ref` exists for reference `&&dyn Error`, but its trait bounds were not satisfied  
  --> signature/src/error.rs:85:34  
   |  
85 |             .map(|source| source.as_ref() as &(dyn core::error::Error + 'static))  
   |                                  ^^^^^^  
   |  
   = note: the following trait bounds were not satisfied:  
           `dyn StdError: AsRef<_>`  
           which is required by `&dyn StdError: AsRef<_>`  
           `&dyn StdError: AsRef<_>`  
           which is required by `&&dyn StdError: AsRef<_>`
```

signature/src/error.rs
```diff
@@ -71,18 +71,17 @@ impl Display for Error {
...
 
-#[cfg(feature = "std")]
-impl std::error::Error for Error {
-    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+impl core::error::Error for Error {
+    fn source(&self) -> Option<&(dyn core::error::Error + 'static)> {
         self.source
             .as_ref()
-            .map(|source| source.as_ref() as &(dyn std::error::Error + 'static))
+            .map(|source| source.as_ref() as &(dyn core::error::Error + 'static))
```
joncinque: The issue is here -- since `source` only exists when the `alloc` feature is enabled, we need to gate this properly:
```suggestion
        #[cfg(feature = "alloc")]
        {
            self.source.as_ref()
            .map(|source| source.as_ref() as &(dyn core::error::Error + 'static))
        }
        #[cfg(not(feature = "alloc"))]
        {
            None
        }
```

joncinque:
Looks like there's still a formatting issue and a build issue. If you can, please run the commands before committing. And if you don't have time, I can take the PR from here

divyaranjan1905:
@joncinque The commands somehow don't run on my system. And it's probably my system's issue. But yeah please take the PR from here and make the necessary formatting fixes.

apfitzge (approved):

Issue #117, opened 2025-04-02
Target architecture `wasm32` should not assume a Javascript environment


gregorydemay:
# Problem

When crates of the `solana-sdk` are compiled to `wasm32`, the dependency `wasm-bindgen` is included. 
For example for `solana-keypair` ([here](https://github.com/anza-xyz/solana-sdk/blob/5cd7a9c9e7acdd02a6fccc3c7970f796d5d6577d/keypair/Cargo.toml#L25)):
```rust
[target.'cfg(target_arch = "wasm32")'.dependencies]
wasm-bindgen = { workspace = true }
```

The `wasm-bindgen` crate only makes sense if the compiled wasm is to be used in a Javascript environment (typically a browser); however, there are many other use cases for compiling to wasm and running the compiled wasm [outside a web environment](https://webassembly.org/docs/non-web/). Currently this fails due to the presence of  `wasm-bindgen`.

Affected crates:

1. `solana-hash` ([here](https://github.com/anza-xyz/solana-sdk/blob/5cd7a9c9e7acdd02a6fccc3c7970f796d5d6577d/hash/Cargo.toml#L38))
2. `solana-instruction` ([here](https://github.com/anza-xyz/solana-sdk/blob/5cd7a9c9e7acdd02a6fccc3c7970f796d5d6577d/instruction/Cargo.toml#L28))
3. `solana-keypair` ([here](https://github.com/anza-xyz/solana-sdk/blob/5cd7a9c9e7acdd02a6fccc3c7970f796d5d6577d/keypair/Cargo.toml#L25))
4. `solana-message` ([here](https://github.com/anza-xyz/solana-sdk/blob/5cd7a9c9e7acdd02a6fccc3c7970f796d5d6577d/message/Cargo.toml#L34))
5. `solana-pubkey` ([here](https://github.com/anza-xyz/solana-sdk/blob/5cd7a9c9e7acdd02a6fccc3c7970f796d5d6577d/pubkey/Cargo.toml#L45))
6. `solana-sdk` ([here](https://github.com/anza-xyz/solana-sdk/blob/5cd7a9c9e7acdd02a6fccc3c7970f796d5d6577d/sdk/Cargo.toml#L180))
7. `solana-transaction` ([here](https://github.com/anza-xyz/solana-sdk/blob/5cd7a9c9e7acdd02a6fccc3c7970f796d5d6577d/transaction/Cargo.toml#L36))

# Proposal

[...]

Issue #84, opened 2025-03-24
Tracking issue for upcoming breaking changes


joncinque:
#### Problem

Now that the sdk is decoupled from agave, we have a lot more freedom to make breaking changes in individual crates.

However, there are still a lot of users of the omnibus crates `solana-sdk` and `solana-program`, and although we can simply stop updating them ever again, that would cause a lot of churn for downstream developers.

This issue aims to bundle together all of the breaking changes for the next breaking release of `solana-sdk` and `solana-program`.

#### Breaking changes in solana-program

* #12 

#### Breaking changes in solana-sdk

All the changes in solana-program, plus:

* #46
* #74
* #77 
* #62
* keypair: remove APIs that expose dalek types directly
* #26 
* feature-set: deprecation (more info at #82)
* transaction / precompiles / ed25519 / reserved-account-keys / secp256k1 / secp256r1: Use feature-set-interface instead of feature-set

jstarry:
Is it possible to release a patch update for solana-sdk still? I'd like to land this: https://github.com/anza-xyz/solana-sdk/pull/94

joncinque:
Yes, we'll do some patch releases to mark everything as deprecated before doing the breaking release

jstarry:
the transaction-context module was re-added in #95 so we can remove #77 from the list of breaking changes fyi

joncinque:
> the transaction-context module was re-added in https://github.com/anza-xyz/solana-sdk/pull/95 so we can remove https://github.com/anza-xyz/solana-sdk/pull/77 from the list of breaking changes fyi

I left in the list as a reminder to remove it once more 😅 I updated the wording though, thanks!

joncinque:
Closing this issue since v3 is out

PR #118, opened 2025-04-02
build: Add `js` Cargo feature flag for WASM builds

lpahlavi:
Previously, the wasm32 target implicitly assumed a browser environment, which caused issues when building for non-browser WASM environments due to the unconditional inclusion of `wasm-bindgen`.

This commit introduces an explicit `js` feature flag, making `wasm-bindgen` and `js-sys` conditional dependencies. This allows greater flexibility for different WASM execution environments.

Related to #117

### DIFF
--- a/hash/Cargo.toml
+++ b/hash/Cargo.toml
@@ -20,6 +20,7 @@ bs58 = { workspace = true, default-features = false }
 bytemuck = { workspace = true, optional = true }
 bytemuck_derive = { workspace = true, optional = true }
 serde = { workspace = true, optional = true }
+js-sys = { workspace = true, optional = true }
 serde_derive = { workspace = true, optional = true }
 solana-atomic-u64 = { workspace = true }
 solana-frozen-abi = { workspace = true, optional = true, features = [
@@ -29,6 +30,7 @@ solana-frozen-abi-macro = { workspace = true, optional = true, features = [
     "frozen-abi",
 ] }
 solana-sanitize = { workspace = true }
+wasm-bindgen = { workspace = true, optional = true }
 
 [dev-dependencies]
 solana-hash = { path = ".", features = ["dev-context-only-utils"] }
@@ -47,6 +49,7 @@ frozen-abi = [
     "dep:solana-frozen-abi-macro",
     "std"
 ]
+js = ["dep:wasm-bindgen", "dep:js-sys"]
 serde = ["dep:serde", "dep:serde_derive"]
 std = []
 
--- a/hash/src/lib.rs
+++ b/hash/src/lib.rs
@@ -19,7 +19,7 @@ use {
     },
     solana_sanitize::Sanitize,
 };
-#[cfg(target_arch = "wasm32")]
+#[cfg(feature = "js")]
 use {
     js_sys::{Array, Uint8Array},
     std::{boxed::Box, format, string::String, vec},
@@ -38,7 +38,7 @@ pub const MAX_BASE58_LEN: usize = 44;
 ///
 /// [SHA-256]: https://en.wikipedia.org/wiki/SHA-2
 /// [blake3]: https://github.com/BLAKE3-team/BLAKE3
-#[cfg_attr(target_arch = "wasm32", wasm_bindgen)]
+#[cfg_attr(feature = "js", wasm_bindgen)]
 #[cfg_attr(feature = "frozen-abi", derive(solana_frozen_abi_macro::AbiExample))]
 #[cfg_attr(
     feature = "borsh",
@@ -151,7 +151,7 @@ impl Hash {
     }
 }
 
-#[cfg(target_arch = "wasm32")]
+#[cfg(feature = "js")]
 #[allow(non_snake_case)]
 #[wasm_bindgen]
 impl Hash {
--- a/pubkey/src/lib.rs
+++ b/pubkey/src/lib.rs
@@ -31,7 +31,7 @@ use {
     num_traits::{FromPrimitive, ToPrimitive},
     solana_decode_error::DecodeError,
 };
-#[cfg(target_arch = "wasm32")]
+#[cfg(feature = "js")]
 use {
     js_sys::{Array, Uint8Array},
     wasm_bindgen::{prelude::wasm_bindgen, JsCast, JsValue},
@@ -149,7 +149,7 @@ impl From<u64> for PubkeyError {
 /// [ed25519]: https://ed25519.cr.yp.to/
 /// [pdas]: https://solana.com/docs/core/cpi#program-derived-addresses
 /// [`Keypair`]: https://docs.rs/solana-sdk/latest/solana_sdk/signer/keypair/struct.Keypair.html
-#[cfg_attr(target_arch = "wasm32", wasm_bindgen)]
+#[cfg_attr(feature = "js", wasm_bindgen)]
 #[repr(transparent)]
 #[cfg_attr(feature = "frozen-abi", derive(solana_frozen_abi_macro::AbiExample))]
 #[cfg_attr(
@@ -1079,7 +1079,7 @@ macro_rules! impl_borsh_serialize {
 #[cfg(feature = "borsh")]
 impl_borsh_serialize!(borsh0_10);
 
-#[cfg(all(target_arch = "wasm32", feature = "curve25519"))]
+#[cfg(all(feature = "js", feature = "curve25519"))]
 fn js_value_to_seeds_vec(array_of_uint8_arrays: &[JsValue]) -> Result<Vec<Vec<u8>>, JsValue> {
     let vec_vec_u8 = array_of_uint8_arrays
         .iter()
@@ -1097,12 +1097,12 @@ fn js_value_to_seeds_vec(array_of_uint8_arrays: &[JsValue]) -> Result<Vec<Vec<u8
     }
 }
 
-#[cfg(target_arch = "wasm32")]
+#[cfg(feature = "js")]
 fn display_to_jsvalue<T: fmt::Display>(display: T) -> JsValue {
     std::string::ToString::to_string(&display).into()
 }
 
-#[allow(non_snake_case)]
+#[cfg(feature = "js")]
 #[cfg(target_arch = "wasm32")]
 #[wasm_bindgen]
 impl Pubkey {
--- a/instruction/Cargo.toml
+++ b/instruction/Cargo.toml
@@ -12,20 +12,18 @@ edition = { workspace = true }
 [dependencies]
 bincode = { workspace = true, optional = true }
 borsh = { workspace = true, optional = true }
+getrandom = { workspace = true, optional = true, features = ["js", "wasm-bindgen"] }
+js-sys = { workspace = true, optional = true }
 num-traits = { workspace = true }
 serde = { workspace = true, optional = true }
 serde_derive = { workspace = true, optional = true }
 solana-frozen-abi = { workspace = true, optional = true }
 solana-frozen-abi-macro = { workspace = true, optional = true }
 solana-pubkey = { workspace = true, default-features = false }
+wasm-bindgen = { workspace = true, optional = true }
 
 [target.'cfg(target_os = "solana")'.dependencies]
 solana-define-syscall = { workspace = true }
-
-[target.'cfg(target_arch = "wasm32")'.dependencies]
-getrandom = { workspace = true, features = ["js", "wasm-bindgen"] }
-js-sys = { workspace = true }
-wasm-bindgen = { workspace = true }
 
 [dev-dependencies]
 solana-instruction = { path = ".", features = ["borsh"] }
@@ -40,6 +38,7 @@ frozen-abi = [
     "serde",
     "std",
 ]
+js = ["dep:getrandom", "dep:wasm-bindgen", "dep:js-sys"]
 serde = [
     "dep:serde",
     "dep:serde_derive",
--- a/instruction/src/lib.rs
+++ b/instruction/src/lib.rs
@@ -26,7 +26,7 @@ pub use account_meta::AccountMeta;
 pub mod error;
 #[cfg(target_os = "solana")]
 pub mod syscalls;
-#[cfg(all(feature = "std", target_arch = "wasm32"))]
+#[cfg(all(feature = "std", feature = "js"))]
 pub mod wasm;
 
 /// A directive for a single invocation of a Solana program.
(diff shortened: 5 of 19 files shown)


apfitzge:
@joncinque requesting your review on this since you just did the feature powerset change. I don't have a strong opinion, if you're okay w/ it I will sign-off on it for the `tx-metadata` team.

hash/Cargo.toml
```diff
@@ -29,6 +30,7 @@ solana-frozen-abi-macro = { workspace = true, optional = true, features = [
     "frozen-abi",
 ] }
 solana-sanitize = { workspace = true }
+wasm-bindgen = { workspace = true, optional = true }
 
 [dev-dependencies]
 solana-hash = { path = ".", features = ["dev-context-only-utils"] }
```
joncinque, line 48: I can't comment lower down, but can you remove `js-sys` and `wasm-bindgen` from the wasm32 dependencies on lines 38-40?

hash/src/lib.rs
```diff
@@ -19,7 +19,7 @@ use {
     },
```
joncinque, line 19: I can't comment further up on line 6, but does a non-JS wasm build always need std? We might be able to change the std usage to:

```
#[cfg(feature = "std")]
```

Maybe more specifically -- does a wasm build like yours need std by default? And if we simplify that feature, then we shoudl also remove the `target_arch` bit on line 12.

pubkey/src/lib.rs
```diff
@@ -31,7 +31,7 @@ use {
     num_traits::{FromPrimitive, ToPrimitive},
```
joncinque, line 31: Same bit at the top, lines 7 and 15, changing to

```
#[cfg(feature = "std")]
```

pubkey/src/lib.rs
```diff
@@ -149,7 +149,7 @@ impl From<u64> for PubkeyError {
 /// [ed25519]: https://ed25519.cr.yp.to/
 /// [pdas]: https://solana.com/docs/core/cpi#program-derived-addresses
 /// [`Keypair`]: https://docs.rs/solana-sdk/latest/solana_sdk/signer/keypair/struct.Keypair.html
-#[cfg_attr(target_arch = "wasm32", wasm_bindgen)]
+#[cfg_attr(feature = "js", wasm_bindgen)]
 #[repr(transparent)]
 #[cfg_attr(feature = "frozen-abi", derive(solana_frozen_abi_macro::AbiExample))]
 #[cfg_attr(
```
joncinque, line 155: I can't comment there, but there's a `TryFrom<Vec<u8>> for Pubkey` impl around line 429 and some gating around `new_unique()` (lines 483), and we should change the gating to just:

```
#[cfg(feature = "std")]
```

pubkey/src/lib.rs
```diff
@@ -1097,12 +1097,12 @@ fn js_value_to_seeds_vec(array_of_uint8_arrays: &[JsValue]) -> Result<Vec<Vec<u8
     }
 }
 
-#[cfg(target_arch = "wasm32")]
+#[cfg(feature = "js")]
 fn display_to_jsvalue<T: fmt::Display>(display: T) -> JsValue {
     std::string::ToString::to_string(&display).into()
 }
 
-#[allow(non_snake_case)]
```
joncinque, line 1105: We'll still need the `allow(non_snake_case)` directive

pubkey/src/lib.rs
```diff
@@ -1097,12 +1097,12 @@ fn js_value_to_seeds_vec(array_of_uint8_arrays: &[JsValue]) -> Result<Vec<Vec<u8
...
+#[cfg(feature = "js")]
 #[cfg(target_arch = "wasm32")]
```
joncinque: We can probably remove this since the `js` feature gates it

hash/src/lib.rs
```diff
@@ -19,7 +19,7 @@ use {
...
     solana_sanitize::Sanitize,
 };
-#[cfg(target_arch = "wasm32")]
+#[cfg(feature = "js")]
```
joncinque: Assuming that we gate std just on the std feature, let's gate this properly with:

```
#[cfg(all(feature = "std", feature = "js"))]
```

For reference, I'm looking at how uuid declares something similar here: https://github.com/uuid-rs/uuid/blob/cb19a46cf17cd9840205e0d67a32ea8a18e58374/src/timestamp.rs#L308

I don't think we need the full target_arch part too, but the std + js feature gate seems to make sense

pubkey/src/lib.rs
```diff
@@ -1079,7 +1079,7 @@ macro_rules! impl_borsh_serialize {
 #[cfg(feature = "borsh")]
 impl_borsh_serialize!(borsh0_10);
 
-#[cfg(all(target_arch = "wasm32", feature = "curve25519"))]
+#[cfg(all(feature = "js", feature = "curve25519"))]
```
joncinque: This one will also need the std feature

pubkey/src/lib.rs
```diff
@@ -1097,12 +1097,12 @@ fn js_value_to_seeds_vec(array_of_uint8_arrays: &[JsValue]) -> Result<Vec<Vec<u8
...
 }
 
-#[cfg(target_arch = "wasm32")]
+#[cfg(feature = "js")]
 fn display_to_jsvalue<T: fmt::Display>(display: T) -> JsValue {
     std::string::ToString::to_string(&display).into()
 }
 
-#[allow(non_snake_case)]
+#[cfg(feature = "js")]
```
joncinque: This will also need the std feature

pubkey/src/lib.rs
```diff
@@ -1097,12 +1097,12 @@ fn js_value_to_seeds_vec(array_of_uint8_arrays: &[JsValue]) -> Result<Vec<Vec<u8
     }
 }
 
-#[cfg(target_arch = "wasm32")]
+#[cfg(feature = "js")]
```
joncinque: This will also need the std feature

joncinque (commented):
Thanks for your contribution! I'm in agreement for this change, makes total sense to me.

The main issue: I believe this is a breaking change for people currently using the packages in wasm builds, since they're expecting all of the JS stuff to come with it. That's unfortunately going to complicate / slow down the release process a bit.

Also, in all the Cargo.tomls, we should keep gating the `js-sys` and `wasm-bindgen` crates on:

```
[target.'cfg(target_family = "wasm")'.dependencies]
```

As another note, I looked around at different feature names used for enabling JS for wasm builds, and here are some popular crates:

* `getrandom` uses `wasm_js`
* `time` uses `wasm-bindgen` as the feature name
* `chrono` uses `wasmbind`
* `uuid` and `jiff` use `js`

So your choice of `js` is good, but I could also be convinced to do something like `wasm-js` if you liked that more.

pubkey/src/lib.rs
lpahlavi: Ah I mistakenly seem to have deleted `#[allow(non_snake_case)]` instead of `#[cfg(target_arch = "wasm32")]`. Thanks for noticing this! It is now fixed.

pubkey/src/lib.rs
lpahlavi: Absolutely! Done.

hash/Cargo.toml
lpahlavi: As per your suggestion, I've moved all the dependencies for the new `js` feature back to a section gated by the `target_arch = "wasm32"`.

hash/src/lib.rs
lpahlavi: Good question! In our particular case, we actually do always use `std` for non-JS builds, but I think what makes most sense here is probably to change most usages of `target_arch = "wasm32"` to `feature = "js"`. I tried to have a look in each affected file where the `std` dependencies where needed and I tried to make it so that when the `std` imports were only needed in code gated by the `js` feature, I changed `target_arch = "wasm32"` to `feature = "js"`, and otherwise I removed the gate on the target architecture. WDYT?

pubkey/src/lib.rs
lpahlavi: Done.

pubkey/src/lib.rs
lpahlavi: Good catch! Done.

pubkey/src/lib.rs
lpahlavi: Good catch! Done.

pubkey/src/lib.rs
lpahlavi: Good catch! Done.

pubkey/src/lib.rs
lpahlavi: Good catch! Done.

lpahlavi (commented):
@joncinque Thank you so much for reviewing this PR so quickly!

I am aware that this will unfortunately be breaking... I believe due to Cargo's limitation to opt-in only features, there's not much we can do to mitigate this. I thought about adding the feature to `default` since the dependencies are anyways gated by the `wasm32` target architecture, but this seems a little bit awkward. WDYT?

I am also wondering if it would be theoretically best to add `target_arch = "wasm32"` to all the places where we have `feature = "js"` or if we are OK with the code simply not compiling when having the `js` feature enabled for another target architecture than `wasm32`. Do you have an opinion on this?

One final question: do you have any idea how long it might take before this is released given that it involves a breaking change?

Thanks again and let me know if you have any further feedback!

joncinque:
> I am aware that this will unfortunately be breaking... I believe due to Cargo's limitation to opt-in only features, there's not much we can do to mitigate this. I thought about adding the feature to default since the dependencies are anyways gated by the wasm32 target architecture, but this seems a little bit awkward. WDYT?

That would still be breaking for users of `default-features = false`, but since that likely narrows the scope of users down a lot, maybe we can do it just this once. We'd be breaking wasm users with `default-features = false`, which should be a tiny fraction.

The other option, which is unfortunately a lot of work, is to use the semver trick for all these crates: https://github.com/dtolnay/semver-trick. We can have the v2 crates depend on the v3 crates with the correct feature, and then re-export everything, but to be totally honest, I'm not too keen on doing all of that.

> I am also wondering if it would be theoretically best to add target_arch = "wasm32" to all the places where we have feature = "js" or if we are OK with the code simply not compiling when having the js feature enabled for another target architecture than wasm32. Do you have an opinion on this?

Yes, after thinking about it a bit more, let's do what you suggested and *also* gate all js usage on the wasm32 check.

> One final question: do you have any idea how long it might take before this is released given that it involves a breaking change?

I'm still figuring out the exact release process, and I need to coordinate with some of our downstream users, so it might take a month before the crates are released.

lpahlavi:
@joncinque Thanks a lot for your feedback! 

Given your feedback, I've now:
* added the `js` feature to the `default` features for all affected crates
* made sure that dependencies are included with the `js` feature _only if_ the `js` feature is enabled
* gated all `wasm_bingen` related functionality by both the `js` feature _and_ the `wasm32` target architecture

I believe this way we should minimize the number of affected users (i.e. only the ones using `wasm_bindgen` _with_ `default-features = false`) while still avoiding having to resort to the semver trick (I'm also not too keen on resorting to that...).

> I'm still figuring out the exact release process, and I need to coordinate with some of our downstream users, so it might take a month before the crates are released.

Thank you for the info! That should be completely fine for us as we can rely on the forked repository in the meantime. Let me know if there's anything I can do to support on that side.

Let me know if you have any other feedback concerning the code!

joncinque:
We've been discussing this a bit offline and in Discord (https://discord.com/channels/428295358100013066/476811830145318912/1369891812797452288), and I'm starting to lean towards a slightly different solution, of just removing all of the wasm-bindgen code completely from the component crates, and instead factoring it all into a new crate, as in https://github.com/anza-xyz/solana-sdk/pull/138

We've been sidetracked by some other things, so at the latest, the breaking changes should be released by the end of the month.

lpahlavi:
Thanks a lot for the quick reply!

> We've been discussing this a bit offline and in Discord (https://discord.com/channels/428295358100013066/476811830145318912/1369891812797452288), and I'm starting to lean towards a slightly different solution, of just removing all of the wasm-bindgen code completely from the component crates, and instead factoring it all into a new crate, as in #138

I see, thanks for the link! Unfortunately it seems I don't have access to the `#sdk` channel so I can't check out the details, but in principle the idea sounds good as well to me and I don't have a particularly strong preference for one or the other option. 

Would someone from Anza then take the lead in this case?

> We've been sidetracked by some other things, so at the latest, the breaking changes should be released by the end of the month.

I see no problem, thank you for the update! Do you know if there is somewhere I can follow this more closely?

joncinque:
I'll repost the most relevant comment:

> I really don’t think there is a single person who throws wasm bindgen into an isomorphic library and thinks “you know what my binary needs? 50 random unused additional functions that only work in the runtime”
context: https://x.com/deanmlittle/status/1917987491614974329?s=46&t=Y1-eTu6GZzzEk5uEf3Rs1Q

And yes, we'll take the lead on this from here, most likely in #138 -- that'll be the PR to follow. It should end up being much simpler since we're just removing everything. Otherwise, you can see updates on the breaking changes issue #84

lpahlavi:
Fantastic, thanks a lot! Looking forward to the release!

Do you have any idea if the [`solana-program/system`](https://github.com/solana-program/system) repo will head in the same direction (see this [issue](https://github.com/solana-program/system/issues/47), essentially a clone of #117)? I ask because it seems that you are also a major contributor there and the problem is essentially the same as here.

joncinque:
> Do you have any idea if the [solana-program/system](https://github.com/solana-program/system) repo will head in the same direction (see this https://github.com/solana-program/system/issues/47, essentially a clone of https://github.com/anza-xyz/solana-sdk/issues/117)?

Yeah, we should probably do the same thing everywhere

lpahlavi:
Great, thank you! I will update the issue there to reflect this.

lpahlavi:
Hi @joncinque! Do you know if there are any updates on the progress for the next release including the breaking changes? Perhaps I could get an invite to the Discord channel to follow the news there? Thanks!

joncinque:
Sorry for the late reply, we'll start landing some breaking changes first thing next week and probably releasing some `3.0.0-rc.1` crates for people to test with by the end of the week

lpahlavi:
Thanks a lot for the heads up! I've been monitoring the repo a bit and see that the first breaking changes look to be soon merged. Looking forward to testing the first release candidates!

lpahlavi:
@joncinque @LucasSte : I've noticed that the PR to add a WASM crate ([here](https://github.com/anza-xyz/solana-sdk/pull/138)) doesn't quite allow removing all `wasm-bindgen` dependencies (yet). Is the idea to also merge this PR here adding a feature flag, or do you expect to create some new PRs to remove the other `wasm-bindgen` dependencies?

joncinque:
That's correct, after that PR lands, the idea is to just move all of the wasm logic from the individual crates and into `solana-sdk-wasm-js`

lpahlavi:
> That's correct, after that PR lands, the idea is to just move all of the wasm logic from the individual crates and into `solana-sdk-wasm-js`

Great, thanks a lot for the confirmation @joncinque! Just to clarify the timeline though, is this planned to be all part of the upcoming release? Or is this a longer term plan?

joncinque:
Sorry for the wait! #244 will finally take care of this

PR #120, opened 2025-04-02
program-memory: Add missing unsafe

febo:
### Program

Currently there are a few functions on `program-memory` crate that are missing the "unsafe" qualifier.

### Solution

Add the missing "unsafe". This PR also adds the `no_std` attribute to the crate since it is already a `no_std` "friendly" crate.


joncinque (commented):
Looks great! Thanks for adding the breaking tag, we'll wait to merge this once all of the deprecations have been published.

joncinque (approved):

PR #122, opened 2025-04-02
rust: Bump nightly to 2025-04-01

joncinque:
#### Problem

The stable toolchain was bumped to 1.85.1, but the nightly toolchain is still behind.

#### Summary of changes

Update the nightly version to 2025-04-01 and fix issues that come up.

* io::Error::new(io::ErrorKind::Other ...) -> io::Error::other
* map_or(true, ...) -> is_none_or
* identation fixes
* using div_ceil

I wasn't sure what version Agave would land on, so let me konw if I should use another one.


yihau (approved):
it reminds me that Agave is still using the outdated one 🫠

apfitzge (approved):

PR #123, opened 2025-04-02
add config-interface crate

kevinheavey:
Problem: some crates (particularly solana-account-decoder) depend on solana-config-program, which is very heavy because it depends on solana-program-runtime.

Solution:

- rip out everything except the processor and date_instruction.rs (deprecated and doesn't seem to be used elsewhere).
- make serde and bincode optional in the new crate

Once this is published we would then replace the relevant code in agave/programs/config with re-exports

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2741,6 +2741,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "solana-config-interface"
+version = "1.0.0"
+dependencies = [
+ "bincode",
+ "serde",
+ "serde_derive",
+ "solana-account",
+ "solana-instruction",
+ "solana-pubkey",
+ "solana-sdk-ids",
+ "solana-short-vec",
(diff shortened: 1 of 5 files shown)


joncinque:
This makes sense to me, thanks! Can you move this PR to https://github.com/solana-program/config instead though?

buffalojoec:
The existing client in https://github.com/solana-program/config does not depend on `solana-program-runtime`.

kevinheavey:
It depends on solana-program though

febo:
I think we can follow the same approach of [solana-program/system]( https://github.com/solana-program/system) and [solana-program/stake]( https://github.com/solana-program/stake) – i.e., have an interface crate alongside the clients and program.

joncinque:
Can this be closed since https://github.com/solana-program/config/pull/47 landed?

Issue #124, opened 2025-04-03
[solana-pubkey] enable std or add lightweight mem::size_of


therealdannzor:
Attempting to upgrade to Solana `2.x` and `Anchor 0.31.0` by patching crates.io (not ideal, I know). 

I get this error, any plans on enabling this for other crates that are dependent on `solana-pubkey`?

```
error[E0425]: cannot find function `size_of` in this scope
   --> src/lib.rs:481:38
    |
481 |         const COUNTER_BYTES: usize = size_of::<T>();
    |                                      ^^^^^^^ not found in this scope
    |
help: consider importing one of these items
    |
11  + use core::mem::size_of;
    |
11  + use crate::mem::size_of;
    |
11  + use std::mem::size_of;
```

Please note that I'm not the direct consumer of this dependency and would ideally try to avoid forking it.

Thanks in advance 🥲

joncinque:
Thanks for reporting the issue. `size_of` is part of the rust prelude: https://doc.rust-lang.org/core/prelude/v1/index.html, which is why this is a bit strange.

Do you have a simple reproduction that shows the problem?

therealdannzor:
Thanks for your quick response @joncinque - much appreciated

I think it might be difficult to reproduce since I'm manually patching `crates.io` with all the various Solana crates that are necessary to upgrade to both `2.x` *and* with anchor `0.31.0`. As far as I can tell, this is the only way forward.

I don't have a minimal reproducible example of this right now but  some suspicions would be using `resolver = 2` that results in stricter feature resolution which means that the crate `solana-pubkey` builds in a `no_std` environment. My dependency tree is clean and only relies on the commit hash `1971070` to the `solana-sdk` workspace.

All in all, there are a lot of moving parts, but I suspect it will be needed at some point when all the crates need to use the same versions in order to upgrade to latest Solana.

Is a MRE of the compile error a blocker for you, in order to consider this addition?

kevinheavey:
size_of only made it into the prelude in Rust 1.80 and solana-program's MSRV is 1.79

therealdannzor:
Thanks @kevinheavey for the reply, I forgot to add that I'm also upgrading to `rustc 1.85.0 (4d91de4e4 2025-02-17)` as part of this endeavor.

joncinque:
> size_of only made it into the prelude in Rust 1.80 and solana-program's MSRV is 1.79

Ah gotcha, thanks! How were you able to see that by the way?

[...]

PR #125, opened 2025-04-04
pubkey: fix compile error by importing size_of

therealdannzor:
Resolves https://github.com/anza-xyz/solana-sdk/issues/124

The crate misses core::mem::size_of (or std) causing build failures when other dependencies require it. Add import to resolve, opting for `core` as opposed to `std` such that it works with existing `#![no_std]` declaration.

As part of upgrading to anchor 0.31.0 and solana 2.x


therealdannzor:
cc @febo @joncinque

joncinque (approved):
Thanks for the contribution!

PR #126, opened 2025-04-04
secp256r1: Deprecate openssl types in public API

joncinque:
#### Problem

This is #114, but for secp256r1 instead, removing external types from the public interface.

#### Summary of changes

This one was a bit more complicated because it uses openssl types, which aren't as straightforward as simple bytes, so it uses bytes encoded in DER as the interface. This seemed like the most standard format from simple web searching, but I'm certainly no expert in this area.

There seemed to also be a lot of vecs in the new instruction implementation, which shouldn't be necessary, so I changed those to arrays.

### DIFF
--- a/secp256r1-program/src/lib.rs
+++ b/secp256r1-program/src/lib.rs
@@ -83,21 +83,40 @@ mod target_arch {
     // Field size in bytes
     pub const FIELD_SIZE: usize = 32;
 
+    #[deprecated(
+        since = "2.2.3",
+        note = "Use `new_secp256r1_instruction_with_signature` and `sign_message` instead"
+    )]
     pub fn new_secp256r1_instruction(
         message: &[u8],
         signing_key: EcKey<Private>,
     ) -> Result<Instruction, Box<dyn std::error::Error>> {
-        let group = EcGroup::from_curve_name(Nid::X9_62_PRIME256V1)?;
-        if signing_key.group().curve_name() != Some(Nid::X9_62_PRIME256V1) {
-            return Err(("Signing key must be on the secp256r1 curve".to_string()).into());
-        }
+        let signature = sign_message(message, &signing_key.private_key_to_der()?)?;
 
         let mut ctx = BigNumContext::new()?;
+        let group = EcGroup::from_curve_name(Nid::X9_62_PRIME256V1)?;
         let pubkey = signing_key.public_key().to_bytes(
             &group,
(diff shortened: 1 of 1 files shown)


secp256r1-program/src/lib.rs
```diff
@@ -83,21 +83,40 @@ mod target_arch {
...
+        Ok(new_secp256r1_instruction_with_signature(
+            message,
+            &signature,
+            &pubkey.try_into().unwrap(), // len was just checked, this is safe
+        ))
+    }
+
+    pub fn sign_message(
+        message: &[u8],
+        der: &[u8],
```
samkim-crypto: This looks fine, but should we rename the `der` variable to `priv_key_bytes_der` or `priv_key_der` to be more specific? `der` can mean an encoding for signatures as well, so it might be worth making it clear from just the function syntax (though we immediately parse as a signing key below).

secp256r1-program/src/lib.rs
```diff
@@ -129,16 +148,20 @@ mod target_arch {
...
             new_padded_s[FIELD_SIZE.saturating_sub(new_s_bytes.len())..]
                 .copy_from_slice(&new_s_bytes);
 
             signature[FIELD_SIZE..].copy_from_slice(&new_padded_s);
         }
+        Ok(signature)
+    }
 
-        assert_eq!(pubkey.len(), COMPRESSED_PUBKEY_SERIALIZED_SIZE);
-        assert_eq!(signature.len(), SIGNATURE_SERIALIZED_SIZE);
```
samkim-crypto, line 140: It seems that the pubkey length check was added above, but the signature length check was removed. Is this intended?

samkim-crypto (commented):
Looks good to me! Just two minor comments below.

secp256r1-program/src/lib.rs
joncinque: Makes sense to me! I went with `priv_key_bytes_der`

secp256r1-program/src/lib.rs
joncinque: That's correct -- `signing_key.public_key().to_bytes()` returns `Vec<u8>`, so we aren't 100% sure the size is correct, whereas `sign_message` returns `[u8; SIGNATURE_SERIALIZED_SIZE]`, so we already know statically that the size is correct, which means it's unnecessary to check the size.

samkim-crypto (approved):
Looks good to me!

PR #130, opened 2025-04-06
ci: Add build-sbf step

### LINKED ISSUE #70: CI: Add step for `cargo build-sbf`
#### Problem

As seen in issues like #69, we don't have checks in CI to make sure that BPF-compatible crates can actually build in BPF.

#### Proposed solution

Add a CI step to run `cargo build-sbf` on all applicable crates.

febo:
### Problem

In general, the workspace rust toolchain has a different (newer) version than the built-in one used by `build-sbf`. This could potentially lead to compilation problems not currently detected by the CI when using crates in a program context – e.g., #124.

### Solution

Add a "check-build-sbf" step to the CI to build crates using `cargo build-sbf`.

Fixes #70

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,7 +134,7 @@ tag-message = "Publish {{crate_name}} v{{version}}"
 consolidate-commits = false
 
 [workspace.metadata.cli]
-solana = "2.2.0"
+solana = "2.2.7"
 
 [workspace.dependencies]
 ahash = "0.8.11"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -133,6 +133,9 @@ pre-release-commit-message = "Publish {{crate_name}} v{{version}}"
 tag-message = "Publish {{crate_name}} v{{version}}"
 consolidate-commits = false
 
+[workspace.metadata.cli]
+solana = "2.2.0"
+
 [workspace.dependencies]
 ahash = "0.8.11"
 anyhow = "1.0.96"
--- a/scripts/check-build-sbf.sh
+++ b/scripts/check-build-sbf.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+here="$(dirname "$0")"
+src_root="$(readlink -f "${here}/..")"
+cd "${src_root}"
+
+./cargo build-sbf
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -133,6 +133,9 @@ pre-release-commit-message = "Publish {{crate_name}} v{{version}}"
 tag-message = "Publish {{crate_name}} v{{version}}"
 consolidate-commits = false
 
+[workspace.metadata.cli]
+solana = "2.2.0"
```
joncinque: We can at least go to 2.2.7!

scripts/check-build-sbf.sh
joncinque, line 1: Nit: since this is actually building, I think we can just call the file `build-sbf.sh`

scripts/check-build-sbf.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+here="$(dirname "$0")"
+src_root="$(readlink -f "${here}/..")"
+cd "${src_root}"
+
+./cargo build-sbf
```
joncinque: I think we can make this better to include all crates that we expect to work on-chain.

`cargo-build-sbf` only builds `cdylib` targets in the workspace, and I've heard that we should even remove that target in order to perform LTO. And later on, as we remove re-exported crates from `solana-program` and `solana-sdk`, those crates will no longer be included in this job.

I came up with this little script to *exclude* certain directories, since most of them actually work! What do you think?

```
#!/usr/bin/env bash

set -eo pipefail
here="$(dirname "$0")"
cd "${src_root}"

exclude_list=(
  ".github"
  "scripts"
  "client-traits"
  "ed25519-program"
  "example-mocks"
  "feature-set"
  "feature-set-interface"
  "file-download"
  "genesis-config"
  "keypair"
  "logger"
  "offchain-message"
  "precompiles"
  "presigner"
  "quic-definitions"
  "rent-collector"
  "reserved-account-keys"
  "secp256k1-program"
  "secp256r1-program"
  "system-transaction"
  "transaction"
)
for dir in $(git ls-tree -d --name-only HEAD .); do
  if [[ ! " ${exclude_list[*]} " =~ [[:space:]]${dir}[[:space:]] ]]; then
    (cd $dir && cargo build-sbf)
  fi
done
```

joncinque: Oh sorry, I didn't run the script through shellcheck :pray:

joncinque (approved):
Beautiful, thanks!

PR #132, opened 2025-04-08
CI: Add minimum supported rust version check

joncinque:
#### Problem

We broke pubkey by assuming that `size_of` was included in the prelude, which was added after 1.79, the minimum supported rust version for solana-program. This happened because we're not explicitly testing the minimum supported rust version in crates that expose it.

More info at #124

#### Summary of changes

Add a "minimum" toolchain in the repo, and add a CI step + script to check that `cargo check` passes for this toolchain.

While working on this, I found a few more places that were broken, so fix those up.


alexpyattaev (approved):
Thank you! This was a sneaky one!

Issue #121, opened 2025-04-02
Compatibility Issues Between Solana SDK and Agave in Newer Versions?


YalDan:
## Overview
I encountered several compatibility issues when trying to build Agave v2.2.6 with different versions of Solana SDK using the [patch-crates-no-header.sh](https://github.com/anza-xyz/solana-sdk/blob/master/scripts/patch-crates-no-header.sh) script. This causes build failures due to some recent changes.

## Environment
- Agave version: v2.2.6
- Solana SDK versions tested: v2.2.1, v2.2.2
- Script: patch-crates-no-header.sh

### 1. Missing transaction-context directory
Initially encountered errors because the patching script looks for `transaction-context` in the SDK repo, but it was moved:

```
error: failed to select a version for the requirement `solana-transaction-context = "=2.2.6"`
candidate versions found which didn't match: 2.2.1, 2.2.0, 0.0.2, ...
```

This appears to be related to PR #77 "Move solana-transaction-context into mono repo" and PR #95 "Add back the solana_sdk::transaction_context module" where:
- First the directory was removed from the repo
- Then added back as a module re-exporting from crates.io

The patching script doesn't handle this properly and creates invalid syntax (`solana-transaction-context = "="`) in Cargo.toml files.

### 2. Private function issue
After fixing the first issue, encountered another error:
```
error[E0603]: function `is_nonoverlapping` is private
  --> programs/bpf_loader/src/syscalls/mod.rs:44:28
   |
44 |     solana_program_memory::is_nonoverlapping,
   |                            ^^^^^^^^^^^^^^^^^ private function
```

This is related to PR #15 "make is_nonoverlapping private and remove num-traits dep from program-memory", where a previously public function was made private.

### 3. Type mismatch issue
After trying to fix the private function by making it public again:
```
error[E0308]: arguments to this function are incorrect
  --> programs/bpf_loader/src/syscalls/mem_ops.rs:31:13
   |
31 |         if !is_nonoverlapping(src_addr, n, dst_addr, n) {
   |             ^^^^^^^^^^^^^^^^^ --------  -  --------  - expected `usize`, found `u64`
```

The function signature was changed to use `usize` instead of `u64`.

## Workaround
Downgrading to Agave v2.2.4 works with Solana SDK v2.2.1, but not with v2.2.2.
It appears that either:
- Agave versions > 2.2.4 OR
- Solana SDK versions > v2.2.1
break compatibility when used together with the patching script.

## Related Documentation
[...]

PR #133, opened 2025-04-08
scripts: Fix downstream patch script

joncinque:
#### Problem

As noted in #121, the patching script in this repo doesn't work on Agave v2.2, because solana-transaction-context was moved back into the Agave repo, but the script is still trying to patch it.

#### Summary of changes

Remove transaction-context from one of the scripts to patch. This is one part of fixing the issue.


yihau (approved):
🫡

PR #134, opened 2025-04-08
instruction: Add syscalls helpers

febo:
### Problem

Currently the syscalls module on the `instruction` crate is behind a `#[cfg(target_os = "solana")]` attribute. Therefore, they are not "visible" on non-solana environment. This makes it difficult to write code that depends on them. Additionally, the crate does not include helpers to wrap the syscalls into friendlier functions.

### Solution

Add a `syscalls` feature to the crate to conditionally include the syscalls module and add wrapper functions for the syscalls. This PR also update the `solana-program` crate to use these helpers to avoid duplication of logic.

### DIFF
--- a/instruction/src/lib.rs
+++ b/instruction/src/lib.rs
@@ -24,7 +24,7 @@ pub mod account_meta;
 #[cfg(feature = "std")]
 pub use account_meta::AccountMeta;
 pub mod error;
-#[cfg(target_os = "solana")]
+#[cfg(feature = "syscalls")]
 pub mod syscalls;
 #[cfg(all(feature = "std", target_arch = "wasm32"))]
 pub mod wasm;
--- a/instruction/src/lib.rs
+++ b/instruction/src/lib.rs
@@ -24,7 +24,7 @@ pub mod account_meta;
 #[cfg(feature = "std")]
 pub use account_meta::AccountMeta;
 pub mod error;
-#[cfg(feature = "syscalls")]
+#[cfg(any(feature = "syscalls", target_os = "solana"))]
 pub mod syscalls;
 #[cfg(all(feature = "std", target_arch = "wasm32"))]
 pub mod wasm;
--- a/instruction/src/syscalls.rs
+++ b/instruction/src/syscalls.rs
@@ -1,8 +1,80 @@
-pub use solana_define_syscall::definitions::sol_get_stack_height;
-use {
+use crate::Instruction;
+#[cfg(target_os = "solana")]
+pub use {
     crate::{AccountMeta, ProcessedSiblingInstruction},
-    solana_define_syscall::define_syscall,
+    solana_define_syscall::{define_syscall, definitions::sol_get_stack_height},
     solana_pubkey::Pubkey,
 };
 
+#[cfg(target_os = "solana")]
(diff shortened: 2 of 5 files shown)


program/src/instruction.rs
```diff
@@ -1,78 +1,12 @@
 pub use {
     crate::message::compiled_instruction::CompiledInstruction,
     solana_instruction::{
-        error::InstructionError, AccountMeta, Instruction, ProcessedSiblingInstruction,
-        TRANSACTION_LEVEL_STACK_HEIGHT,
+        error::InstructionError,
+        syscalls::{get_processed_sibling_instruction, get_stack_height},
```
joncinque: The new functions change the behavior since they don't use the program-stubs, so there should still be the old functions which calls the new one for solana builds, and fall back to the program stubs for non-solana builds.

instruction/src/lib.rs
```diff
@@ -24,7 +24,7 @@ pub mod account_meta;
 #[cfg(feature = "std")]
 pub use account_meta::AccountMeta;
 pub mod error;
-#[cfg(target_os = "solana")]
+#[cfg(feature = "syscalls")]
```
joncinque: It's up to you, but I'd recommend avoiding the breaking change in this PR, and we can always create another PR after this that includes the breaking change. What do you think?

instruction/src/syscalls.rs
```diff
@@ -1,8 +1,82 @@
-pub use solana_define_syscall::definitions::sol_get_stack_height;
-use {
+use crate::Instruction;
```
joncinque, line 2: I think this also breaks people that were using the crate without the std feature, so we should additionally gate the `Instruction` usage on `std` and not enable `std` if `syscalls` is enabled

instruction/src/lib.rs
febo: Sounds good.

instruction/src/syscalls.rs
febo: Hmmm, not sure if I fully understand this. I think the issue is that we need `std` for the syscall helpers (there is a `Vec` there). It also needs the `Instruction` for the helper signature.

I think we need to gate the helpers and `Instruction` behind the feature `syscalls`. So we preserve the current behaviour (no syscall helpers) and when the `syscalls` feature is used, the helpers are present.

program/src/instruction.rs
febo: Good point!

instruction/src/lib.rs
febo: Made the module "syscalls" conditional to any of `target_os = "solana"` and `feature = "syscalls"`, so the current behaviour is maintained but we also get it visible when the `syscalls` feature is explicit enabled.

instruction/src/syscalls.rs
febo: Pushed a change that should maintain the current behaviour, but it will provide the helpers when the `syscalls` feature is enabled.

febo:
> Makes sense to me overall, mostly some bits about avoiding breaking and supporting program-stubs properly

I removed the "breaking" label since it does not contain breaking changes anymore.

instruction/src/syscalls.rs
joncinque: The changes you've made work too, thanks!

PR #135, opened 2025-04-09
Use only rlib as crate type

LucasSte:
**Problem**

We cannot run LTO natively for the SBF target, since the Rust compiler complains that dependencies are declared as both cdylib and rlib crate types.

**Summary of changes**

1. Rewrite documentation to mention the fact that we should use different crate types depending on the Rust package intended usage.
2. Set the crate type as rlib for SDK and `program`, since they are not supposed to be programs for deployment, but instead libraries to be imported.

### DIFF
--- a/program/Cargo.toml
+++ b/program/Cargo.toml
@@ -127,7 +127,7 @@ all-features = true
 rustdoc-args = ["--cfg=docsrs"]
 
 [lib]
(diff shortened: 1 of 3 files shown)


joncinque:
I believe this can be closed because the rlib changes were made in https://github.com/anza-xyz/solana-sdk/pull/138

PR #136, opened 2025-04-11
CI: Add a comment if a PR touches sysvars

joncinque:
#### Problem

As noted at
https://github.com/anza-xyz/agave/pull/428#issuecomment-2795213465, it's possible to update the Rust client and forget to update the JS client.

#### Summary of changes

Copy the solution at https://github.com/orgs/community/discussions/69944#discussioncomment-10796276, but adapt it to our needs.

~~NOTE: I'll try updating some paths to see if this works.~~ Testing failed, oh well

### DIFF
--- a/.github/workflows/sysvar-comment.yml
+++ b/.github/workflows/sysvar-comment.yml
@@ -2,8 +2,8 @@ name: Sysvar Auto-Comment
 
 on:
   pull_request:
-    types:
-      - opened
+    #types:
+    #  - opened
     paths:
       - 'clock/src/**'
       - 'epoch-rewards/src/**'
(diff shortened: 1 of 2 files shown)


joncinque:
Taking a look at https://github.com/marketplace/actions/comment-pull-request#permissions, I don't think I'll be able to properly test this, but I think the perms will be correct once we land this. I can open a test PR after this

steveluscher (approved):
I think that `synchronized` is the event type that you're looking for to test whether this puppy runs on updates, but meh; it probably works.

Thanks!

steveluscher:
But yeah also, forks.

joncinque:
> I think that synchronized is the event type that you're looking for to test whether this puppy runs on updates, but meh; it probably works.

Oh yeah, good call, updated to add `synchronize`, thanks

steveluscher (approved):
image

PR #138, opened 2025-04-11
Create solana-sdk-wasm-js crate

LucasSte:
**Problem**

`solana-program` and `solana-sdk` have both cdylib and rlib declared as crate types. Both crates are supposed to be used as libraries by other projects. This setting is specially problematic for SBF programs that import `solana-program`, since it does not permit us to run link-time optimizations:

```
   Compiling solana-program v2.2.1
error: lto can only be run for executables, cdylibs and static library outputs

error: could not compile `solana-program` (lib) due to 1 previous error
```

The problem lies in using cdylib and rlib together, since that prevents Rust from running LTO, when using Solana-program as a dependency. See https://github.com/rust-lang/rust/issues/51009 for more information.

**Summary of changes**

1. Since the cdylib and the rlib types are only here to be compatible with the wasm target for wasm-pack, I created a new crate `solana-sdk-wasm` to hold all the items necessary for such a target.
2. Use only rlib for crate types in both `solana-program` and `solana-sdk`.
3. Updated the CI jobs accordingly.
4. Update the comments in `program/src/lib.rs` to correctly mention how to use the crate types.

**Basic benchmark**

This is not a comprehensive benchmark. I was just curious on the impact on a small `Hello, Solana!` program, like this:

```rust
entrypoint!(process_instruction);
fn process_instruction(
    _program_id: &Pubkey,
    accounts: &[AccountInfo],
    instruction_data: &[u8],
) -> ProgramResult {
    msg!("Hello, Solana!");
    Ok(())
}
```

These are the results:

No LTO: 18kb (18504 bytes) with 728 CUs consumed.
LTO enabled: 17kb (17576 bytes) with 638 CUs consumed.


program/Cargo.toml
```diff
@@ -122,12 +119,12 @@ solana-pubkey = { workspace = true, features = ["dev-context-only-utils"] }
 solana-sysvar = { workspace = true, features = ["dev-context-only-utils"] }
 
 [package.metadata.docs.rs]
-targets = ["x86_64-unknown-linux-gnu", "wasm32-unknown-unknown"]
```
joncinque: Since there's still wasm code in this crate, probably fine to run the docs build against wasm too

program/src/lib.rs
```diff
@@ -512,7 +517,6 @@ pub mod syscalls;
 pub mod system_instruction;
 pub mod system_program;
 pub mod sysvar;
-pub mod wasm;
```
joncinque, line 492: We shouldn't need to make this breaking change -- the wasm code can still live in `solana-program`, to be used by downstream builds.

program/src/lib.rs
```diff
@@ -567,8 +571,6 @@ pub use solana_stable_layout as stable_layout;
 pub use solana_sysvar::program_stubs;
 #[deprecated(since = "2.2.0", note = "Use `solana-vote-interface` crate instead")]
 pub use solana_vote_interface as vote;
-#[cfg(target_arch = "wasm32")]
-pub use wasm_bindgen::prelude::wasm_bindgen;
```
joncinque, line 532: We should avoid this breaking change too

sdk-wasm/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+console_error_panic_hook = { workspace = true }
+console_log = { workspace = true }
+wasm-bindgen = { workspace = true }
+log = { workspace = true }
+js-sys = { workspace = true }
+
+[lib]
+crate-type = ["cdylib", "rlib"]
```
joncinque: Does this need `rlib` at all?

sdk-wasm/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "solana-sdk-wasm"
+description = "Solana SDK Wasm"
+documentation = "https://docs.rs/solana-sdk"
```
joncinque: nit: since this is really specific to using `wasm-pack` to generate an npm package, let's call out the js part:

```suggestion
name = "solana-sdk-wasm-js"
description = "Solana SDK Wasm JS"
documentation = "https://docs.rs/solana-sdk-wasm-js"
```

sdk-wasm/src/lib.rs
```diff
@@ -2,7 +2,25 @@
...
+        entrypoint,
+        entrypoint_deprecated,
+        example_mocks,
+        feature,
+        hash,
+        program_utils,
+        pubkey,
+        *,
+    },
+};
```
joncinque: This is probably all that we need from the new crate, everything else should live in its original crate.

sdk/Cargo.toml
```diff
@@ -192,12 +190,12 @@ solana-program = { workspace = true, features = ["dev-context-only-utils"] }
 solana-sdk = { path = ".", features = ["dev-context-only-utils"] }
 
 [package.metadata.docs.rs]
-targets = ["x86_64-unknown-linux-gnu", "wasm32-unknown-unknown"]
+targets = ["x86_64-unknown-linux-gnu"]
```
joncinque: We can keep the docs build for wasm

sdk/src/lib.rs
joncinque, line 1: Let's revert all of the changes in the code in this crate

sdk/Cargo.toml
```diff
@@ -176,8 +176,6 @@ thiserror = { workspace = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 getrandom = { version = "0.1.1", features = ["wasm-bindgen"] }
-js-sys = { workspace = true }
-wasm-bindgen = { workspace = true }
```
joncinque, line 114: Huh, it looks like we don't actually need these, so we can keep this change!

sdk-wasm/README.md
```diff
@@ -0,0 +1,16 @@
...
+</p>
+
+# Solana SDK
+
+Use the Solana SDK Crate to be consumed by Javascript libraries. Sse the [Solana Program Crate](https://crates.io/crates/solana-program) instead for
+on-chain programs, and, the [Solana SDK Crate](https://crates.io/crates/solana-sdk) for client-side applications.
+
+More information about Solana is available in the [Solana documentation](https://solana.com/docs).
+
+The [Solana Program Library](https://github.com/solana-labs/solana-program-library) provides examples of how to use this crate.
```
joncinque: nit: May as well call out its new home!
```suggestion
The [Solana Program Organization](https://github.com/solana-program) provides examples of how to use this crate.
```

sdk-wasm/README.md
```diff
@@ -0,0 +1,16 @@
+<p align="center">
+  <a href="https://solana.com">
+    <img alt="Solana" src="https://i.imgur.com/IKyzQ6T.png" width="250" />
+  </a>
+</p>
+
+# Solana SDK
```
joncinque: nit
```suggestion
# Solana SDK WASM JS
```

sdk-wasm/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+Use the Solana SDK Crate to be consumed by Javascript libraries. Sse the [Solana Program Crate](https://crates.io/crates/solana-program) instead for
+on-chain programs, and, the [Solana SDK Crate](https://crates.io/crates/solana-sdk) for client-side applications.
```
joncinque:
```suggestion
Use the Solana SDK WASM JS crate to build a JS package with `wasm-pack` usable in Node or browser environments. Sse the [Solana Program Crate](https://crates.io/crates/solana-program) instead for
on-chain programs, and, the [Solana SDK Crate](https://crates.io/crates/solana-sdk) for client-side applications.
```

joncinque (commented):
The main change introduced in this PR is that people need to run `wasm-pack` against your new package to get a binary that's usable in JS, and that should be the only breakage.

It's as if `solana-sdk` had a binary output and a library output (like [solana-gossip](https://github.com/anza-xyz/agave/blob/6865683ba7f1a600aeb07ecb0391798bcf4b9f9f/gossip/Cargo.toml#L123)), but we're just factoring the binary part into a new crate.

This means we can keep all of the code as is, but just re-export functions from the new crate, along with instructions for building. Does that make sense?

LucasSte:
> This means we can keep all of the code as is, but just re-export functions from the new crate, along with instructions for building. Does that make sense?

I didn't quite get the idea. Can I still maintain the wasm-specifc code in the new crate, but then re-export it in Solana-program and in solana-sdk? Or is it the opposite?

joncinque:
> I didn't quite get the idea. Can I still maintain the wasm-specifc code in the new crate, but then re-export it in Solana-program and in solana-sdk? Or is it the opposite?

Ah, sorry it wasn't clear. The wasm-specific code should still live in the original crate, and the new crate just re-exports everything and can be built into a JS package with `wasm-pack`.

sdk-wasm/Cargo.toml
LucasSte: Should I rename the folder to `sdk-wasm-js` as well?

LucasSte: Ok, I've renamed the folder too.

ilya-bobyr:
> **Problem**
> 
> `solana-program` and `solana-sdk` have both cdylib and rlib declared as crate types. Both crates are supposed to be used as libraries by other projects. This crate setting is specially problematic for SBF programs that import `solana-program`, since it does not permit us to run link-time optimizations:
> 
> ```
>    Compiling solana-program v2.2.1
> error: lto can only be run for executables, cdylibs and static library outputs
> 
> error: could not compile `solana-program` (lib) due to 1 previous error
> ```

Based on the above, it looks like in order to run LTO, you need the `solana-program` crate marked with `cdylib` or `staticlib`(?).
Though, it is unclear if this applies just to the final target type, or also to all the dependencies.
The error says "could not compile `solana-program` (lib)", so it seems that the dependencies need to be one of the specified types as well. 

> **Summary of changes**
> 
>     1. Since the cdylib and the rlib types are only here to be compatible with the wasm target for wasm-pack, I created a new crate `solana-sdk-wasm` to hold all the items necessary for such a target.
> 
>     2. Use only rlib for crate types in both `solana-program` and `solana-sdk`.
> [...]

And yet, later, you are saying that `rlib` should be used for `solana-program`.
And it is in the diff, that `lib.crate-type` is changed from `["cdylib", "rlib"]` to `["rlib"]` (https://github.com/anza-xyz/solana-sdk/pull/138/files#diff-3c9d25a189ead32ebcc2ba81fb9b1a6ff81f14631141bbaf1202e0ddf98fe716R129-R131).

You are saying that you do see a size reduction.  So, I'm probably missing something.
But I wonder if I am not the only one who might be confused here.

Or is it that the type of the dependent crates does not matter, and only the type of the final artifact matters?
If this is the case, then `solana-program` crate type should not be affecting the type of the crate that holds a specific program.
But you are narrowing the `solana-program` crate type.

LucasSte:
@ilya-bobyr The problem is the usage of multiple crate types together. A short answer is that dependencies should be `rlib` and final objects should be cdylib, executables or staticlibs for LTO to run.

The option `RUSTFLAGS= -C lto=fat` is applied to all steps of the compilation, including the dependencies, so rustc will try to run LTO to generate the intermediate objects specified in the dependencies. In the case of Solana-program, rustc will generate an .so file for the crate, since it is specified in the crate types on its toml file. While rlib requires all functions to be present (no LTO needed), cdylib only requires the exported ones, so Rust doesn't really know how to act when you have two conflicting types of crates (this is a known issue in the compiler: https://github.com/rust-lang/rust/issues/51009).

The plan is to support LTO via cargo-build-sbf, in which case we need the correct configuration of crate types. Since most solana programs depend on our own library, setting it to rlib facilitates it.

program/src/lib.rs
```diff
@@ -94,16 +94,21 @@
...
-//! they can be linked to other Rust crates.
+//! Note that a Solana program must specify its crate-type as "cdylib", to
+//! be discovered and built by the `cargo-build-sbf` command as a deployable program.
+//! Solana programs also often have crate-type "rlib" so they can be linked to other Rust crates.
+//! Avoid using "rlib" and "cdylib" crates together, since their combined usage precludes
+//! compiler optimizations that may decrease program size and CU usage.
+//!
+//! Prefer writing a separate package if it is supposed to be used as a library for other Solana
+//! programs (i.e. a "rlib" only crate). When created a Rust project intended to be a program
+//! ready for deployment, ues only the "cdylib" crate type.
```
febo: typo:
```suggestion
//! ready for deployment, use only the "cdylib" crate type.
```

program/src/lib.rs
```diff
@@ -94,16 +94,21 @@
...
-//! build-bpf` command. Solana programs also often have crate-type "rlib" so
-//! they can be linked to other Rust crates.
+//! Note that a Solana program must specify its crate-type as "cdylib", to
+//! be discovered and built by the `cargo-build-sbf` command as a deployable program.
+//! Solana programs also often have crate-type "rlib" so they can be linked to other Rust crates.
+//! Avoid using "rlib" and "cdylib" crates together, since their combined usage precludes
+//! compiler optimizations that may decrease program size and CU usage.
+//!
+//! Prefer writing a separate package if it is supposed to be used as a library for other Solana
+//! programs (i.e. a "rlib" only crate). When created a Rust project intended to be a program
```
febo: Wondering if we can expand a bit this to add something along the lines:

```rust
//! This would be normally the case for defining account types and helpers that are used by both clients and program.
```

febo: possible typo:
```suggestion
//! programs (i.e. a "rlib" only crate). When creating a Rust project intended to be a program
```

program/src/lib.rs
LucasSte: Fixed in https://github.com/anza-xyz/solana-sdk/pull/138/commits/7c469054cb4d49540c9a1499b19aa3f03da02dc6

program/src/lib.rs
LucasSte: Fixed in https://github.com/anza-xyz/solana-sdk/pull/138/commits/7c469054cb4d49540c9a1499b19aa3f03da02dc6

LucasSte: Fixed in https://github.com/anza-xyz/solana-sdk/pull/138/commits/7c469054cb4d49540c9a1499b19aa3f03da02dc6

LucasSte:
As per the discussion on Discord (https://discord.com/channels/428295358100013066/476811830145318912/1370445804078235908), I reverted the PR back to its original form. Let me know if there is any other feedback to be addressed.

program/Cargo.toml
```diff
@@ -109,10 +109,7 @@ arbitrary = { workspace = true, features = ["derive"] }
 solana-logger = { workspace = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-console_error_panic_hook = { workspace = true }
-console_log = { workspace = true }
 getrandom = { workspace = true, features = ["js", "wasm-bindgen"] }
```
joncinque: Is this still needed? I thought we'd only need it in the sdk-wasm-js crate

sdk-wasm-js/src/lib.rs
```diff
@@ -2,7 +2,25 @@
 #![cfg(target_arch = "wasm32")]
 #[deprecated(since = "2.2.0", note = "Use solana_instruction::wasm instead.")]
 pub use solana_instruction::wasm as instructions;
-use wasm_bindgen::prelude::*;
+use {::log::Level, wasm_bindgen::prelude::*};
```
joncinque, line 5: Is there a reason to use the initial `::` on the log import?
```suggestion
use {log::Level, wasm_bindgen::prelude::*};
```

sdk-wasm-js/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "solana-sdk-wasm-js"
+description = "Solana SDK Wasm JS"
+documentation = "https://docs.rs/solana-sdk-wasm-js"
+version = "2.2.2"
```
joncinque: Let's just set this to 1.0.0 to start

sdk/Cargo.toml
```diff
@@ -197,7 +195,7 @@ all-features = true
 rustdoc-args = ["--cfg=docsrs"]
 
 [lib]
-crate-type = ["cdylib", "rlib"]
+crate-type = ["rlib"]
```
joncinque: :partying_face:

sdk-wasm-js/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+# Solana SDK WASM JS
+
+Use the Solana SDK WASM JS crate to build a JS package with `wasm-pack` usable in Node or browser environments.
+See the [Solana Program Crate](https://crates.io/crates/solana-program) instead for on-chain programs, and, the [Solana SDK Crate](https://crates.io/crates/solana-sdk) for client-side
+applications.
+
+More information about Solana is available in the [Solana documentation](https://solana.com/docs).
+
+The [Solana Program Organization](https://github.com/solana-program) provides examples of how to use this crate.
```
joncinque: Does it? I can't think of any places where we use the wasm crate :thinking:

joncinque (commented):
Looks good overall, and will be an important first step to eventually extracting the wasm-bindgen bits out of the component crates.

sdk-wasm-js/src/lib.rs
LucasSte:
```
error[E0432]: unresolved import `log::Level`
 --> sdk-wasm-js/src/lib.rs:5:6
  |
5 | use {log::Level, wasm_bindgen::prelude::*};
  |      ^^^^^^^^^^ no `Level` in `log`
  |
  = help: consider importing this enum instead:
          ::log::Level

error[E0659]: `log` is ambiguous
 --> sdk-wasm-js/src/lib.rs:5:6
  |
5 | use {log::Level, wasm_bindgen::prelude::*};
  |      ^^^ ambiguous name
  |
  = note: ambiguous because of a conflict between a name from a glob import and an outer scope during import or macro resolution
  = note: `log` could refer to a crate passed with `--extern`
  = help: use `::log` to refer to this crate unambiguously
  ```

program/Cargo.toml
LucasSte: I did have a problem before with `getrandom` not being compatible with wasm, since it does not implement the random number generation for wasm unless we explicitly pass on the feature. I wasn't able to replicate the issue right now, so I'll remove the import.

sdk-wasm-js/src/lib.rs
LucasSte: Both the `log` crate and the `console_log` crate are dependencies.

sdk-wasm-js/README.md
LucasSte: That was just copy and paste. I removed that in https://github.com/anza-xyz/solana-sdk/pull/138/commits/fa316d5d7051f7d9bdefc33954e564588d322ce5.

sdk-wasm-js/Cargo.toml
LucasSte: Fixed in https://github.com/anza-xyz/solana-sdk/pull/138/commits/fa316d5d7051f7d9bdefc33954e564588d322ce5.

program/Cargo.toml
LucasSte: I removed the dependency in https://github.com/anza-xyz/solana-sdk/pull/138/commits/dcab0211ebc601cdd0658f674b7c0e6ccaca4603. Changes got eaten up during rebase.

sdk-wasm-js/src/lib.rs
joncinque: Ah right, makes sense, thanks!

joncinque (commented):
Looks good to me! Just needs to fix the merge conflict, then we can land this

LucasSte:
> Looks good to me! Just needs to fix the merge conflict, then we can land this

Thanks! I've just rebased the PR.

joncinque (approved):
Great work, thanks!

PR #140, opened 2025-04-15
vote-interface: move vote state v3 to new module

jstarry:
Before adding vote state v4, v3 should be moved to a separate module


vote-interface/src/state/vote_state_v3.rs
```diff
@@ -0,0 +1,757 @@
...
+    // However, it can be arbitrary slot, when being used inside Tower
+    pub root_slot: Option<Slot>,
+
+    /// the signer for vote transactions
+    pub authorized_voters: AuthorizedVoters,
+
+    /// history of prior authorized voters and the epochs for which
+    /// they were set, the bottom end of the range is inclusive,
+    /// the top of the range is exclusive
+    pub prior_voters: CircBuf<(Pubkey, Epoch, Epoch)>,
```
jstarry, line 54: The only other change besides moving code around was change the vis of these two fields from private to public.

AshwinSekar (approved):

PR #141, opened 2025-04-16
CI: Fix miri job to actually run tests

joncinque:
#### Problem

The miri job is assuming that the `hash` module and the `account_info` module in `solana-program` have tests, but those were moved into `solana-hash` and `solana-account-info` awhile ago.

#### Summary of changes

Specify the correct packages to test.


yihau (approved):
🫡

PR #143, opened 2025-04-21
program-entrypoint: Export used types by no_alloc macro

febo:
### Problem

Currently, the `entrypoint_no_alloc` macro uses `AccountInfo` and `Pubkey` types, which requires the client code to make them available to avoid compilation errors.

image

### Solution

Export `AccountInfo` and `Pubkey` types and qualified their use in the macro with `$crate::`.

### DIFF
--- a/program-entrypoint/src/lib.rs
+++ b/program-entrypoint/src/lib.rs
@@ -7,6 +7,8 @@
 extern crate alloc;
 use {
     alloc::vec::Vec,
+    solana_account_info::AccountInfo,
+    solana_pubkey::Pubkey,
     std::{
         alloc::Layout,
         cell::RefCell,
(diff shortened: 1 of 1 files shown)


program-entrypoint/src/lib.rs
```diff
@@ -20,8 +18,8 @@ use {
 };
 // need to re-export msg for custom_heap_default macro
 pub use {
-    solana_account_info::MAX_PERMITTED_DATA_INCREASE, solana_msg::msg as __msg,
-    solana_program_error::ProgramResult,
+    solana_account_info::AccountInfo, solana_account_info::MAX_PERMITTED_DATA_INCREASE,
+    solana_msg::msg as __msg, solana_program_error::ProgramResult, solana_pubkey::Pubkey,
```
joncinque: nit: I don't feel too strongly about this, but there's an inconsistency here since we're re-exporting `msg` as `__msg` to avoid outside usage -- should we do the same for the new types? ie. `AccountInfo` -> `__AccountInfo` and `Pubkey` -> `__Pubkey`?

joncinque (dismissed):
Feel free to merge as is, just a nit that I noticed

joncinque (approved):

PR #144, opened 2025-04-21
security: Add policy

joncinque:
#### Problem

The SDK is used by Agave, and could potentially contain security issues, but there is not security policy in the repo.

#### Summary of changes

Similar to the Agave security policy, add one for the SDK. The bounty bit links to Agave's bounties.


t-nelson (approved):
seems like a good start. thanks!

PR #145, opened 2025-04-21
CI: Add scripts for GHSA management

joncinque:
#### Problem

The sdk has security advisories enabled, but there's no job to automatically add the security incident teams to them.

#### Summary of changes

Copy the scripts from the agave repo and rename `agave` to `solana-sdk` everywhere.

This will require `secrets.GHSA_ADD_SECURITY_INCIDENT_RESPONSE`, and I'm not sure if that's a repo secret or an org secret -- it might be easier to make it an org secret if it isn't setup that way already.


yihau (approved):
🪖

Issue #148, opened 2025-04-26
Package native-token does not convert SOL amounts to lamports correctly


svenski123:
The function solana_native_token::sol_to_lamports(f64) -> u64 is widely used in CLI tools and elsewhere to convert double precision floating point SOL amounts to unsigned sixty four bit integer lamport amounts.

The function multiplies the SOL amount by 1e9 and converts to an unsigned integer by truncation which returns an incorrect value surprisingly often, for example in the range 0.1289 to 0.1291 SOL, approximately 23.16% of the unique lamport values are not converted correctly, including 0.1289, 0.12893, 0.12896, 0.12904, 0.12907 and 0.1291. Other "simple" SOL values that aren't converted correctly include 4.1, 8.2, 16.4 and 16.9.

The function is also unsafe as casting negative floating point values to unsigned integers is undefined, as is casting values exceeding the maximum value representable by a u64 and casting NaNs.

This issue relates to the following issue raised in the agave repo.

https://github.com/anza-xyz/agave/issues/5930

joncinque:
Closed by #149

PR #149, opened 2025-04-26
sol_to_lamports(): convert negative and NaN to 0, clip positive value…

svenski123:
Handle NaNs, negative and large SOL values safely.
Round lamports to integer rather than truncate.
Add test cases.
Addresses issue #148.

### DIFF
--- a/native-token/src/lib.rs
+++ b/native-token/src/lib.rs
@@ -4,15 +4,25 @@
 
 /// There are 10^9 lamports in one SOL
 pub const LAMPORTS_PER_SOL: u64 = 1_000_000_000;
+const LAMPORTS_PER_SOL_F64: f64 = LAMPORTS_PER_SOL as f64;
 
 /// Approximately convert fractional native tokens (lamports) into native tokens (SOL)
 pub fn lamports_to_sol(lamports: u64) -> f64 {
-    lamports as f64 / LAMPORTS_PER_SOL as f64
+    lamports as f64 / LAMPORTS_PER_SOL_F64
 }
 
 /// Approximately convert native tokens (SOL) into fractional native tokens (lamports)
 pub fn sol_to_lamports(sol: f64) -> u64 {
-    (sol * LAMPORTS_PER_SOL as f64) as u64
+    const U64_MAX_PLUS_1: f64 = 18446744073709551616.; // 0x1p64
+    let lamports = (sol * LAMPORTS_PER_SOL_F64).round();
+    if lamports <= 0.0 {
+        0
+    } else if lamports >= U64_MAX_PLUS_1 {
+        u64::MAX
+    } else {
+        // NaNs fallthrough to here and cast to zero
+        lamports as u64
+    }
 }
 
 use std::fmt::{Debug, Display, Formatter, Result};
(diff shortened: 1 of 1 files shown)


svenski123:
Removed is_nan() relying on u64 cast forcing NaNs to zero; unit test checks this.

svenski123:
@joncinque @diman-io This solana-sdk PR fixing sol_to_lamports() will also fix the off-by-1 lamport bug in SOL amount parsing of CLI tools.

joncinque:
@rustopian I tagged you to review, can you take a look?

The best option would be to deprecate this function entirely, since we shouldn't be in the business of floating point numbers, but that will cause a lot of churn, and it's easier to make this less worse in the short term

native-token/src/lib.rs
rustopian, line 1: Good coverage and defensive handling. Agreed with everyone that long term we should eliminate floats entirely, but this is robust in the interim.

rustopian (approved):
Good coverage and defensive handling. I'm unable to cause any undesired behavior (other than expected tiny lamport drifts). I'd rather see floor(), and tests need clippy attributes to pass checks, but otherwise a robust interim solution until we can (someday) get rid of FPs

native-token/src/lib.rs
rustopian: Typical practice is to floor() so you don't mint/send more than you ask for, any reason to choose not to floor @svenski123 ?

native-token/src/lib.rs
```diff
@@ -4,15 +4,25 @@
...
 pub fn lamports_to_sol(lamports: u64) -> f64 {
-    lamports as f64 / LAMPORTS_PER_SOL as f64
+    lamports as f64 / LAMPORTS_PER_SOL_F64
 }
 
 /// Approximately convert native tokens (SOL) into fractional native tokens (lamports)
 pub fn sol_to_lamports(sol: f64) -> u64 {
-    (sol * LAMPORTS_PER_SOL as f64) as u64
+    const U64_MAX_PLUS_1: f64 = 18446744073709551616.; // 0x1p64
+    let lamports = (sol * LAMPORTS_PER_SOL_F64).round();
```
rustopian: Typical practice is to floor() so you don't mint/send more than you ask for (and maybe erroring if you're thus exceeding available balance), any reason to choose not to floor @svenski123 ?

native-token/src/lib.rs
```diff
@@ -40,3 +50,74 @@ impl Debug for Sol {
...
+        assert_eq!(0.00001, lamports_to_sol(10000));
+        assert_eq!(0.0001, lamports_to_sol(100000));
+        assert_eq!(0.001, lamports_to_sol(1000000));
+        assert_eq!(0.01, lamports_to_sol(10000000));
+        assert_eq!(0.1, lamports_to_sol(100000000));
+        assert_eq!(1., lamports_to_sol(1000000000));
+        assert_eq!(4.1, lamports_to_sol(4_100_000_000));
+        assert_eq!(8.2, lamports_to_sol(8_200_000_000));
+        assert_eq!(8.50228288, lamports_to_sol(8_502_282_880));
+        assert_eq!(18446744073.70955276489257812500, lamports_to_sol(u64::MAX));
```
rustopian, line 68: consider
#[allow(clippy::excessive_precision)]
for the lines like this or checks will fail
@svenski123

rustopian (changes requested):
Need to consider floor() and make clippy happy in the test block.

native-token/src/lib.rs
svenski123: floor() leads to minting/sending strictly less than you ask for quite often i.e. 4.1*1e9 == 4099999999.999999523162841796875 exactly so flooring will get you one lamport less than you asked for.

Rounding allows the identity X == sol_to_lamports(lamports_to_sol(X)) to hold up to roughly 1 million SOL given the precision limits of double precision floating point numbers.

All non-test uses of sol_to_lamports() in agave are to convert floating point values that have parsed from strings (i.e. command line arguments) or to convert SOL constants (500, 1, 0.5) in ledger tool , genesis and genesis utils.

I have not come across any floating point calculations (i.e. such as epoch inflationary rewards, etc.) that are converted to integer lamports using sol_to_lamports().

native-token/src/lib.rs
svenski123: Putting that attribute on the floating point literals causes `error[E0658]: attributes on expressions are experimental` errors so have put them on the test functions instead.

svenski123:
There is no ISO standard for Rust as far as I am aware however the Rust Reference specifies the following logic for float to int type casts which includes NaN to zero handling and mirrors what I had explicitly so I have changed the PR to rely on language level type casting.

Should the Rust language change in future such that this no longer holds, the unit tests should fail.

[[expr.as.numeric.float-as-int]](https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.as.numeric.float-as-int)

```
Casting from a float to an integer will round the float towards zero

NaN will return 0
Values larger than the maximum integer value, including INFINITY, will saturate to the maximum value of the integer type.
Values smaller than the minimum integer value, including NEG_INFINITY, will saturate to the minimum value of the integer type.

assert_eq!(42.9f32 as i32, 42);
assert_eq!(-42.9f32 as i32, -42);
assert_eq!(42_000_000f32 as i32, 42_000_000);
assert_eq!(std::f32::NAN as i32, 0);
assert_eq!(1_000_000_000_000_000f32 as i32, 0x7fffffffi32);
assert_eq!(std::f32::NEG_INFINITY as i32, -0x80000000i32);
```

native-token/src/lib.rs
rustopian: Got it. I'm used to absolute under-mint safety, but in this context and given the limited uses of sol_to_lamports() your reasoning makes sense.

joncinque (approved):
Thanks for your contribution!

diman-io:
I see that I’m late, but I’m opposed to using float arithmetic in finance (in the way it’s implemented in “regular” programming). So I would actually exclude these two conversion functions from the SDK and leave their implementation to the end “users”. Or implement a type with integer math for lamports. Both round and truncate have their side effects — which is bad, in my opinion. My preference for truncate is rather because its effect is consistent (i.e., the number can only decrease).
This might explain why I specifically patched the parsing.

Nevertheless, if we look from the standpoint of keeping float arithmetic, I’m impressed with how it’s currently fixed. The only thing is, I might be in favor of adding a check that the tail after lamports (i.e., starting from the 10th digit after the decimal point) contains a run of X nines or zeros, to guarantee that rounding is truly just a patch for how float is stored in computer memory.

rustopian:
@diman-io agreed on all points.

We're gradually moving away from float math everywhere, balancing patches and deprecations along the way. Changes like yours for `lamports_from_sol` are right in line with that.

([4988](https://github.com/anza-xyz/agave/pull/4988), for any lurkers)

svenski123:
It's not so much the use of binary floating point arithmetic in finance but rather accounting that is problematic, as it cannot represent fractions exactly other than those that are sums of powers of two. It's one thing when computing risk values, greeks, discounting factors, theoretical values, etc. and another when computing actual numbers of lamports to transfer, quote, etc. As far as the core SDK goes, ideally there should be only integer representations of lamports not SOL, and any libraries providing SOL/lamport and floating point convenience conversions should be separate and higher level - i.e. opt-in but not required. IEEE-754 floating point numbers are common enough (and with subtle pitfalls) that it is worthwhile providing some sort of support for them with well specifed sane defaults but again not as part of the core SDK. 

Regarding truncation/floor as opposed to rounding, ceil or something else there is no "safe" choice but rather it is context dependent. Round to even rounding is good for unbiased numerical stability, ceil is conservative for computing offer prices as you favour selling at a higher price, etc. If utility functions implementing this conversion are to be provided, they should allow  for these variations.

Parsing decimal strings directly to uint64_t lamports should be implemented correctly and efficiently. It is understandable why system library parsing routines were used (i.e. strtod() and Rust equivalents) - they are well tested, efficient and handle outlier cases and localisation. Consideration should be given to this in specifying a SOL parsing function (i.e. should '_' characters allowed as spacers (or even spaces themselves), what about use of '.' versus ',' as decimal point in certain locales, etc.)

svenski123:
Regarding converting certain binary floating point numbers to lamports, the original string is not available (as it has already been converted to an IEEE-754 double precision floating point value) and so you can't look at '0' or '9' in the tenth decimal place or later.

However considering the passed in float as computed from <lamports> / 1000000000, then multiplying the number by 1000000000, the result will be a number in the form X = <MANTISSA> * 2**<EXP> closest to the actual decimal. This means that X will either be equal to the decimal number exactly or it will be slightly above or slightly below.

The next higher representable float is obtained by adding 2**<EXP> to the float. If you then truncate the float and the next higher float, you will then obtain two numbers - if the float is equal to or slightly higher than the original decimal then both numbers truncated will be equal however if the float is slightly less than the original decimal then it will truncate to the integer below whereas the next higher float will truncate to the integer above.

In this way you can identify floats that are the closest representation to the integer value - the C code below shows this more clearly. Only floats computed as an integer divided by 1e9 (i.e. parsed from strings) will return true, others will not.

This will only work for SOL*1e9 floating point values up to 2251799813685247 for which the epsilon is 0.25; from 2251799813685248 and higher the episilon is 0.5 and off-by-one (or more the bigger the float gets) lamport errors will arise.

I am not convinced that it would be beneficial to introduce this however it is a way of introducing a strict and sparse mapping of floating point SOL values to integer lamports that rejects floats in between and floats so large that the precision is coarser than one lamport.

```
#include <math.h>
#include <stdbool.h>
inline bool sol2lam(double s, uint64_t *l) {
  const double EPSILON = 0.5;
  const double d = s*1e9;
  const double e = nextafter(d, INFINITY);
  const bool ok = e - d < EPSILON;
  const double f = ok && trunc(d) < trunc(e) ? e : d;
  const double g = f/1e9;
  if (l) *l = f;
  return ok && s == g;
}
```

PR #151, opened 2025-04-30
Deprecate re-exports more loudly

joncinque:
#### Problem

There are many different crates and types that we would like to remove from the sdk, but they are still re-exported from solana-sdk and solana-program. The re-exports have a `#[deprecated(...)]` attribute, but they aren't flagged to downstream users.

#### Summary of changes

Make the re-export deprecations louder by creating a deprecated module and re-exporting from in there.

These are the types affected

* `DecodeError`: this trait isn't all that useful, and is set to be removed entirely with #104, so deprecate it here

These are the crates to no longer re-export from solana-program because they will be moved to program-specific repos:

* `address_lookup_table` -> `solana_address_lookup_table_interface`
* `bpf_loader_upgradeable` -> `solana_loader_v3_interface`
* `loader_upgradeable_instruction` -> `solana_loader_v3_interface::instruction`
* `loader_v4` -> `solana_loader_v4_interface`
* `loader_v4_instruction` -> `solana_loader_v4_interface::instruction`
* `nonce` -> `solana_nonce`
* `feature` -> `solana_feature_gate_interface`
* `loader_instruction` -> `solana_loader_v2_interface`
* `vote` -> `solana_vote_interface`
* `stake` -> `solana_stake_interface`
* `stake_history` -> `solana_stake_interface::stake_history`
* `system_instruction` -> `solana_system_interface`
* `system_program` -> `solana_sdk_ids::system_program` (not sure about this one)

Separately, `solana_message` and `solana_sanitize` are no longer re-exported. They never quite fit in `solana_program`, and caused more annoyance than anything else. The re-exports remain in `solana_sdk`.

`solana-decode-error` will be removed entirely, so its deprecation is louder too.

For `solana-sdk`, there are many types that just don't belong, so stop re-exporting these:

* `commitment_config` -> `solana_commitment_config`
* `genesis_config` -> `solana_genesis_config`
* `hard_forks` -> `solana_hard_forks`
* `rent_collector` -> `solana_rent_collector`
* `alt_bn128` -> `solana_bn254`
* `client` -> `solana_client_traits`
* `compute_budget` -> `solana_compute_budget_interface`
* `derivation_path` -> `solana_derivation_path`
* `ed25519_instruction` -> `solana_ed25519_program`
* `nonce_account` -> `solana_nonce_account`
* `packet` -> `solana_packet`
* `poh_config` -> `solana_poh_config`
* `quic` -> `solana_quic_definitions`
* `rent_debits` -> `solana_rent_debits`
* `secp256k1_instruction` -> `solana_secp256k1_program`
* `secp256k1_recover` -> `solana_secp256k1_recover`
* `system_transaction` -> `solana_system_transaction`
* `exit` -> `solana_validator_exit`

I ran this branch against agave, and there won't be too many changes required. Let me know if you want to add or remove any other re-exports, these seemed like the most sensible.

### DIFF
--- a/decode-error/src/lib.rs
+++ b/decode-error/src/lib.rs
@@ -21,6 +21,7 @@ use num_traits::FromPrimitive;
 /// [`ProgramError`]: https://docs.rs/solana-program-error/latest/solana_program_error/enum.ProgramError.html
 /// [`ProgramError::Custom`]: https://docs.rs/solana-program-error/latest/solana_program_error/enum.ProgramError.html#variant.Custom
 /// [`ToPrimitive`]: num_traits::ToPrimitive
+#[deprecated(since = "2.3.0", note = "Use `num_traits::FromPrimitive` instead")]
 pub trait DecodeError<E> {
     fn decode_custom_error_to_enum(custom: u32) -> Option<E>
     where
@@ -32,6 +33,7 @@ pub trait DecodeError<E> {
 }
 
 #[cfg(test)]
+#[allow(deprecated)]
 mod tests {
     use {super::*, num_derive::FromPrimitive};
 
--- a/program/src/loader_v4.rs
+++ b/program/src/loader_v4.rs
@@ -1,14 +1,14 @@
-#[deprecated(since = "2.2.0", note = "Use solana-loader-v4-interface instead")]
-pub use solana_loader_v4_interface::{
-    instruction::{
-        create_buffer, deploy, deploy_from_source, finalize, is_deploy_instruction,
-        is_finalize_instruction, is_retract_instruction, is_set_program_length_instruction,
-        is_set_program_length_instruction as is_truncate_instruction,
-        is_transfer_authority_instruction, is_write_instruction, retract,
-        set_program_length as truncate, set_program_length as truncate_uninitialized,
-        set_program_length, transfer_authority, write,
+#![deprecated(since = "2.3.0", note = "Use solana-loader-v4-interface instead")]
+pub use {
+    solana_loader_v4_interface::{
+        instruction::{
+            create_buffer, deploy, deploy_from_source, finalize, is_deploy_instruction,
+            is_finalize_instruction, is_retract_instruction, is_set_program_length_instruction,
+            is_transfer_authority_instruction, is_write_instruction, retract,
+            set_program_length as truncate, transfer_authority, write,
+        },
+        state::{LoaderV4State, LoaderV4Status},
+        DEPLOYMENT_COOLDOWN_IN_SLOTS,
     },
-    state::{LoaderV4State, LoaderV4Status},
-    DEPLOYMENT_COOLDOWN_IN_SLOTS,
+    solana_sdk_ids::loader_v4::{check_id, id, ID},
 };
-pub use solana_sdk_ids::loader_v4::{check_id, id, ID};
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -35,26 +35,107 @@
 extern crate self as solana_sdk;
 
 #[cfg(feature = "full")]
-pub use solana_commitment_config as commitment_config;
+#[deprecated(since = "2.3.0", note = "Use `solana-commitment-config` crate instead")]
+pub mod commitment_config {
+    pub use solana_commitment_config::*;
+}
 #[cfg(not(target_os = "solana"))]
 pub use solana_program::program_stubs;
-// These solana_program imports could be *-imported, but that causes a bunch of
-// confusing duplication in the docs due to a rustdoc bug. #26211
-#[allow(deprecated)]
-pub use solana_program::sdk_ids;
 #[cfg(target_arch = "wasm32")]
 pub use solana_program::wasm_bindgen;
 pub use solana_program::{
-    account_info, address_lookup_table, big_mod_exp, blake3, bpf_loader, bpf_loader_deprecated,
-    bpf_loader_upgradeable, clock, config, custom_heap_default, custom_panic_default,
-    debug_account_data, declare_deprecated_sysvar_id, declare_sysvar_id, ed25519_program,
-    epoch_rewards, epoch_schedule, fee_calculator, impl_sysvar_get, incinerator, instruction,
-    keccak, lamports, loader_instruction, loader_upgradeable_instruction, loader_v4,
-    loader_v4_instruction, message, msg, native_token, nonce, program, program_error,
-    program_option, program_pack, rent, secp256k1_program, serialize_utils, slot_hashes,
-    slot_history, stable_layout, stake, stake_history, syscalls, system_instruction,
-    system_program, sysvar, unchecked_div_by_const, vote,
+    account_info, big_mod_exp, blake3, bpf_loader, bpf_loader_deprecated, clock, config,
+    custom_heap_default, custom_panic_default, debug_account_data, declare_deprecated_sysvar_id,
+    declare_sysvar_id, ed25519_program, epoch_rewards, epoch_schedule, fee_calculator,
+    impl_sysvar_get, incinerator, instruction, keccak, lamports, msg, native_token, program,
+    program_error, program_option, program_pack, rent, secp256k1_program, serialize_utils,
+    slot_hashes, slot_history, stable_layout, syscalls, sysvar, unchecked_div_by_const,
 };
+#[deprecated(
+    since = "2.3.0",
+    note = "Use `solana-address-lookup-table-interface` crate instead"
+)]
+pub mod address_lookup_table {
+    #[allow(deprecated)]
+    pub use solana_program::address_lookup_table::*;
+}
+#[deprecated(
+    since = "2.3.0",
+    note = "Use `solana-loader-v3-interface` crate instead"
+)]
+pub mod bpf_loader_upgradeable {
+    #[allow(deprecated)]
+    pub use solana_program::bpf_loader_upgradeable::*;
+}
+#[deprecated(
+    since = "2.3.0",
+    note = "Use `solana-loader-v2-interface` crate instead"
+)]
+pub mod loader_instruction {
+    #[allow(deprecated)]
+    pub use solana_program::loader_instruction::*;
+}
+#[deprecated(
+    since = "2.3.0",
+    note = "Use solana_loader_v3_interface::instruction instead"
+)]
+pub mod loader_upgradeable_instruction {
+    #[allow(deprecated)]
+    pub use solana_program::loader_upgradeable_instruction::*;
+}
+#[deprecated(since = "2.3.0", note = "Use solana-loader-v4-interface instead")]
+pub mod loader_v4 {
+    #[allow(deprecated)]
+    pub use solana_program::loader_v4::*;
+}
+#[deprecated(
+    since = "2.3.0",
+    note = "Use solana_loader_v4_interface::instruction instead"
+)]
+pub mod loader_v4_instruction {
+    #[allow(deprecated)]
+    pub use solana_program::loader_v4_instruction::*;
+}
+#[deprecated(since = "2.2.0", note = "Use `solana-message` crate instead")]
+pub use solana_message as message;
+#[deprecated(since = "2.3.0", note = "Use `solana-nonce` crate instead")]
+pub mod nonce {
+    #[allow(deprecated)]
+    pub use solana_program::nonce::*;
+}
+#[deprecated(since = "2.3.0", note = "Use `solana-sdk-ids` crate instead")]
+pub mod sdk_ids {
+    #[allow(deprecated)]
+    pub use solana_program::sdk_ids::*;
+}
+#[deprecated(since = "2.3.0", note = "Use `solana-stake-interface` instead")]
+pub mod stake {
+    #[allow(deprecated)]
+    pub use solana_program::stake::*;
+}
+#[deprecated(
+    since = "2.3.0",
+    note = "Use `solana_stake_interface::stake_history` instead"
+)]
+pub mod stake_history {
+    #[allow(deprecated)]
+    pub use solana_program::stake_history::*;
+}
+#[deprecated(since = "2.3.0", note = "Use `solana_system_interface` crate instead")]
+pub mod system_instruction {
+    #[allow(deprecated)]
+    pub use solana_program::system_instruction::*;
+}
+#[deprecated(since = "2.3.0", note = "Use `solana_sdk_ids::system_program` instead")]
+pub mod system_program {
+    #[allow(deprecated)]
+    pub use solana_program::system_program::*;
+}
+#[deprecated(since = "2.3.0", note = "Use `solana_vote_interface` crate instead")]
+pub mod vote {
+    #[allow(deprecated)]
+    pub use solana_program::vote::*;
+}
 #[cfg(feature = "borsh")]
 pub use solana_program::{borsh, borsh0_10, borsh1};
 #[cfg(feature = "full")]
@@ -63,13 +144,21 @@ pub use solana_signer::signers;
 pub mod entrypoint;
 pub mod entrypoint_deprecated;
 pub mod example_mocks;
+#[deprecated(
+    since = "2.2.0",
+    note = "Use `solana-feature-gate-interface` crate instead"
+)]
 pub mod feature;
 #[cfg(feature = "full")]
-#[deprecated(since = "2.2.0", note = "Use `solana-genesis-config` crate instead")]
-pub use solana_genesis_config as genesis_config;
+#[deprecated(since = "2.3.0", note = "Use `solana-genesis-config` crate instead")]
+pub mod genesis_config {
+    pub use solana_genesis_config::*;
+}
 #[cfg(feature = "full")]
-#[deprecated(since = "2.2.0", note = "Use `solana-hard-forks` crate instead")]
-pub use solana_hard_forks as hard_forks;
+#[deprecated(since = "2.3.0", note = "Use `solana-hard-forks` crate instead")]
+pub mod hard_forks {
+    pub use solana_hard_forks::*;
+}
 pub mod hash;
 pub mod log;
 pub mod native_loader;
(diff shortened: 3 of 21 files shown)


sdk/src/lib.rs
```diff
@@ -35,26 +35,107 @@
...
+    since = "2.3.0",
+    note = "Use solana_loader_v4_interface::instruction instead"
+)]
+pub mod loader_v4_instruction {
+    #[allow(deprecated)]
+    pub use solana_program::loader_v4_instruction::*;
+}
+#[deprecated(since = "2.2.0", note = "Use `solana-message` crate instead")]
+pub use solana_message as message;
+#[deprecated(since = "2.3.0", note = "Use `solana-nonce` crate instead")]
```
t-nelson, line 101:
```suggestion
#[deprecated(since = "2.3.0", note = "Use `solana-system-interface::nonce` crate instead")]
```

half of me wants this. other half wants to leave it separate to ease removal

sdk/src/lib.rs
```diff
@@ -35,26 +35,107 @@
...
+pub mod nonce {
+    #[allow(deprecated)]
+    pub use solana_program::nonce::*;
+}
+#[deprecated(since = "2.3.0", note = "Use `solana-sdk-ids` crate instead")]
```
t-nelson, line 106: this crate feels like a mistake. I'd rather have individual `solana_x_interface::program` than a junk drawer crate

sdk/src/lib.rs
```diff
@@ -35,26 +35,107 @@
...
+pub mod stake_history {
+    #[allow(deprecated)]
+    pub use solana_program::stake_history::*;
+}
+#[deprecated(since = "2.3.0", note = "Use `solana_system_interface` crate instead")]
+pub mod system_instruction {
+    #[allow(deprecated)]
+    pub use solana_program::system_instruction::*;
+}
+#[deprecated(since = "2.3.0", note = "Use `solana_sdk_ids::system_program` instead")]
```
t-nelson:
```suggestion
#[deprecated(since = "2.3.0", note = "Use `solana_system_interface::program` instead")]
```

sdk/src/lib.rs
```diff
@@ -106,24 +197,36 @@ pub use solana_account as account;
...
-#[deprecated(since = "2.1.0", note = "Use `solana-bn254` crate instead")]
-pub use solana_bn254 as alt_bn128;
+#[deprecated(since = "2.3.0", note = "Use `solana-bn254` crate instead")]
+pub mod alt_bn128 {
+    pub use solana_bn254::*;
+}
 #[cfg(feature = "full")]
-#[deprecated(since = "2.2.0", note = "Use `solana-client-traits` crate instead")]
-pub use solana_client_traits as client;
+#[deprecated(since = "2.3.0", note = "Use `solana-client-traits` crate instead")]
```
t-nelson, line 208: `solana-client*` are prime candidates for eventual monorepo extraction imo

sdk/src/lib.rs
```diff
@@ -106,24 +197,36 @@ pub use solana_account as account;
...
-pub use solana_derivation_path as derivation_path;
+pub mod compute_budget {
+    pub use solana_compute_budget_interface::*;
+}
+#[deprecated(since = "2.3.0", note = "Use `solana-decode-error` crate instead")]
+pub mod decode_error {
+    pub use solana_decode_error::*;
+}
+#[deprecated(since = "2.3.0", note = "Use `solana-derivation-path` crate instead")]
+pub mod derivation_path {
```
t-nelson, line 225: this almost certainly should not exist

sdk/src/lib.rs
```diff
@@ -106,24 +197,36 @@ pub use solana_account as account;
...
+    pub use solana_derivation_path::*;
+}
 #[cfg(feature = "full")]
-#[deprecated(since = "2.2.0", note = "Use `solana-ed25519-program` crate instead")]
-pub use solana_ed25519_program as ed25519_instruction;
+#[deprecated(since = "2.3.0", note = "Use `solana-ed25519-program` crate instead")]
```
t-nelson, line 229: why is there no `solana-ed25519-interface`?

sdk/src/lib.rs
```diff
@@ -145,15 +248,21 @@ pub use solana_inflation as inflation;
     note = "Use `solana_message::inner_instruction` instead"
 )]
 pub use solana_message::inner_instruction;
-#[deprecated(since = "2.2.0", note = "Use `solana-nonce-account` crate instead")]
-pub use solana_nonce_account as nonce_account;
+#[deprecated(since = "2.3.0", note = "Use `solana-nonce-account` crate instead")]
+pub mod nonce_account {
```
t-nelson, line 255: wouldn't `solana_nonce::state` be more consistent?

sdk/src/lib.rs
```diff
@@ -145,15 +248,21 @@ pub use solana_inflation as inflation;
...
+}
 #[cfg(feature = "full")]
 #[deprecated(since = "2.2.0", note = "Use `solana-offchain-message` crate instead")]
 pub use solana_offchain_message as offchain_message;
-#[deprecated(since = "2.1.0", note = "Use `solana-packet` crate instead")]
-pub use solana_packet as packet;
-#[deprecated(since = "2.2.0", note = "Use `solana-poh-config` crate instead")]
-pub use solana_poh_config as poh_config;
+#[deprecated(since = "2.3.0", note = "Use `solana-packet` crate instead")]
+pub mod packet {
```
t-nelson, line 262: 100 SOL bounty to whoever kills this

sdk/src/lib.rs
```diff
@@ -174,10 +283,14 @@ pub use solana_program_memory as program_memory;
 /// ```
 pub use solana_pubkey::pubkey;
 #[cfg(feature = "full")]
-#[deprecated(since = "2.2.0", note = "Use `solana-quic-definitions` crate instead")]
-pub use solana_quic_definitions as quic;
-#[deprecated(since = "2.2.0", note = "Use `solana-rent-debits` crate instead")]
-pub use solana_rent_debits as rent_debits;
+#[deprecated(since = "2.3.0", note = "Use `solana-quic-definitions` crate instead")]
+pub mod quic {
```
t-nelson, line 290: this needs to die. it's just agave defaults

sdk/src/lib.rs
```diff
@@ -215,11 +328,15 @@ pub use solana_sdk_macro::declare_deprecated_id;
...
-pub use solana_secp256k1_program as secp256k1_instruction;
-#[deprecated(since = "2.1.0", note = "Use `solana-secp256k1-recover` crate instead")]
-pub use solana_secp256k1_recover as secp256k1_recover;
+pub mod secp256k1_instruction {
+    pub use solana_secp256k1_program::*;
+}
+#[deprecated(since = "2.3.0", note = "Use `solana-secp256k1-recover` crate instead")]
+pub mod secp256k1_recover {
+    pub use solana_secp256k1_recover::*;
+}
```
t-nelson, line 342: these would also be more consistent as `solana_*_interface`?

sdk/src/lib.rs
```diff
@@ -228,10 +345,12 @@ pub use solana_serde_varint as serde_varint;
 pub use solana_short_vec as short_vec;
 #[cfg(feature = "full")]
 #[deprecated(
-    since = "2.2.0",
+    since = "2.3.0",
     note = "Use `solana-system-transaction` crate instead"
 )]
-pub use solana_system_transaction as system_transaction;
+pub mod system_transaction {
```
t-nelson, line 354: worthless helper is worthless :axe:

t-nelson (dismissed):

t-nelson:
there are also a few other things we should kill

`solana_sdk::saturating_add_assign!()` can be replaced by `std::num::Saturating`

`solana_sdk::program_utils::limited_deserialize` needs to go, but i haven't investigated why we changed the interface. i vaguely remember questioning why it exists at all

`solana_sdk::rpc_port` should never have existed

anything that lives exclusively in `solana-sdk` should be gone

apfitzge (dismissed):
`tx-metadata` changes look good to me.

sdk/src/lib.rs
joncinque: It's already in a separate crate for now, but we could certainly merge them later.

sdk/src/lib.rs
joncinque: I completely agree that theoretically we shouldn't need this, but you'd be surprised how many times someone just needs a pubkey and nothing else. We can certainly work towards deprecating it either way.

sdk/src/lib.rs
joncinque: Ah yes, thanks, done!

sdk/src/lib.rs
joncinque: Big time, I can't wait to take those out

sdk/src/lib.rs
joncinque: Haha agreed

sdk/src/lib.rs
joncinque: This one's a bit annoying. The current crate is `solana-ed25519-program`, and it contains the instruction and the verification code. 

However, the verification code has been deprecated in favor of `agave_precompiles`. Once we remove the verification code from `solana-ed25519-program`, we can just rename the crate to `solana-ed25519-interface`. What do you think?

sdk/src/lib.rs
joncinque: This one's also a bit annoying, it's mostly helpers for working with `solana_account` types, and not actually used by the system program implementation, except in tests.

We could potentially add it to `solana_nonce` in the future though

sdk/src/lib.rs
joncinque: :saluting_face:

sdk/src/lib.rs
joncinque: :100:

sdk/src/lib.rs
joncinque: `solana-secp256k1-program` can do the same as `solana-ed25519-program`.

The recover crate is a bit strange, since it's just one syscall and a fallback implementation for non-solana targets. I'd prefer to keep it as is since it has some crypto deps.

joncinque (commented):
Thanks for the review! To address your points:

> solana_sdk::saturating_add_assign!() can be replaced by std::num::Saturating

Done, but it's gonna be loud in agave (145 uses)

> solana_sdk::program_utils::limited_deserialize needs to go, but i haven't investigated why we changed the interface. i vaguely remember questioning why it exists at all

I'll make the deprecation louder, it's totally unused in agave right now

> solana_sdk::rpc_port should never have existed

It's still used unfortunately, but I can add it to `solana-rpc-client-api` or just locally define them wherever they're used in agave.

decode-error/src/lib.rs
```diff
@@ -21,6 +21,7 @@ use num_traits::FromPrimitive;
 /// [`ProgramError`]: https://docs.rs/solana-program-error/latest/solana_program_error/enum.ProgramError.html
 /// [`ProgramError::Custom`]: https://docs.rs/solana-program-error/latest/solana_program_error/enum.ProgramError.html#variant.Custom
 /// [`ToPrimitive`]: num_traits::ToPrimitive
+#[deprecated(since = "2.3.0", note = "Use `num_traits::FromPrimitive` instead")]
```
febo, line 24: Do we use specific functionalities from `num_traits::FromPrimitive`? For the changes in #12 we are using `TryFrom<u32>` and removed the `num_traits` dependency.

joncinque: Probably not, but we're also removing this crate very soon, so I wanted to go with the simplest option

program/src/loader_v4.rs
```diff
@@ -1,14 +1,14 @@
-#[deprecated(since = "2.2.0", note = "Use solana-loader-v4-interface instead")]
-pub use solana_loader_v4_interface::{
-    instruction::{
-        create_buffer, deploy, deploy_from_source, finalize, is_deploy_instruction,
-        is_finalize_instruction, is_retract_instruction, is_set_program_length_instruction,
-        is_set_program_length_instruction as is_truncate_instruction,
-        is_transfer_authority_instruction, is_write_instruction, retract,
-        set_program_length as truncate, set_program_length as truncate_uninitialized,
-        set_program_length, transfer_authority, write,
+#![deprecated(since = "2.3.0", note = "Use solana-loader-v4-interface instead")]
```
febo, line 1: The deprecated note on `pub use` would still not be "loud" right?

sdk/src/lib.rs
t-nelson: could we not get there with features?

sdk/src/lib.rs
t-nelson: i think precompiles were a mistake. proposed solution will do. just annoying how many similar crate naming patterns contain inconsistent symbol paths

sdk/src/lib.rs
t-nelson: ah right. these are actually my fault iirc. could deprecate them entirely in theory. maybe to `{solana_system_interface::,solana_}nonce::account` perhaps featurized to discourage future usage

sdk/src/lib.rs
t-nelson: imo we need to work toward a consistent interface here. absurd that there's a Precompile trait and that there's like zero consistency between the implementations

t-nelson:
> Thanks for the review! To address your points:
> 
> > solana_sdk::saturating_add_assign!() can be replaced by std::num::Saturating
> 
> Done, but it's gonna be loud in agave (145 uses)

luckily i believe a majority of usage is for metrics accumulations

> > solana_sdk::program_utils::limited_deserialize needs to go, but i haven't investigated why we changed the interface. i vaguely remember questioning why it exists at all
> 
> I'll make the deprecation louder, it's totally unused in agave right now

i have diffs that would suggest otherwise

> > solana_sdk::rpc_port should never have existed
> 
> It's still used unfortunately, but I can add it to `solana-rpc-client-api` or just locally define them wherever they're used in agave.

is it used externally to agave? i'd move the whole crate back to monorepo if so. there we can eliminate it with configs and cli arg defaults as god intended

decode-error/src/lib.rs
t-nelson: iirc we use this trait for some dumb explicit i32 <-> enum variant marshaling that's probably unnecessary (blame the phone guys)

program/src/loader_v4.rs
t-nelson: yeah if we want these warnings raised, we actually need to reverse our current path deprecation policy. leave symbols scheduled to move in place with deprecation warning, `pub use` them at the new, suggested location, then move the symbols with removal of the deprecation warning

joncinque: Note the exclamation point -- this goes on the whole module, and is the same as if I had put `#[deprecated(...)]` on `pub mod loader_v4`

sdk/src/lib.rs
joncinque: Maybe, but it does get tricky because features should be additive. If I just want a pubkey from a crate then, all of the other functionality would be enabled on a feature, and that can get gross

sdk/src/lib.rs
joncinque: Since it's already in a separate crate, we could just stop using it downstream, then we can remove the crate entirely

sdk/src/lib.rs
joncinque: Yep, agreed big-time

sdk/src/lib.rs
t-nelson: i guess i was thinking something like...
```toml
[features]
full = [] # everything but the declare_id!() behind this
default = [ "full" ]
```

then if i only want pubkey...
```toml
[dependencies]
solana_crate = { version = "x.y.z", default-features = false }
```

program/src/loader_v4.rs
t-nelson: ah conversation view trimmed context strikes again!

t-nelson:
> Yikes, it's _very used_: https://github.com/search?q=solana_sdk%3A%3Arpc_port&type=code we haven't factored it into a separate crate, but I could do that in a follow-up

tbf, anyone who thought using this was a good idea deserves to be broken

program/src/loader_v4.rs
febo: > Note the exclamation point -- this goes on the whole module, and is the same as if I had put #[deprecated(...)] on pub mod loader_v4

Nice, perfect!

sdk/src/lib.rs
```diff
@@ -35,26 +35,110 @@
...
+}
+#[deprecated(
+    since = "2.3.0",
+    note = "Use solana_loader_v4_interface::instruction instead"
+)]
+pub mod loader_v4_instruction {
+    #[allow(deprecated)]
+    pub use solana_program::loader_v4_instruction::*;
+}
+#[deprecated(since = "2.2.0", note = "Use `solana-message` crate instead")]
```
febo, line 99: We probably need to tweak this one, since you won't be able to put a `!` here – something like:
```rust
#[deprecated(since = "2.2.0", note = "Use `solana-message` crate instead")]
pub mod message {
    pub use solana_message::*;
}
```

sdk/src/lib.rs
```diff
@@ -35,26 +35,110 @@
...
+pub use solana_message as message;
```
febo, line 100: We probably need to tweak this one – something like:
```rust
#[deprecated(since = "2.2.0", note = "Use `solana-message` crate instead")]
pub mod message {
    pub use solana_message::*;
}
```

joncinque: This one is actually on purpose -- my thinking is that `Transaction` / `Message` / `Keypair` / `Signer` are the most used types in `solana-sdk`, and will be the most annoying to remove, and aren't a problem to keep, so we keep it for now.

febo (approved):
Looks good to me.

joncinque:
@apfitzge @t-nelson would you like to take another look?

apfitzge (approved):
lgtm from tx-metadata side

t-nelson (approved):
good start!

PR #153, opened 2025-05-04
[bls] Add bls signature crate

samkim-crypto:
Creating a BLS signature crate in the bls branch of the sdk for downstream projects to use.

### DIFF
--- a/bls/Cargo.toml
+++ b/bls/Cargo.toml
@@ -1,0 +1,39 @@
+[package]
+name = "solana-bls"
+description = "Solana BLS"
+documentation = "https://docs.rs/solana-bls-keypair"
+version = "0.1.0"
+authors = { workspace = true }
+repository = { workspace = true }
+homepage = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+
+[dependencies]
+bytemuck = { workspace = true }
+serde = { workspace = true, features = ["derive"] }
+serde_with = { workspace = true, features = ["macros"] }
+
+[target.'cfg(not(target_os = "solana"))'.dependencies]
+blst = { workspace = true }
+blstrs = { workspace = true }
+ff = { workspace = true }
+group = { workspace = true }
+rand = { workspace = true }
+solana-signature = { workspace = true, optional = true }
+solana-signer = { workspace = true, optional = true }
+subtle = { workspace = true, optional = true }
+thiserror = { workspace = true }
+
+[dev-dependencies]
+bincode = { workspace = true }
+solana-keypair = { workspace = true }
+
+[features]
+solana-signer-derive = ["dep:solana-signer", "dep:solana-signer", "dep:subtle"]
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[lints]
+workspace = true
--- a/bls/Cargo.toml
+++ b/bls/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "solana-bls"
 description = "Solana BLS"
-documentation = "https://docs.rs/solana-bls-keypair"
+documentation = "https://docs.rs/solana-bls"
 version = "0.1.0"
 authors = { workspace = true }
 repository = { workspace = true }
--- a/bls/Cargo.toml
+++ b/bls/Cargo.toml
@@ -37,7 +37,7 @@ solana-signer-derive = [
 ]
 
 [lib]
-crate-type = ["cdylib", "rlib"]
+crate-type = ["rlib"]
 
 [lints]
 workspace = true
--- a/bls/Cargo.toml
+++ b/bls/Cargo.toml
@@ -30,7 +30,12 @@ bincode = { workspace = true }
 solana-keypair = { workspace = true }
 
 [features]
-solana-signer-derive = ["dep:solana-signer", "dep:solana-signer", "dep:subtle"]
+solana-signer-derive = [
+    "dep:solana-signer",
+    "dep:solana-signature",
+    "dep:solana-signer",
+    "dep:subtle",
+]
 
 [lib]
 crate-type = ["cdylib", "rlib"]
--- a/bls/Cargo.toml
+++ b/bls/Cargo.toml
@@ -33,7 +33,6 @@ solana-keypair = { workspace = true }
 solana-signer-derive = [
     "dep:solana-signer",
     "dep:solana-signature",
-    "dep:solana-signer",
     "dep:subtle",
 ]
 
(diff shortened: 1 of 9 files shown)


bls/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name = "solana-bls"
+description = "Solana BLS"
+documentation = "https://docs.rs/solana-bls-keypair"
```
joncinque: nit: this should be `solana-bls`

bls/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+[dev-dependencies]
+bincode = { workspace = true }
+solana-keypair = { workspace = true }
+
+[features]
+solana-signer-derive = [
+    "dep:solana-signer",
+    "dep:solana-signature",
+    "dep:solana-signer",
+    "dep:subtle",
```
joncinque: nit: solana-signer appears twice here

bls/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+]
+
+[lib]
+crate-type = ["cdylib", "rlib"]
```
joncinque: nit: this shouldn't need `cdylib`

bls/src/proof_of_possession.rs
```diff
@@ -0,0 +1,40 @@
...
+
+impl TryFrom<&ProofOfPossession> for ProofOfPossessionProjective {
+    type Error = BlsError;
+
+    fn try_from(proof: &ProofOfPossession) -> Result<Self, Self::Error> {
+        let maybe_uncompressed: Option<G2Affine> = G2Affine::from_uncompressed(&proof.0).into();
+        let uncompressed = maybe_uncompressed.ok_or(BlsError::PointConversion)?;
+        Ok(Self(uncompressed.into()))
+    }
+}
```
joncinque: nit: it might be possible to re-use some code here, by having `TryFrom<ProofOfPossession>` call into `TryFrom<&ProofOfPossession>`

bls/src/pod.rs
```diff
@@ -0,0 +1,224 @@
...
+);
+
+impl Default for Pubkey {
+    fn default() -> Self {
+        Self([0; BLS_PUBLIC_KEY_AFFINE_SIZE])
+    }
+}
+
+// Directly implement the `Pod` and `Zeroable` traits for types that are
+// wrappers around byte arrays.
```
joncinque: nit: maybe this is a bit pedantic, but can you add a clarifying comment saying this is safe *because* they're all byte arrays, and we can't derive it automatically because bytemuck doesn't derive Pod for all sizes of byte arrays?

bls/src/keypair.rs
```diff
@@ -0,0 +1,249 @@
...
+
+impl TryFrom<&Pubkey> for PubkeyProjective {
+    type Error = BlsError;
+
+    fn try_from(proof: &Pubkey) -> Result<Self, Self::Error> {
+        let maybe_uncompressed: Option<G1Affine> = G1Affine::from_uncompressed(&proof.0).into();
+        let uncompressed = maybe_uncompressed.ok_or(BlsError::PointConversion)?;
+        Ok(Self(uncompressed.into()))
+    }
+}
```
joncinque: same here might be nice to re-use the implementation

bls/src/keypair.rs
```diff
@@ -0,0 +1,249 @@
...
+impl Keypair {
+    /// Constructs a new, random `Keypair` using a caller-provided RNG
+    pub fn generate<R>(csprng: &mut R) -> Self
+    where
+        R: CryptoRng + RngCore,
+    {
+        let secret = SecretKey::generate(csprng);
+        let public = PubkeyProjective::from_secret(&secret);
+        Self { secret, public }
+    }
```
joncinque: This unfortunately ties us to a certain version of `rand` in the public interface. Makes me realize we should probably deprecate these functions elsewhere in the sdk, so that people generate their own random bytes and pass them into a constructor.

What do you think?

bls/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+
+/// Domain separation tag used when hashing public keys to G2 in the proof of
+/// possession signing and verification functions. See
+/// https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#section-4.2.3.
+pub const POP_DST: &[u8] = b"BLS_SIG_BLS12381G1_XMD:SHA-256_SSWU_RO_POP_";
+
+#[cfg(not(target_os = "solana"))]
+pub struct Bls;
+#[cfg(not(target_os = "solana"))]
+impl Bls {
```
joncinque, line 47: nit: I typically find it's clearer to avoid a lot of code in `lib.rs`. Since most of these are related to messages, maybe it can get moved to `message.rs`? Or maybe `pairing.rs`? But I don't love either of these, so we can also leave this as is if you prefer

joncinque (commented):
Looks really good! Mostly small things, the crate and the code is pretty understandable.

If you need to publish this to crates.io and iterate, we can also merge the branch into master

bls/Cargo.toml
samkim-crypto: Thanks for the catch!

bls/Cargo.toml
samkim-crypto: Fixed!

bls/Cargo.toml
samkim-crypto: Yes, removed!

bls/src/proof_of_possession.rs
samkim-crypto: Good idea! Fixed!

bls/src/pod.rs
samkim-crypto: Yeah I followed the style in the zk-sdk and updated the comments.

bls/src/keypair.rs
samkim-crypto: Yes good point. I removed the `generate` functions that use the generators directly.

bls/src/lib.rs
samkim-crypto: Yeah I think this is a good idea, but I think I would actually need to think about whether we can remove the struct `Bls` and just let these functions be defined directly in `Keypair`, `Pubkey`, etc. Let me defer this one to a follow-up PR.

joncinque: Ok that works

joncinque (commented):
Looks good to me! Just a clippy issue to resolve, then we can land this

joncinque (approved):

PR #154, opened 2025-05-06
use saturating_sub in is_writable_index() - fix #150

### LINKED ISSUE #150: is_writable_index() may panic when overflow checking is enabled
### Summary
The `is_writable_index()` function of `solana_message::Message` may panic due to integer subtractions with overflow if overflow checking is enabled in Rust. If overflow checking is not enabled, the code may behave unexpectedly, but we're not aware of a specific impact in that case.

### Reporting Context
We first reported this issue confidentially, following the submission policy for potential security issues. According to our current understanding, the integer overflows do not happen for messages which successfully pass the `::Sanitize()` check, which is used by some consumers of the `solana-sdk` crates such as `Agave` before calling the affected function. The Anza team communicated that they therefore see this as an **informational** level issue in the SDK and asked us to report it publicly instead of a GHSA entry.

### Details
The implementation of `is_writable_index()`  for  `solana_message::Message` contains two separate subtractions on unsigned integer types which can overflow (or more precisely: underflow) depending on message values:
https://github.com/anza-xyz/solana-sdk/blob/89251af222d5b350f944859712d2ff0a3a60f37c/message/src/legacy.rs#L646-L654  
Note that while this message format is designated as legacy, it's still common and actively used. The `is_writable_index()` function is reachable through the public `is_maybe_writable()` and the deprecated public `is_writable()` function.

Based on our initial analysis, the newer `V0::Message` type has a different implementation for `is_writable_index()` which contains deliberate `saturating_sub` instructions and is not affected.

### PoC
```Rust
#[test]
fn solana_is_writable_index_reproducer1() {
    use solana_message::Message;
    use solana_message::MessageHeader;

    let message = Message {
        // the overflow happens due to num_readonly_signed_accounts > num_required_signatures
        header: MessageHeader {
            num_required_signatures: 1,
            num_readonly_signed_accounts: 2,
            ..MessageHeader::default()
        },
        ..Message::default()
    };

    // panics with "attempt to subtract with overflow" if Rust overflow-checks are enabled
    message.is_maybe_writable(0, None);
}
```

```Rust
#[test]
fn solana_is_writable_index_reproducer2() {
    use solana_message::Message;
    use solana_message::MessageHeader;

    let message = Message {
        // the overflow happens due to num_readonly_unsigned_accounts > account_keys.len()
        header: MessageHeader {
            num_readonly_unsigned_accounts: 1,
            ..MessageHeader::default()
        },
        ..Message::default()
    };

    // panics with "attempt to subtract with overflow" if Rust overflow-checks are enabled
    message.is_maybe_writable(0, None);
}
```

### Impact
Programs which use the affected functions on crafted messages are susceptible to a denial-of-service issue if runtime overflow checks are enabled. At the time of writing, this is not the case on Rust release profile builds by default. In the absence of panics, there may be other negative effects through logically incorrect results. 

### Affected Crate

`solana-message` in version `<= 2.3.0`. 
Issue observed with crate versions `0.0.2`,`0.0.3`,`2.2.1`, and `2.3.0`.

### Suspected Mitigation
https://github.com/anza-xyz/solana-sdk/blob/5ee135d96a2dc1a5da7a0c2c0b2010a34f3feb88/message/src/legacy.rs#L211-L245

Specifically, it does the following guard checks: https://github.com/anza-xyz/solana-sdk/blob/d705aac6dd13793a5b6f3f6527996d772cc5348e/message/src/legacy.rs#L214-L216 

https://github.com/anza-xyz/solana-sdk/blob/d705aac6dd13793a5b6f3f6527996d772cc5348e/message/src/legacy.rs#L222

### Suggestions

1. Ensure that no integer overflow is possible even on malformed messages, for example by using wrapping subtraction operations
1. Improve library code documentation on the sanitize functionality to encourage or enforce its usage

### Acknowledgements
Discovered by Christian Reitter during internal security reviews at [Turnkey](https://www.turnkey.com/).

rustopian:
legacy: use saturating_sub in is_writable_index()

Fixes #150.

Problem
-------
`is_writable_index()` subtracts two u8 header counters.  
If the caller skips `Message::sanitize()` the subtraction can underflow, panicking in debug builds or silently wrapping to 255 in release, making ≈all accounts appear writable.

Options considered
------------------
* wrapping_sub  
  + Zero-cost.  
  - Fail-open: an unchecked message marks many indices writable.

* saturating_sub (chosen)  
  + Matches the [v0-message implementation](https://github.com/anza-xyz/solana-sdk/blob/0f9989d0f2f6aa5328124084e20f587446bda79b/message/src/versions/v0/loaded.rs#L116-L136).  
  + Fail-closed: underflow clamps to 0, no extra accounts are writable.  
  - Adds a micro-op, but this should cause no performance impact.

Change
------
Replace both unchecked subtractions with `saturating_sub`.
Add a unit test covering scenarios including those demonstrated in the original PoC.

Result
------
Now immune to underflow; behavior of legacy messages here aligns with behavior of v0 messages.

### DIFF
--- a/message/src/legacy.rs
+++ b/message/src/legacy.rs
@@ -1044,7 +1044,7 @@ mod tests {
         };
         assert!(!message1.is_writable_index(0));
 
-        // Matching issue #150 PoC 2 concept - num_readonly_unsigned_accounts > account_keys.len()
+        // Matching issue #150 PoC 2 - num_readonly_unsigned_accounts > account_keys.len()
         let key_for_poc2 = Pubkey::new_unique();
         let message2 = Message {
             header: MessageHeader {
--- a/message/src/legacy.rs
+++ b/message/src/legacy.rs
@@ -646,11 +646,13 @@ impl Message {
     /// Returns true if the account at the specified index was requested to be
     /// writable. This method should not be used directly.
     pub(super) fn is_writable_index(&self, i: usize) -> bool {
-        i < (self.header.num_required_signatures - self.header.num_readonly_signed_accounts)
-            as usize
+        i < (self.header.num_required_signatures as usize)
+            .saturating_sub(self.header.num_readonly_signed_accounts as usize)
             || (i >= self.header.num_required_signatures as usize
-                && i < self.account_keys.len()
-                    - self.header.num_readonly_unsigned_accounts as usize)
+                && i < self
+                    .account_keys
+                    .len()
+                    .saturating_sub(self.header.num_readonly_unsigned_accounts as usize))
     }
 
     /// Returns true if the account at the specified index is writable by the
--- a/message/src/legacy.rs
+++ b/message/src/legacy.rs
@@ -1025,4 +1025,81 @@ mod tests {
     fn test_inline_all_ids() {
         assert_eq!(solana_sysvar::ALL_IDS.to_vec(), ALL_IDS.to_vec());
     }
+
+    #[test]
+    fn test_is_writable_index_saturating_behavior() {
+        // Directly matching issue #150 PoC 1:
+        // num_readonly_signed_accounts > num_required_signatures
+        // This now results in the first part of the OR condition in is_writable_index effectively becoming `i < 0`.
+        let key0 = Pubkey::new_unique();
+        let message1 = Message {
+            header: MessageHeader {
+                num_required_signatures: 1,
+                num_readonly_signed_accounts: 2, // 2 > 1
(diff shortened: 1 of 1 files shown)


joncinque (approved):
The fix and the tests make sense to me, but it does beg the question about whether we should change this interface altogether since it'll just give `false` in broken situations. But maybe that's ok -- I'll let someone from tx-metadata give the final :heavy_check_mark: .

PR #155, opened 2025-05-06
Feature - Adds `UpgradeableLoaderInstruction::ExtendProgramChecked`

Lichtso:
For [SIMD-0164](https://github.com/solana-foundation/solana-improvement-documents/pull/164).

PR #156, opened 2025-05-07
Marks the authority in `extend_program_checked()` as signer

Lichtso:
Ironically #155 missed the vital part: The authority being a signer.


joncinque (approved):
Hehe, oops.

PR #160, opened 2025-05-09
[bls] Add `serde`, `bytemuck` and `frozen-abi` features

samkim-crypto:
Just reorganizing some trait implementations for BLS pod types
- Derived `Ord` and `PartialOrd` for all the pod types
- Created serde feature and made the serde serialization/deserialization optional
- Created bytemuck feature and made the bytemuck implementation optional
- Created `frozen-abi` feature and implemented the `AbiExample` traits for the bls pod types
- Implemented `Display` and `FromStr` for bls pod types

### DIFF
--- a/bls/src/pod.rs
+++ b/bls/src/pod.rs
@@ -1,5 +1,9 @@
 #[cfg(feature = "bytemuck")]
 use bytemuck::{Pod, PodInOption, Zeroable, ZeroableInOption};
+use {
+    base64::{prelude::BASE64_STANDARD, Engine},
+    std::fmt,
+};
 #[cfg(feature = "serde")]
 use {
     serde::{Deserialize, Serialize},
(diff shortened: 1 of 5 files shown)


bls/src/pod.rs
```diff
@@ -24,12 +33,27 @@ impl Default for SignatureCompressed {
...
+impl fmt::Display for SignatureCompressed {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", BASE64_STANDARD.encode(self.0))
+    }
+}
+
+impl_from_str!(
+    TYPE = SignatureCompressed,
+    BYTES_LEN = BLS_SIGNATURE_COMPRESSED_SIZE,
+    BASE64_LEN = 128
```
wen-coding: Do you think we should define 128 and 256 as constants as well?

samkim-crypto: Oh oops, I meant to define these, but I must have missed it.

wen-coding (approved):

PR #162, opened 2025-05-09
sysvar-instructions: check data length before storing current instruction idx

mjain-jump:
There are no checks for the instructions sysvar data length before trying to subtract / store data into the account. This does not cause issues in production, but causes issues with Firedancer's fuzzing efforts as this code triggers a panic if the sysvar account is not correctly initialized by the fuzzer when testing UB.

### DIFF
--- a/instructions-sysvar/src/lib.rs
+++ b/instructions-sysvar/src/lib.rs
@@ -161,9 +161,13 @@ pub fn load_current_index_checked(
 }
 
 /// Store the current `Instruction`'s index in the instructions sysvar data.
-pub fn store_current_index(data: &mut [u8], instruction_index: u16) {
+pub fn store_current_index(data: &mut [u8], instruction_index: u16) -> Result<(), ProgramError> {
+    if data.len() < 2 {
+        return Err(ProgramError::AccountDataTooSmall);
+    }
     let last_index = data.len() - 2;
     data[last_index..last_index + 2].copy_from_slice(&instruction_index.to_le_bytes());
+    Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


buffalojoec (commented):
I really don't think it's a good idea to silently do nothing in the case of a bad data length. This could violate consensus for two nodes using otherwise compatible Agave versions.

Is there a safer way we can address the fuzzing mismatches caused by this?

mjain-jump:
> I really don't think it's a good idea to silently do nothing in the case of a bad data length. I would say this could cause a consensus mismatch on `data.len() < 2`, but the original implementation would cause a panic.
> 
> Maybe it's better to just make it fallible with an `InstructionError`?

In production it's never possible for the serialized size to be < 2 based on the sysvar serialization logic. I don't think this would cause a consensus break but I'm open to returning an instruction error instead of panicking, just would prefer not to crash the client here.

buffalojoec:
> > I really don't think it's a good idea to silently do nothing in the case of a bad data length. I would say this could cause a consensus mismatch on `data.len() < 2`, but the original implementation would cause a panic.
> > Maybe it's better to just make it fallible with an `InstructionError`?
> 
> In production it's never possible for the serialized size to be < 2 based on the sysvar serialization logic. I don't think this would cause a consensus break but I'm open to returning an instruction error instead of panicking, just would prefer not to crash the client here.

Yeah, I think the `InstructionError` is the better move, since on the Agave side, we can just bubble the error up through `InvokeContext::push`.

buffalojoec (approved):
Thanks!

PR #166, opened 2025-05-12
keypair: Deprecate `generate` to avoid public rand dep

joncinque:
#### Problem

The `generate()` function on `Keypair` exposes a public dependency on `rand` because it requires certain types from the crate. Using the public type from an external crate in our interface means that we can't upgrade without breaking downstream users.

#### Summary of changes

Mark the function as deprecated, and provide a new way to create a random keypair using `new_from_array`.

Unfortunately, we can't mark the function as `const` because the dalek types don't have const constructors.


samkim-crypto (approved):
Yep, makes sense to me!

PR #167, opened 2025-05-13
add functions to read and write keypairs from files

samkim-crypto:
#### Problem
There is no support to read and write bls keypairs from files yet.

#### Summary of Changes
Added support to read and write bls keyparis from files. I put it under the `fs` feature so that users don't have to import serde, but if there is a more conventional way to do this, then please let me know.


bls/src/keypair.rs
```diff
@@ -229,4 +359,19 @@ mod tests {
...
+        let outfile = format!("{}/tmp/test_write_keypair_file.json", out_dir);
+        let original_keypair = Keypair::new();
+        original_keypair.write_json_file(&outfile).unwrap();
+        assert!(Path::new(&outfile).exists());
+
+        let read_keypair = Keypair::read_json_file(&outfile).unwrap();
+        assert_eq!(original_keypair, read_keypair);
+
+        fs::remove_file(&outfile).unwrap();
+    }
```
joncinque: For this test to run by default, you'll need to add this crate as a `dev-dependency` and enable the `fs` feature. Example at https://github.com/anza-xyz/solana-sdk/blob/c39b21fa43d142fe129e57593a150853df87fa9a/vote-interface/Cargo.toml#L46

bls/src/keypair.rs
```diff
@@ -229,4 +359,19 @@ mod tests {
         assert_eq!(keypair.secret, secret);
         assert_eq!(keypair.public, public);
     }
+
+    #[test]
+    #[cfg(feature = "fs")]
+    fn test_keypair_file() {
+        let out_dir = std::env::var("FARF_DIR").unwrap_or_else(|_| "farf".to_string());
+        let outfile = format!("{}/tmp/test_write_keypair_file.json", out_dir);
```
joncinque: nit: It'd be better to use `tempfile` for these  rather than the `FARF` directory

bls/Cargo.toml
```diff
@@ -40,6 +41,7 @@ solana-keypair = { workspace = true }
 [features]
 bytemuck = ["dep:bytemuck"]
 frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro"]
+fs = ["dep:serde_json"]
```
joncinque: nit: I looked around briefly, and the only example I could see to gate `fs` usage was the ed25519_dalek crate, and it used the `std` feature for that.

I'm not sure if you plan on making this crate `no-std` compatible, but it might actually be possible in this crate, in which case, maybe calling the feature `std` does make more sense.

bls/src/keypair.rs
```diff
@@ -75,6 +88,26 @@ impl SecretKey {
...
+        let scalar: Option<Scalar> = Scalar::from_bytes_le(bytes.try_into().unwrap()).into();
+        scalar.ok_or(BlsError::FieldDecode).map(Self)
+    }
+}
+
+impl From<&SecretKey> for [u8; BLS_SECRET_KEY_SIZE] {
+    fn from(secret_key: &SecretKey) -> Self {
+        let mut bytes = [0u8; BLS_SECRET_KEY_SIZE];
+        bytes.copy_from_slice(secret_key.0.to_bytes_le().as_slice());
+        bytes
```
joncinque: This implementation is a bit redundant -- `to_bytes_le()` already returns a properly-sized array, so you can just do
```suggestion
        secret_key.0.to_bytes_le()
```

bls/src/keypair.rs
```diff
@@ -157,6 +190,28 @@ impl TryFrom<&Pubkey> for PubkeyProjective {
...
+        public_affine.try_into()
+    }
+}
+
+impl From<&PubkeyProjective> for [u8; BLS_PUBLIC_KEY_AFFINE_SIZE] {
+    fn from(pubkey: &PubkeyProjective) -> Self {
+        let mut bytes = [0u8; BLS_PUBLIC_KEY_AFFINE_SIZE];
+        let pubkey_affine: Pubkey = (*pubkey).into();
+        bytes.copy_from_slice(pubkey_affine.0.as_slice());
+        bytes
```
joncinque: Same with this one, since the inner type is public, you can just do
```suggestion
        let pubkey_affine: Pubkey = (*pubkey).into();
        pubkey_affine.0
```

joncinque (commented):
Just a few little comments, looks good otherwise!

bls/src/keypair.rs
samkim-crypto: Oh oops. Thanks for the catch!

bls/Cargo.toml
samkim-crypto: Yeah good idea! I renamed it to `std` and took a stab at making the crate `no-std` compatible. Please let me know if my changes make sense.

joncinque (approved):
Looks great! Note that while the crate code is no-std, there's still some work to make it fully no-std, since some dependencies still use it. For example, base64 has std enabled by default: https://github.com/marshallpierce/rust-base64/blob/6243bd11f82eb27e6880e69844977e7226e7378b/Cargo.toml#L50

But that can always get refined over time, let's land this as is!

PR #168, opened 2025-05-13
feature-gate-interface: check account data len on deserialize

buffalojoec:
The feature gate interface's `Feature` API allows for serializing and deserializing a `Feature { activated_at: Option<u64> }` within account state. However, this can be a little flaky when it comes to attempting to set that `Option<u64>` value to `Some`.

For example, if you create an account under `Feature111...` with only 1 byte, it will deserialize just fine into `Feature { activated_at: None }`, but if you attempt to set the slot value to `Some`, it will fail to serialize.

Since the API already enforces some special-casing about account ownership, and declares a constant `size` function, it makes sense for it to also enforce account data length on either side of the serialization workflow.

### DIFF
--- a/feature-gate-interface/src/lib.rs
+++ b/feature-gate-interface/src/lib.rs
@@ -42,6 +42,9 @@ impl Feature {
         if *account_info.owner != id() {
             return Err(ProgramError::InvalidAccountOwner);
         }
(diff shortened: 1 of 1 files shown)


jstarry (approved):

joncinque (approved):
Looks great, thanks!

PR #171, opened 2025-05-14
native-token: Deprecate float usage, add str usage

joncinque:
#### Problem

The `sol_to_lamports` and `lamports_to_sol` functions are problematic because they use floats, which are not precise.

#### Summary of changes

Deprecate the functions, and add versions that use strings instead.

Added @tao-stones to check the `FeeStructure` change -- it looks like `FeeStructure::new` is only used in one test.

### DIFF
--- a/fee-structure/src/lib.rs
+++ b/fee-structure/src/lib.rs
@@ -4,7 +4,7 @@
 
 #[cfg(not(target_os = "solana"))]
 use solana_message::SanitizedMessage;
-use {solana_native_token::sol_to_lamports, std::num::NonZeroU32};
+use std::num::NonZeroU32;
 
 /// A fee and its associated compute unit limit
 #[derive(Debug, Default, Clone, Eq, PartialEq)]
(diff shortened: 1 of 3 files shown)


rustopian (approved):
Correct, good tests, all works great.

rustopian:
Oops, @tao-stones looks like your review is requested directly despite the auto-merge, my bad

t-nelson:
i do this far too often :facepalm: 

Screenshot from 2025-05-14 19-30-55

joncinque:
Ah darn, sorry -- for the first point, we'll get a warning on removal, which will fail CI, so we can't forget to remove the const.

And for the second point, I wasn't aware of that function... I guess the options are to either include a bool flag (edit: in this new function), or just remove it and tell people to use `build_balance_message`. All of the users of `lamports_to_sol` are CLIs / user-facing, so dependencies aren't an issue. I'd lean towards the second option. What do you think?

tao-stones (commented):
Changes to fee-structure is good.

t-nelson:
> Ah darn, sorry

no worries! i'm the idiot that didn't hit the comment button

> for the first point, we'll get a warning on removal, which will fail CI, so we can't forget to remove the const.

👍 

> And for the second point, I wasn't aware of that function... I guess the options are to either include a bool flag (edit: in this new function), or just remove it and tell people to use `build_balance_message`. All of the users of `lamports_to_sol` are CLIs / user-facing, so dependencies aren't an issue. I'd lean towards the second option. What do you think?

yeah agreed. i think we should stick one implementation to avoid all of the caveats of replicode

PR #172, opened 2025-05-14
instructions-sysvar: refactor #162 to be non-breaking

buffalojoec:
Whoopsie, changes #162 to be non-breaking so we can just cut a patch version. We can still make Agave use the new, non-deprecated method.


joncinque (approved):
Makes sense, thanks!

PR #176, opened 2025-05-30
Deprecates `AccountInfo::realloc()`

Lichtso:
Only a minuscule amount of transactions grows accounts. Zeroing out 10 KiB for every instruction account in every instruction (including CPI) of every transaction just in case is quite wasteful. It would make more sense to only do this on demand and have programs pay for what they use.

This change makes sure that programs always call `sol_memset()` to communicate to the program runtime that they wish to grow an account and access the new space even before the next CPI or the end of the current instruction.

### DIFF
--- a/account-info/src/lib.rs
+++ b/account-info/src/lib.rs
@@ -136,11 +136,8 @@ impl<'a> AccountInfo<'a> {
     /// Note:  Account data can be increased within a single call by up to
     /// `solana_program::entrypoint::MAX_PERMITTED_DATA_INCREASE` bytes.
     ///
-    /// Note: Memory used to grow is already zero-initialized upon program
-    /// entrypoint and re-zeroing it wastes compute units.  If within the same
-    /// call a program reallocs from larger to smaller and back to larger again
-    /// the new space could contain stale data.  Pass `true` for `zero_init` in
-    /// this case, otherwise compute units will be wasted re-zero-initializing.
+    /// Note: `zero_init` being `false` will no longer be supported by the
+    /// program runtime.
     ///
     /// # Safety
     ///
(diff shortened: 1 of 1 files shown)


Lichtso:
Seems like Token 2022 is also using this:
https://github.com/solana-program/token-2022/blob/f8359372316509f332a9973634d64d28dccb29e6/program/src/extension/reallocate.rs#L73C1-L73C60

https://github.com/anza-xyz/agave/actions/runs/15403664409/job/43341856516?pr=5871

account-info/src/lib.rs
```diff
@@ -185,6 +183,22 @@ impl<'a> AccountInfo<'a> {
...
+    /// Note:  Account data can be increased within a single call by up to
+    /// `solana_program::entrypoint::MAX_PERMITTED_DATA_INCREASE` bytes.
+    ///
+    /// # Safety
+    ///
+    /// This method makes assumptions about the layout and location of memory
+    /// referenced by `AccountInfo` fields. It should only be called for
+    /// instances of `AccountInfo` that were created by the runtime and received
+    /// in the `process_instruction` entrypoint of a program.
+    pub fn resize(&self, new_len: usize) -> Result<(), ProgramError> {
```
joncinque, line 197: How about marking this `unsafe` in preparation for https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html for when `realloc` is removed?

joncinque (commented):
I put in https://github.com/solana-program/token-2022/pull/479 for token-2022

account-info/src/lib.rs
Lichtso: The function could be implemented safely, it is just that the way the SDK currently abuses the length field causes it to require an unsafe block inside. Don't see a reason why we should expose that to all call sites.

joncinque: Could we have `resize` implement the non-unsafe version in this PR then?

Lichtso: Correcting that hacky bit is non-trivial because the issue already starts with the way we deserialize in the entrypoint and that we use `Rc<RefCell<>>` for the account data. Talked with @febo how to avoid this in pinocchio.

joncinque: Gotcha, I'll take that to mean that there's a plan :grin:

PR #177, opened 2025-05-30
deps: Run cargo-sort on the whole workspace

joncinque:
#### Problem

`cargo-sort` had a new breaking release, which changes how things are sorted. This new version is run in CI, and reports new differences.

#### Summary of changes

Run `./cargo nightly sort --workspace` and commit the changes.


yihau (approved):
🫶

PR #180, opened 2025-06-04
scripts: Remove bad directories from patch script

joncinque:
#### Problem

The patch script is supposed to patch all crates in the SDK repo, but some of the entries are incorrect, mainly "scripts" which is used for the scripts, and "target" which is used for build artifacts.

#### Summary of changes

Remove the entries.


yihau (approved):
🫡

Issue #175, opened 2025-05-17
solana-keypair 2.2.2 compilation error


juchiast:
Look like solana-keypair something went wrong in the release process, it is using solana-signature 2.2.1, need 2.3.0 to works

```
error[E0432]: unresolved import `solana_signature::error`
  --> ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/solana-keypair-2.2.2/src/lib.rs:10:24
   |
10 |     solana_signature::{error::Error as SignatureError, Signature},
   |                        ^^^^^ could not find `error` in `solana_signature`
```

joncinque:
Thanks for pointing this out, you're absolutely right. solana-keypair and solana-secp256k1-program v2.2.2 have both been yanked, since they need the new error type. I'll update the version on the repo and publish a new version

joncinque:
Note for any onlookers: I ran a test for all of the recent publishes to make sure they properly work against all published versions, and everything else checks out.

joncinque:
v2.2.3 has been released with the fix of #182 , and #181 added a check before publishing to make sure it doesn't happen again.

Thanks again for your quick reporting!

PR #181, opened 2025-06-04
CI: Add step to check minimal versions on detached crate

joncinque:
#### Problem

Although CI checks that we satisfy the minimal versions declared in Cargo.toml, it doesn't work for dependencies in the workspace, as noticed in #175. There, solana-keypair should depend on a newer version of solana-signature, but the normal minimal-versions check doesn't catch it.

#### Summary of changes

The script is a bit brittle, but it detaches a crate from the workspace and then runs `cargo-minimal-versions` on it as a pre-publish test. Once this lands, I'll do a dry-run publish to make sure solana-keypair fails.

Also, bumped up two dependencies that the tests revealed.


yihau (approved):
🪖

joncinque:
It worked! https://github.com/anza-xyz/solana-sdk/actions/runs/15464728499/job/43534100682

PR #182, opened 2025-06-04
Bump versions for publish

joncinque:
#### Problem

As noticed in #175, some of the intra-repo dependencies are too low, so the crates don't properly declare their minimum supported versions.

#### Summary of changes

Bump the required dependencies. I will wait to land this until after #181 and a failing publish on solana-keypair.


yihau (approved):
🫶

PR #184, opened 2025-06-04
sbpf: Fix various build warnings

joncinque:
#### Problem

When building certain crates with `cargo build-sbf`, we generate warnings currently, which isn't great for downstream users.

#### Summary of changes

Fix the warnings, and make CI stricter by forbidding warnings.

### DIFF
--- a/secp256r1-program/src/lib.rs
+++ b/secp256r1-program/src/lib.rs
@@ -831,12 +831,17 @@ mod target_arch {
 
 #[cfg(any(target_arch = "wasm32", target_os = "solana"))]
 mod target_arch {
-    use {solana_feature_set::FeatureSet, solana_precompile_error::PrecompileError};
+    use solana_precompile_error::PrecompileError;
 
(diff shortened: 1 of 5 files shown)


secp256r1-program/src/lib.rs
```diff
@@ -831,12 +831,17 @@ mod target_arch {
...
+    #[deprecated(
+        since = "2.2.4",
+        note = "Use agave_precompiles::secp256r1::verify instead"
+    )]
+    #[allow(deprecated)]
     pub fn verify(
         _data: &[u8],
         _instruction_datas: &[&[u8]],
-        _feature_set: &FeatureSet,
+        _feature_set: &solana_feature_set::FeatureSet,
```
febo, line 844: Wondering if you meant to use the full qualified type here.

joncinque: Mostly laziness to avoid adding another `#[allow(deprecated)]` above line 834

PR #185, opened 2025-06-05
publish: Fix git-cliff command

joncinque:
#### Problem

With the new version of `git-cliff-action`, the "args" is no longer intrepreted as a list, and instead as a string, causing a failure during a publish at
https://github.com/anza-xyz/solana-sdk/actions/runs/15466949268/job/43541245695

#### Summary of changes

Put all of the args in one line


yihau (approved):
🪖

PR #187, opened 2025-06-05
Update solana-sdk-macro to use solana_pubkey::Pubkey

### LINKED ISSUE #179: Update `solana-sdk-macro` to use small Pubkey crate
Several macros introduce in `solana-sdk-macro` introduce `solana_sdk::pubkey::Pubkey` into calling code:
https://github.com/anza-xyz/solana-sdk/blob/1276772ee61fbd1f8a60cfec7cd553aa4f6a55f3/sdk-macro/src/lib.rs#L103

These should be updated to use `solana_pubkey::Pubkey` instead

RaghavenderSingh:
- Replace ::solana_sdk::pubkey::Pubkey with ::solana_pubkey::Pubkey in SdkPubkey, Id, IdDeprecated, and Pubkeys structs
- Keep ::solana_program::pubkey::Pubkey for ProgramSdk* structs
- Add solana-pubkey dependency to Cargo.toml

Fixes #179

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3688,9 +3688,6 @@ dependencies = [
  "bs58",
  "proc-macro2",
  "quote",
- "solana-program",
- "solana-pubkey",
- "solana-sdk",
  "syn 2.0.98",
 ]
 


sdk-macro/Cargo.toml
```diff
@@ -20,3 +20,4 @@ bs58 = { workspace = true, features = ["alloc"] }
 proc-macro2 = { workspace = true }
 quote = { workspace = true }
 syn = { workspace = true, features = ["full"] }
+solana-pubkey = { workspace = true }
```
buffalojoec: You don't need this dependency. All `solana_pubkey` lines in this macro lib are converted to tokens.

buffalojoec (commented):
I wonder if there's a better way to do this, since it's going to be a breaking change anyway.

With the current implementation, anyone using `solana_sdk` would have to also add `solana_pubkey` to use these macros, which could be a bit annoying, since `Pubkey` is already offered through `solana_sdk`.

We could follow an explicit import pattern, where the macro lib defaults to `solana_pubkey`, but developers can override it to use `solana_sdk::pubkey` instead. Here are a few examples:
* `num_derive`: https://docs.rs/num-derive/latest/num_derive/#explicit-import
* `spl_program_error`:
  * Impl: https://github.com/solana-program/libraries/blob/fcd6052feccb74b5ae4f7a8a7858e85d7f4adc93/program-error/derive/src/parser.rs#L24-L29
  * Example: https://github.com/solana-program/libraries/blob/fcd6052feccb74b5ae4f7a8a7858e85d7f4adc93/program-error/tests/spl.rs#L77

Of course, if any maintainers think this is too much, and we should just break it, I will concede. 😀

RaghavenderSingh:
@buffalojoec you're absolutely right, thanks for catching that!
I've removed the dependency from Cargo.toml - much cleaner since we're just generating the tokens. Updated the PR and all tests are still passing.
Is there anything else I should address before this is ready for merge?

buffalojoec:
@RaghavenderSingh I'm not sure if you saw my top-level comment [here](https://github.com/anza-xyz/solana-sdk/pull/187#pullrequestreview-2903354887).

sdk-macro/src/lib.rs
```diff
@@ -18,6 +17,86 @@ use {
     },
 };
 
+fn parse_solana_pubkey_param(input: ParseStream) -> Result<proc_macro2::TokenStream> {
```
buffalojoec: This function and the one below it are identical, with the exception of the returned tokens.

buffalojoec: In fact, I think _all_ of this duplicated code you've got in these new functions can be compacted onto an enum:

```rust
enum PubkeyImport {
    SolanaPubkey,
    SolanaProgram,
    SolanaSdk,
}

impl Parse for PubkeyImport { /* ... */ }

impl ToTokens for PubkeyImport { /* ... */ }
```

RaghavenderSingh: Hey @buffalojoec!
Yeah you're totally right, that's way too much duplicated code. I was so focused on getting the feature working that I didn't step back and see the bigger picture. The enum approach is much cleaner - thanks for the suggestion!
I'll refactor this to use the PubkeyImport enum pattern you outlined. Should be pretty straightforward to consolidate all that parsing logic into one place and just have different default behaviors.
Also need to add some tests for the explicit import feature - completely forgot about that part. Will get both of those done.
Thanks for keeping me honest on the code quality! 😅

PR #188, opened 2025-06-06
Add AbiExample for dashmap and boxcar

alessandrod:
Needed by https://github.com/anza-xyz/agave/pull/3796

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -140,6 +140,7 @@ bitflags = { version = "2.8.0" }
 blake3 = "1.5.5"
 borsh = { version = "1.5.5", features = ["derive", "unstable__schema"] }
 borsh0-10 = { package = "borsh", version = "0.10.3" }
+boxcar = "0.2.12"
 bs58 = { version = "0.5.1", default-features = false }
 bv = "0.11.1"
 bytemuck = "1.21.0"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -140,6 +140,7 @@ bitflags = { version = "2.8.0" }
 blake3 = "1.5.5"
 borsh = { version = "1.5.5", features = ["derive", "unstable__schema"] }
 borsh0-10 = { package = "borsh", version = "0.10.3" }
+boxcar = "0.2.7"
```
joncinque: I know this is coming from an older PR, but any reason to not use the newest version which is committed to the lockfile?

joncinque (dismissed):
Looks great! Just a nit on the declared version, which you can take or leave

Cargo.toml
alessandrod: no reason! I'll update

alessandrod: done

alessandrod:
Thanks for reviewing!

So... what now? Do we need to make a release to unblock https://github.com/anza-xyz/agave/pull/3796 ?

joncinque:
That's correct, you can typically just publish the crate yourself using the GitHub Action. In this case, I did it for you -- enjoy! https://crates.io/crates/solana-frozen-abi/2.3.0

A bit more info at https://github.com/anza-xyz/solana-sdk?tab=readme-ov-file#publishing-a-crate-from-this-repository

PR #189, opened 2025-06-06
program-error!: Improve ToStr ergonomics

joncinque:
#### Problem

The `ToStr` trait is a bit difficult to use, forcing a lot of redundant uses of types, when programs will just want to call `err.to_str::<MyError>()`.

#### Summary of changes

Simplify the trait to not require anything, then add a generic `to_str` function directly on `ProgramError` that allows us to get the error message for a particular error type.

Also, clarify the deprecation warning on `DecodeError`.

BREAKING CHANGE: the `ToStr` trait requires fewer bounds

### DIFF
--- a/program-error/src/lib.rs
+++ b/program-error/src/lib.rs
@@ -189,22 +189,54 @@ impl PrintProgramError for ProgramError {
     }
 }
 
-/// A trait for converting a program error to a `&str`.
+/// A trait for converting a program's specific error type to a `&str`.
+///
+/// Can be used with `ProgramError::to_str::<E>()` to get an error string
+/// belonging to a specific program's error if the variant is
+/// `ProgramError::Custom(...)`, or generic strings from the contained
+/// `ProgramError` for all other variants.
+///
+/// The `ProgramError::to_str::<E>()` function also requires implementing
+/// `TryFrom<u32>` on an error type, which can be done easily using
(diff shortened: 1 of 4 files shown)


joncinque:
#191 will address the CI failure

program-error/src/lib.rs
```diff
@@ -189,22 +189,54 @@ impl PrintProgramError for ProgramError {
...
+/// `ProgramError::Custom(...)`, or generic strings from the contained
+/// `ProgramError` for all other variants.
+///
+/// The `ProgramError::to_str::<E>()` function also requires implementing
+/// `TryFrom<u32>` on an error type, which can be done easily using
+/// `num_enum::TryFromPrimitive`.
 pub trait ToStr {
-    fn to_str<E>(&self) -> &'static str
-    where
-        E: 'static + ToStr + TryFrom<u32>;
```
febo, line 196: Super nice!

febo (approved):
Looks great!

febo:
> BREAKING CHANGE: the `ToStr` trait requires fewer bounds

Is requiring fewer bounds a breaking change?

febo:
> > BREAKING CHANGE: the `ToStr` trait requires fewer bounds
> 
> Is requiring fewer bounds a breaking change?

Ah, it is on the `to_str` method.

PR #194, opened 2025-06-09
program!: Remove all re-exports, update changelog

joncinque:
#### Problem

There are many re-exports in solana-program that don't make sense, and should instead use component crates.

#### Summary of changes

Remove those re-exports, and document it in the README.

BREAKING CHANGE: re-exports removed.


transaction/Cargo.toml
```diff
@@ -80,5 +80,5 @@ solana-pubkey = { workspace = true, features = ["rand"] }
 solana-sdk = { path = "../sdk" }
 solana-sha256-hasher = { workspace = true }
 solana-transaction = { path = ".", features = ["dev-context-only-utils"] }
-solana-vote-interface = { workspace = true }
+solana-vote-interface = { workspace = true, features = ["bincode"] }
```
febo, line 83: Just checking whether this was intentional or not.

joncinque: Yep! The tests don't build without it. It's probably because solana-transaction depends on solana-sdk which depends on solana-program, which used to enable the `bincode` feature on `solana-vote-interface`. Now that we've removed that dep from solana-program, we need to enable the feature explicitly

PR #198, opened 2025-06-16
sysvar: Improve error handling

febo:
### Problem

Currently the implementation of `impl_sysvar_get` macro and `get_sysvar` return a `e.into()` when the syscall fails to return success. This generated a `ProgramError::Custom(_)` error, although there are no custom errors associated with them.

### Solution

There are only two situations where a sysvar get will return an error code:
1. if you pass a wrong combination of length and offset, which tries to copy data outside the sysvar length
2. If you pass an invalid sysvar id

This PR updates the error handling to return a `ProgramError::InvalidArgument` for the first case and `ProgramError::UnsupportedSysvar` for the second.

### DIFF
--- a/sysvar/src/lib.rs
+++ b/sysvar/src/lib.rs
@@ -110,6 +110,10 @@ pub mod slot_hashes;
 pub mod slot_history;
 pub mod stake_history;
 
+/// Return value to indicate that the `offset + length` is greater than the length of
+/// the sysvar data.
+const OFFSET_LENGTH_EXCEEDS_SYSVAR: u64 = 1;
+
 #[deprecated(
     since = "2.0.0",
     note = "please use `solana_sdk::reserved_account_keys::ReservedAccountKeys` instead"
@@ -206,7 +210,7 @@ macro_rules! impl_sysvar_get {
 
             match result {
                 $crate::__private::SUCCESS => Ok(var),
-                e => Err(e.into()),
+                _ => Err($crate::__private::ProgramError::UnsupportedSysvar),
             }
         }
     };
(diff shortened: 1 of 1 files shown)


sysvar/src/lib.rs
```diff
@@ -206,7 +217,8 @@ macro_rules! impl_sysvar_get {
 
             match result {
                 $crate::__private::SUCCESS => Ok(var),
-                e => Err(e.into()),
+                // Unexpected errors are folded into `UnsupportedSysvar`.
+                _ => Err($crate::__private::ProgramError::UnsupportedSysvar),
```
joncinque, line 221: Just to check, is this actually reachable by the syscall? Or just by the stub call? I don't see any other error code returned by the syscall implementation

sysvar/src/lib.rs
```diff
@@ -239,7 +251,10 @@ fn get_sysvar(
 
     match result {
         solana_program_entrypoint::SUCCESS => Ok(()),
-        e => Err(e.into()),
+        OFFSET_LENGTH_EXCEEDS_SYSVAR => Err(ProgramError::InvalidArgument),
+        SYSVAR_NOT_FOUND => Err(ProgramError::UnsupportedSysvar),
+        // Unexpected errors are folded into `UnsupportedSysvar`.
+        _ => Err(ProgramError::UnsupportedSysvar),
```
joncinque: Same with this, it's only reachable by the stub implementation, correct?

joncinque (commented):
This could be a breaking change for people customizing the syscalls with program-stubs, but I think it makes sense. Once CI passes, I'll approve

sysvar/src/lib.rs
febo: No, it should not happen. It is here only because we are matching against an `u64` value.

febo: And yes, it will happen in the case of the stub call since it returns `UNSUPPORTED_SYSVAR`.

sysvar/src/lib.rs
febo: Yes, only the stub implementation will reach here.

PR #200, opened 2025-06-18
[base3-encoding] Add solana base3 encoding crate

samkim-crypto:
This PR introduces the solana-base3-encoding crate, created to efficiently encode Alpenglow vote certificates.

Alpenglow certificates contain aggregate signatures, which don't specify individual signers. For simple certificate types like Finalize or Notarize, a bit vector is sufficient to identify the voters.

However, more complex certificates like NotarizeFallback or Skip can contain two different vote types. Using a separate bit vector for each type is not scalable. As the validator set grows to 4,000, the two bit vectors would exceed 1 KB, making certificates harder to fit in a single network packet. 

This crate provides a more compact encoding scheme. The design is tailored to the [CertificateMessage](https://github.com/solana-program/alpenglow-vote/blob/master/program/src/bls_message.rs#L32), taking two BitVec inputs and combining them into a base-3 representation.

This does seem to make the crate quite specific for alpenglow certificates, so maybe we can either have this crate in the `alpenglow-vote` repo or think of ways to make the interface more generic.

### DIFF
--- a/base3-encoding/src/lib.rs
+++ b/base3-encoding/src/lib.rs
@@ -218,8 +218,7 @@ pub use bitvec_support::*;
 
 #[cfg(feature = "bitvec")]
 mod bitvec_support {
-    use super::*;
-    use bitvec::prelude::*;
+    use {super::*, bitvec::prelude::*};
 
     /// A wrapper to encode two `BitVec`s into a single `Vec<u8>`.
     pub fn encode(
@@ -275,9 +274,10 @@ mod tests {
 
     #[cfg(feature = "bitvec")]
     mod bitvec_tests {
-        use super::super::bitvec_support::*;
-        use super::*;
-        use bitvec::prelude::*;
+        use {
+            super::{super::bitvec_support::*, *},
+            bitvec::prelude::*,
+        };
 
         fn create_test_data(len: usize) -> (BitVec<u8, Lsb0>, BitVec<u8, Lsb0>) {
             let mut base = BitVec::with_capacity(len);
--- a/base3-encoding/src/lib.rs
+++ b/base3-encoding/src/lib.rs
@@ -88,8 +88,13 @@ pub fn encode_from_bytes(
 
     for chunk_index in 0..total_u8_length {
         let mut block_num: u8 = 0;
-        let start_bit = chunk_index * BASE3_SYMBOL_PER_CHUNK;
-        let end_bit = (start_bit + BASE3_SYMBOL_PER_CHUNK).min(num_bits);
+        let start_bit = chunk_index
+            .checked_mul(BASE3_SYMBOL_PER_CHUNK)
+            .ok_or(EncodeError::ArithmeticOverflow)?;
+        let end_bit = start_bit
+            .checked_add(BASE3_SYMBOL_PER_CHUNK)
+            .ok_or(EncodeError::ArithmeticOverflow)?
+            .min(num_bits);
 
         for i in (start_bit..end_bit).rev() {
             let byte_idx = i / 8;
@@ -128,6 +133,9 @@ pub enum DecodeError {
 }
 
 /// Decodes an encoded byte slice back into two byte vectors.
+///
+/// This function returns a tuple containing the base byte vector, the fallback
+/// byte vector, and the exact number of valid bits in those vectors.
 pub fn decode_to_bytes(
     bytes: &[u8],
     max_len: usize,
@@ -175,7 +183,11 @@ pub fn decode_to_bytes(
             let remainder = block_num % 3;
             block_num /= 3;
 
-            let bit_index = i * BASE3_SYMBOL_PER_CHUNK + j;
+            let bit_index = i
+                .checked_mul(BASE3_SYMBOL_PER_CHUNK)
+                .ok_or(DecodeError::ArithmeticOverflow)?
+                .checked_add(j)
+                .ok_or(DecodeError::ArithmeticOverflow)?;
             let byte_idx = bit_index / 8;
             let bit_idx = bit_index % 8;
 
@@ -193,7 +205,9 @@ pub fn decode_to_bytes(
                 fallback_bytes[byte_idx] |= 1 << bit_idx;
             }
         }
-        bits_remaining -= bits_in_this_chunk;
+        bits_remaining = bits_remaining
+            .checked_sub(bits_in_this_chunk)
+            .ok_or(DecodeError::ArithmeticOverflow)?;
     }
 
     Ok((base_bytes, fallback_bytes, total_bits))
@@ -247,8 +261,8 @@ mod tests {
     #[test]
     fn test_bytes_round_trip_small() {
         let num_bits = 10;
-        let base_bytes = vec![0b0100_1001, 0b0000_0001]; // Corresponds to bits [1,0,0,1,0,0,1,0,1,0]
-        let fallback_bytes = vec![0b1010_0110, 0b0000_0000]; // Corresponds to bits [0,1,1,0,0,1,0,1,0,0]
+        let base_bytes = vec![0b0100_1001, 0b0000_0001];
+        let fallback_bytes = vec![0b1010_0110, 0b0000_0000];
 
         let encoded = encode_from_bytes(&base_bytes, &fallback_bytes, num_bits).unwrap();
         let (decoded_base, decoded_fallback, decoded_bits) =
@@ -256,11 +270,12 @@ mod tests {
 
         assert_eq!(decoded_bits, num_bits);
         assert_eq!(base_bytes, decoded_base);
-        assert_eq!(fallback_bytes, decoded_fallback); // Compare relevant part
+        assert_eq!(fallback_bytes, decoded_fallback);
     }
 
     #[cfg(feature = "bitvec")]
     mod bitvec_tests {
+        use super::super::bitvec_support::*;
         use super::*;
         use bitvec::prelude::*;
 
--- a/base3-encoding/src/lib.rs
+++ b/base3-encoding/src/lib.rs
@@ -141,6 +141,10 @@ pub enum DecodeError {
 /// This function reverses the process of the `encode` function, reconstructing
 /// the original two bit vectors from a byte slice.
 ///
+/// # Parameters
+/// - `bytes`: The byte slice containing the encoded data.
+/// - `max_len`: The maximum expected length of the decoded bit vectors.
+///
 /// # Algorithm
 /// 1.  Reads the first 2 bytes to determine the `total_bits` of the original data.
 /// 2.  Validates that the remaining data payload is a multiple of 16 bytes.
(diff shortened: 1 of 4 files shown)


samkim-crypto:
@0x0ece It would be great to get your feedback on this!

samkim-crypto:
Had a private discussion with @0x0ece. As per discussion, I updated the encoding/decoding to use little-endian instead of big-endian.

Currently I use `u128` to encode a block of symbols. @0x0ece pointed out that `u128` might not be the fastest in terms of performance. So I created https://github.com/samkim-crypto/base3-encoding-tests and compared the three implementations using `u8`, `u64`, and `u128`. Indeed it seems encoding using `u64` is faster than using `u128`.

In terms of the length of the encoded bytes for length 4096 bit-vectors, we have
```
u8 -> 822 bytes
u64 -> 826 bytes
u128 -> 834 bytes
```
Using u8 gives the more compact encoding, but with just 4 additional bytes, an implementation using `u64` seems to be the way to go, but I am interested to hear people's thoughts.

wen-coding:
Yeah, I think we can spare 4 bytes for better performance.

base3-encoding/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+    /// The combination `(true, true)` was found, which is not allowed.
+    InvalidBitCombination,
+    /// The length of the input vectors exceeds `u16::MAX` (65,535).
+    LengthExceedsLimit,
+    /// Arithmetic overflow.
+    ArithmeticOverflow,
+}
+
+const BASE3_SYMBOL_PER_CHUNK: usize = 80;
+const ENCODED_BYTES_PER_CHUNK: usize = 16; // std::mem::size_of::<u128>()
```
wen-coding: I think we decided to use u64?

base3-encoding/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+///
+/// # Algorithm
+/// 1.  Validates that the input vectors have the same length and do not exceed
+///     `u16::MAX`.
+/// 2.  Writes the length of the vectors as a 2-byte big-endian `u16` to the
+///     output buffer.
+/// 3.  Iterates through the input vectors in chunks of 80 bits.
+/// 4.  For each chunk, it converts the 80 bit-pairs into 80 ternary digits and
+///     constructs a `u128` number from them.
+/// 5.  This `u128` number is appended to the output buffer as 16 big-endian bytes.
```
wen-coding: I think we use little-endian now?

0x0ece:
Two minor things (again, sorry...)

1. We could keep the ternary-bits in ascending order (same order as the original bits) - this also seems slightly faster (in all impl)
https://github.com/0x0ece/base3-encoding-tests/commit/5caeafcd4780ae9bb0a552047962825560284e3c

2. The reason I was proposing u8 was not just to save 4 bytes, but mostly for consistency of representation. Think for example to 5 ternary-bits: (0, 2, 1, 0, 0) -> 3+9 -> 0x0C. If we use the u8 implementation, the same sequence repeated would always translate into 0x0C, e.g. (0, 2, 1, 0, 0, 0, 2, 1, 0, 0, 0, 2, 1, 0, 0, 0, 2, 1, 0, 0) -> 0x0C0C0C0C. If we use the other representations the numbering will be more confusing to read. Imo the perf speed up of u64 is negligible (considering we're doing BLS when we do this), so I'd rather keep it simple with u8. (I'm sure there's ways to use avx and speed it up, but not sure it's worth it)

base3-encoding/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+
+    let capacity = total_byte_length
+        .checked_add(2)
+        .ok_or(EncodeError::ArithmeticOverflow)?;
+    let mut result = Vec::with_capacity(capacity);
+
+    result.extend_from_slice(&(bit_vec_base.len() as u16).to_le_bytes());
+
+    // Process the bits of `bit_vec_base` and `bit_vec_fallback` in chunks of 80 bits
+    for (base_chunk, fallback_chunk) in bit_vec_base.chunks(80).zip(bit_vec_fallback.chunks(80)) {
```
wen-coding: You probably want to use BASE3_SYMBOL_PER_CHUNK instead of 80 here?

base3-encoding/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+    if data_bytes
+        .len()
+        .checked_rem(ENCODED_BYTES_PER_CHUNK)
+        .unwrap()
+        != 0
+    {
+        return Err(DecodeError::CorruptDataPayload);
+    }
+
+    let mut bit_vec_base = BitVec::with_capacity(total_bits);
```
wen-coding: Should we allow passing in the maximum expected bit vector length (probably 4096) so we can return CorruptDataPayload early?

samkim-crypto:
> The reason I was proposing u8 was not just to save 4 bytes, but mostly for consistency of representation. Think for example to 5 ternary-bits: (0, 2, 1, 0, 0) -> 3+9 -> 0x0C. If we use the u8 implementation, the same sequence repeated would always translate into 0x0C, e.g. (0, 2, 1, 0, 0, 0, 2, 1, 0, 0, 0, 2, 1, 0, 0, 0, 2, 1, 0, 0) -> 0x0C0C0C0C. If we use the other representations the numbering will be more confusing to read. Imo the perf speed up of u64 is negligible (considering we're doing BLS when we do this), so I'd rather keep it simple with u8. (I'm sure there's ways to use avx and speed it up, but not sure it's worth it)

Yeah that is actually a nice point. Seems like u8 will give a more readable implementation. If we use u64 or u128, then the symbol will depend on the position within the 40 or 80 symbol chunk. I am happy to go with the simpler u8 route. 

@wen-coding, are you okay with using u8? If so, I can make the change and push. I think we can always optimize it further or experiment with u64 again before the full alpenglow launch if it turns out to be too slow.

base3-encoding/src/lib.rs
samkim-crypto: Good catch. Fixed!

base3-encoding/src/lib.rs
samkim-crypto: Fixed!

wen-coding:
Sure, I think u8 also works

base3-encoding/src/lib.rs
samkim-crypto: Yeah good point! I added a `max_len` parameter to the decode function, which we can set to be 4096 for alpenglow certificates.

samkim-crypto:
Okay, I addressed all the comments from the review. I also ended up making the encoding and decoding scheme to work with byte arrays and made the `bitvec` create optional to make the crate more general and easier to maintain on future `bitvec` version updated.

base3-encoding/src/lib.rs
```diff
@@ -0,0 +1,401 @@
...
+    if num_bits > u16::MAX as usize {
+        return Err(EncodeError::LengthExceedsLimit);
+    }
+    let required_bytes = num_bits
+        .checked_add(7)
+        .ok_or(EncodeError::ArithmeticOverflow)?
+        .checked_div(8)
+        .ok_or(EncodeError::ArithmeticOverflow)?;
+
+    if base_bytes.len() < required_bytes || fallback_bytes.len() < required_bytes {
```
wen-coding, line 70: Is it okay if base_bytes.len() and fallback_bytes.len() > required_bytes?

base3-encoding/src/lib.rs
```diff
@@ -0,0 +1,401 @@
...
+        .ok_or(EncodeError::ArithmeticOverflow)?
+        .checked_div(BASE3_SYMBOL_PER_CHUNK)
+        .ok_or(EncodeError::ArithmeticOverflow)?;
+
+    let total_byte_length = total_u8_length
+        .checked_mul(ENCODED_BYTES_PER_CHUNK)
+        .ok_or(EncodeError::ArithmeticOverflow)?;
+
+    let capacity = total_byte_length
+        .checked_add(2)
```
wen-coding: nit: add a note here saying we use 2 bytes to hold the byte length?

base3-encoding/src/lib.rs
```diff
@@ -0,0 +1,401 @@
...
+        let end_bit = start_bit
+            .checked_add(BASE3_SYMBOL_PER_CHUNK)
+            .ok_or(EncodeError::ArithmeticOverflow)?
+            .min(num_bits);
+
+        for i in (start_bit..end_bit).rev() {
+            let byte_idx = i / 8;
+            let bit_idx = i % 8;
+
+            let base_bit = (base_bytes[byte_idx] >> bit_idx) & 1 == 1;
```
wen-coding, line 105: Sorry I have a dumb question, why did we switch to raw bytes instead of using bitvec as input?

base3-encoding/src/lib.rs
```diff
@@ -0,0 +1,401 @@
...
+        let mut block_num: u8 = 0;
+        let start_bit = chunk_index
+            .checked_mul(BASE3_SYMBOL_PER_CHUNK)
+            .ok_or(EncodeError::ArithmeticOverflow)?;
+        let end_bit = start_bit
+            .checked_add(BASE3_SYMBOL_PER_CHUNK)
+            .ok_or(EncodeError::ArithmeticOverflow)?
+            .min(num_bits);
+
+        for i in (start_bit..end_bit).rev() {
```
wen-coding, line 101: Another dumb question: is it maybe easier to loop on bit in the outer loop and pack until you hit a full chunk then push the full chunk to the output array?

base3-encoding/src/lib.rs
samkim-crypto: Yeah the way the function signature is defined is that the `num_bits` represent the exact number of bits that should be encoded from `base_bytes` and `fallback_bytes`. If `base_bytes.len()` or `fallback_bytes.len()` is greater than the required bytes, the extra bits will just be ignored. I'll write a comment to clarify that.

base3-encoding/src/lib.rs
samkim-crypto: Done!

base3-encoding/src/lib.rs
samkim-crypto: Yeah I thought `bitvec` is quite specific, so I made the `bitvec` crate an optional dependency. In theory, projects that do not use the `bitvec` crate, but still wants to use the `solana-base3-encoding` crate can choose to use it without adding a dependency to it.

It is a little hard to imagine a project needing to use this very specific logic, but in the future, maybe there can be another rust based Solana client that wants to implement alpenglow, but does not want to use the `bitvec` crate for their bit-vector implementation. They can choose to import `solana-base3-encoding` without the `bitvec` feature.

It does make the implementation slightly tedious though I agree... If you think working directly with `bitvec` is a more preferable approach, then I can change it back.

base3-encoding/src/lib.rs
samkim-crypto: Yeah that is possible. I guess the code would just be a little more verbose since it requires work to manage the state for the chunk manually. The iteration through chunks seem more readable and the compiler should optimize the chunk loop pretty well, but I can explore more if the manual implementation is faster.

base3-encoding/src/lib.rs
wen-coding: I'm a little bit worried that right now the caller uses bitvec and we need to do conversion, maybe we should just switch back to raw bits in the caller as well when we add base 3 encoding there.

base3-encoding/src/lib.rs
wen-coding: We can check this in for now and optimize if it's a problem.

base3-encoding/src/lib.rs
samkim-crypto: Yeah this is a valid concern. According to the docs, [from_vec](https://docs.rs/bitvec/latest/bitvec/vec/struct.BitVec.html#method.from_vec) and [as_raw_slice](https://docs.rs/bitvec/latest/bitvec/vec/struct.BitVec.html#method.as_raw_slice), which is what we use for the conversion [here](https://github.com/anza-xyz/solana-sdk/pull/200/files#diff-5a635532243f6bed35508bf62d24cf6037b940f2ed0df550921002eed68bcbaaR234) and [here](https://github.com/anza-xyz/solana-sdk/pull/200/files#diff-5a635532243f6bed35508bf62d24cf6037b940f2ed0df550921002eed68bcbaaR234), are in-place functions that does not require any memory allocations, so this should be very cheap. Let's try things out with bitvec and then if it turns out too slow, we can try raw bytes.

PR #201, opened 2025-06-19
Increase quic heartbeat interval to 45 second to reduce overhead

lijunwangs:
As the change to increase QUIC connection timeout has been released. Now we can afford to reduce the frequency to send QUIC keep-alive messages.

The interval is changed from 1 s to 45 seconds

This to follow up on https://github.com/anza-xyz/solana-sdk/pull/29.

### DIFF
--- a/quic-definitions/src/lib.rs
+++ b/quic-definitions/src/lib.rs
@@ -22,9 +22,7 @@ pub const QUIC_MAX_TIMEOUT: Duration = Duration::from_secs(60);
 
 /// To avoid idle timeout, the QUIC endpoint sends a ping every
 /// QUIC_KEEP_ALIVE. This shouldn't be too low to avoid unnecessary ping traffic.
-/// For upgrade purpose, we keep the original one. Once the network is upgraded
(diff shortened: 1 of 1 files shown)


alexpyattaev:
Generally this is a good idea, just for the record, in which version did the increased value for the timeouts on the streamer side land?

lijunwangs:
> Generally this is a good idea, just for the record, in which version did the increased value for the timeouts on the streamer side land?

it is released in solana-quic-definitions 2.3.0. -- which I will update our validator to take in another PR in agave repo. After this PR lands we will release solana-quic-definitions  with another version 2.3.1 or whatever number.

alessandrod (approved):

alessandrod:
why 45s and not 60s or another number?

lijunwangs:
> why 45s and not 60s or another number?

60s could cause timeout due to latency in scheduling and on the wire. 45 s is a heuristic value to ensure with 60 second we see at least 1 timeout and no more than 2.

PR #202, opened 2025-06-20
[bls] Improve signature and proof of possession verification interface so that it works with any projective, affine, and compressed types

samkim-crypto:


joncinque (commented):
The changes look good to me overall! My only question: is it possible to use `TryInto<...>` instead of the new `As...` traits? That would be more idiomatic and require less code

samkim-crypto:
> The changes look good to me overall! My only question: is it possible to use `TryInto<...>` instead of the new `As...` traits? That would be more idiomatic and require less code

Yeah that was what I tried first actually and it is certainly possible to do that. I went with the more specific trait route because:
- I thought using a more specific trait felt more natural to read overall. For example, to define a `.verify()` function for projective, affine, and compressed types at the same time, we have to define `VerifiablePubkey`, `VerifiableSignature`, and `VerifiableProofOfPossession` traits, but to have these traits extend from `TryInto<...>` felt a little unnatural. Having a specific trait like `AsProjective` kind of makes the intent a bit more clear.
- I guess a more practical reason could be extensibility. If a downstream project defines its own representation type for a BLS pubkey `CustomBLSPubkey(...)`, it can easily implement `AsProjective` for the custom pubkey type. But it can't implement `TryFrom<CustomBLSPubkey> for PubkeyProjective` because both `TryFrom` and `PubkeyProjective` are foreign types and rust's orphan rule shouldn't allow this. It can implement `TryInto<PubkeyProjective> for CustomBLSPubkey`, but this does break symmetry and less idiomatic to implement `TryInto` without implementing `TryFrom`.

I guess the two reasons above are relatively small, so if you prefer a `TryInto<...>` approach, then I can take that route as well.

joncinque:
Traits can be annoying to reason about, so my vote would be to use generic functions rather than new traits. For example, you can eliminate `VerifiableProofOfPossession` in favor of a function that works on existing traits, ie:

```
fn verify_proof_of_possession<T: TryInto<ProofOfPosessionProjective>, P: TryInto<PubkeyProjective>>(pop: T, p: P) -> ...
```

The `VerifiableProofOfPossession` trait has a generic implementation over all T anyway, so you're not really gaining anything over a generic function except for more indirection when understanding the code.

As for the extensibility bit, that was relaxed in Rust 1.41. Nowadays you can write:

```
  use solana_pubkey::Pubkey;
  use std::convert::TryFrom;
  struct MyPubkey;

  impl TryFrom<MyPubkey> for Pubkey {
      type Error = ();
      fn try_from(p: MyPubkey) -> Result<Self, Self::Error> {
          Ok(Self::default())
      }
  }
```

See https://doc.rust-lang.org/std/convert/trait.From.html for the info

With that said, I think the error type returned by `TryInto` will be the only annoying part, since you might need to constrain it to return a `BlsError`. Or you could maybe have it work generically too over `<T as TryFrom<U> >::Error`! I probably got that syntax wrong, but I hope the idea is clear.

samkim-crypto:
Oh yeah I actually did not know that this worked! 
```
  use solana_pubkey::Pubkey;
  use std::convert::TryFrom;
  struct MyPubkey;

  impl TryFrom<MyPubkey> for Pubkey {
      type Error = ();
      fn try_from(p: MyPubkey) -> Result<Self, Self::Error> {
          Ok(Self::default())
      }
  }
```

The implementation `impl ForeignTrait<LocalType> for ForeignType { ... }` seems to actually not violate the orphan rule. This is good to know.

Okay, I took a stab at removing the `As...Projective` and `Verifiable...` traits. Let me know what you think. I think the disadvantage of this approach is that (1) the syntax gets a little complicated with lifetimes (I couldn't get this work without explicit lifetimes, but please let me know if you have ideas).
```
    /// Verify the proof of possession against any convertible public key type
    pub fn verify<P>(&self, pubkey: &P) -> Result<bool, BlsError>
    where
        for<'a> &'a P: TryInto<PubkeyProjective>,
        for<'a> <&'a P as TryInto<PubkeyProjective>>::Error: Into<BlsError>,
    { ... }
```

and that (2) to get the ergonomic syntax of being able to call `proof.verify(&pubkey)`, we need to implement things for each of the three type variants. Maybe we can save on some lines using a macro, but seems like that would complicate things even more. But this does remove unnecessary lifetimes, which is good.

After the changes, I do feel that using traits makes the code a bit more readable, so I would marginally prefer using the dedicated traits, but I am happy with this approach as well.

Yet another option would be to keep the `Verifiable...` traits and just replace the `As...Projective` traits with `TryInto`. It seems like this could allow us implement the verify function once for the projective, affine, and compressed variants while being a middle ground for traits.

joncinque:
Sorry, I wasn't clear enough about this. By using `TryInto` even more aggressively and defining free functions `verify_signature` and `verify_proof_of_possession`, we can remove a lot of the replicode. I took a crack at it at https://github.com/joncinque/solana-sdk/tree/bls-interface-jon, let me know what you think!

samkim-crypto:
> Sorry, I wasn't clear enough about this. By using `TryInto` even more aggressively and defining free functions `verify_signature` and `verify_proof_of_possession`, we can remove a lot of the replicode. I took a crack at it at https://github.com/joncinque/solana-sdk/tree/bls-interface-jon, let me know what you think!

Oh yeah sorry, I understand your point. It is just that my preference would be to have an ergonomic interface like
```
pubkey.verify_signature(&signature, &message);
signature.verify(&pubkey, &message);
proof_of_possession.verify(&pubkey);
```

We can definitely define a function first and then have three implementations for each of the three types that take in `&self` to make it more ergonomic. I am just saying that if we define a trait, then we can implement the function for the three types at the same time.

Beside this point, just having a dedicate trait seems to allow us to create a much cleaner function syntax without specific lifetime specifiers, so I'd prefer that direction...

joncinque:
That's fine then, a trait works too!

samkim-crypto:
Okay! I organized the signature verification using traits. I also noticed that the `TryFrom` and `From` implementations between the projective, affine, and compressed types of pubkeys, signatures, and proof of possessions are quite repetitive, so I ended up adding a macro to prevent repetitive code as much as possible. Thanks for your patience and please let me know what you think 🙏!

bls-signatures/src/macros.rs
```diff
@@ -22,3 +22,122 @@ macro_rules! impl_from_str {
...
+        // Fallible conversions from serialized types (affine, compressed)
+        // back to the projective type.
+        // ---
+        impl TryFrom<&$affine> for $projective {
+            type Error = $error_type;
+
+            fn try_from(affine: &$affine) -> Result<Self, Self::Error> {
+                let maybe_point: Option<$point_type> =
+                    <$point_type>::from_uncompressed(&affine.0).into();
+                let point = maybe_point.ok_or(<$error_type>::PointConversion)?;
```
joncinque: This depends on `$error_type` having a `PointConversion` variant. May as well just remove `$error_type` and instead use `BlsError` directly

bls-signatures/src/proof_of_possession.rs
```diff
@@ -30,51 +30,53 @@ pub const BLS_PROOF_OF_POSSESSION_COMPRESSED_BASE64_SIZE: usize = 128;
...
+/// A trait for types that can be converted into a `ProofOfPossessionProjective`.
 #[cfg(not(target_os = "solana"))]
-impl ProofOfPossessionProjective {
-    /// Verify a proof of possession against a public key
-    pub fn verify(&self, public_key: &PubkeyProjective) -> bool {
-        public_key.verify_proof_of_possession(self)
-    }
+pub trait AsProofOfPossessionProjective {
+    /// Attempt to convert the type into a `ProofOfPossessionProjective`.
+    fn try_as_projective(&self) -> Result<ProofOfPossessionProjective, BlsError>;
```
joncinque, line 39: nit-up-to-you: we've been back and forth on this already quite a bit, so it's fine if you want to keep it, but there shouldn't be any need for the `As...` traits, since they just wrap `TryFrom`. The `Verifiable...` traits can then wrap all the conversions and maintain the object-oriented interface.

And on the other hand, if you really want to keep the `As...` traits, may as well add their impl blocks to the impl_bls_conversions macro!

joncinque (commented):
Just a few last things, then this can go in!

bls-signatures/src/macros.rs
samkim-crypto: Yep, good idea! This is updated.

bls-signatures/src/proof_of_possession.rs
samkim-crypto: Yeah I was really torn about this... It is true that `As...` trait is a wrapper for `TryFrom`, but if I use `TryFrom` directly, then the function signature becomes quite verbose

```
where
    for<'a> &'a P: TryInto<PubkeyProjective>,
    for<'a> <&'a P as TryInto<PubkeyProjective>>::Error: Into<BlsError>,
```

and this has to be repeated for all pubkey, signature, and proof of possession types. It just seems like adding a dedicated trait can wrap this complexity out cleanly.

I need to clean up the function signature for the pubkey and signature aggregation as well on a follow-up PR. Maybe some new issue could arise there, so maybe we can have a second look regarding this issue on the follow-up PR?

PR #203, opened 2025-06-20
Up the version for solana-quic-definitions

lijunwangs:
Update solana-quic-definitions to 2.3.1 from 2.3.0 for the QUIC keep-alive change (https://github.com/anza-xyz/solana-sdk/pull/201)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3658,7 +3658,7 @@ dependencies = [
 
 [[package]]
 name = "solana-quic-definitions"
-version = "2.3.0"
+version = "2.3.1"
 dependencies = [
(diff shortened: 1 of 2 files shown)


joncinque:
No need to bump the version manually, the auto-publish job takes care of it. I ran it for you in this case, so you can go ahead and use `solana-quic-definitions` v2.3.1: https://github.com/anza-xyz/solana-sdk/releases/tag/quic-definitions%40v2.3.1

We should consider moving the crate back to Agave, since it has no purpose in the SDK

joncinque:
Closing per comment

PR #205, opened 2025-06-26
Create `solana-epoch-stake` crate

buffalojoec:
#### Problem

I noticed the API that wraps the `sol_get_epoch_stake` syscall still lives in `solana_program`. It would be nice if program devs didn't have to break the beloved á la carte pattern for their dependencies if they wanted to use this syscall API.

#### Summary of Changes

Create a new crate for this API - `solana-epoch-stake` - however, first create a new crate `solana-program-stubs` to decouple it from `solana-sysvar`, allowing the new `solana-epoch-stake` crate to depend on just `solana-program-stubs` instead of all of `solana-sysvar`.

EDIT: We nixed the program-stubs crate. This PR just creates a standalone `solana-epoch-stake` for on-chain usage of the `sol_get_epoch_stake` sysvar.


buffalojoec:
Hey @joncinque let me know if you get a chance to review this!

epoch-stake/src/lib.rs
```diff
@@ -4,14 +4,14 @@
...
+use solana_pubkey::Pubkey;
 
 fn get_epoch_stake(var_addr: *const u8) -> u64 {
     #[cfg(target_os = "solana")]
-    let result = unsafe { crate::syscalls::sol_get_epoch_stake(var_addr) };
+    let result = unsafe { solana_define_syscall::definitions::sol_get_epoch_stake(var_addr) };
 
     #[cfg(not(target_os = "solana"))]
-    let result = crate::program_stubs::sol_get_epoch_stake(var_addr);
+    let result = solana_program_stubs::sol_get_epoch_stake(var_addr);
```
joncinque: We've avoided having `program_stubs` used outside of `solana_program` and `solana_sysvar`, since it's a bit rough to use and shouldn't be encouraged.

In this case, some sort of fall-through or black_box would be better, and we can leave a version of `get_epoch_stake` that uses program-stubs in solana-program

buffalojoec: Okay, well in that case there's no need to decouple it from `solana-program` in this PR.

buffalojoec:
@joncinque I decided to just make this on-chain only, similar to how much of the Pinocchio API works. If you insist on the black-box/fall-through let me know. However, I felt like this API actually made sense to be on-chain only, given its purpose.

joncinque:
The pinocchio APIs will typically have a `black_box` for the off-chain behavior, ie https://github.com/anza-xyz/pinocchio/blob/3fecb3e569442e219185e5fe35854bfbf971e23a/sdk/pinocchio/src/pubkey.rs#L149

I'd like to do that here as well for consistency, or to have it panic in the offchain case. Otherwise, users of the crate need to add `target_os = "solana"` to all uses of this crate.

joncinque (approved):
Looks great! I reserved `solana-epoch-stake` -- @yihau can you accept ownership of the crate?

PR #206, opened 2025-06-27
Implement `VoteStateV4`

buffalojoec:
Includes the new interface, full serialization support, and the flipping of "current" to be v4.


vote-interface/src/state/vote_state_v3.rs
```diff
@@ -241,9 +200,33 @@ impl VoteStateV3 {
...
+                                .map(|versioned| versioned.convert_to_current())
+                                .map_err(|_| InstructionError::InvalidAccountData)?,
+                        );
+                    }
+                    Ok(())
+                }
+                #[cfg(target_os = "solana")]
+                {
+                    // For SBPF targets, for now we're just going to bail.
+                    Err(InstructionError::InvalidAccountData)
```
buffalojoec: This one I'm not sure about, but we chatted offline @jstarry. The alternative is to modify the `deserialize_vote_state_into_v3` function to handle v4 state in sBPF context.

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -89,31 +162,34 @@ impl VoteStateVersions {
...
 
+            VoteStateVersions::V3(vote_state) => vote_state.authorized_voters.is_empty(),
             VoteStateVersions::Current(vote_state) => vote_state.authorized_voters.is_empty(),
         }
     }
 
     pub fn vote_state_size_of(is_current: bool) -> usize {
         if is_current {
-            VoteState::size_of()
+            VoteStateV4::size_of() // Same as v3
```
jstarry: The meaning of "current" has changed so this doesn't look right. Maybe we can get away with renaming `is_current` to `is_expanded` and have that mean either v3 or v4?

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -89,31 +162,34 @@ impl VoteStateVersions {
 
             VoteStateVersions::V1_14_11(vote_state) => vote_state.authorized_voters.is_empty(),
 
+            VoteStateVersions::V3(vote_state) => vote_state.authorized_voters.is_empty(),
             VoteStateVersions::Current(vote_state) => vote_state.authorized_voters.is_empty(),
         }
```
jstarry: the vote state v4 variant cannot exist in an uninitialized state so we can return false here

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -73,6 +78,74 @@ impl VoteStateVersions {
...
+                root_slot: state.root_slot,
+                authorized_voters: state.authorized_voters,
+                prior_voters: CircBuf::default(),
+                epoch_credits: state.epoch_credits,
+                last_timestamp: state.last_timestamp,
+            },
+        }
+    }
+
+    pub fn convert_to_current(self) -> VoteStateV4 {
```
jstarry: How about naming this `convert_to_v4`? Then we don't have to break the API for each new version

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -73,6 +78,74 @@ impl VoteStateVersions {
                 last_timestamp: state.last_timestamp,
             },
 
+            VoteStateVersions::V3(state) => *state,
+            VoteStateVersions::Current(state) => VoteStateV3 {
```
jstarry: We never want to convert from v4 to v3 because it would be a lossy conversion. Would it make sense to return an `Option` or `Result` here?

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -0,0 +1,200 @@
...
+    ) -> Result<(), InstructionError> {
+        bincode::serialize_into(output, versioned).map_err(|err| match *err {
+            bincode::ErrorKind::SizeLimit => InstructionError::AccountDataTooSmall,
+            _ => InstructionError::GenericError,
+        })
+    }
+
+    pub fn is_correct_size_and_initialized(data: &[u8]) -> bool {
+        const VERSION_OFFSET: usize = 4;
+        data.len() == VoteStateV4::size_of() && data.get(VERSION_OFFSET).copied() == Some(3)
```
jstarry: I prefer this to match all of the first 4 version bytes. If for some reason we used the first few byes for something else in the future, this might erroneously match

vote-interface/src/state/mod.rs
```diff
@@ -22,9 +22,13 @@ pub use vote_state_1_14_11::*;
 pub mod vote_state_versions;
 pub use vote_state_versions::*;
 pub mod vote_state_v3;
-pub use vote_state_v3::VoteState;
+pub use vote_state_v3::VoteStateV3;
```
jstarry, line 25: Should we do an intermediate minor release which deprecates `VoteState` and introduces `VoteStateV3`?

vote-interface/src/state/vote_state_v3.rs
jstarry: I think it's best to not allow converting v3 to v4 because it's lossy

vote-interface/src/state/vote_state_v4.rs
buffalojoec: Ok can do. I think I just copied from the other one.

vote-interface/src/state/vote_state_versions.rs
buffalojoec: I'd prefer `Result` and rename it to `try_convert_to_v3`.

vote-interface/src/state/vote_state_versions.rs
buffalojoec: Yes, can do, but if we anticipate future versions, there's probably a fair amount of further refactoring I can do to make this less painful for v5 and beyond.

Just a quick and dirty idea is to have a small enum like this, and just roll a `try_convert_to` generic function:

```rust
enum SupportedVersions {
    #[deprecated(..)]
    V3,
    V4,
    V5,
}
```

I _think_ you can deprecate variants(?).

vote-interface/src/state/vote_state_versions.rs
buffalojoec: Yeah that's a good point. You could just make the argument `is_v3_or_v4`, but that's starting to feel a little overloaded. We could just move this to the vote state layouts themselves, ie. `VoteStateV3::size_of` and `VoteStateV4::size_of`.

buffalojoec: I think that's probably best, and just make them constants.

vote-interface/src/state/mod.rs
buffalojoec: Yeah we can do that. I did this in separate commits so that would be very easy to cherry pick.

buffalojoec: I put this up as a separate PR to see what the best flow would be: #221

vote-interface/src/state/vote_state_versions.rs
jstarry: Hmm could be interesting.. I think I prefer specific conversion functions for each version though

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -76,24 +77,14 @@ impl VoteStateVersions {
...
-                epoch_credits: state.epoch_credits,
-                last_timestamp: state.last_timestamp,
-            },
+            VoteStateVersions::V3(state) => Ok(*state),
+            VoteStateVersions::V4(_) => Err(InstructionError::InvalidArgument),
         }
     }
 
-    pub fn convert_to_current(self) -> VoteStateV4 {
+    pub fn convert_to_v4(self) -> VoteStateV4 {
```
jstarry: should probably be `try_` prefixed too because adding v5 would cause this to have a failure condition in the future

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -163,15 +154,7 @@ impl VoteStateVersions {
...
-            VoteStateVersions::Current(vote_state) => vote_state.authorized_voters.is_empty(),
-        }
-    }
-
-    pub fn vote_state_size_of(is_current: bool) -> usize {
-        if is_current {
-            VoteStateV4::size_of() // Same as v3
-        } else {
-            VoteState1_14_11::size_of()
+            VoteStateVersions::V4(_) => false,
```
jstarry, line 166: Can you add a comment explaining why `false` is sufficient here?

vote-interface/src/state/vote_state_versions.rs
buffalojoec: Ok, sure, we can just make this the de facto API for now.

buffalojoec:
@jstarry addressed all your feedback and also added updates based on SIMD changes.

In order to pull this out of draft, I'll have to do a rebase, where I'll resolve the conflicts and probably also rebase on top of #221.

vote-interface/src/instruction.rs
```diff
@@ -272,7 +272,7 @@ pub struct CreateVoteAccountConfig<'a> {
 impl Default for CreateVoteAccountConfig<'_> {
     fn default() -> Self {
         Self {
-            space: VoteStateVersions::vote_state_size_of(false) as u64,
+            space: VoteState1_14_11::size_of() as u64,
```
jstarry: We should default to the "current" vote state size now while we're in the area. We only needed to create vote accounts with the smaller size before the latency feature gate was activated.

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -0,0 +1,320 @@
...
+    /// Compressed BLS pubkey for Alpenglow.
+    #[cfg_attr(feature = "serde", serde(with = "serde_bls_pubkey_compressed"))]
+    pub bls_pubkey_compressed: Option<[u8; BLS_PUBKEY_COMPRESSED_BYTES]>,
+
+    pub votes: VecDeque<LandedVote>,
+    pub root_slot: Option<Slot>,
+
+    /// The signer for vote transactions.
+    /// Unlike `VoteStateV3`, contains entries for the current epoch and the
+    /// previous epoch.
```
jstarry:
```suggestion
    /// Unlike `VoteStateV3`, contains entries for the previous epoch
```
v3 has an entry for the current epoch already

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -0,0 +1,320 @@
...
+    pub fn is_correct_size_and_initialized(data: &[u8]) -> bool {
+        data.len() == VoteStateV4::size_of() && data[..4] == [3, 0, 0, 0] // little-endian 3u32
+                                                                          // Always initialized
+    }
+}
+
+// Based off of serde's array impls. See:
+// https://github.com/serde-rs/serde/blob/babafa54d283fb087fa94f50a2cf82fa9e582a7c/serde/src/de/impls.rs#L1268
+#[cfg(feature = "serde")]
+mod serde_bls_pubkey_compressed {
```
jstarry: I think it's best to directly impl serde traits on a struct wrapped `[u8; 48]` type rather than this option wrapped type

vote-interface/src/state/mod.rs
```diff
@@ -518,8 +649,12 @@ mod tests {
             // so we only check that the parser does not panic and that it succeeds or fails exactly in line with bincode
             let mut test_vote_state = MaybeUninit::uninit();
             let test_res = VoteStateV3::deserialize_into_uninit(&raw_data, &mut test_vote_state);
+
+            // `deserialize_into_uninit` will eventually call into
+            // `try_convert_to_v3`, so we have alignment in the following map.
```
jstarry: We don't call `try_convert_to_v3` for all variants downstream of `deserialize_into_uninit`, not sure we need this comment

vote-interface/src/state/mod.rs
```diff
@@ -481,18 +566,64 @@ mod tests {
...
+            .unwrap();
             let test_vote_state = unsafe { test_vote_state.assume_init() };
 
             assert_eq!(target_vote_state, test_vote_state);
         }
     }
 
     #[test]
-    fn test_vote_deserialize_into_uninit_nopanic() {
+    fn test_vote_deserialize_into_uninit_nopanic_v3() {
```
jstarry, line 626: should probably change the random tag to `let tag = rng.gen::<u8>() % 4` to cover v4

vote-interface/src/state/mod.rs
```diff
@@ -531,7 +666,50 @@ mod tests {
...
+        // variant
+        let serialized_len_x4 = serialized_size(&VoteStateV4::default()).unwrap() * 4;
+        let mut rng = rand::thread_rng();
+        for _ in 0..1000 {
+            let raw_data_length = rng.gen_range(1..serialized_len_x4);
+            let mut raw_data: Vec<u8> = (0..raw_data_length).map(|_| rng.gen::<u8>()).collect();
+
+            // pure random data will ~never have a valid enum tag, so lets help it out
+            if raw_data_length >= 4 && rng.gen::<bool>() {
+                let tag = rng.gen::<u8>() % 3;
```
jstarry:
```suggestion
                let tag = rng.gen::<u8>() % 4;
```

vote-interface/src/state/mod.rs
```diff
@@ -542,6 +720,56 @@ mod tests {
                 VoteStateVersions::arbitrary(&mut unstructured).unwrap();
             let original_buf = bincode::serialize(&original_vote_state_versions).unwrap();
 
+            // Skip any v4 since they can't convert to v3.
+            if original_vote_state_versions.try_convert_to_v3().is_ok() {
```
jstarry: I wouldn't assume that the only v3 conversion failure is from v4. How about checking with `matches!`?

vote-interface/src/state/mod.rs
```diff
@@ -907,7 +1139,7 @@ mod tests {
     #[test]
     fn test_vote_state_size_of() {
         let vote_state = VoteStateV3::get_max_sized_vote_state();
-        let vote_state = VoteStateVersions::new_current(vote_state);
+        let vote_state = VoteStateVersions::new_v3(vote_state);
```
jstarry, line 1141: for completeness we should assert that the max sized v4 state is less than its `size_of` value

vote-interface/src/state/vote_state_v3.rs
```diff
@@ -241,9 +200,33 @@ impl VoteStateV3 {
...
             }
             // V1_14_11. substantially different layout and data from V0_23_5
-            1 => deserialize_vote_state_into(&mut cursor, vote_state, false),
-            // Current. the only difference from V1_14_11 is the addition of a slot-latency to each vote
-            2 => deserialize_vote_state_into(&mut cursor, vote_state, true),
+            1 => deserialize_vote_state_into_v3(&mut cursor, vote_state, false),
+            // V3. the only difference from V1_14_11 is the addition of a slot-latency to each vote
+            2 => deserialize_vote_state_into_v3(&mut cursor, vote_state, true),
+            // Current (V4). Convert to V3 through bincode
+            3 => {
```
jstarry: I thought we weren't deserializing v4 to v3? I don't think we need this arm at all

vote-interface/src/state/vote_state_v4.rs
buffalojoec: I think I just worded this a bit weird. Updated.

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -0,0 +1,242 @@
...
+#[derive(Debug, Default, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "dev-context-only-utils", derive(Arbitrary))]
+pub struct VoteStateV4 {
+    /// The node that votes in this account.
+    pub node_pubkey: Pubkey,
+    /// The signer for withdrawals.
+    pub authorized_withdrawer: Pubkey,
+
+    /// The collector account for inflation rewards.
+    pub inflation_rewards_collector: Pubkey,
```
HaoranYi, line 37: so, the intention is that we will update vote rewards payment to this account for votestatev4?
will this be defaulted to vote-pubkey?

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -0,0 +1,242 @@
...
+    /// The collector account for block revenue.
+    pub block_revenue_collector: Pubkey,
+
+    /// Basis points (0-10,000) that represent how much of the inflation
+    /// rewards should be given to this vote account.
+    pub inflation_rewards_commission_bps: u16,
+    /// Basis points (0-10,000) that represent how much of the block revenue
+    /// should be given to this vote account.
+    pub block_revenue_commission_bps: u16,
```
HaoranYi, line 46: 👍

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    /// Reward amount pending distribution to stake delegators.
+    pub pending_delegator_rewards: u64,
```
HaoranYi, line 49: 👍

vote-interface/src/state/vote_state_v4.rs
buffalojoec: Yep! https://github.com/anza-xyz/solana-sdk/pull/206/files#diff-bf0a4efa24b4c64ebb5e1e8116bbe64dfbe1d0f21eebb6c1d9db026bf7dd145eR100

jstarry:
Do you mind doing the deserialization code move (into new module) in a separate PR to reduce the diff size of this one and made it easier to review?

buffalojoec:
> Do you mind doing the deserialization code move (into new module) in a separate PR to reduce the diff size of this one and made it easier to review?

Yeah sure, can do. I also am going to consider refactoring or replacing `VoteStateVersions`. I'll move this into draft for now.

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    /// Compressed BLS pubkey for Alpenglow.
+    #[cfg_attr(
+        feature = "serde",
+        serde_as(as = "Option<[_; BLS_PUBKEY_COMPRESSED_BYTES]>")
+    )]
+    pub bls_pubkey_compressed: Option<[u8; BLS_PUBKEY_COMPRESSED_BYTES]>,
```
ksn6, line 56: For Alpenglow, we'll want to have a `bls_pubkey_compressed` associated with each authorized voter. This way, an authorized voter change results in an appropriate change to the corresponding BLS pubkey.

Accordingly, we'll probably either want to:

(1) rewrite `bls_pubkey_compressed` as `bls_pubkeys: Option<BTreeMap<Epoch, [u8; BLS_PUBKEY_COMPRESSED_BYTES]>>`

(2) modify `AuthorizedVoters` to have the value of the `BTreeMap` include `Option<[u8; BLS_PUBKEY_COMPRESSED_BYTES]>`.

Slightly partial to option (1), as it seems a bit cleaner / more space-efficient.

ksn6: Not suggesting that we change within this SIMD, though - why do we need to keep around a `BTreeMap<Epoch, _>` rather than "current epoch" and "next epoch?"

jstarry: Oh this is news to me. Sounds like we will want to amend SIMD-0185 with this change. And I'm not opposed to removing the btreemap. In SIMD-0185, I also added an entry for the "previous epoch" authorized voter as well so that when transitioning into a new epoch we can still process votes from the previous authorized voter because votes for slots near the end of the epoch signed by the previous authorized voter might land in the next epoch. But I'm thinking this isn't necessary for alpenglow?

0x0ece: @samkim-crypto I think we also need a proof of ownership for the BLS pubkey (a BLS signature):
```
pub const BLS_SIGNATURE_COMPRESSED_BYTES: usize = 96;
```

Two options are:
1. we add it to the vote account
2. we verify it in the tx that adds/updates the bls pubkey

(1 doesn't exclude 2)
I think it's cleaner to do it in 2), but it won't be possible in BPF without a syscall, so the vote program needs to remain native (until at least we add the syscall).

jstarry: @0x0ece fyi we are discussing some of this in the #consensus discord channel

samkim-crypto: Yes, we do need to verify BLS pubkeys. @0x0ece can you expand on adding the verification to the vote account? Option 2 is essentially required, no?

0x0ece: Yes I agree option 2 is better. (I was thinking vote program in bpf, no syscall to do bls sigverify, then we could store the proof in the vote account and do the verification when we boot / load the keys. Unnecessary complication.)

buffalojoec:
I'll be breaking this PR up and handling the evolving state changes in other PRs.

PR #208, opened 2025-06-30
CI: Fixup git-cliff step for non-master branches

joncinque:
#### Problem

The publish job against maintenance/v2.x failed to generate a changelog because we look for changes to `master`, but we really want changes up to the current HEAD: https://github.com/anza-xyz/solana-sdk/actions/runs/15971057246/job/45042369237

#### Summary of changes

Use `HEAD` instead of `master`.

NOTE: this isn't technically needed because it's targeting master, but the next PR will target the maintenance branch.


yihau (approved):
🪖

PR #209, opened 2025-06-30
v2: CI: Fixup git-cliff step for non-master branches

joncinque:
#### Problem

The publish job against maintenance/v2.x failed to generate a changelog because we look for changes to `master`, but we really want changes up to the current HEAD.

#### Summary of changes

Use `HEAD` instead of `master`.


yihau (approved):
🪖

PR #210, opened 2025-06-30
Bump Rust to 1.88, nightly to 2025-06-29

joncinque:
#### Problem

Rust has been upgrading, but not the version used in the sdk repo. Also, with the upcoming breaking changes to the sdk, it's a good time to bump the rust edition to 2024, but we should upgrade to the newest Rust first.

#### Summary of changes

Bump the version in rust-toolchain.toml and scripts/rust-version.sh, then fix any issues reported.


apfitzge (approved):

febo (approved):
Looks good!

PR #211, opened 2025-06-30
account / account-info / rent-collector: Remove `rent_epoch`

joncinque:
#### Problem

The concept of rent-epoch doesn't exist anymore in the runtime, but the base account type in Agave and account-info type for on-chain programs still have the field.

#### Summary of changes

Remove the field, and clean up RentCollector at the same time.

BREAKING: a few traits and types are broken with this change

* `InheritableAccountFields` is removed in favor of a u64
* `create_account_with_fields` -> `create_account_with_lamports`
* `solana_account_info::Account` trait removes `Epoch`
* The entrypoint still assumes that rent-epoch is provided
* `RentCollector` reduced to the only parts needed by Agave, most functions were unused

I'm aware that this might be controversial, but I wanted to get some eyes on it as soon as possible if we do want to make the change.

### DIFF
--- a/genesis-config/src/lib.rs
+++ b/genesis-config/src/lib.rs
@@ -53,7 +53,7 @@ pub const UNUSED_DEFAULT: u64 = 1024;
 #[cfg_attr(
     feature = "frozen-abi",
     derive(AbiExample),
-    frozen_abi(digest = "D9VFRSj4fodCuKFC9omQY2zY2Uw8wo6SzJFLeMJaVigm")
+    frozen_abi(digest = "FbARREDu5nb5nuTJKEeATfZUVtdGU98aSwH3hpPhFzBU")
 )]
 #[cfg_attr(
     feature = "serde",


account-info/src/lib.rs
```diff
@@ -331,7 +320,7 @@ impl<'a, T: Account> IntoAccountInfo<'a> for &'a mut (Pubkey, T) {
 /// # let p = Pubkey::new_unique();
 /// # let l = &mut 0;
 /// # let d = &mut [0u8];
-/// # let a = AccountInfo::new(&p, false, false, l, d, &p, false, 0);
+/// # let a = AccountInfo::new(&p, false, false, l, d, &p, false, );
```
jstarry, line 323: nit: remove trailing comma (also on line 373)

account/src/lib.rs
```diff
@@ -51,8 +50,6 @@ pub struct Account {
     pub owner: Pubkey,
     /// this account's data contains a loaded program (and is now read-only)
     pub executable: bool,
-    /// the epoch at which this account will next owe rent
-    pub rent_epoch: Epoch,
```
jstarry, line 55: I don't think this is a good idea because any account serialized with `solana-account:v3` would not be deserializable by `solana-account:v2` and possibly vice versa as well. This presents a problem in how we handle the genesis config in agave and likely would cause problems for downstream users as well.

account/src/lib.rs
```diff
@@ -142,8 +136,6 @@ pub struct AccountSharedData {
     owner: Pubkey,
     /// this account's data contains a loaded program (and is now read-only)
     executable: bool,
-    /// the epoch at which this account will next owe rent
-    rent_epoch: Epoch,
```
jstarry, line 146: Similarly, we cannot make this breaking change to `AccountSharedData` because the serialized stakes cache and epoch stakes in snapshots both rely on serialization being stable.

account-info/src/lib.rs
```diff
@@ -28,8 +28,6 @@ pub struct AccountInfo<'a> {
     pub data: Rc<RefCell<&'a mut [u8]>>,
     /// Program that owns this account
     pub owner: &'a Pubkey,
-    /// The epoch at which this account will next owe rent
-    pub rent_epoch: u64,
```
jstarry, line 32: I'm in favor of removing rent epoch from `AccountInfo` because programs shouldn't be reading this value anymore. Also we don't directly serialize or deserialize this struct. So I think it should be fine but I'm curious if you've been able to think of any potential downstream issues with this?

program-entrypoint/src/lib.rs
```diff
@@ -444,7 +444,7 @@ unsafe fn deserialize_account_info<'a>(
     offset += (offset as *const u8).align_offset(BPF_ALIGN_OF_U128); // padding
 
     #[allow(clippy::cast_ptr_alignment)]
-    let rent_epoch = *(input.add(offset) as *const u64);
+    let _rent_epoch = *(input.add(offset) as *const u64);
```
jstarry, line 447: why not just remove this?

rent-collector/src/lib.rs
```diff
@@ -1,15 +1,14 @@
 //! Calculate and collect rent from accounts.
+//!
+//! The `RentCollector` type is only used for its `rent` field by Agave, but
```
jstarry, line 3: I think we should EOL the `solana-rent-collector` crate in the SDK and introduce a new crate or module somewhere in the agave repo if this is just used internally. And I think this change should be independently of the `rent_epoch` account field changes

account-info/src/lib.rs
joncinque: Ah thanks, I thought cargo fmt would take care of that for me

account/src/lib.rs
joncinque: We could write an adapters between the two types, where we fill in u64::MAX for the rent-epoch if doing v3 -> v2, and ignore rent-epoch when doing v2 -> v3. But I do think the concern is valid

account-info/src/lib.rs
joncinque: I couldn't think of anything, but I tagged @Lichtso to see if there might be issues in the VM, maybe with the `sol_invoke_signed_rust` syscall

program-entrypoint/src/lib.rs
joncinque: The rent-epoch is still part of the serialization format expected by on-chain programs, so we need to add those extra 8 bytes to make sure our pointer is in the right place

rent-collector/src/lib.rs
joncinque: That should be easy to do with sdk v3, since we've removed the rent-collector re-export in solana-sdk.

program-entrypoint/src/lib.rs
joncinque: oh never mind, I thought this was modifying offset at the same time, you're right!

account/src/lib.rs
brooksprumo: Echoing jstarry. Anything that is repr-C probably cannot be modified.

It may be the case that `Account` is not actually ser/de, but rather `account_serialize::Account` is always used instead. If that's actually true, we can remove repr-C here, and then modify the custom de/serialization in `mod account_serialize` to handle adapting between the two.

genesis-config/src/lib.rs
```diff
@@ -53,7 +53,7 @@ pub const UNUSED_DEFAULT: u64 = 1024;
 #[cfg_attr(
     feature = "frozen-abi",
     derive(AbiExample),
-    frozen_abi(digest = "D9VFRSj4fodCuKFC9omQY2zY2Uw8wo6SzJFLeMJaVigm")
+    frozen_abi(digest = "FbARREDu5nb5nuTJKEeATfZUVtdGU98aSwH3hpPhFzBU")
```
brooksprumo, line 56: I'm guessing the digest changed here due to `Account` changing?

account-info/src/lib.rs
febo: I think we can't remove the field from the struct, since it will change the size of the type. This will break the account translation [here](https://github.com/anza-xyz/agave/blob/master/programs/bpf_loader/src/syscalls/mod.rs#L618) for CPI. This is also true for `sol_invoke_signed_c`. We could make the field private and renamed it.

Lichtso: `AccountInfo` has to stay as is, changing it is an ABI breaking change.

joncinque: Ok great, thanks for the feedback!

What if we renamed it to `unused` and avoided deserializing it from the entrypoint? Meaning, does the value inside it need to be correct for CPI? I don't see it being used anywhere, but I want to be sure.

Or if you think it's silly, feel free to let me know

febo: Pinocchio does not deserialize the `rent_epoch` on the entrypoint and sets it to `0` for CPI.

joncinque:
Thanks to everyone for the feedback here! To sum everything up:

* account can't be touched currently, unless there's a lot of work done in agave to create a separate type. I'll leave it as is
* account-info cannot change size, so I'll put in a PR to deprecate the rent-epoch field, then a second (breaking) change to rename it to `unused`
* rent-collector can just go away, so I'll put in a PR for that

I will tag the appropriate people for those two follow-up PRs.

PR #213, opened 2025-07-02
[bls-signature] Update the public key and signature aggregation function to be generic

samkim-crypto:
#### Problem
The signature and proof of possession verification interface was made more generic and flexible in https://github.com/anza-xyz/solana-sdk/pull/202, but the public key and signature aggregation still only works for the `PubkeyProjective` and `SignatureProjective` types.

#### Summary of Changes

I made two improvements to the BLS aggregation interface:

[1736017](https://github.com/anza-xyz/solana-sdk/pull/213/commits/17360174ed800deb2bdc1fba7deedd3abdeea0b3): The `aggregate` and `aggregate_with` functions for `PubkeyProjective` and `SignatureProjective` have been updated to support any of projective, affine, and compressed types. I added a `?Sized` bound to also support `Vec<&dyn AsPubkeyProjective>` and `Vec<&dun AsSignatureProjective>` input as well, so this should be quite ergonomic and flexible to use now.

https://github.com/anza-xyz/solana-sdk/pull/213/commits/91f32b59a162042ab309e4446cbe446e6e077fb3: The `impl Default` for `PubkeyProjective` and `SignatureProjective` has been removed. I originally added the Default implementation to work as a starting point for aggregation: a user can run `let mut agg = PubkeyProjective::default()` and then start aggregating public keys from there. But `Default` is a bit misleading because the identity element of the group is not actually a valid public key or signature and should not really be treated as a "default". Instead, I added a new `identity()` constructor on both types to make the intent clear that this function should only be used as a starting point for aggregation.

### DIFF
--- a/bls-signatures/src/pubkey.rs
+++ b/bls-signatures/src/pubkey.rs
@@ -262,7 +262,7 @@ mod tests {
             signature::{Signature, SignatureCompressed},
         },
         core::str::FromStr,
-        std::{string::ToString, vec, vec::Vec},
+        std::{string::ToString, vec::Vec},
     };
 
     #[test]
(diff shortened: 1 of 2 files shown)


joncinque (approved):
Clear and clean, looks great!

PR #216, opened 2025-07-03
[bn254] Refactor out integration tests

samkim-crypto:
#### Problem
As I was doing a pass over the bn254 crate, I noticed that the test suite was a bit cluttered. The big JSON data blobs are embedded directly in the source code, making the test functions hard to read. Also, all our tests for both internal logic and the public API are mixed together in the src directory. 

#### Summary of Changes
Moved out all the test data: I pulled the huge JSON strings out of lib.rs and compression.rs and placed them into their own files in a new tests/data/ directory. The tests now just load them with include_str!.

Created a proper integration test: I moved all the tests that check the public functions into a new tests/curve_operations.rs file and left only a few of the smaller, internal-logic checks in src/ as unit tests. I think this is a more idiomatic way to do things in Rust.


bn254/src/compression.rs
```diff
@@ -483,6 +403,9 @@ mod tests {
 
         test_cases.iter().for_each(|test| {
             let input = array_bytes::hex2bytes_unchecked(&test.input);
+            if input.len() < 192 {
+                return;
+            }
```
LucasSte: Has the testing changed in any way by removing the input? I just wonder why you needed to check the length now.

LucasSte (commented):
This is not something needed for this PR, but a suggestion for the future. I believe having separate json files for each test case is neater.

bn254/src/compression.rs
samkim-crypto: Yeah sorry for the confusion. I completely forgot about this myself and I should have mentioned it in the PR description. Before, there were two test cases, one for the `alt_bn128_pairing_test` in `lib.rs` and one for the `alt_bn128_compression_pairing_test_input` in `compression.rs`. They are actually the same set of tests except that the `alt_bn128_pairing_test` had an [empty](https://github.com/anza-xyz/solana-sdk/pull/216/files#diff-6ae2a2cde7e2bb582e7dd42bf93edffc1720aacf1225b0d75283333ec8f27770R45) case that does not make sense for compression. I ended up merging the two test cases into a single data file that include the empty case and then added a check to skip the empty case in the compression tests. I wrote a comment to clarify this.

samkim-crypto:
> This is not something needed for this PR, but a suggestion for the future. I believe having separate json files for each test case is neater.

Yep makes sense! I am planning on writing more exhaustive tests after this refactor, so maybe I can experiment with having a separate json file for each test case then.

LucasSte (approved):

PR #219, opened 2025-07-04
[big-mod-exp] Refactor test data

samkim-crypto:
#### Summary of Changes
This is an analogue of https://github.com/anza-xyz/solana-sdk/pull/216 for the `big-mod-exp` crate. I refactored the test data and also moved the bench logic that lived in `sdk/benches` to the `big-mod-exp` crate. I separated out the test data that are used in benches to `benches/data` as well to clean up the bench logic as well.

### DIFF
--- a/big-mod-exp/benches/big_mod_exp_bench.rs
+++ b/big-mod-exp/benches/big_mod_exp_bench.rs
@@ -52,38 +52,38 @@ fn all_benches(c: &mut Criterion) {
 
     group_exp_3.bench_function("512 bits odd", |b| {
         b.iter(|| {
-            big_mod_exp(
-                black_box(&const_exp_data.bits_512.base),
-                black_box(&exponent_3),
-                black_box(&const_exp_data.bits_512.modulus_odd),
-            )
+            black_box(big_mod_exp(
+                &const_exp_data.bits_512.base,
+                &exponent_3,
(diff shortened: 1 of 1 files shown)


big-mod-exp/benches/big_mod_exp_bench.rs
```diff
@@ -0,0 +1,157 @@
...
+    let exponent_3 = [3u8];
+    let const_exp_data = &data.exp_3_and_65537;
+
+    group_exp_3.bench_function("512 bits odd", |b| {
+        b.iter(|| {
+            big_mod_exp(
+                black_box(&const_exp_data.bits_512.base),
+                black_box(&exponent_3),
+                black_box(&const_exp_data.bits_512.modulus_odd),
+            )
```
LucasSte: Maybe you could wrap the function call in a `black_box`?

```
            black_box(big_mod_exp(
               &const_exp_data.bits_512.base,
               &exponent_3,
               &const_exp_data.bits_512.modulus_odd,
            ))
```

Does this work? It is conciser.

samkim-crypto: Ah thanks for the catch! For some reason I couldn't get the outer wrap to work, but probably due to some trivial mistake that I made somewhere else. It works now, so it is updated!

PR #221, opened 2025-07-09
vote-interface: rename `VoteState` to `VoteStateV3`

buffalojoec:
In preparation for introducing a new vote state (v4), this PR renames `vote_state_v3::VoteState` to `vote_state_v3::VoteStateV3`.

Normally, the module path could serve as the differentiator (ie. `vote_state_v3::VoteState` and `vote_state_v3::VoteState`), but that's brittle and we re-exported `VoteState`.

@joncinque I wanted to get your opinion on how best to handle this breaking change, as well as subsequent breaking changes to `VoteStateVersions`. We could introduce this change here as a minor release, where we mark the previous re-export as deprecated:

```rust
#[deprecated(..)]
pub use vote_state_v3::VoteStateV3 as VoteState;
pub use vote_state_v3::VoteStateV3;
```

However, since we plan to nuke `VoteStateVersions`, I wonder if the whole thing is ok to do in a major.

See #206

### DIFF
--- a/vote-interface/src/state/mod.rs
+++ b/vote-interface/src/state/mod.rs
@@ -25,6 +25,9 @@ pub mod vote_state_v3;
 pub use vote_state_v3::VoteStateV3;
 mod vote_instruction_data;
 pub use vote_instruction_data::*;
+// The struct's name has changed.
+#[deprecated(since = "2.2.6", note = "Use vote_state_v3::VoteStateV3 instead")]
+pub use vote_state_v3::VoteStateV3 as VoteState;
 
 // Maximum number of votes to keep around, tightly coupled with epoch_schedule::MINIMUM_SLOTS_PER_EPOCH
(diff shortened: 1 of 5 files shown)


buffalojoec:
> I don't see how `VoteStateVersions` fits into the rename here, but it doesn't seem too important for this PR. Is that correct?

I was alluding to the fact that everything we do to `VoteStateVersions` is also going to be a breaking change. So, I was wondering if it wasn't totally insane to just do all of the breakage in one major. There's not really a good deprecation strategy for adding a new variant to `VoteStateVersions`.

However, since the first change in Agave and likely most annoying breakage for everyone downstream is this rename, I agree with your points above.

joncinque (approved):
Ah gotcha, yeah we can certainly remove `VoteStateVersions` in the next major release!

We should just deprecate it first, similar to what this PR is doing.

PR #222, opened 2025-07-10
docs: Fix all warnings in docs build

joncinque:
#### Problem

Although we check that all of the rustdocs build correctly for the whole sdk, there are warnings in the build.

#### Summary of changes

Fix the warnings, and forbid warnings in CI.

Closes #152

cc @kevinheavey thanks for suggesting this!


bls-signatures/src/hash.rs
```diff
@@ -6,7 +6,7 @@ use {
 /// Domain separation tag used for hashing messages to curve points to prevent
 /// potential conflicts between different BLS implementations. This is defined
 /// as the ciphersuite ID string as recommended in the standard
-/// https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#section-4.2.1.
+/// <https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#section-4.2.1>.
```
samkim-crypto: Maybe we can change this and the one below in bls-signatures to a named link as well for consistency, but I can probably just take care of that on a next PR on BLS.

samkim-crypto (approved):
Looks good to me!

bls-signatures/src/hash.rs
joncinque: I had to rebase anyway, so I made the change, let me know what you think!

samkim-crypto (approved):
Looks good to me!

PR #224, opened 2025-07-11
docs: Fix tests, run tests in CI

joncinque:
#### Problem

The doctests are currently very broken, but it wasn't detected because we only run checks on the docs.

#### Summary of changes

Fix all of the doctests, and add a step to test them in CI.


febo (approved):
Looks great!

PR #225, opened 2025-07-11
account-info: Rename `rent_epoch` to `unused`

joncinque:
#### Problem

The concept of rent-epoch doesn't exist anymore in the runtime, but the account-info type for on-chain programs requires some 8-byte field where the rent-epoch is currently.

#### Summary of changes

Rename the field, skip it in the program entrypoint.

BREAKING: `solana_account_info::Account` trait removes `Epoch`


account-info/src/lib.rs
```diff
@@ -28,8 +28,10 @@ pub struct AccountInfo<'a> {
     pub data: Rc<RefCell<&'a mut [u8]>>,
     /// Program that owns this account
     pub owner: &'a Pubkey,
-    /// The epoch at which this account will next owe rent
-    pub rent_epoch: u64,
+    /// Formerly, the epoch at which this account will next owe rent. A field
+    /// must remain because the runtime depends on the exact layout of this
+    /// struct.
+    pub unused: u64,
```
febo: Could this field be private?
```suggestion
    unused: u64,
```

joncinque: I mainly did it to avoid breaking users that create `AccountInfo` by hand, ie. without using `new`. I was also worried that some tests in Agave would still want to check the field, so I thought it was better to keep it around.

On the flip side, there's an argument that a test that requires reading the field could use a different type, or that the test should be rewritten to not look at the field at all.

I'll let @Lichtso weigh on whatever he thinks is best

Lichtso: Renaming the field will break them anyway, right? Only question is if we also force them though the constructor method. For the sake of consistency I would leave it public, it does not matter as the field is unused after all, so writing to it can cause no harm.

joncinque: Yeah it's true that it's breaking no matter what. Happy to leave it public though!

buffalojoec: Is a more explicit `_unused` better? Can we also mark it deprecated?

joncinque: Sure! I didn't do that in case someone wanted to be clever and reuse the space to store other things, like how the duplicate flag is used as a reference counter in no-std-entrypoint / pinocchio, but those people can just define their own type if they really want to use that space :sweat_smile:

joncinque:
@apfitzge can I get a re-approval please?

PR #226, opened 2025-07-14
rent-collector: Deprecate everything except the struct

joncinque:
#### Problem

The rent collector is still being used in Agave to hold onto the `Rent` for the cluster, but is otherwise unused.

#### Summary of changes

Deprecate all functions and constants that should be redefined elsewhere.

I decided to not deprecate the whole thing since Agave still needs the `RentCollector` struct, but full deprecation is also an option.

This will be backported to maintenance/v2.x. If we decide on full deprecation, then we can also remove the crate entirely.


jstarry:
I think we should migrate Agave away from using the sdk rent collector crate and use an internal agave crate instead (if at all). Then just deprecate the whole crate here

joncinque:
Closing in favor of #262

PR #228, opened 2025-07-15
Remove some unused dependencies

LStan:
This PR removes some unused dependencies.

For now I left `getrandom` dependency in `solana-program` crate because there is a vague comment above it. Most probably it can be removed, but need some conformation.
@joncinque


joncinque:
Take a look at some of the CI errors -- looks like frozen-abi needs solana-logger, so we can have that crate only included with the `frozen-abi` feature

LStan:
Should be fixed. What about `getrandom` dependency in `solana-program`?

joncinque (approved):
Looks great, thanks! We can probably move the `getrandom` dependency directly into sdk-wasm-js, but I'll take care of that in some follow-up work.

LStan:
`getrandom` in `solana-program` is for `solana` target, not `wasm32`. Adding this crate for on-chain looks strange. There are no errors if I remove it. I didn't remove it only because of a very vague comment above it that I don't understand

joncinque:
Oh yeah that one was really annoying to figure out. The comment should make it clear-ish why it's still there. Most likely it's still possible to trip that error using the component crates, but we'll clean this all up in upcoming work anyway

LStan:
Is there something I can help with? Something simple

joncinque:
Thanks for offering, but it'll be a bit specific to what was discussed in #118. I'll be sure to create any issues if something does come up though :smile:

PR #229, opened 2025-07-15
all!: Remove deprecated functions and structs

joncinque:
#### Problem

There are a lot of deprecated functions and structs all over the SDK that can be removed now that we're making breaking changes.

#### Summary of changes

Remove it all!


account-info/src/lib.rs
```diff
@@ -130,23 +130,18 @@ impl<'a> AccountInfo<'a> {
...
-    ///
     /// # Safety
     ///
     /// This method makes assumptions about the layout and location of memory
     /// referenced by `AccountInfo` fields. It should only be called for
     /// instances of `AccountInfo` that were created by the runtime and received
     /// in the `process_instruction` entrypoint of a program.
-    #[deprecated(since = "2.3.0", note = "Use AccountInfo::resize() instead")]
-    pub fn realloc(&self, new_len: usize, zero_init: bool) -> Result<(), ProgramError> {
+    pub fn resize(&self, new_len: usize) -> Result<(), ProgramError> {
```
febo, line 144: I am in two minds on this one. My understanding is that this behaviour will be introduced with ABIv2, where to resize will be done via a syscall. It will most likely require a different `AccountInfo`-like implementation.

If we mimic the same behaviour now, it will cost more to resize an account since we are always zeroing the data. At the same time, the current behaviour will continue to be supported so in a way it is unnecessary to zero the data.

What do you think if we keep the deprecated so people are aware that changes are coming but still allow the current `realloc` to be used?

joncinque: Ah ok, I thought this was going to be broken much sooner, ie before ABIv2. @Lichtso what do you think? If it won't be broken soon, then we can definitely keep it around for longer.

febo (commented):
Looks great – only have a question about the changes on `AccountInfo`.

account-info/src/lib.rs
Lichtso: @febo This will come in the global direct mapping feature, before the opt-in ABIV2. This is simply a case where we under-metered programs and we are fixing that now.

febo: Oh, how is it going to work for existing (deployed) programs that expect the area to be zeroed? E.g., will loader v3 stop zeroing the account data "padding" memory?

Lichtso: All loaders will stop zeroing the padding. Programs have to write to it first to use it (which can be done using the memset syscall or normal store instructions). The padding is uninitialized anyway because CPI exists. This option to skip the zeroing was added in the SDK before CPI was introduced, and it should have been removed back then but was forgotten.

febo: Isn't there a risk that it will break existing programs? For example, a program that resizes an account and then loads a type using a zero-copy library (e.g., bytemuck) assuming that the memory is zeroed.

Lichtso: Yep, these programs are on the direct mapping naughty list. There is a bunch of them, but the instructions doing resizing don't have much traffic and almost all of these programs are upgradeable. Thus, they will have to use the new SDK version, remove their usesites of this deprecated parameter and then should fully work again.

febo: Given that this behaviour is going to change sooner than ABIv2 and the fact that `realloc` should really not be used anymore, I agree that removing it is the best way forward.

joncinque: In that case, can I get a :heavy_check_mark: ?

Lichtso (approved):

febo (approved):

transaction/src/lib.rs
```diff
@@ -1285,12 +1285,15 @@ mod tests {
     }
 
     fn create_sample_transaction() -> Transaction {
-        let keypair = Keypair::from_bytes(&[
```
apfitzge, line 1288: weird, wonder why this wasn't complaining about us using a deprecated fn previously.

apfitzge (approved):
looks good on tx-metadata changes

transaction/src/lib.rs
joncinque: line 1136 has `#![allow(deprecated)]` for the whole testing module -- i'll fix that up in a follow-up

PR #230, opened 2025-07-15
borsh!: Remove borsh v0.10 support

joncinque:
#### Problem

The sdk has implemented the borsh v0.10 traits for years, even though v1 has been out for awhile. This was to avoid breaking people who were still on the older crate version.

#### Summary of changes

It's been a long time, and we're making breaking changes, so remove all borsh 0.10 support!

cc @jacobcreech


febo (approved):
Looks good!

PR #231, opened 2025-07-16
[bls-signatures] Add bench code for bls signatures

samkim-crypto:


bls-signatures/benches/bls_signatures.rs
```diff
@@ -0,0 +1,93 @@
+use {
+    criterion::{black_box, criterion_group, criterion_main, Criterion},
```
joncinque: nit: `black_box` from criterion has been deprecated in favor of `std::hint::black_box` -- maybe that should be used instead?

https://docs.rs/criterion/latest/criterion/fn.black_box.html

bls-signatures/benches/bls_signatures.rs
```diff
@@ -0,0 +1,93 @@
...
+                assert!(verification_result);
+            });
+        });
+    }
+    group.finish();
+}
+
+// Benchmark for generating a new keypair
+fn bench_key_generation(c: &mut Criterion) {
+    c.bench_function("key_generation", |b| b.iter(Keypair::new));
```
joncinque: Is there a reason to not black_box the creation here?

bls-signatures/benches/bls_signatures.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// Benchmark for creating and verifying a proof of possession
+fn bench_proof_of_possession(c: &mut Criterion) {
+    let keypair = Keypair::new();
+    let pop = keypair.proof_of_possession();
+
+    c.bench_function("proof_of_possession_creation", |b| {
+        b.iter(|| keypair.proof_of_possession());
```
joncinque: Same here, does this need a black_box?

joncinque (commented):
Looks good overall! Just some small things

bls-signatures/benches/bls_signatures.rs
samkim-crypto: Good point. Updated!

bls-signatures/benches/bls_signatures.rs
samkim-crypto: I was reading a blog post some time ago about this, but I can't seem to find it now. My understanding was that if the output is an object, then criterion will handle the dead code elimination for you, but if the result is a simple type like boolean, then it is not guaranteed.

But to be safe, I just wrapped these with black_box. I also initially wrapped black_box on inputs instead of outputs on some functions, but I updated these too!

joncinque: Ok great!

PR #233, opened 2025-07-16
deps: Update env_logger to v0.11

joncinque:
#### Problem

The SDK and Agave are still using env_logger v0.9, which has a dependency on an unmaintained crate, atty.

#### Summary of changes

Update to the newest version of env_logger.

I'm not sure why this wasn't done earlier, but I didn't see any discussion in the old labs or in the agave repo.


yihau:
I guess it’s just because no one really checks the dep list haha. I'm gonna steal your PR to agave 😈

yihau:
oh btw, why not just assign 0.11.8 to the Cargo.toml 🤔

joncinque:
It made the minimal-versions check fail, so I decided to keep the version older. Agave should use 0.11.8 though

PR #236, opened 2025-07-17
hash!: Unify all hash types, introduce "offchain" feature

joncinque:
#### Problem

There are many repeated `Hash` types in all of the different hasher crates, which brings in unnecessary dependencies and copied code.

#### Summary of changes

Unify all hashes under `solana_hash::Hash`, by re-exporting that type from keccak and blake3.

While doing this work, I also cleaned up the dependencies, which were unnecessarily bringing in the crate for the offchain implementation.

When integrating the SDK with Pinocchio, we'll have many situations in which Pinocchio doesn't want to have an offchain fallback implementation available, but that fallback is still used by other downstream users.

To satisfy both uses, this PR also introduces a general `"offchain"` crate feature.  For example, `Pubkey::find_program_address` could use the offchain feature for deriving the address without using the syscall, which is useful for most offchain applications.

I could also split off the offchain feature in a different PR if that's clearer.

cc @kevinheavey in case you want to weigh in on this too


blake3-hasher/src/lib.rs
```diff
@@ -152,71 +62,19 @@ pub fn hash(val: &[u8]) -> Hash {
     hashv(&[val])
 }
 
-#[cfg(feature = "std")]
-/// Return the hash of the given hash extended with the given value.
-pub fn extend_and_hash(id: &Hash, val: &[u8]) -> Hash {
-    let mut hash_data = id.as_ref().to_vec();
-    hash_data.extend_from_slice(val);
-    hash(&hash_data)
-}
```
joncinque, line 155: I removed this function because it's entirely unused in Agave, and seems easy to reproduce elsewhere if needed. What do you think?

keccak-hasher/src/lib.rs
```diff
@@ -153,11 +63,3 @@ pub fn hashv(vals: &[&[u8]]) -> Hash {
...
     hashv(&[val])
 }
-
-#[cfg(feature = "std")]
-/// Return the hash of the given hash extended with the given value.
-pub fn extend_and_hash(id: &Hash, val: &[u8]) -> Hash {
-    let mut hash_data = id.as_ref().to_vec();
-    hash_data.extend_from_slice(val);
-    hash(&hash_data)
-}
```
joncinque, line 157: Same with this, it's entirely unused and allows us to remove the std feature

sha256-hasher/src/lib.rs
joncinque, line 1: I could also remove `extend_and_hash` here, which won't work in actual no-std environments. It's only used here: https://github.com/anza-xyz/agave/blob/58d7da606d8583bd1c447e3cf3e602a8b6ff7f70/runtime/src/bank.rs#L4458

What do you think?

joncinque: https://github.com/anza-xyz/agave/pull/7028, we can discuss there

febo: > What do you think?

Just for curiosity, it could be re-written as this right?
```rust
pub fn extend_and_hash(id: &Hash, val: &[u8]) -> Hash {
  hashv(&[id.as_ref(), val])
}
```

joncinque: Yep that's correct! I tested to be sure:
```
    fn test_extend_and_hash() {
        let val = "gHiljKpq";
        let val_hash = hash(val.as_bytes());
        let ext = "lM890t";
        let ext_hash = extend_and_hash(&val_hash, ext.as_bytes());
        let other_ext_hash =  hashv(&[val_hash.as_ref(), ext.as_bytes()]);
        assert!(ext_hash == other_ext_hash);
    }
```

keccak-hasher/src/lib.rs
```diff
@@ -124,9 +35,17 @@ pub fn hashv(vals: &[&[u8]]) -> Hash {
...
+        #[cfg(feature = "offchain")]
+        {
+            let mut hasher = Hasher::default();
+            hasher.hashv(vals);
+            hasher.result()
+        }
+        #[cfg(not(feature = "offchain"))]
+        {
+            core::hint::black_box(vals);
+            panic!("hashv is only available on target `solana`")
```
febo: Does this message should say feature "offchain" should be enabled instead of target "solana"?

keccak-hasher/src/lib.rs
```diff
@@ -2,50 +2,19 @@
...
-
-#[cfg(any(feature = "sha3", not(target_os = "solana")))]
 #[derive(Clone, Default)]
+#[cfg(all(feature = "offchain", not(target_os = "solana")))]
 pub struct Hasher {
     hasher: Keccak256,
 }
 
-#[cfg(any(feature = "sha3", not(target_os = "solana")))]
+#[cfg(all(feature = "offchain", not(target_os = "solana")))]
```
febo: I am bit confused about the "offchain" feature. It seems highly correlated with `not(target_os = "solana")`. For example, when `not(target_os = "solana")`, you need to have the "offchain" flag enabled, otherwise you get "nothing". So it seems that having  `not(target_os = "solana")` should be enough? I get the feeling that I am missing something.

blake3-hasher/src/lib.rs
febo: Agree, the client can use `hashv` directly.

keccak-hasher/src/lib.rs
febo: Oh oh, I think I got it – we need the feature so the dependencies are not pulled in.

joncinque: The concept is to make these kinds of libraries usable in pinocchio environments, where people want to minimize dependencies in their build.

Currently, things like `create_program_address` just fail in non-solana builds: https://github.com/anza-xyz/pinocchio/blob/3fecb3e569442e219185e5fe35854bfbf971e23a/sdk/pinocchio/src/pubkey.rs#L203

If we want pinocchio users to use these crates too, we should offer the opportunity to opt in of having these functions work if `target_os != solana`. This way, pinocchio can just not enable the `offchain` feature and never pull in the crates that implement the offchain behavior. Does that make sense?

blake3-hasher/src/lib.rs
```diff
@@ -122,9 +33,17 @@ pub fn hashv(vals: &[&[u8]]) -> Hash {
...
+        #[cfg(feature = "offchain")]
+        {
+            let mut hasher = Hasher::default();
+            hasher.hashv(vals);
+            hasher.result()
+        }
+        #[cfg(not(feature = "offchain"))]
+        {
+            core::hint::black_box(vals);
+            panic!("hashv is only available on target `solana`")
```
febo: I think the message should mentioned the "offchain" feature right?

keccak-hasher/src/lib.rs
joncinque: Ah yes, sorry!

blake3-hasher/src/lib.rs
```diff
@@ -146,71 +65,19 @@ pub fn hash(val: &[u8]) -> Hash {
...
-    #[test]
-    fn test_extend_and_hash() {
+    fn test_hashv() {
         let val = "gHiljKpq";
         let val_hash = hash(val.as_bytes());
+
         let ext = "lM890t";
-        let hash_ext = [&val_hash.0, ext.as_bytes()].concat();
-        let ext_hash = extend_and_hash(&val_hash, ext.as_bytes());
+        let ext_hash = hashv(&[&val_hash.to_bytes(), ext.as_bytes()]);
```
febo, line 79: Will this work when the feature "offchain" is not enabled? Perhaps we need to put a feature cfg on the test?

keccak-hasher/src/lib.rs
febo: Yep, totally. Took me a second look to figure out the logic. 😅

sha256-hasher/src/lib.rs
```diff
@@ -1,15 +1,15 @@
 #![no_std]
```
febo, line 1: I can't comment on the specific [lines](https://github.com/anza-xyz/solana-sdk/pull/236/files#diff-be8b223a57abdbd0bcf2507146cb5d6e570ed443f52b1911679133a8c95803d4R69-R74), but this files also has an `extend_and_hash` implementation, which can be removed - same behaviour can be achieved with the `hashv`.

joncinque: I wanted to deprecate and remove it in a separate PR through the normal process, since it does appear to have some downstream usage.

blake3-hasher/src/lib.rs
joncinque: Good point, done!

joncinque:
> At the same time, I can't come up with an alternative name, apart from using a name related to the functionality/crate being enabled by the feature.

We can go with that too, as the crates have been since the breakup. I was thinking a common name would be better for downstream users, so they don't need to read through our source code. 

On the flip-side, maybe that's not a big deal, since the compiler will tell you if you're trying to use something and it's not available due to a feature-gate, or our panic messages will tell that too.

So if you don't like it, I'll revert. I don't feel that strongly about it.

blake3-hasher/src/lib.rs
```diff
@@ -146,71 +65,20 @@ pub fn hash(val: &[u8]) -> Hash {
...
-#[cfg(feature = "std")]
-/// Return the hash of the given hash extended with the given value.
-pub fn extend_and_hash(id: &Hash, val: &[u8]) -> Hash {
-    let mut hash_data = id.as_ref().to_vec();
-    hash_data.extend_from_slice(val);
-    hash(&hash_data)
-}
-
 #[cfg(test)]
+#[cfg(feature = "blake3")]
```
febo, line 69: It seems that the feature is enabled on the dev-dependencies, so we probably don't need this. I might have missed that on my first review. Not sure if the idea is to have this pattern, but it can get a bit verbose for crates that have many features.

febo:
> > At the same time, I can't come up with an alternative name, apart from using a name related to the functionality/crate being enabled by the feature.
> 
> We can go with that too, as the crates have been since the breakup. I was thinking a common name would be better for downstream users, so they don't need to read through our source code.
> 
> On the flip-side, maybe that's not a big deal, since the compiler will tell you if you're trying to use something and it's not available due to a feature-gate, or our panic messages will tell that too.
> 
> So if you don't like it, I'll revert. I don't feel that strongly about it.

Yeah, I had the new `address` crate ([PR #243](#243)) in mind, since that one there could be a few things that you might enable for "offchain".

PR #237, opened 2025-07-18
[bls-signatures] Use batched pairing to improve signature verification times

samkim-crypto:


deanmlittle:
lgtm. this is also how i implemented it for altbn128 BLS signatures: https://github.com/deanmlittle/solana-alt-bn128-bls/blob/master/src/g2_point.rs#L23

joncinque (approved):
Looks like it does what it says on the tin, mostly trusting Dean on this!

PR #239, opened 2025-07-18
message / transaction: Remove dependency on solana-system-interface

joncinque:
#### Problem

While debugging changes to the wasm-js build, I noticed that `solana-transaction` and `solana-message` were pulling in `solana-system-interface` in a lot of situations, even though there's a lot of code in place to make sure that doesn't happen.

#### Summary of changes

Totally remove the dependency on solana-system-interface by inlining some nonce-related code in solana-message.

### DIFF
--- a/message/Cargo.toml
+++ b/message/Cargo.toml
@@ -17,8 +17,6 @@ rustdoc-args = ["--cfg=docsrs"]
 [features]
 bincode = [
     "dep:bincode",
-    "dep:solana-bincode",
-    "dep:solana-system-interface",
     "serde",
 ]
 blake3 = ["dep:blake3"]
@@ -45,7 +43,6 @@ blake3 = { workspace = true, features = ["traits-preview"], optional = true }
 lazy_static = { workspace = true }
 serde = { workspace = true, optional = true }
 serde_derive = { workspace = true, optional = true }
-solana-bincode = { workspace = true, optional = true }
 solana-frozen-abi = { workspace = true, optional = true }
 solana-frozen-abi-macro = { workspace = true, optional = true }
 solana-hash = { workspace = true }
@@ -55,9 +52,6 @@ solana-pubkey = { workspace = true }
 solana-sanitize = { workspace = true }
 solana-sdk-ids = { workspace = true }
 solana-short-vec = { workspace = true, optional = true }
-solana-system-interface = { workspace = true, optional = true, features = [
-    "bincode",
-] }
 solana-transaction-error = { workspace = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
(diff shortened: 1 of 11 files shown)


message/src/inline_nonce.rs
```diff
@@ -0,0 +1,66 @@
...
+pub(crate) fn advance_nonce_account_instruction(
+    nonce_pubkey: &Pubkey,
+    nonce_authority_pubkey: &Pubkey,
+) -> Instruction {
+    Instruction::new_with_bytes(
+        system_program::id(),
+        &ADVANCE_NONCE_DATA,
+        vec![
+            AccountMeta::new(*nonce_pubkey, false),
+            #[allow(deprecated)]
```
apfitzge, line 33: do we hae a replacement for the deprecated thing here? seems weird to use deprecated in new code to me

transaction/src/lib.rs
```diff
@@ -1122,12 +1117,7 @@ pub fn uses_durable_nonce(tx: &Transaction) -> Option<&CompiledInstruction> {
             matches!(
                 message.account_keys.get(instruction.program_id_index as usize),
                 Some(program_id) if system_program::check_id(program_id)
-            )
-            // Is a nonce advance instruction
-            && matches!(
-                limited_deserialize(&instruction.data, PACKET_DATA_SIZE as u64),
-                Ok(SystemInstruction::AdvanceNonceAccount)
-            )
+            ) && is_advance_nonce_instruction_data(&instruction.data)
```
apfitzge, line 1118: with this change, this function no longer requires the `bincode` feature, right?

transaction/src/versioned/mod.rs
```diff
@@ -216,12 +215,7 @@ impl VersionedTransaction {
                 matches!(
                     message.static_account_keys().get(instruction.program_id_index as usize),
                     Some(program_id) if system_program::check_id(program_id)
-                )
-                // Is a nonce advance instruction
-                && matches!(
-                    limited_deserialize(&instruction.data, crate::PACKET_DATA_SIZE as u64,),
-                    Ok(SystemInstruction::AdvanceNonceAccount)
-                )
+                ) && is_advance_nonce_instruction_data(&instruction.data)
```
apfitzge, line 217: same here - we don't need this fn to be gated by the `bincode` feature now, I think.

message/src/inline_nonce.rs
joncinque: Nope, we deprecated the sysvar in v1.9.0 without a plan for how it would be replaced.

transaction/src/lib.rs
joncinque: Correct. I was worried about adding another function to the API surface for all builds, but that concern seems silly now that I write it out. I'll remove the gating!

transaction/src/versioned/mod.rs
joncinque: Will remove, thanks! I also found an instance in solana-message that no longer needed the gating

message/Cargo.toml
```diff
@@ -70,11 +61,13 @@ borsh = { workspace = true }
 itertools = { workspace = true }
 serde_json = { workspace = true }
 solana-address-lookup-table-interface = { workspace = true, features = ["bincode", "bytemuck"] }
+solana-bincode = { workspace = true }
```
apfitzge: I see the test does actually depend on bincode, but I'm wondering if we need this.

Most of the other optional deps are still gated even in the tests.

joncinque: Do you mean gating the test on the `bincode` feature in the crate? The test does rely on that feature being enabled, but it's already enabled via `dev-context-only-utils`. I'd prefer to leave it as is, because we'll get failures if someone messes with the feature in some way.

If we gate the test, there's a possibility that someone breaks something, and we just never find out because the test is never compiled.

And in case I misunderstood, I removed the `solana-bincode` usage in the test and the dependency totally in 835ac7c7

apfitzge (approved):

PR #240, opened 2025-07-18
wasm-js!: Remove system program dependency

joncinque:
#### Problem

While working on the refactor of all wasm-js logic into sdk-wasm-js, I noticed that no matter what I did, `SystemInstruction` still made it into the final build. That made it impossible to build the package with breaking changes.

#### Summary of changes

Remove the usage of `SystemInstruction` in the tests. This required exposing `Instruction` and `AccountMeta` through wasm-bindgen to be able to create instructions.

The API doesn't need to be perfect, just usable, so I went with a builder pattern for adding `AccountMeta`s to an `Instruction`, which hugely simplifies the logic. Otherwise we'd need to perform conversions between JS arrays and `Vec<AccountMeta>`, which I was too lazy to do properly because wasm-js crate is more of a proof-of-concept.

While doing that, I moved the wasm-js deps to the component crates that were required for the tests.

### DIFF
--- a/instruction/src/wasm.rs
+++ b/instruction/src/wasm.rs
@@ -1,7 +1,12 @@
 //! The `Instructions` struct is a legacy workaround
 //! from when wasm-bindgen lacked Vec<T> support
 //! (ref: https://github.com/rustwasm/wasm-bindgen/issues/111)
-use {crate::Instruction, wasm_bindgen::prelude::*};
+#![allow(non_snake_case)]
+use {
+    crate::{AccountMeta, Instruction},
+    solana_pubkey::Pubkey,
+    wasm_bindgen::prelude::*,
+};
 
 #[wasm_bindgen]
 #[derive(Default)]
@@ -24,5 +29,32 @@ impl Instructions {
 impl From<Instructions> for std::vec::Vec<Instruction> {
     fn from(instructions: Instructions) -> Self {
         instructions.instructions
+    }
+}
+
+#[wasm_bindgen]
+impl Instruction {
+    /// Create a new `Instruction`
+    #[wasm_bindgen(constructor)]
+    pub fn constructor(program_id: Pubkey, data: &[u8]) -> Self {
+        Instruction::new_with_bytes(program_id, data, std::vec::Vec::new())
+    }
+
+    pub fn pushAccount(mut self, account_meta: AccountMeta) -> Self {
+        self.accounts.push(account_meta);
+        self
+    }
+}
+
+#[wasm_bindgen]
+impl AccountMeta {
+    /// Create a new writable `AccountMeta`
+    pub fn newWritable(pubkey: Pubkey, is_signer: bool) -> Self {
+        AccountMeta::new(pubkey, is_signer)
+    }
+
+    /// Create a new readonly `AccountMeta`
+    pub fn newReadonly(pubkey: Pubkey, is_signer: bool) -> Self {
+        AccountMeta::new_readonly(pubkey, is_signer)
     }
 }
--- a/instruction/src/wasm.rs
+++ b/instruction/src/wasm.rs
@@ -36,13 +36,16 @@ impl From<Instructions> for std::vec::Vec<Instruction> {
 impl Instruction {
     /// Create a new `Instruction`
     #[wasm_bindgen(constructor)]
-    pub fn constructor(program_id: Pubkey, data: &[u8]) -> Self {
-        Instruction::new_with_bytes(program_id, data, std::vec::Vec::new())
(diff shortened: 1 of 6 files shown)


sdk-wasm-js/Cargo.toml
```diff
@@ -18,13 +18,16 @@ rustdoc-args = ["--cfg=docsrs"]
...
-solana-program = { workspace = true }
-solana-sdk = { workspace = true }
+solana-keypair = { workspace = true }
+solana-pubkey = { workspace = true, features = ["curve25519", "sha2"] }
+solana-transaction = { workspace = true, features = ["bincode", "verify"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 console_error_panic_hook = { workspace = true }
 console_log = { workspace = true }
+getrandom = { version = "0.1.1", features = ["wasm-bindgen"] }
```
grod220, line 30: Looks like `getRandom` does not inherit the newer version in the workspace. Any reason for that?

sdk-wasm-js/src/lib.rs
```diff
@@ -1,23 +1,12 @@
...
-        declare_deprecated_id,
-        declare_id,
-        entrypoint,
-        entrypoint_deprecated,
-        example_mocks,
-        hash,
-        program_stubs,
-        pubkey,
-        *,
-    },
```
grod220, line 20: Guessing this is a breaking change that requires a major version bump?

instruction/src/wasm.rs
```diff
@@ -1,7 +1,12 @@
 //! The `Instructions` struct is a legacy workaround
 //! from when wasm-bindgen lacked Vec<T> support
 //! (ref: https://github.com/rustwasm/wasm-bindgen/issues/111)
-use {crate::Instruction, wasm_bindgen::prelude::*};
+#![allow(non_snake_case)]
```
grod220, line 4: You can also use the attribute if you wanted:

```rust
    #[wasm_bindgen(js_name = pushAccount)]
    pub fn push_account(mut self, account_meta: AccountMeta) -> Self
```

Not sure why one is better than the other though tbh

instruction/src/wasm.rs
```diff
@@ -26,3 +31,30 @@ impl From<Instructions> for std::vec::Vec<Instruction> {
...
+    /// Create a new `Instruction`
+    #[wasm_bindgen(constructor)]
+    pub fn constructor(program_id: Pubkey, data: &[u8]) -> Self {
+        Instruction::new_with_bytes(program_id, data, std::vec::Vec::new())
+    }
+
+    pub fn pushAccount(mut self, account_meta: AccountMeta) -> Self {
+        self.accounts.push(account_meta);
+        self
+    }
```
grod220: Do you think someone familiar with javascript APIs would expect to be able to do this?:

```js
const instruction = new Instruction(programId, instructionData);
instruction.pushAccount(metadataA);
instruction.pushAccount(metadataB);
```

When I tried that in your test I got a `Error: Attempt to use a moved value`. 

One idea is to be more explicit with the builder and require a `build()`. Something that would look like:

```js
const ix = new InstructionBuilder(programId)
    .addAccount(srcAccountMeta)
    .addAccount(dstAccountMeta)
    .setData(instructionData)
    .build();
```
What do you think?

sdk-wasm-js/Cargo.toml
joncinque: Yeah, we need it for randomness to work in wasm with the current versions that we use. At least, we needed it a year or so ago, so maybe all of the newer versions work now. I can take a look at it in follow-up work!

sdk-wasm-js/src/lib.rs
joncinque: yep!

instruction/src/wasm.rs
joncinque: Typically I've found that `js_name` is more useful for renaming functions totally. Last time I checked this, maybe a year ago, there was no default way to just have functions renamed into camel-case, so I'd prefer to go with this.

instruction/src/wasm.rs
joncinque: I agree that a builder API would be neater. However, a builder pattern would be inconsistent with all of the other APIs offered in the SDK. If this were an important API, then we could consider adopting builder patterns all over the SDK, but since sdk-wasm-js is more of a proof-of-concept, I would prefer to keep things simple and avoid adding wasm-js-only types.

Unfortunately, wasm-bindgen doesn't allow for returning references, so I went with just normal mutating functions instead of any kind of builder pattern. Let me know how it looks!

grod220 (approved):
LGTM 👍

PR #242, opened 2025-07-18
Optimized x86 SHA-NI implementation for single 32 byte sha256 hashing

kagren:
SHA256 operates on block sizes of 64 bytes (512 bits), so when hashing something smaller (32 bytes in this case, which is common in Solana), the input block will be padded with 32 bytes to make a full 64 byte block. The padding applied is deterministic for the first block, and as such we can precompute and optimize the x86 SHA-NI code slightly as long as we are hashing 32 bytes and only a single block.

If multiple blocks are hashed, or larger input sizes than 32, then the default implementation, the RustCrypto hashers/sha2, is used (as before).

### DIFF
--- a/sha256-hasher/src/lib.rs
+++ b/sha256-hasher/src/lib.rs
@@ -3,6 +3,10 @@
 use sha2::{Digest, Sha256};
 use solana_hash::Hash;
 
+// Only include x86_shani_single_hash_optimized if we are on x86_64 (and not on chain)
+#[cfg(all(target_arch = "x86_64", not(target_os = "solana")))]
+mod x86_shani_single_hash_optimized;
+
 #[cfg(any(feature = "sha2", not(target_os = "solana")))]
 #[derive(Clone, Default)]
 pub struct Hasher {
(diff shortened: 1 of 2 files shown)


joncinque:
@samkim-crypto can you take a look at this?

samkim-crypto:
Thinking about this PR again, it does seem that touching the way PoH is implemented does seem to introduce quite a lot of risk. Given that PoH is ultimately on the way out after alpenglow, I would recommend trying to push these changes upstream rather than making direct changes in the solana sdk.

kagren:
Thanks for the feedback @samkim-crypto! Understand your concerns and agree this would be more suited to be implemented upstream, at least long term. I mainly offered this as a way for validators to speed up their PoH hash rate, and since posting the source on the SF tech discord several validators have implemented and seem to be running it successfully on mainnet with improved hash rates as a result. By including this optimization in tsolana-sdk, all validators could benefit.

But ofc with alpenglow the PoH aspect goes away as you say. Regardless I will make a PR to the upstream RustCrypto hashers repo.

joncinque:
Closing since we won't be integrating the change, but thanks for your contribution!

Issue #204, opened 2025-06-24
Rename to this crate to solana-address on next major release?


mikemaccana:
As the cargo docs and code go out of their way to note repeatedly, as as we already know, `solana_pubkey::Pubkey`s are often not pubkeys. 

"Some pubkeys aren't pubkeys" creates difficultly:

 - For people using Anza's new front end tools (which correctly refer to addresses as Addressees) that have to use onchain data from this crate
 - For people learning Solana, where we know that public keys always have matching private keys
 - For having conversation with the cryptography community who wonder WTF we're doing. 

This is a breaking change, but if we rip the bandaid off now and schedule it for the next breaking release I think it will save us all more pain in the long run.

joncinque:
It'll be very disruptive to flat-out deprecate `Pubkey` considering we'll be doing the v3 release in the next month or two, but we can consider creating `solana-address` and re-exporting everything from there in `solana-pubkey`.

EDIT: and maybe with v4 next year or so, we remove pubkey

@febo has been thinking about this a bit too, but in the context of pinocchio.

mikemaccana:
@joncinque Same here, I support for V4.

mikemaccana:
Thanks @febo ! ❤️

PR #243, opened 2025-07-22
Rename `Pubkey` to `Address`

febo:
### Problem

While a Solana address is represented by a `[u8; 32]` array, addresses are currently represented by a "wrapper" `Pubkey` type over the `[u8; 32]` array. Pinocchio on the other hand, defines its own `Pubkey` type as an alias to `[u8; 32]`, which is considered a different type. The main motivation for that is to avoid bringing unnecessary dependencies to programs.

Additionally, there is confusion to what a pubkey represent (#204) since "some pubkeys aren't pubkeys."

### Solution

This PR addresses both concerns by moving the current `Pubkey` type and renaming it to `Address` into a new crate `solana-address`. All dependencies in the new crate are activated by features, therefore a program has a fine-grained control of which dependencies are active.

To mitigate the impact of this refactoring, `solana-pubkey` re-exports everything from `solana-address`.

cc: @kevinheavey


address/Cargo.toml
```diff
@@ -0,0 +1,87 @@
...
+solana-frozen-abi = { workspace = true, features = ["frozen-abi"], optional = true }
+solana-frozen-abi-macro = { workspace = true, features = ["frozen-abi"], optional = true }
+solana-program-error = { workspace = true, optional = true }
+solana-sanitize = { workspace = true, optional = true }
+
+[target.'cfg(not(target_os = "solana"))'.dependencies]
+curve25519-dalek = { workspace = true, optional = true }
+solana-sha256-hasher = { workspace = true, features = ["sha2"], optional = true }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
```
febo: If we don't make the wasm dependencies optional, it seems they are pulled by default.

address/Cargo.toml
```diff
@@ -0,0 +1,87 @@
...
+    "dep:solana-frozen-abi-macro",
+    "dep:solana-program-error",
+    "std",
+]
+rand = ["dep:rand", "atomic", "std"]
+sanitize = ["dep:solana-sanitize"]
+serde = ["dep:serde", "dep:serde_derive"]
+sha2 = ["dep:solana-sha256-hasher", "error"]
+std = ["decode"]
+syscalls = ["dep:solana-define-syscall", "error"]
```
febo, line 32: The "error" feature is needed for syscalls since the helpers return an `AddressError` (used to be `PubkeyError`) instead of a `ProgramError`. I think the differentiation is there since these helpers can be used offchain. We could make the on-chain "version" of those helpers return a `ProgramError` so programs don't need the "error" feature in a future PR.

address/src/hasher.rs
```diff
@@ -0,0 +1,140 @@
...
+            [self.offset..self.offset + mem::size_of::<u64>()]
+            .try_into()
+            .unwrap();
+        self.state = u64::from_ne_bytes(*chunk);
+    }
+}
+
+/// A builder for faster, but less collision resistant hasher for pubkeys.
+///
+/// Initializes `PubkeyHasher` instances that use an 8-byte
```
febo: Typo:

```suggestion
/// Initializes `AddressHasher` instances that use an 8-byte
```

joncinque:
I'll be taking over the PR from here since Febo's out -- @buffalojoec can you take a look when you have a moment?

address/Cargo.toml
joncinque: Resolved since all wasm code was moved to sdk-wasm-js

address/Cargo.toml
buffalojoec, line 1: Cool! I like how every dependency is optional! However I think you're going to annoy a lot of developers if you don't include a default. I recommend at least `decode`, but you could also make a case for `std`.

We can build hyper-modular tooling to empower devs who want to optimize their programs, but that doesn't mean we need to increase the burden on the majority of developers.

address/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+        }
+        Self::from(b)
+    }
+
+    // If target_os = "solana", then the solana_sha256_hasher crate will use
+    // syscalls which bring no dependencies.
+    // When target_os != "solana", this should be opt-in so users
+    // don't need the sha2 dependency.
+    #[cfg(feature = "sha2")]
+    pub fn create_with_seed(
```
buffalojoec, line 236: Following the comment, shouldn't this function be `any(feature = "sha2", target_os = "solana")`?

It states that when `target_os = "solana"`, the `solana_sha256_hasher` will use the syscall, therefore no dependency is used. However, that means that in order to compile the program for sBPF you must enable this feature which will enable `decode` which will bring in `five8`.

address/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+/// #   use solana_address::declare_id;
+/// declare_id!("My11111111111111111111111111111111111111111");
+/// # }
+/// # use item_wrapper::id;
+///
+/// let my_id = Address::from_str("My11111111111111111111111111111111111111111").unwrap();
+/// assert_eq!(id(), my_id);
+/// ```
+#[macro_export]
+macro_rules! declare_id {
```
buffalojoec: Is this thing even useful still? It has always seemed kind of silly.

All you really need is `const MY_ADDRESS: Address = address!("...");`.

address/src/lib.rs
```diff
@@ -0,0 +1,720 @@
...
+        fn test_id() {
+            assert!(check_id(&id()));
+        }
+    };
+}
+
+#[cfg(feature = "decode")]
+/// Same as [`declare_id`] except that it reports that this ID has been deprecated.
+#[macro_export]
+macro_rules! declare_deprecated_id {
```
buffalojoec: Same with this one. Not super useful IMO.

buffalojoec (commented):
The changes look clean to me, just few things and a question about the compatibility!

> To mitigate the impact of this refactoring, solana-pubkey re-exports everything from solana-address.

I wonder if this is even worth doing. We're essentially publishing a pass-through crate, when we can just not publish a new `solana-pubkey`. If SDK v3 has a bunch of breaking changes, why not just make the migration from `Pubkey` to `Address` part of that upgrade process for program devs?

address/src/lib.rs
joncinque: The function is only available with the `sha2` feature. If building for solana, then `solana_sha256_hasher` has no features enabled, which means no deps. If building for non-solana, then `solana_sha256_hasher` will have the extra feature, which brings in the extra dep.

If you want to call `create_with_seed`, you just enable `sha2`, and you get it. No matter what target you're building for, we figure out the deps underneath for you.

I might be missing something -- how does this work with the `decode` feature?

address/src/lib.rs
joncinque: That's a good point, let's just keep this back in `Pubkey` for now. We can always bring it down here if it's really needed.

I have heard that this code doesn't always run at compile-time, but we'll need to fix that with a real proc-macro instead.

joncinque:
> If SDK v3 has a bunch of breaking changes, why not just make the migration from Pubkey to Address part of that upgrade process for program devs?

That's a non-starter in my eyes. It would mean having to replace _every_ instance of `Pubkey` with `Address` in order to upgrade to v3, which is already going to be a pain. If we publish both, then people have time to port things over to `Address` before we remove `Pubkey` with the next breaking change.

address/Cargo.toml
joncinque: Yeah that's a good point. My thinking was that most people will use `solana-pubkey` for a while, so that'll act like the "default" feature as people port over.

Since it's a breaking change to remove features from the default, I'd prefer to keep this lean for now. We can always add default features later as people struggle. What do you think?

joncinque:
I also moved `new_rand` back to `solana_pubkey` at the same time

address/Cargo.toml
buffalojoec: I guess I'll leave it up to you, but if you have a small set of sensible defaults, and power users aren't happy with those, what's the reason you'd ever have to remove a default feature because someone couldn't get around it with `default-features = false`?

address/src/lib.rs
buffalojoec: Actually I think I traced the wrong feature tree for this one (`sha2`/`std`).

> If you want to call `create_with_seed`, you just enable `sha2`, and you get it. No matter what target you're building for, we figure out the deps underneath for you.

Ok yeah this works for me. Cool!

address/src/lib.rs
```diff
@@ -307,114 +307,6 @@ impl core::fmt::Display for Address {
...
-///
-/// let my_id = Address::from_str("My11111111111111111111111111111111111111111").unwrap();
-/// assert_eq!(ID, my_id);
-/// ```
-#[macro_export]
-macro_rules! address {
-    ($input:literal) => {
-        $crate::Address::from_str_const($input)
-    };
-}
```
buffalojoec: Oh I would still keep this one! I just think the module-based ones aren't something we need to continue supporting unless everyone loves them. I think this one is fine, though, and useful.

buffalojoec (commented):
> > If SDK v3 has a bunch of breaking changes, why not just make the migration from Pubkey to Address part of that upgrade process for program devs?
> 
> That's a non-starter in my eyes. It would mean having to replace _every_ instance of `Pubkey` with `Address` in order to upgrade to v3, which is already going to be a pain. If we publish both, then people have time to port things over to `Address` before we remove `Pubkey` with the next breaking change.

Okay, works for me. The churn is inevitable if `solana-pubkey` is doomed, but I won't argue against giving people more time.

address/src/lib.rs
joncinque: Oops gotcha, done!

buffalojoec (approved):

scripts/check-nits.sh
```diff
@@ -17,10 +17,11 @@ declare prints=(
 # Parts of the tree that are expected to be print free
 declare print_free_tree=(
   ':**.rs'
+  ':^address/src/hasher.rs'
+  ':^address/src/syscalls.rs'
   ':^logger/src/lib.rs'
   ':^msg/src/lib.rs'
   ':^program-option/src/lib.rs'
-  ':^pubkey/src/lib.rs'
```
apfitzge, line 23: even though it's been reduced, we should probably still have this crate be print free, right?

joncinque: The `^` in this format actually means to _exclude_ that file, so removing it is a good thing, and means it will be println free. Before it was in the exception list.

joncinque:
>  That benefit doesn't seem worth the trouble this will likely cause though. Is there something I'm missing in here?

The motivation is really to unite the `Pubkey` type used by both Pinocchio and the SDK, which will make it much easier to write framework-agnostic code.

There was repeated feedback from audits that we should use a newtype instead of a type alias, and since `solana_pubkey::Pubkey` is already a newtype, Pinocchio can just use that instead.

apfitzge:
> There was repeated feedback from audits that we should use a newtype instead of a type alias, and since solana_pubkey::Pubkey is already a newtype, Pinocchio can just use that instead.

It can only use the same type by adding a dependency - wasn't (at least initially) part of the point of pinocchio to have no strings attached? Maybe that was just a remnant of our mono-sdk?

A downside I see of keeping a newtype is that different versions will still be "different", i.e. v3.0 Address is NOT the same as v3.1 Address, so still requires devs to find/use the correct versions of crates that have consistent versions of `Address`.
Maybe with split sdk crates these versions will not need to change often though.
Are we going to actually use separate crate versions in 3.0+?

joncinque:
> A downside I see of keeping a newtype is that different versions will still be "different", i.e. v3.0 Address is NOT the same as v3.1 Address, so still requires devs to find/use the correct versions of crates that have consistent versions of Address

What do you mean by this? Although the types can be different based on how versions are declared, ie. one crate asks for `solana-address = "=3.0"` and another asks for `solana-address = "=3.1"`, but if people are using default dependency declarations, `solana-address = "3.0"` and `solana-address = "3.1"` will resolve to the same package and work out fine.

>  Maybe with split sdk crates these versions will not need to change often though.

That's one of our big hopes. It's only because we made breaking changes in the wasm side that we actually need to publish v3 of all the crates, otherwise we could have ridden just fine with v2.

>  Are we going to actually use separate crate versions in 3.0+?

That's my hope!

apfitzge:
> What do you mean by this? Although the types can be different based on how versions are declared, ie. one crate asks for solana-address = "=3.0" and another asks for solana-address = "=3.1", but if people are using default dependency declarations, solana-address = "3.0" and solana-address = "3.1" will resolve to the same package and work out fine.

I don't recall what (possibly community) libraries I was using, but I've hit something like this with a useless error message:
```expected struct solana_sdk::pubkey::Pubkey, got struct solana_sdk::pubkey::Pubkey```

(might have been Hash, but it was one of the really fundamental types). came down to different minor versions being used in the crates. Just want to avoid that since it's a huge dev pain to hit these things. In hindsight it was **probably** because `sdk` was a mono and they needed to pin to a specific version because we broke something.

If we're committed to using correct versioning for our crates, not opposed to using newtype!

PR #247, opened 2025-07-24
[bls-signatures] Add `parallel` feature and use `rayon` for aggregating pubkeys and signatures

samkim-crypto:
#### Problem
The bls-signatures crate do not yet support parallelization.

#### Summary of Changes
I took a stab at adding parallelization support via rayon.
- I added the `parallel` feature
- under the `parallel` feature, I added the `par_aggregate`, `par_aggregate_with`, and `par_aggregate_verify` functions that uses rayon for summing up the public keys or signatures
- in the process, the lifetimes and the iterator syntax was getting out of hand, so I ended up simplifying the api so that these functions take in slices of pubkeys or signatures instead of iterators
- one thing to note is that the standard `aggregate`, `aggregate_with`, and `aggregate_verify` functions have `dyn` type support while their parallel analogue do not. I did this on purpose to really make the parallel versions as fast as possible.

I ran some benches and I am getting the following numbers on my machine.

```
64 signatures / public keys:
Signature Aggregation: Parallel is ~2% slower (76.6 µs vs 74.5 µs).
Public Key Aggregation: Parallel is ~2.4x slower (67.2 µs vs 27.9 µs).
Aggregate Verification: Parallel is ~7% slower (1.06 ms vs 996 µs).

256 signatures / public keys
Signature Aggregation: Parallel is ~2.1x faster (144 µs vs 300 µs).
Public Key Aggregation: Parallel is ~1.1x faster (100 µs vs 111 µs).
Full Verification: Parallel is ~1.1x faster (1.16 ms vs 1.30 ms).

1024 signatures / public keys
Signature Aggregation: Parallel is ~4.1x faster (294 µs vs 1.21 ms).
Public Key Aggregation: Parallel is ~2.8x faster (160 µs vs 457 µs).
Full Verification: Parallel is ~1.9x faster (1.35 ms vs 2.54 ms).

2048 signatures / public keys
Signature Aggregation: Parallel is ~5.2x faster (466 µs vs 2.42 ms).
Public Key Aggregation: Parallel is ~3.8x faster (236 µs vs 896 µs).
Full Verification: Parallel is ~2.7x faster (1.55 ms vs 4.22 ms).
```

For smaller number of signatures and public keys, the sequential functions perform better due to the initial overhead of parallelization. But after about ~64 signatures and ~128 public keys, the parallel aggregation starts to outperform the sequential functions (of course, this should be pretty hardware dependent).

### DIFF
--- a/bls-signatures/src/pubkey.rs
+++ b/bls-signatures/src/pubkey.rs
@@ -437,7 +437,6 @@ mod tests {
     }
 
     #[test]
-    #[cfg(not(feature = "parallel"))]
     fn test_pubkey_aggregate_dyn() {
         let keypair0 = Keypair::new();
         let keypair1 = Keypair::new();
--- a/bls-signatures/src/pubkey.rs
+++ b/bls-signatures/src/pubkey.rs
@@ -1,5 +1,7 @@
 #[cfg(feature = "bytemuck")]
 use bytemuck::{Pod, PodInOption, Zeroable, ZeroableInOption};
+#[cfg(all(feature = "parallel", not(target_os = "solana")))]
+use rayon::prelude::*;
 #[cfg(all(not(target_os = "solana"), feature = "std"))]
 use std::sync::LazyLock;
 #[cfg(not(target_os = "solana"))]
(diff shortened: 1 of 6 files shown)


bls-signatures/src/signature.rs
```diff
@@ -66,58 +68,92 @@ impl SignatureProjective {
...
+    #[cfg(feature = "parallel")]
+    pub fn par_aggregate<S: AsSignatureProjective + Sync>(
+        signatures: &[&S],
+    ) -> Result<SignatureProjective, BlsError> {
+        if signatures.is_empty() {
+            return Err(BlsError::EmptyAggregation);
+        }
+        signatures
+            .into_par_iter()
+            .map(|sig| (*sig).try_as_projective())
```
wen-coding, line 130: I don't care too much about error handling in par_aggregate of Pubkey because we should have verification of Pubkeys when they are written to vote accounts.

But do you think it helps if we return which signature failed here so we can kick it out? Does it help much if we return multiple failures?

bls-signatures/src/signature.rs
```diff
@@ -66,58 +68,92 @@ impl SignatureProjective {
...
+            .reduce(
+                || Ok(SignatureProjective::identity()),
+                |a, b| {
+                    let mut a = a?;
```
wen-coding, line 134: I guess if `let mut a = a?;` fails we have a big trouble.... But I'm fine with it written this way.

bls-signatures/src/signature.rs
samkim-crypto: Yeah for the alpenglow application, we can always assume that the public keys and the signatures for the purpose of aggregation. For verification, we should have a separate logic to identify the invalid public keys or signatures. I was planning on adding that in a follow-up to this PR.

wen-coding: Sounds good. We should almost never have invalid pubkeys, but invalid signatures is definitely possible (although should be rare in most cases).

bls-signatures/src/signature.rs
samkim-crypto: Yeah assuming that all the signatures have been pre-verified and valid, this should never fail. We have to unwrap the logic here somehow and I think this is probably the cleanest way to write it.

wen-coding (approved):

bls-signatures/src/pubkey.rs
```diff
@@ -353,6 +389,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(not(feature = "parallel"))]
```
joncinque: Was this change intended? We should probably run this test always unless there's something I'm missing

bls-signatures/src/signature.rs
```diff
@@ -33,7 +35,7 @@ pub const BLS_SIGNATURE_AFFINE_BASE64_SIZE: usize = 256;
 
 /// A trait for types that can be converted into a `SignatureProjective`.
 #[cfg(not(target_os = "solana"))]
-pub trait AsSignatureProjective {
+pub trait AsSignatureProjective: Send + Sync {
```
joncinque: nit: up to you, but I imagine the `Send + Sync` requirement is only needed if `parallel` is enabled, so you could define two different versions of the trait.

I didn't find any great options to do this, but https://stackoverflow.com/questions/61392550/rust-trait-bound-depending-on-feature seemed reasonable

bls-signatures/src/signature.rs
```diff
@@ -319,51 +355,52 @@ mod tests {
...
+            &[&keypair0.public, &keypair1.public],
+            &[] as &[&SignatureProjective],
             test_message,
         )
         .unwrap_err();
         assert_eq!(err, BlsError::EmptyAggregation);
     }
 
     #[test]
+    #[cfg(not(feature = "parallel"))]
```
joncinque: Same with this, can we enable the test?

bls-signatures/src/pubkey.rs
samkim-crypto: Oh good point. There is no harm in letting this test run for any features. Fixed!

bls-signatures/src/signature.rs
samkim-crypto: Oh yeah sorry, I meant to not add `Send` and `Sync`. I first started out by adding `dyn` type support even for parallel aggregation, but I decided against it as mentioned in the PR description. After removing support for `dyn`, I missed removing the trait `Send` and `Sync`. I now updated to remove these traits. Thanks for the catch!

joncinque (dismissed):
Looks like there's a conflict to resolve, but after that, looks good to me!

joncinque (approved):
All good to me!

PR #248, opened 2025-07-24
[base3-encoding] Add support for either base2 or base3 encoding

samkim-crypto:
Instead of taking care of the regular binary vector and base3 case separately, I thought it would be convenient to combine the logic in a single crate. I made updates to the base3-encoding crate so that it handles both the regular binary vector and base3 case.

### Encoding Scheme

This implementation uses a versioned byte format to support two distinct encoding schemes, identified by the first byte of the output.

The general structure of the encoded data is:
1.  **Version Byte (1 byte)**: A `u8` indicating the encoding scheme (`0` for Base2, `1` for Base3).
2.  **Length Prefix (2 bytes)**: A `u16` in **little-endian** format storing the number of bits in the original vector(s).
3.  **Data Payload**: The encoded data, whose format depends on the version.

---

#### Base2 Encoding (`Version: 0`)

* **Purpose**: Encodes a single boolean vector.
* **Payload**: The raw byte data of the `BitVec`, with no additional packing. This is primarily a "pass-through" format with the version and length header.

---

#### Base3 Encoding (`Version: 1`)

* **Purpose**: Compresses two boolean vectors (`base` and `fallback`) of the same length into a single byte stream.
* **Constraint**: It assumes that for any given index, the bits in the `base` and `fallback` vectors will not both be `1`. The combination `(true, true)` is invalid.
* **Principle**: Each pair of booleans is mapped to a single ternary (base-3) digit:
    * `(false, false)` -> `0`
    * `(true, false)` -> `1`
    * `(false, true)` -> `2`
* **Packing**: 5 ternary digits are packed into a single `u8`. The packing is achieved by processing the bits in **descending (reverse) order** within each 5-bit chunk.

#### Encoding Functions

Two separate functions are provided for each encoding scheme:

* **`encode_base2(bit_vec: &BitVec) -> Result<Vec<u8>, EncodeError>`**
    This function takes a single boolean vector and serializes it using the **Base2** scheme.

* **`encode_base3(base: &BitVec, fallback: &BitVec) -> Result<Vec<u8>, EncodeError>`**
    This function takes two boolean vectors and compresses them together using the **Base3** scheme. It returns an error if the input vectors have mismatched lengths or contain an invalid `(true, true)` bit pair.

---

#### Decoding Function

A single, unified function handles decoding:

* **`decode(bytes: &[u8], max_len: usize) -> Result<Decoded, DecodeError>`**
    This function inspects the first byte of the input slice to determine the version and decodes the data accordingly. It returns a `Result` containing the `Decoded` enum.

    The **`Decoded`** enum has two variants, allowing you to handle the output with a `match` statement:
    * `Decoded::Base2(BitVec<...>)`
    * `Decoded::Base3(BitVec<...>, BitVec<...>)`

---


base3-encoding/src/lib.rs
```diff
@@ -1,203 +1,251 @@
-//! Provides a space-efficient encoding scheme for two boolean vectors.
+//! Provides a space-efficient encoding scheme for one or two boolean vectors.
 //!
-//! This module implements a compression algorithm to encode two boolean vectors
-//! of the same length into a single byte vector (`Vec<u8>`). It achieves
-//! this by mapping each pair of corresponding bits from the input vectors into a
-//! single ternary (base-3) digit.
+//! This module implements compression algorithms to encode boolean vectors
+//! into a single byte vector (`Vec<u8>`). It supports two distinct schemes
```
wen-coding: nit: we might support more than two in the future right?

0x0ece (approved):

base3-encoding/src/lib.rs
```diff
@@ -1,203 +1,251 @@
...
     if num_bits > u16::MAX as usize {
         return Err(EncodeError::LengthExceedsLimit);
     }
-    let required_bytes = num_bits
-        .checked_add(7)
-        .ok_or(EncodeError::ArithmeticOverflow)?
-        .checked_div(8)
+
+    let raw_slice = bit_vec.as_raw_slice();
+    let capacity = 3usize
```
wen-coding: nit: should we make this 3 a constant maybe?

base3-encoding/src/lib.rs
```diff
@@ -1,203 +1,251 @@
...
-//!
-//! # Encoded Format
+//! The combination `(true, true)` is considered invalid. These ternary digits are
+//! packed five at a time into a single `u8`, since `3^5 < 2^8`.
 //!
-//! The resulting `Vec<u8>` has a simple structure:
-//! 1.  **Length Prefix (2 bytes)**: A `u16` in little-endian format storing the
+//! The format is:
+//! 1.  **Version Byte (1 byte)**: `Version::Base3` as a `u8`.
+//! 2.  **Length Prefix (2 bytes)**: A `u16` in little-endian format storing the
```
wen-coding: nit: should we specify this is number of bits in the raw vector, not number of bytes in the final vector?

samkim-crypto: Okay, added clarification!

base3-encoding/src/lib.rs
samkim-crypto: Done!

base3-encoding/src/lib.rs
samkim-crypto: Okay, I added the word "currently". I also renamed the crate from `base3-encoding` to `signer-store` to emphasize the main use-case of the crate and since this crate is not limited to just base3 encoding anymore.

wen-coding (approved):

PR #249, opened 2025-07-25
sdk!: Remove `program` feature

joncinque:
#### Problem

The `program` feature has existed since v1.3 so that people can use solana-sdk in on-chain programs.

`cargo-build-sbf` even had a special case where it would enable the `program` feature when building solana-sdk, but it was removed in https://github.com/anza-xyz/agave/pull/5870.

This means that our current build-sbf script fails when using v2.3 of the tools.

#### Summary of changes

Remove the `program` feature from the sdk, and fix the build-sbf script to exclude the sdk, and then try to build without default features.

Rather than adding more excludes, I also changes the build-sbf script to use `cargo hack`, which is a bit cleaner.

### DIFF
--- a/scripts/build-sbf.sh
+++ b/scripts/build-sbf.sh
@@ -27,4 +27,5 @@ build_sbf_excludes=(
 
 ./cargo nightly hack --workspace "${build_sbf_excludes[@]}" build-sbf
 
-./cargo nightly build-sbf --manifest-path sdk/Cargo.toml --no-default-features
+# This can be added back in once the SDK upgrades to v2.3 of Agave tools
+#./cargo nightly build-sbf --manifest-path sdk/Cargo.toml --no-default-features


joncinque:
Well, this actually shows that it's dangerous to remove the `program` feature, since older versions of `cargo build-sbf` will expect to enable it, and fail. This means we'll either need to update the solana-cli version to v2.3 with this PR, or keep the program feature.

What do you think is a better approach?

LucasSte:
> Well, this actually shows that it's dangerous to remove the `program` feature, since older versions of `cargo build-sbf` will expect to enable it, and fail. This means we'll either need to update the solana-cli version to v2.3 with this PR, or keep the program feature.
> 
> What do you think is a better approach?

Newer versions of `cargo-build-sbf` will ignore the flag. What about removing `solana-sdk` from the tests by now (i.e insert it in the exclude list) and keeping the feature flag until we update the CLI version?

Perhaps, it would be worth it to add a comment explaining that the flag is deprecated.

joncinque:
> What about removing solana-sdk from the tests by now (i.e insert it in the exclude list) and keeping the feature flag until we update the CLI version?

That probably makes more sense, since we're essentially saying "don't use solana-sdk on-chain" by removing the "program" feature. Done!

LucasSte (approved):

PR #250, opened 2025-07-25
sha256-hasher: Deprecate `extend_and_hash`

joncinque:
#### Problem

As noted in https://github.com/anza-xyz/agave/pull/7028, the `extend_and_hash` function is unnecessary and creates another allocation, so we inlined it in Agave.

#### Summary of changes

Deprecate it formally in the crate. This PR will be backported to v2.x and published, so we can remove the function in a follow-up PR.

### DIFF
--- a/sha256-hasher/src/lib.rs
+++ b/sha256-hasher/src/lib.rs
@@ -67,6 +67,7 @@ pub fn hash(val: &[u8]) -> Hash {
 }
 
 /// Return the hash of the given hash extended with the given value.
+#[deprecated(since = "2.3.0", note = "Use `hashv(&[hash.as_ref(), val])` directly")]
 pub fn extend_and_hash(id: &Hash, val: &[u8]) -> Hash {
     let mut hash_data = id.as_ref().to_vec();
     hash_data.extend_from_slice(val);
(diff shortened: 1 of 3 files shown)


sha256-hasher/src/lib.rs
```diff
@@ -67,6 +67,7 @@ pub fn hash(val: &[u8]) -> Hash {
 }
 
 /// Return the hash of the given hash extended with the given value.
+#[deprecated(since = "2.3.0", note = "Use `hashv(&[hash.as_ref(), val])` directly")]
 pub fn extend_and_hash(id: &Hash, val: &[u8]) -> Hash {
```
febo, line 71: Would it make sense to replace the implementation with `hashv(&[id.as_ref(), &val])` given that it is what people should be doing directly?

febo (approved):
Looks great! Just left a comment that might not be applicable.

sha256-hasher/src/lib.rs
joncinque: That's a good point. If there were more users of this function, we probably should. But in this case, I'm happier to limit the API surface :sweat_smile:

joncinque:
Eh, I'll just do the backport manually then.

PR #251, opened 2025-07-25
[Backport maintenance/v2.x] sha256-hasher: Deprecate `extend_and_hash`

joncinque:
#### Problem

As noted in https://github.com/anza-xyz/agave/pull/7028, the `extend_and_hash` function is unnecessary and creates another allocation, so we inlined it in Agave.

#### Summary of changes

Deprecate it formally in the crate.


joncinque:
CI issue will be addressed with #252

febo (approved):
👌

PR #255, opened 2025-07-28
[solana-bls] Remove `once-cell`

samkim-crypto:
#### Problem
An optimization in https://github.com/anza-xyz/solana-sdk/pull/237 requires pre-computing a fixed value that is re-used for all verification. I first tried using `std::sync::LazyLock`, but I wanted the bls crate to be no-std compatible, so I used the `once-cell` crate. But I realized that `once-cell` also uses `std` under the hood.

#### Summary of Changes
I updated the code so that the pre-computation step happens if the `std` feature is set and computes it on-the-fly (before the optimization) when the `std` feature is not set. I also removed the `once-cell` dependency and just use the `std::sync::LazyLock`.


joncinque (approved):
Looks great!

PR #256, opened 2025-07-30
{instruction,program}error!: remove unused std features

kevinheavey:
The instruction-error and program-error crates have `std` features in Cargo.toml that don't do anything

### DIFF
--- a/instruction-error/src/lib.rs
+++ b/instruction-error/src/lib.rs
@@ -5,8 +5,6 @@ use core::fmt;
 use num_traits::ToPrimitive;
 #[cfg(feature = "frozen-abi")]
 use solana_frozen_abi_macro::{AbiEnumVisitor, AbiExample};
-#[cfg(feature = "std")]
-extern crate std;
 use solana_program_error::ProgramError;
 pub use solana_program_error::{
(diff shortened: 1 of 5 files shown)


joncinque (approved):
Very nice, thanks!

instruction-error/src/lib.rs
```diff
@@ -1,13 +1,11 @@
 #![no_std]
 #![cfg_attr(feature = "frozen-abi", feature(min_specialization))]
-use core::fmt;
 #[cfg(feature = "num-traits")]
 use num_traits::ToPrimitive;
 #[cfg(feature = "frozen-abi")]
 use solana_frozen_abi_macro::{AbiEnumVisitor, AbiExample};
-#[cfg(feature = "std")]
+#[cfg(feature = "frozen-abi")]
 extern crate std;
```
apfitzge, line 8: will this be confusing since std is included even without a `std` feature?

i guess my expectation for a `no-std` crate is that it doesn't use `std` unless you pass `std` feature.

kevinheavey: frozen-abi doesn't count imo, if someone pulls that in they have other problems

joncinque: I'd agree with that -- the frozen-abi feature is only used for a specific test build, and the macros happen to require std, so we have to pull in std somehow.

Issue #235, opened 2025-07-16
sysvar: Use `sol_get_sysvar` instead of `sol_get_<NAME>_sysvar`


joncinque:
#### Problem

All of the `get` functions exposed by the `Sysvar` trait use sysvar-specific syscalls. For example, the getter for rent uses `sol_get_rent_sysvar`: https://github.com/anza-xyz/solana-sdk/blob/8eef25b054c4e5ca6d8b879744456297a187db92/sysvar/src/rent.rs#L134

Currently, the sysvar-specific syscalls cost the sysvar base cost + `size_of::<T>()`. On the other hand, the generic syscall costs the `sysvar_base_cost + id_cost (size_of::<Pubkey> / bytes_per_cu) + max(size_of::<T> / bytes_per_cu, mem_op_base_cost)`.

Canceling out the common term, we compare `size_of::<T>()` to `max(size_of::<T>() / 250, 10)` (id_cost is 0). This means that for any sysvar 10 bytes or greater, the general syscall is always cheaper in terms of CUs.

#### Proposed solution

Update `impl_sysvar_get!` to use the generic syscall instead. There's two ways of doing this:

* if #87 goes forward, then we can just implement it on the simplified `Sysvar` trait
* otherwise, create a new trait, e.g. `SysvarGet`, which implements its `get` function via the generic syscall

cc @rustopian

joncinque:
Fixed with #257

PR #257, opened 2025-07-30
Use sol_get_sysvar instead of sol_get_<NAME>_sysvar

rustopian:
Issue #235


rustopian:
CI fix waiting on [264](https://github.com/anza-xyz/solana-sdk/pull/264)

sysvar/src/lib.rs
```diff
@@ -161,6 +161,27 @@ pub trait SysvarSerialize:
...
+            let mut uninit = core::mem::MaybeUninit::<Self>::uninit();
+            let size = core::mem::size_of::<Self>() as u64;
+            // Safety: we build a mutable slice pointing to the uninitialized
+            // buffer.  The `get_sysvar` syscall will fill exactly `size`
+            // bytes, after which the buffer is fully initialised.
+            let dst = unsafe {
+                core::slice::from_raw_parts_mut(uninit.as_mut_ptr() as *mut u8, size as usize)
+            };
+            // Attempt to load the sysvar data.
+            $crate::get_sysvar(dst, &id(), 0, size)?;
```
febo: You probably don't need to create a slice, you can pass the mutable pointer from the `MaybeUninit`.
```suggestion
            $crate::get_sysvar(uninit.as_mut_ptr() as *mut _ as *mut u8, &id(), 0, size)?;
```

rustopian: Sounds good, but doesn't work, assuming I understand you correctly. `get_sysvar` [is defined](https://github.com/rustopian/solana-sdk/blob/0a7699de0f6eeefb8cf8f6b19c65ad7b7d435430/sysvar/src/lib.rs#L207) to take a mutable slice. I'm not sure it's worth changing that, as the change would basically shuffle some things around without saving any cycles.

febo: My bad, did not see that the `get_sysvar` is a wrapper over the syscall. We probably want to tweak that since it seems odd to convert something to a slice and do a length check when you already allocated the correct memory space for it. In any case, something for a future PR.

define-syscall/src/definitions.rs
```diff
@@ -33,12 +33,12 @@ define_syscall!(fn sol_get_sysvar(sysvar_id_addr: *const u8, result: *mut u8, of
...
 define_syscall!(fn sol_panic_(filename: *const u8, filename_len: u64, line: u64, column: u64));
 
-// these are to be deprecated once they are superceded by sol_get_sysvar
-define_syscall!(fn sol_get_clock_sysvar(addr: *mut u8) -> u64);
-define_syscall!(fn sol_get_epoch_schedule_sysvar(addr: *mut u8) -> u64);
-define_syscall!(fn sol_get_rent_sysvar(addr: *mut u8) -> u64);
-define_syscall!(fn sol_get_last_restart_slot(addr: *mut u8) -> u64);
-define_syscall!(fn sol_get_epoch_rewards_sysvar(addr: *mut u8) -> u64);
+// these are deprecated - use sol_get_sysvar instead
+define_syscall!(#[deprecated(since = "3.0.0", note = "Use `sol_get_sysvar` instead")] fn sol_get_clock_sysvar(addr: *mut u8) -> u64);
```
joncinque: nit: not too sure about this, but how about specifying "Use `sol_get_sysvar` with the clock sysvar's address" to be super clear?

sysvar/src/epoch_rewards.rs
```diff
@@ -163,8 +163,43 @@ pub use {
...
+
+    fn to_bytes<T>(value: &T) -> Vec<u8> {
+        unsafe {
+            let size = core::mem::size_of::<T>();
+            let ptr = (value as *const T) as *const u8;
+            let mut data = vec![0u8; size];
+            std::ptr::copy_nonoverlapping(ptr, data.as_mut_ptr(), size);
+            data
+        }
+    }
```
joncinque: nit: how about using `crate::tests::to_bytes` instead?

sysvar/src/fees.rs
```diff
@@ -76,4 +76,52 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    #[serial]
+    fn test_fees_get_deprecated_syscall_path() {
+        let prev = crate::program_stubs::set_syscall_stubs(Box::new(MockFeesSyscall));
+        let got = Fees::get().unwrap();
+        // restore original stubs to avoid side effects on other tests
+        let _ = crate::program_stubs::set_syscall_stubs(prev);
```
joncinque: Is this needed? It looks like all of the tests in this PR properly set the syscall stubs.

I'd consider it a bug if a test expects a particular set of stubs to be set without setting it itself.

sysvar/src/lib.rs
joncinque: nit: assuming that `id()` is available (and correct) in the scope of the macro could cause unexpected errors -- how about passing the sysvar id explicitly in the macro definition?

that way callers would call the macro like `impl_sysvar_get(id())` instead

sysvar/src/last_restart_slot.rs
```diff
@@ -45,8 +45,36 @@ pub use {
...
+
+    fn to_bytes<T>(value: &T) -> Vec<u8> {
+        unsafe {
+            let size = core::mem::size_of::<T>();
+            let ptr = (value as *const T) as *const u8;
+            let mut data = vec![0u8; size];
+            std::ptr::copy_nonoverlapping(ptr, data.as_mut_ptr(), size);
+            data
+        }
+    }
```
joncinque: nit: how about using `crate::tests::to_bytes` instead?

sysvar/src/lib.rs
rustopian: Oof, good note. Brittle, and tests overfitted.

Passing as you suggest is clearly the best option, as opposed to specifying/checking in macro. Done in 57bf78e5

sysvar/src/fees.rs
rustopian: Good note; it's no longer needed. Also removed in 5ba7eb55

sysvar/src/epoch_rewards.rs
rustopian: Used in 5ba7eb55

define-syscall/src/definitions.rs
rustopian: Done for all in 5ba7eb55

jamie-osec:
I'm not sure exactly why, but this appears to be breaking. While [updating Anchor](https://github.com/solana-foundation/anchor/pull/4031) to use the 3.0 SDK, we encountered strange errors:
```
  Program log: program : AccountInfo
  Program log: Instruction: IdlCreateAccount
  Program log: ProgramError occurred. Error Code: InvalidArgument. Error Number: 8589934592. Error Message: The arguments provided to a program instruction were invalid.
  Program 2cLanTyE8xvwEGSApMxVpX13zcdhiDGK5fRcjniE3T6Z consumed 10334 of 200000 compute units
  Program 2cLanTyE8xvwEGSApMxVpX13zcdhiDGK5fRcjniE3T6Z failed: invalid program argument
```
which came from `Rent::get()?`. Locking `sysvar` to 3.0.x resolved this.

rustopian:
> I'm not sure exactly why, but this appears to be breaking. While [updating Anchor](https://github.com/solana-foundation/anchor/pull/4031) to use the 3.0 SDK, we encountered strange errors:
> 
> ```
>   Program log: program : AccountInfo
>   Program log: Instruction: IdlCreateAccount
>   Program log: ProgramError occurred. Error Code: InvalidArgument. Error Number: 8589934592. Error Message: The arguments provided to a program instruction were invalid.
>   Program 2cLanTyE8xvwEGSApMxVpX13zcdhiDGK5fRcjniE3T6Z consumed 10334 of 200000 compute units
>   Program 2cLanTyE8xvwEGSApMxVpX13zcdhiDGK5fRcjniE3T6Z failed: invalid program argument
> ```
> 
> which came from `Rent::get()?`. Locking `sysvar` to 3.0.x resolved this.

Thanks for bringing this up; I'll look into it as soon as possible.

sysvar/src/lib.rs
```diff
@@ -179,6 +182,24 @@ macro_rules! impl_sysvar_get {
...
+            // Allocate uninitialized memory for the sysvar struct
+            let mut uninit = core::mem::MaybeUninit::<Self>::uninit();
+            let size = core::mem::size_of::<Self>() as u64;
+            // Safety: we build a mutable slice pointing to the uninitialized
+            // buffer.  The `get_sysvar` syscall will fill exactly `size`
+            // bytes, after which the buffer is fully initialised.
+            let dst = unsafe {
+                core::slice::from_raw_parts_mut(uninit.as_mut_ptr() as *mut u8, size as usize)
+            };
+            // Attempt to load the sysvar data using the provided sysvar id.
```
jamie-osec, line 196: This is UB. The data here is uninitialized, and slices must be initialized. From the [docs](https://doc.rust-lang.org/std/slice/fn.from_raw_parts_mut.html):
> `data` must point to `len` consecutive properly initialized values of type `T`.

PR #258, opened 2025-07-30
deps: Use k256 instead of libsecp256k1

### LINKED ISSUE #241: Build fails for `aarch64-apple-watchos` watchOS  target because `getrandom` is too old
I''m building a watchOS app and trying to use a Rust shared library with `uniffi-rs` and `cargo swift-package`. Build fails because of `getrandom` is too old and target is not supported.

solana-sdk `Cargo.toml`:
```
getrandom = { version = "0.1.1", features = ["wasm-bindgen"] }
```

Error running `cargo swift package`:
```
   Compiling getrandom v0.2.16
   Compiling rand_core v0.6.4
   Compiling zeroize v1.8.1
   Compiling bytemuck v1.23.1
   Compiling five8_core v0.1.2
   Compiling five8 v0.2.1
   Compiling solana-sanitize v2.2.1
   Compiling solana-atomic-u64 v2.2.1
   Compiling solana-decode-error v2.3.0
   Compiling five8_const v0.1.4
   Compiling lazy_static v1.5.0
   Compiling itoa v1.0.15
   Compiling log v0.4.27
   Compiling ppv-lite86 v0.2.21
   Compiling opaque-debug v0.3.1
   Compiling rand_chacha v0.3.1
error: target is not supported, for more information see: https://docs.rs/getrandom/#unsupported-targets
   --> /Users/setoelka/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/getrandom-0.1.16/src/lib.rs:267:9
    |
267 | /         compile_error!("\
268 | |             target is not supported, for more information see: \
269 | |             https://docs.rs/getrandom/#unsupported-targets\
270 | |         ");
    | |__________^

   Compiling block-buffer v0.9.0
   Compiling hashbrown v0.13.2
   Compiling bitflags v2.9.1
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `imp`
   --> /Users/setoelka/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/getrandom-0.1.16/src/lib.rs:291:5
    |
291 |     imp::getrandom_inner(dest)
    |     ^^^ use of unresolved module or unlinked crate `imp`
    |
    = help: if you wanted to use a crate named `imp`, use `cargo add imp` to add it to your `Cargo.toml`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `getrandom` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
```

In their latest version however, it's [supported](https://docs.rs/getrandom/latest/getrandom/#supported-targets):

`iOS, tvOS, watchOS | *‑apple‑{ios,tvos,watchos} | CCRandomGenerateBytes`

joncinque:
#### Problem

As noted in https://github.com/anza-xyz/agave/pull/4744 and #241, there are a few issues with libsecp256k1, mainly that it's unmaintained and has old dependencies, which causes some builds to break.

#### Summary of changes

Port everything over to `k256` instead. It has a lot of similar interfaces, but I wasn't sure about certain functions, such as `to_encoded_point(false)`, and whether we're using the message hash correctly.

To bring more confidence, I expanded one of the doctests to actually run the code.

Fixes #241 

cc @kevinheavey thanks for bringing this up in Agave!


secp256k1-recover/src/lib.rs
```diff
@@ -419,14 +426,15 @@ pub fn secp256k1_recover(
...
             .map_err(|_| Secp256k1RecoverError::InvalidRecoveryId)?;
-        let signature = libsecp256k1::Signature::parse_standard_slice(signature)
-            .map_err(|_| Secp256k1RecoverError::InvalidSignature)?;
-        let secp256k1_key = libsecp256k1::recover(&message, &signature, &recovery_id)
+        let signature = k256::ecdsa::Signature::from_slice(signature)
             .map_err(|_| Secp256k1RecoverError::InvalidSignature)?;
-        Ok(Secp256k1Pubkey::new(&secp256k1_key.serialize()[1..65]))
+        let secp256k1_key =
+            k256::ecdsa::VerifyingKey::recover_from_prehash(hash, &signature, recovery_id)
+                .map_err(|_| Secp256k1RecoverError::InvalidSignature)?;
```
samkim-crypto, line 439: I think this is one breaking change. Before, if the `hash` is not 32 bytes, then the function will return `InvalidHash` Now, this case will return `InvalidSignature` error. It is your call, but we can introduce a length check if we wanted to preserve the error behavior previously. I think the `InvalidHash` is actually not used anywhere else other than here.

```
    if hash.len() != 32 {
        return Err(Secp256k1RecoverError::InvalidHash);
    }
```

secp256k1-program/src/lib.rs
```diff
@@ -322,6 +323,7 @@
 //! #             }))
 //! #     }
 //! # }
+//! use k256::elliptic_curve::scalar::IsHigh;
```
samkim-crypto, line 326: I can't comment on the unchanged lines, but in line 273 we say `libsecp256k1` is used while in the example, we use `k256`. Should this line also be updated?

secp256k1-recover/src/lib.rs
joncinque: Good point, done!

secp256k1-program/src/lib.rs
joncinque: I wasn't sure about this one... it says:

> The Solana program. Note that it uses `libsecp256k1` version 0.7.0 to parse the secp256k1 signature to prevent malleability.

The runtime is still using libsecp256k1, so I thought it was referencing that. You're probably right though, done!

PR #260, opened 2025-08-01
sdk-wasm-js: Make wrapper types public

joncinque:
#### Problem

While working on extracting the wasm js code from solana-system-interface, I noticed that it's impossible to create the wrapper types, since we don't expose any constructors.

#### Summary of changes

Make the wrapped type `pub` instead of `pub(crate)` so that other wasm-js crates can properly create types compatible with wasm_bindgen.

### DIFF
--- a/sdk-wasm-js/src/hash.rs
+++ b/sdk-wasm-js/src/hash.rs
@@ -7,7 +7,7 @@ use {
 
 #[wasm_bindgen]
 #[derive(Debug, PartialEq, Eq, Clone)]
-pub struct Hash(pub(crate) solana_hash::Hash);
+pub struct Hash(pub solana_hash::Hash);
 
 #[allow(non_snake_case)]
 #[wasm_bindgen]
--- a/sdk-wasm-js/src/instruction.rs
+++ b/sdk-wasm-js/src/instruction.rs
@@ -10,7 +10,7 @@ use {crate::pubkey::Pubkey, wasm_bindgen::prelude::*};
 /// is fixed. This must not diverge from the regular non-wasm Instruction struct.
 #[wasm_bindgen]
 #[derive(Debug, PartialEq, Eq, Clone)]
-pub struct Instruction(pub(crate) solana_instruction::Instruction);
+pub struct Instruction(pub solana_instruction::Instruction);
 
 #[wasm_bindgen]
 impl Instruction {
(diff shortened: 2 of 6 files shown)


joncinque:
I went with a slightly different approach after seeing the tests fail, not totally confident that this is the best option, so let me know if you have any better ideas!

sdk-wasm-js/src/hash.rs
```diff
@@ -7,7 +7,19 @@ use {
...
 #[derive(Debug, PartialEq, Eq, Clone)]
-pub struct Hash(pub(crate) solana_hash::Hash);
+pub struct Hash {
+    pub(crate) inner: solana_hash::Hash,
+}
+
+impl Hash {
+    pub fn new(inner: solana_hash::Hash) -> Self {
+        Self { inner }
+    }
```
grod220: I wonder if a

```rust
impl From<solana_hash::Hash> for Hash {}
```

would be more idiomatic here if the behavior is more akin to conversions?

sdk-wasm-js/src/hash.rs
```diff
@@ -7,7 +7,19 @@ use {
...
+
+    pub fn inner(&self) -> &solana_hash::Hash {
+        &self.inner
+    }
```
grod220: Think inner is fine, but you could also consider `impl Deref for Hash` so the consumer could call right into the inner type's methods directly.

grod220 (commented):
Could we add tests to validate the new constructors work? Though, I'm not sure it would work outside of a test-crate that imports these and runs wasm-pack on that test crate.

sdk-wasm-js/src/hash.rs
joncinque: That's a good point -- I've noticed that people are typically split about whether using `From` is better or worse than constructors. Since it's not really a converter, I went with a constructor in this case, but I have no issues with using a `From` impl instead

sdk-wasm-js/src/hash.rs
joncinque: That's also a good point and would likely make this type easier to use. I've noticed people are also split on this one, because they're not always sure what type they have, but the usability win is probably worth it.

joncinque:
Ok, I believe I added what you asked for, and refactored a bit to use a macro instead -- let me know what you think!

sdk-wasm-js-tests/package.json
```diff
@@ -0,0 +1 @@
+../sdk-wasm-js/package.json
```
grod220, line 1: Does this automatically inherit the other? Didn't know you could do that. It doesn't feel like valid json lol.

joncinque: This is actually a symlink -- GitHub shows them weirdly it seems :sweat_smile:

PR #263, opened 2025-08-05
CI: Don't run clippy during publish

joncinque:
#### Problem

Since we're starting to release v3 crates, the repo is in a half-upgraded state, which means that the circular dependencies like solana-system-interface and solana-stake-interface will fail to build due to the patches.

#### Summary of changes

Keep things simple, and don't run the repo-wide clippy before publish.

### DIFF
--- a/.github/workflows/publish-rust.yml
+++ b/.github/workflows/publish-rust.yml
@@ -77,6 +77,7 @@ jobs:
     name: Clippy
     needs: [sanity]
     runs-on: ubuntu-latest
+    if: false
     steps:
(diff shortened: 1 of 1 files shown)


joncinque:
Ah darn, the whole repo build is going to be messed up for a little bit while we publish everything.

joncinque:
I could bring solana-system-interface back into the repo and remove solana-stake-interface from solana-sysvar to resolve everything, but I'm not sure if that's worth it

joncinque:
Let's just do the right thing: #264

PR #264, opened 2025-08-05
Fixup build: remove stake-interface, bring system-interface back

joncinque:
#### Problem

The build is currently failing due to circular dependencies between the sdk, system-interface, and stake-interface. This is totally normal since we're publishing breaking changes, but in general, these circular dependencies can be tricky to manage. See #263 for more info.

#### Summary of changes

Remove the circular dependencies for good:

* BREAKING: remove stake-interface from the sysvar crate
* bring system-interface back into the repo
* update system-interface based on breaking changes currently in the sdk
* remove all patches from the sdk toml

Work to do after:

* add back the wasm-js code for the system interface via a new crate
* fix up solana-stake-interface to implement the sysvar trait once all the crates have been published


example-mocks/Cargo.toml
```diff
@@ -11,6 +11,11 @@ edition = { workspace = true }
 
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
+all-features = true
+rustdoc-args = ["--cfg=docsrs"]
+
+[features]
+bincode = ["solana-system-interface/bincode"]
```
joncinque, line 18: This change looks unrelated, but it's unfortunately needed.

The docs build fails without it, because example-mocks is re-exporting from system-interface, but doesn't have the bincode feature enabled, which means that the instructions referenced in the docs don't exist.

Cargo.toml
```diff
@@ -331,38 +330,3 @@ opt-level = 1
...
-solana-clock = { path = "clock" }
-solana-cpi = { path = "cpi" }
-solana-frozen-abi = { path = "frozen-abi" }
-solana-frozen-abi-macro = { path = "frozen-abi-macro" }
-solana-instruction = { path = "instruction" }
-solana-program-error = { path = "program-error" }
-solana-pubkey = { path = "pubkey" }
-solana-rent = { path = "rent" }
-solana-signature = { path = "signature" }
-solana-sysvar-id = { path = "sysvar-id" }
```
joncinque, line 368: This is the best part of this change, and the whole motivation

system-interface/src/instruction.rs
```diff
@@ -0,0 +1,1718 @@
...
+//! for the [`SystemInstruction`] variants for each System program instruction,
+//! and these variants are linked from the documentation for their constructors.
+//!
+//! [`RpcClient`]: https://docs.rs/solana-client/latest/solana_client/rpc_client/struct.RpcClient.html
+//! [cpi]: https://docs.rs/solana-cpi/latest/solana_cpi/index.html
+//! [`invoke`]: https://docs.rs/solana-cpi/latest/solana_cpi/fn.invoke.html
+//! [`invoke_signed`]: https://docs.rs/solana-cpi/latest/solana_cpi/fn.invoke_signed.html
+//! [`AccountInfo`]: https://docs.rs/solana-account-info/latest/solana_account_info/struct.AccountInfo.html
+//! [`Instruction`]:
+//! https://docs.rs/solana-instruction/latest/solana_instruction/struct.Instruction.html
```
joncinque, line 40: This is the only change in this file

sysvar/src/lib.rs
```diff
@@ -184,7 +183,7 @@ macro_rules! impl_sysvar_get {
 
 /// Handler for retrieving a slice of sysvar data from the `sol_get_sysvar`
 /// syscall.
-fn get_sysvar(
+pub fn get_sysvar(
```
joncinque, line 186: We'll need this function when reimplementing the moved logic in `solana-stake-interface`, so make it public

sysvar/Cargo.toml
```diff
@@ -52,7 +47,7 @@ solana-fee-calculator = { workspace = true }
 solana-frozen-abi = { workspace = true, optional = true }
 solana-frozen-abi-macro = { workspace = true, optional = true }
 solana-hash = { workspace = true, features = ["bytemuck"] }
-solana-instruction = { workspace = true }
+solana-instruction = { workspace = true, features = ["std"] }
```
joncinque: This was being enabled through solana-stake-interface

system-interface/README.md
```diff
@@ -0,0 +1,104 @@
+<p align="center">
+  <a href="https://solana.com">
+    <img alt="Solana" src="https://github.com/user-attachments/assets/534af75d-6347-48dc-8943-129423b2ba63" height="80" />
+  </a>
+</p>
+
+# Solana System Interface
+
+This crate contains instructions and constructors for interacting with the [System program](https://docs.solanalabs.com/runtime/programs#system-program).
```
rustopian: Not really essential to this PR, but

```suggestion
This crate contains instructions and constructors for interacting with the [System program](https://solana.com/docs/core/programs#core-programs).
```

The old link has an awkward redirect notice...  which arguably takes you to the wrong page anyway (account model).

rustopian (approved):
Looks good. Some improvements to error.rs, rm wasm as you mentioned, otherwise diffs are clean.

I noticed one sort-of-broken link in the docs. Up to you.

system-interface/README.md
rustopian: Hmm, this doesn't seem like quite the right place either, even though it's where the redirect ends up. I'll find better.

joncinque: Works for me!

joncinque:
For what it's worth, I know it kind of stinks to not use the program-specific repo for the system interface, but it's just too tied into the rest of the sdk unfortunately.

I prefer the sanity of a self-contained repo over splitting things up. It'll also mean that tools like `cargo-semver-checks` just work, instead of requiring fixes or hacks on our side.

Also, since the plans for a BPF system program have been put on hold for a long time, the program-specific repo becomes a bit less necessary for the rust crates. We can certainly keep hosting the idl and kit library though.

PR #265, opened 2025-08-05
signature: add optional bytemuck support

kevinheavey:
I currently find myself converting to the raw bytes to use bytemuck, so this would be nice to have


joncinque:
Can you rebase?

PR #266, opened 2025-08-05
sysvar: Deprecate stake-history and functions

joncinque:
#### Problem

The stake-history module in solana-sysvar will be removed with #264, but there hasn't been a deprecation warning.

#### Summary of changes

Put in a deprecation warning.

### DIFF
--- a/sysvar/src/stake_history.rs
+++ b/sysvar/src/stake_history.rs
@@ -74,7 +74,7 @@ impl Sysvar for StakeHistory {
 // we do not provide Default because this requires the real current epoch
 #[deprecated(
     since = "2.3.0",
-    note = "Use solana_stake_interface::sysvar::stake_history::StakeHistorySysvar in v2 instead"
+    note = "Use solana_stake_interface::sysvar::stake_history::StakeHistorySysvar in solana-stake-interface v2 instead"
 )]
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct StakeHistorySysvar(pub Epoch);


sysvar/src/stake_history.rs
```diff
@@ -68,12 +72,17 @@ impl Sysvar for StakeHistory {
 }
 
 // we do not provide Default because this requires the real current epoch
+#[deprecated(
+    since = "2.3.0",
+    note = "Use solana_stake_interface::sysvar::stake_history::StakeHistorySysvar in v2 instead"
+)]
```
buffalojoec: What does "in v2 instead" mean?

buffalojoec (approved):

sysvar/src/stake_history.rs
joncinque: We're inverting which crate implements the sysvar code and trait from `solana-sysvar` to `solana-stake-interface`, but since only one crate can implement the trait, we have to do it in a breaking change, ie in `solana-stake-interface` v2 and `solana-sysvar` v3. 

With normal deprecation warnings, there's typically an alternative that you can use right away, but we don't have the alternative ready in `solana-sysvar` v2, so I wanted to make clear that you need to use `solana-stake-interface` v2. Is it clearer if it says this?

```
Use solana_stake_interface::sysvar::stake_history::StakeHistorySysvar in solana-stake-interface v2 instead
```

buffalojoec: Yep! That's a little more clear, thanks!

PR #267, opened 2025-08-06
Add Alt-BN128 Little Endian operations for SIMD-0284

LStan:
[SIMD-0284](https://github.com/solana-foundation/solana-improvement-documents/pull/284) suggests extending the Alt-BN128 syscalls to add little-endian support. In this PR I added the required functions.

Questions to consider:
1. `alt_bn128_pairing` for `solana` target returns a vector, but only one byte is actually significant. I made `alt_bn128_pairing_le` to return `u8`. Should I change it to return a vector, as in `alt_bn128_pairing`?
2. There are `ALT_BN128_SUB` and `ALT_BN128_SUB_LE` constants, but actual functions for them are not implemented. Should I keep the constants, remove them, or implement the functions?


deanmlittle:
implement them

> [SIMD-0284](https://github.com/solana-foundation/solana-improvement-documents/pull/284) suggests extending the Alt-BN128 syscalls to add little-endian support. In this PR I added the required functions.
> 

At first glance, this looks pretty good to me aside from the 2 issues mentioned. Leaving a comment so I can subscribe to the updates for this PR. I will have a proper review when I get a chance. Also tagging @samkim-crypto @0x0ece for review.

> Questions to consider:
> 
> 1. `alt_bn128_pairing` for `solana` target returns a vector, but only one byte is actually significant. I made `alt_bn128_pairing_le` to return `u8`. Should I change it to return a vector, as in `alt_bn128_pairing`?

You should return the Vec<u8> to maintain the existing API and style of `solana-sdk`. In reality, the point of this change is so we don't need to use `solana-sdk` to interact with these programs and can just use `ark-bn254` directly. We can always do better in a `nostd` or `pinocchio` SDK too.

> 2. There are `ALT_BN128_SUB` and `ALT_BN128_SUB_LE` constants, but actual functions for them are not implemented. Should I keep the constants, remove them, or implement the functions?

Imo, don't bother to implement them for now. While they will be implemented in the syscall, the less code for us to review, the easier it will be to get this PR merged.

LStan:
I think that at first [SIMD-0334](https://github.com/solana-foundation/solana-improvement-documents/pull/334) should be considered, accepted or rejected, implemented or not, than this PR

0x0ece:
I approved SIMD-0334 on the FD side, I think we should just do it. I agree with @deanmlittle recommendations for this PR, but of course need @samkim-crypto or someone else from Anza to approve here.

bn254/benches/bn254.rs
```diff
@@ -24,6 +30,27 @@ fn bench_addition(c: &mut Criterion) {
...
+        70, 145, 97, 12, 93, 59, 116, 3, 78, 9, 61, 201, 6, 60, 144, 156, 71, 32, 132, 12, 181, 19,
+        76, 185, 245, 159, 167, 73, 117, 87, 150, 129, 150, 88, 211, 46, 252, 13, 40, 129, 152,
+        243, 114, 102,
+    ]);
+    let q_bytes = convert_endianness::<32, 64>(&[
+        7, 194, 183, 245, 138, 132, 189, 97, 69, 240, 12, 156, 43, 192, 187, 26, 24, 127, 32, 255,
+        44, 146, 150, 58, 136, 1, 158, 124, 106, 1, 78, 237, 6, 97, 78, 32, 193, 71, 233, 64, 242,
+        215, 13, 163, 247, 76, 154, 23, 223, 54, 23, 6, 164, 72, 92, 116, 43, 214, 120, 132, 120,
+        250, 23, 215,
+    ]);
```
samkim-crypto: This is a nit, but do you think we can refactor out constants from the functions to something like
```
const ADD_P_BYTES_BE: [u8; 64] = [
    24, 177, 138, 207, 180, 194, 195, 2, 118, 219, 84, 17, 54, 142, 113, 133, 179, 17, 221, 18,
    70, 145, 97, 12, 93, 59, 116, 3, 78, 9, 61, 201, 6, 60, 144, 156, 71, 32, 132, 12, 181, 19,
    76, 185, 245, 159, 167, 73, 117, 87, 150, 129, 150, 88, 211, 46, 252, 13, 40, 129, 152,
    243, 114, 102,
];

const ADD_Q_BYTES_BE: [u8; 64] = [
    7, 194, 183, 245, 138, 132, 189, 97, 69, 240, 12, 156, 43, 192, 187, 26, 24, 127, 32, 255,
    44, 146, 150, 58, 136, 1, 158, 124, 106, 1, 78, 237, 6, 97, 78, 32, 193, 71, 233, 64, 242,
    215, 13, 163, 247, 76, 154, 23, 223, 54, 23, 6, 164, 72, 92, 116, 43, 214, 120, 132, 120,
    250, 23, 215,
];
```
and then use these constants instead inside the function? I think we just need to define the BE constants and then in the LE functions, we can do something like:

```suggestion
    let p_bytes_le = convert_endianness::<32, 64>(&ADD_P_BYTES_BE);
    let q_bytes_le = convert_endianness::<32, 64>(&ADD_Q_BYTES_BE);
```

Seems like this will make things a bit readable with all these long constants.

bn254/src/lib.rs
```diff
@@ -152,6 +156,15 @@ mod target_arch {
             reverse_copy(&be_bytes[FIELD_SIZE..], &mut pod_bytes[FIELD_SIZE..])?;
             Ok(Self(pod_bytes))
         }
+
+        #[inline(always)]
+        fn from_le_bytes(le_bytes: &[u8]) -> Result<Self, AltBn128Error> {
```
samkim-crypto, line 163: I can't comment on non-diff lines, but can you update the docs on top of `pub struct PodG1(pub [u8; G1_POINT_SIZE]);` especially regarding
```
/// The Solana syscalls still expect the inputs to be encoded in big-endian as
/// specified in EIP-197. The type `PodG1` is an intermediate type that
/// facilitates the translation between the EIP-197 encoding and the arkworks
/// implementation encoding.
```

bn254/src/lib.rs
```diff
@@ -192,6 +205,15 @@ mod target_arch {
             )?;
             Ok(Self(pod_bytes))
         }
+
+        #[inline(always)]
+        fn from_le_bytes(le_bytes: &[u8]) -> Result<Self, AltBn128Error> {
```
samkim-crypto, line 214: Likewise, can you update the docs on top of `pub struct PodG2(pub [u8; G2_POINT_SIZE]);`? Here, I think it is important to make clear of the ordering convention of the polynomial variables.

bn254/src/lib.rs
```diff
@@ -39,6 +39,10 @@ mod consts {
     pub const ALT_BN128_SUB: u64 = 1;
     pub const ALT_BN128_MUL: u64 = 2;
     pub const ALT_BN128_PAIRING: u64 = 3;
+    pub const ALT_BN128_ADD_LE: u64 = ALT_BN128_ADD | 0x80;
```
samkim-crypto: Nit: Can we define `0x80` as a constant?

```
const LE_FLAG: u64 = 0x80;
...
pub const ALT_BN128_ADD_LE: u64 = ALT_BN128_ADD | LE_FLAG;
pub const ALT_BN128_SUB_LE: u64 = ALT_BN128_SUB | LE_FLAG;
...
```

samkim-crypto (commented):
Thanks for making the changes! Things look good to me. Just some pretty minor stylistic suggestions below. I agree with @deanmlittle's suggestions to the two questions.

> alt_bn128_pairing for solana target returns a vector, but only one byte is actually significant. I made alt_bn128_pairing_le to return u8. Should I change it to return a vector, as in alt_bn128_pairing?

Yeah let's stick with the original vector output for now for consistency. This can be updated later on a follow-up.

> There are ALT_BN128_SUB and ALT_BN128_SUB_LE constants, but actual functions for them are not implemented. Should I keep the constants, remove them, or implement the functions?

For this as well, let's keep the constants for consistency.

LStan:
Question about input length checks in add and mul. Currently, the input is checked that it is not greater than some length. If it is shorter, it is padded with zeros. But does it make sense for LE? Maybe for LE it would be better to check for strict equality?

0x0ece:
Yes, strict please. If it were for me, I'd change to strict also big endian and remove the padding.

samkim-crypto:
Yeah I agree with @0x0ece. Let's do strict equality for LE in this PR. It would be nice to update the BE implementation as well, but that would require a feature gate, so let's consider this on a follow-up.

samkim-crypto (approved):
Looks good to me. I'll give @deanmlittle or @0x0ece a day or two to sanity check this and then merge it in.

0x0ece (approved):
LGTM. Just for planning purposes, this will go in 3.1, or do you expect it in 3.0?

deanmlittle:
> LGTM. Just for planning purposes, this will go in 3.1, or do you expect it in 3.0?

We haven't discussed release schedule, but as soon as you are ready on the FD side, I think we can push this into whatever is the next release.

LStan:
I still think that SIMD-0334 should be implemented before merging this

samkim-crypto:
> LGTM. Just for planning purposes, this will go in 3.1, or do you expect it in 3.0?

v3.0 is already published, so this will go into v3.1.

> I still think that SIMD-0334 should be implemented before merging this

SIMD-0334 requires a breaking change on the existing function. Patching breaking changes are a bit annoying in that we need to create a new function with the patch in order to feature gate the update in the agave repo (like `alt_bn128_multiplication` and `alt_bn128_multiplication_128`).

After this PR is merged, I was going to do some refactoring in this crate to make future breaking patches easier to apply. Basically update these functions to take in an enum for the versioning.

So let's first merge this PR, refactor, and then apply SIMD-0334.

LStan:
But then you'll have to create new functions for both BE and LE pairing.
And SIMD-0334 fix can be something simple like [this](https://github.com/anza-xyz/solana-sdk/compare/master...LStan:solana-sdk:pairing-check-fix)

samkim-crypto:
> But then you'll have to create new functions for both BE and LE pairing. And SIMD-0334 fix can be something simple like [this](https://github.com/anza-xyz/solana-sdk/compare/master...LStan:solana-sdk:pairing-check-fix)

Yeah I was thinking that we merge as is and then we can hold off on publishing the crate until SIMD-0334 is merged/implemented (both SIMD-0284 and 0334 can be included in v3.1). In this case, we can just apply SIMD-0334 directly to the LE pairing and we just have to define a new function for BE.

The main reason I want to merge is that I can work on the refactoring while we wait for SIMD-0334 to be merged. I'd be fine to wait for SIMD-0334 too, but I feel like it might be cleaner to make the interface a bit more modular before we implement SIMD-0334, which is another breaking change. What do you think?

LStan:
Ok, let's do that. I've just added `#[inline(always)]` for pairings

samkim-crypto (approved):

PR #268, opened 2025-08-06
Fix sign offchain message in solana-offchain-message

rustopian:
## Problem
Offline signing CLI doesn't use up-to-date spec and thus does not work with Ledger ([agave issue](https://github.com/anza-xyz/agave/issues/3340)).

## Solution
This is a backwards-compatible fix to align the solana-offchain-message crate with [the offchain message specification](https://github.com/anza-xyz/agave/blob/master/docs/src/proposals/off-chain-message-signing.md).

`OffchainMessage::new()` supplies default parameters for the missing items so that the crate immediately works with any current dependents: this includes application domain as well as dummy signer count and signer pubkeys. `sign()` replaces signer information inline.

`OffchainMessage::new_with_domain()` allows the application domain to be passed in. The CLI in Agave does not currently use this, but it will be trivial to implement in a later PR.

`OffchainMessage::new_with_params()` allows both application domain and signer information to be passed in.

Most of the added code is unit tests.

## Testing in CLI with Ledger

Update agave to include this patch in Cargo.toml:
```
solana-offchain-message = { version = "2.2.1", git = "https://github.com/rustopian/solana-sdk", branch = "fix-sign-offchain-message" }
```

Compile the CLI:
```
cargo build --package solana-cli --release
```

Set config to ledger, plug in a Ledger and open the Solana app, sanity-check by querying address, and sign message:
```
./target/release/solana config set --keypair usb://ledger
./target/release/solana address
./target/release/solana sign-offchain-message "test"
```


offchain-message/Cargo.toml
```diff
@@ -20,15 +20,15 @@ verify = ["dep:solana-pubkey", "solana-signature/verify"]
...
-solana-sha256-hasher = { workspace = true, features = ["sha2"] }
-solana-signature = { workspace = true }
-solana-signer = { workspace = true }
+solana-hash = "2.2.1"
+solana-packet = "2.2.1"
+solana-pubkey = { version = "2.4.0", optional = true }
+solana-sanitize = "2.2.1"
+solana-sha256-hasher = { version = "2.2.1", features = ["sha2"] }
+solana-signature = "2.3.0"
+solana-signer = "2.2.1"
```
rustopian, line 29: Needed to match the agave CLI versions to avoid type errors so that people can easily test the Ledger signing. Probably can be reverted in the future.

PR #269, opened 2025-08-06
Fix offline message signing in CLI (align to spec)

rustopian:
[ WIP: back to draft pending implementation of updated spec, see [7628](https://github.com/anza-xyz/agave/pull/7628) ]

## Problem
Offline message signing in CLI doesn't align with up-to-date spec and thus does not work with Ledger hardware wallets ([agave issue](https://github.com/anza-xyz/agave/issues/3340)).

## Solution
This is a fix to align the solana-offchain-message crate with [the offchain message specification](https://github.com/anza-xyz/agave/blob/master/docs/src/proposals/off-chain-message-signing.md).

Breakage is not expected, but if any caller formerly relied on this crate to sign non-compliant messages and then implemented some other function which assumes the old non-compliant message format for some reason, they may encounter an issue, since default values for the specification's required parameters are now added prior to signing.

For compatibility with current callers, `OffchainMessage::new()` supplies default parameters for the missing items so that the crate immediately works with any current dependents: this includes application domain as well as dummy signer count and signer pubkey. `sign()` replaces signer information with the correct values.

`OffchainMessage::new_with_domain()` allows the application domain to be passed in. The CLI in Agave does not currently use this, but it will be trivial to implement in a later PR.

`OffchainMessage::new_with_params()` allows both application domain and signer information to be passed in. This is especially important when signing a message with multiple signers.

`Envelope` is not essential and is not implemented.

## Testing in CLI with Ledger

1. Add this to agave Cargo.toml dependencies:
```
solana-offchain-message = { version = "3.0.0", git = "https://github.com/rustopian/solana-sdk", branch = "cli-sign-offchain-message-testing" }
```
Ensure that cli/Cargo.toml uses either the same line or `{ workspace = true }`.

**Note that this branch `cli-sign-offchain-message-testing` is slightly different**; it modifies the offchain-message Cargo.toml to use 2.x versions for dependencies so that agave will not complain about a host of mismatches while the transition to 3.0.0 is still ongoing.

Once agave [7492](https://github.com/anza-xyz/agave/pull/7492) is merged, testing will work with this PR branch  instead:

```
solana-offchain-message = { version = "3.0.0", git = "https://github.com/rustopian/solana-sdk", branch = "cli-sign-offchain-message" }
```

2. Update solana-offchain-message:
```
cargo update -p solana-offchain-message
```

3. Compile the CLI:
```
cargo build --package solana-cli --release
```

4. Plug in a Ledger and open the Solana app
5. Set config to ledger, sanity-check by querying address, and sign message:
```
./target/release/solana config set --keypair usb://ledger
./target/release/solana address
./target/release/solana sign-offchain-message "test"
```

### DIFF
--- a/offchain-message/src/lib.rs
+++ b/offchain-message/src/lib.rs
@@ -59,6 +59,9 @@ pub fn is_utf8(data: &[u8]) -> bool {
 }
 
 #[allow(clippy::arithmetic_side_effects)]
+/// `v0::OffchainMessage` is shared by header versions 0 and 1 to preserve
+/// API compatibility; version-specific policy (e.g., newline allowance) is
+/// enforced at construction and (de)serialization via `serialization::Spec`.
 pub mod v0 {
     use {
         super::{MessageFormat, OffchainMessage as Base},


rustopian:
CI waiting on #264

offchain-message/src/lib.rs
```diff
@@ -245,12 +427,182 @@ impl OffchainMessage {
...
+/// Envelope for off-chain messages with multiple signatures
+/// All signers listed in the message must provide signatures (no threshold logic)
+/// This implements the envelope format from the spec:
+/// | Field | Start offset | Length (bytes) | Description |
+/// | Signature Count | 0x00 | 1 | Number of signatures |
+/// | Signatures | 0x01 | `SIG_COUNT` * 64 | ed25519 signatures |
+/// | Message Preamble | 0x01 + `SIG_COUNT` * 64 | variable | The message preamble |
+/// | Message Body | varies | variable | The message content |
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct Envelope {
```
rustopian, line 439: I'm not sure whether this is implemented elsewhere and I can just pull it in; will look more later. Any tips appreciated if so.

apfitzge:
@rustopian I noticed this is still being pushed to quite often. Is this ready for review at this point, or should I hold off?

rustopian:
> @rustopian I noticed this is still being pushed to quite often. Is this ready for review at this point, or should I hold off?

TL;DR - code is ready for review.

I was working on getting Ledger testing working in agave with this crate prior to its 3.0.0 update (which isn't yet merged), but at last I pushed a patch to a dedicated testing branch as noted in the PR descrip.

So if you want to test agave CLI with a Ledger, you'll need to use that branch with its different deps for this crate. (Only Cargo files differ there.)

Once agave is merged to 3.0.0 I'll update the PR descrip, but this PR itself shouldn't change.

offchain-message/src/lib.rs
```diff
@@ -38,109 +32,40 @@ pub enum MessageFormat {
...
+        pub(crate) format: MessageFormat,
+        pub(crate) signers: Vec<[u8; 32]>,
+        pub(crate) message: Vec<u8>,
     }
 
     impl OffchainMessage {
-        // Header Length = Message Format (1) + Message Length (2)
-        pub const HEADER_LEN: usize = 3;
-        // Max length of the OffchainMessage
+        pub const HEADER_LEN: usize = 36;
```
joncinque: It took me a while to figure out what this corresponds to, can we document it here? Ie.

```
/// Encoded byte length for the application domain, format, number of signers, and message length
```

offchain-message/src/lib.rs
```diff
@@ -8,27 +11,18 @@ use {
...
+
+// Assertions to prevent accidental drift
 #[cfg(test)]
-static_assertions::const_assert_eq!(OffchainMessage::HEADER_LEN, 17);
-#[cfg(test)]
-static_assertions::const_assert_eq!(v0::OffchainMessage::MAX_LEN, 65515);
+static_assertions::const_assert_eq!(v0::OffchainMessage::MAX_LEN, 65482);
 #[cfg(test)]
-static_assertions::const_assert_eq!(v0::OffchainMessage::MAX_LEN_LEDGER, 1212);
+static_assertions::const_assert_eq!(v0::OffchainMessage::MAX_LEN_LEDGER, 1179);
```
joncinque, line 20: I'm not sure if this is correct, since the preamble contains the pubkey of the signer, which means that a ledger can only sign a message with a length of 1147, and not 1179

offchain-message/src/lib.rs
```diff
@@ -38,109 +32,40 @@ pub enum MessageFormat {
...
+        pub(crate) signers: Vec<[u8; 32]>,
+        pub(crate) message: Vec<u8>,
     }
 
     impl OffchainMessage {
-        // Header Length = Message Format (1) + Message Length (2)
-        pub const HEADER_LEN: usize = 3;
-        // Max length of the OffchainMessage
+        pub const HEADER_LEN: usize = 36;
         pub const MAX_LEN: usize = u16::MAX as usize - Base::HEADER_LEN - Self::HEADER_LEN;
```
joncinque, line 86: nit: to be clearer, what do you think about renaming this to something like `FORMAT_2_MAX_LEN`?

offchain-message/src/serialization.rs
```diff
@@ -0,0 +1,334 @@
...
+    use {
+        super::SanitizeError,
+        crate::{serialization, v0::OffchainMessage as V0OffchainMessage, MessageFormat},
+        solana_serialize_utils::{
+            append_slice, append_u16, append_u8, read_slice, read_u16, read_u8,
+        },
+    };
+
+    /// Components of a v0 message: (application_domain, format, signers, message)
+    pub type V0MessageComponents = ([u8; 32], MessageFormat, Vec<[u8; 32]>, Vec<u8>);
```
joncinque: nit: tuples past 2 types can be a bit cumbersome -- is this actually needed though? It looks just like a tuple version of `v0::OffchainMessage`, and that way we can also remove the `new_with_params` in here

Also, I don't feel too strongly about this, but since this is already in a `v0` mod, we might not need the `V0` prefix in the typename too.

offchain-message/src/lib.rs
```diff
@@ -169,37 +109,81 @@ impl OffchainMessage {
...
+        // Use dummy signer that will be replaced during signing
+        Self::new_with_params(version, application_domain, &[[0u8; 32]], message)
+    }
+
+    /// Construct a new OffchainMessage object with all parameters from the spec
+    ///
+    /// # Usage Patterns:
+    /// - **Single-signer with custom domain**: You may pass`&[[0u8; 32]]` for signers,
+    ///   in which case actual signer will be filled in when `sign()` is called
+    /// - **Multi-signer predefined**: Pass real signer pubkeys, all signers must provide signatures
```
joncinque, line 158: I know this is an annoying situation, since you get into a chicken-and-egg situation easily. You need the signer pubkeys in order to construct the message, but you might not have them right away, so you want to put in some sort of placeholder. Transactions have the same issue.

Unfortunately, this comment highlights an unintuitive usage pattern that's prone to error.

Maybe we can type these properly to make it easier to use, ie:

* Construct an `OffchainMessage` with the message / version / domain. Users can also provide the pubkeys for the expected signers.
* Add signatures with `sign()` or `add_signature()`
* Try to get a `SignedOffchainMessage` from an `OffchainMessage`, which is immutable, or can give you back an `OffchainMessage` for modifying.

What do you think?

offchain-message/src/lib.rs
```diff
@@ -230,11 +214,27 @@ impl OffchainMessage {
...
+                Self::new_with_params(0, application_domain, &[signer_pubkey], message)?;
+            return Ok(signer.sign_message(&rebuilt_message.serialize()?));
+        }
+        if !message_signers.contains(&signer_pubkey) {
+            return Err(SanitizeError::InvalidValue);
+        }
         Ok(signer.sign_message(&self.serialize()?))
     }
 
-    #[cfg(feature = "verify")]
```
joncinque, line 237: Why was this removed?

offchain-message/src/lib.rs
```diff
@@ -230,11 +214,27 @@ impl OffchainMessage {
     }
 
     /// Sign the message with provided keypair
+    /// If message was created with dummy signer, update it with actual signer.
+    /// For spec compliance, verify signer matches expected pubkey in message.
     pub fn sign(&self, signer: &dyn Signer) -> Result<Signature, SanitizeError> {
```
joncinque, line 269: We can change this in a breaking change in the future, but `sign` is quite heavy since it requires `Signer` and `Signature`.

People might just want the message serde functions and sign separately, but they'll need to pull everything else in.

offchain-message/Cargo.toml
```diff
@@ -15,17 +15,18 @@ all-features = true
 rustdoc-args = ["--cfg=docsrs"]
 
 [features]
-dev-context-only-utils = ["verify"]
-verify = ["dep:solana-pubkey", "solana-signature/verify"]
+dev-context-only-utils = []
+verify = []
```
joncinque, line 19: Why was this changed?

offchain-message/src/serialization.rs
```diff
@@ -0,0 +1,334 @@
...
+        let signers = parse_signers(data, &mut offset, signer_count)?;
+        serialization::reject_zero_pubkey_in_multi_signer(&signers)?;
+        let message_len = parse_message_length(data, &mut offset)?;
+        let message = parse_message_body(data, &mut offset, message_len)?;
+
+        let total_size = preamble_and_body_size(signers.len(), message_len);
+        let expected_format = serialization::detect_format(total_size, &message)?;
+        if expected_format != format {
+            return Err(SanitizeError::InvalidValue);
+        }
```
joncinque: This check looks prone to false negatives because someone could use a higher format than required in their message.

The code here just needs to make sure that the deserialized format doesn't violate the message contents.

offchain-message/src/serialization.rs
```diff
@@ -0,0 +1,334 @@
...
+) -> Result<(), SanitizeError> {
+    if signers.is_empty() || signers.len() > u8::MAX as usize {
+        return Err(SanitizeError::ValueOutOfBounds);
+    }
+    reject_zero_pubkey_in_multi_signer(signers)?;
+    if message.is_empty() {
+        return Err(SanitizeError::InvalidValue);
+    }
+    Ok(())
+}
```
joncinque, line 31: nit: up to you, but separating this into two functions for `validate_signers` and `validate_message` might be clearer

offchain-message/src/serialization.rs
```diff
@@ -0,0 +1,334 @@
...
+        if data.len() < V0OffchainMessage::HEADER_LEN {
+            return Err(SanitizeError::ValueOutOfBounds);
+        }
+
+        let mut offset = 0;
+        let application_domain = parse_application_domain(data, &mut offset)?;
+        let format = parse_message_format(data, &mut offset)?;
+        let signer_count = parse_signer_count(data, &mut offset)?;
+        let signers = parse_signers(data, &mut offset, signer_count)?;
+        serialization::reject_zero_pubkey_in_multi_signer(&signers)?;
```
joncinque: nit: up to you, but since most of the validation is happening inside of the individual `parse_*` functions, this check could go into `parse_signers` instead

offchain-message/src/serialization.rs
rustopian: Good note, I've replaced this with a `required_minimal_format` so that "higher" formats are allowed in 6bd77154.

Note that this section is prone to a tiny change once we ascertain something related to Ledger implementation's support for LF chars and related incoming updates to the spec.

offchain-message/src/serialization.rs
rustopian: Removed in 39b23e43.

I considered removing alias `V0OffchainMessage` in favor of specifying `v0::OffchainMessage`, but this leads to specifying `crate::v0::` repeatedly within `serialization::v0`.

So `use crate::v0::OffchainMessage as V0OffchainMessage` remains just to make those instances a little more resilient.

rustopian: Removed in bee1b96d.

The similarly named `use crate::v0::OffchainMessage as V0OffchainMessage` remains in order to prevent repeated confusing uses of `crate::v0::` within `serialization::v0`.

offchain-message/src/lib.rs
rustopian: Thanks for pushing on this.

Agreed, the placeholder-signer pattern is error-prone. I wanted to avoid breaking the current `new` → `sign` → `verify` path used by agave CLI, but I can deprecate that path and add something like your suggestion without too many LoC.

offchain-message/src/serialization.rs
```diff
@@ -0,0 +1,331 @@
...
+        let declared_format = parse_message_format(data, &mut offset)?;
+        let signer_count = parse_signer_count(data, &mut offset)?;
+        let signers = parse_signers(data, &mut offset, signer_count)?;
+        serialization::reject_zero_pubkey_in_multi_signer(&signers)?;
+        let message_len = parse_message_length(data, &mut offset)?;
+        let message = parse_message_body(data, &mut offset, message_len)?;
+
+        let total_size = preamble_and_body_size(signers.len(), message_len);
+        let required_minimal_format = serialization::detect_minimal_format(total_size, &message)?;
+        if u8::from(declared_format) < u8::from(required_minimal_format) {
```
joncinque: nit: This makes me slightly worried in case we ever add a new format that isn't a superset of the formats that came before it.

I know it's a bit pedantic, but we could have a function for this check, like `Format::implies` or `Format::includes`?

rustopian: I like this. Changed so that now we have:
`if !declared_format.includes(implied_format) { ... }`
in 716d75b

rustopian:
Back to draft pending [7628](https://github.com/anza-xyz/agave/pull/7628)

steveluscher:
Do you have any feedback on https://github.com/solana-foundation/SRFCs/discussions/3?

PR #272, opened 2025-08-06
address: Gate `ParseAddressError` on `decode` feature

joncinque:
#### Problem

The `ParseAddressError` enum is only needed if the `decode` feature is enabled, but it's currently included in all cases.

#### Summary of changes

Gate `ParseAddressError` on the `decode` feature.


febo (approved):
Great!

PR #275, opened 2025-08-08
SIMD-0194: Deprecate rent exemption threshold

deanmlittle:
Implemented SDK changes outlined in [SIMD-0194](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0194-deprecate-rent-exemption-threshold.md).

This allows us to remove `f64` math from Rent calculation, allowing us to save a bunch of CUs in all newly deployed programs and making the rent_exeption_threshold value inert, simplifying any future changes to rent.

NOTE: This PR should not be merged before feature gate `rent6iVy6PDoViPBeJ6k5EJQrkj62h7DPyLbWGHwjrC` is activated on all networks. The `[deprecated(since = "2.3.X")` message will also need to be updated when a release candidate is decided upon.

Tagging @joncinque to review.

### DIFF
--- a/rent/src/lib.rs
+++ b/rent/src/lib.rs
@@ -30,11 +30,12 @@ static_assertions::const_assert_eq!(
 )]
 #[derive(PartialEq, CloneZeroed, Debug)]
 pub struct Rent {
-    /// Rental rate in lamports/byte-year.
-    pub lamports_per_byte_year: u64,
+    /// Rental rate in lamports/byte.
+    pub lamports_per_byte: u64,
 
     /// Amount of time (in years) a balance must include rent for the account to
     /// be rent exempt.
+    #[deprecated(since = "2.3.X", note = "Use `lamports_per_byte` instead")]
     pub exemption_threshold: f64,
 
     /// The percentage of collected rent that is burned.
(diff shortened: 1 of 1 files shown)


febo:
@deanmlittle What do you think if we remove completely any reference to `f64`? We could do this in 2 steps. First a PR to add the deprecation messages followed by a patch release; a second PR with the refactoring to be used when the feature is activated, which eventually is published as a new major release.

deanmlittle:
> @deanmlittle What do you think if we remove completely any reference to `f64`? We could do this in 2 steps. First a PR to add the deprecation messages followed by a patch release; a second PR with the refactoring to be used when the feature is activated, which eventually is published as a new major release.

I would be *very* happy to replace it with a [u8;8] instead and place a clippy rule in to disallow floating point arithmetic, but I was originally trying to move as few things as possible. If there's appetite from your end, let's do it? I also think we can deprecate like half of the rent-related convenience functions in there too.

febo:
> > @deanmlittle What do you think if we remove completely any reference to `f64`? We could do this in 2 steps. First a PR to add the deprecation messages followed by a patch release; a second PR with the refactoring to be used when the feature is activated, which eventually is published as a new major release.
> 
> I would be _very_ happy to replace it with a [u8;8] instead and place a clippy rule in to disallow floating point arithmetic, but I was originally trying to move as few things as possible. If there's appetite from your end, let's do it? I also think we can deprecate like half of the rent-related convenience functions in there too.

I think we should and we can do it in two steps. The first one just deprecate stuff, so we don't actually change anything; then in the refactoring we remove all the stuff that won't be needed anymore after the feature activation.

Somewhat related to this, I had to [remove](https://github.com/anza-xyz/pinocchio/blob/febo/remove-f64-from-rent/sdk/pinocchio/src/sysvars/rent.rs#L35-L50) all the `f64` things so Alessandro could compile p-token with the upstream rustc ebpf target.

cc: @joncinque

joncinque:
> I think we should and we can do it in two steps. The first one just deprecate stuff, so we don't actually change anything; then in the refactoring we remove all the stuff that won't be needed anymore after the feature activation.

Yes, let's follow what Pinocchio did with https://github.com/anza-xyz/pinocchio/pull/243 first, by deprecating the field and all rent-paying functions, so that we can rename the field and remove the concept of rent-paying in the next breaking change.

PR #278, opened 2025-08-11
vote-interface: rename current to v3

buffalojoec:
Simply renames "current" in each of the `VoteStateVersions` API to "v3".

Paves the way for the introduction of v4 in `VoteStateVersions`.

### DIFF
--- a/vote-interface/src/state/vote_state_versions.rs
+++ b/vote-interface/src/state/vote_state_versions.rs
@@ -20,15 +20,15 @@ use {
 pub enum VoteStateVersions {
     V0_23_5(Box<VoteState0_23_5>),
     V1_14_11(Box<VoteState1_14_11>),
-    Current(Box<VoteStateV3>),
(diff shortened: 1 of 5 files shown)


joncinque:
I'm right about to publish the new major version for vote interface, and since this PR contains breaking changes, I can wait until it lands. Let me know!

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -20,15 +20,15 @@ use {
 pub enum VoteStateVersions {
     V0_23_5(Box<VoteState0_23_5>),
     V1_14_11(Box<VoteState1_14_11>),
-    Current(Box<VoteStateV3>),
+    V3(Box<VoteStateV3>),
```
jstarry, line 23: Normally we would have to deprecate and introduce a whole new struct for a variant rename like this but since it's a major version release I think it's ok to skip that step.

jstarry (approved):

PR #279, opened 2025-08-11
system-wasm-js: Introduce crate for exposing wasm-js instructions

joncinque:
#### Problem

The system interface used to include wasm-js bindings for the instruction creators, which was a problem for non-JS wasm builds, as described at https://github.com/solana-program/system/issues/47.

#### Summary of changes

Add the wasm-js bindings as an external crate, similar to the work done with sdk-wasm-js.

Fixes https://github.com/solana-program/system/issues/47
Closes https://github.com/solana-program/system/pull/48

### DIFF
--- a/system-wasm-js/package.json
+++ b/system-wasm-js/package.json
@@ -1,0 +1,6 @@
+{
+  "scripts": {
+    "postinstall": "npm run build",
+    "build": "wasm-pack build --target nodejs --dev --out-dir node_modules/crate --out-name crate"
+  }
+}
--- a/system-wasm-js/src/lib.rs
+++ b/system-wasm-js/src/lib.rs
@@ -1,0 +1,116 @@
+//! `SystemInstruction` Javascript interface
+#![cfg(target_arch = "wasm32")]
+#![allow(non_snake_case)]
+use {
+    solana_sdk_wasm_js::{address::Address, instruction::Instruction},
+    solana_system_interface::instruction::{
+        advance_nonce_account, allocate, allocate_with_seed, assign, assign_with_seed,
+        authorize_nonce_account, create_account, create_account_with_seed, create_nonce_account,
+        transfer, transfer_with_seed, withdraw_nonce_account,
+    },
+    wasm_bindgen::prelude::*,
+};
+
+#[wasm_bindgen]
+struct SystemInstruction;
+
+#[wasm_bindgen]
+impl SystemInstruction {
+    pub fn createAccount(
+        from: &Address,
+        to: &Address,
+        lamports: u64,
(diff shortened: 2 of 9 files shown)


system-wasm-js/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+    solana_system_interface::instruction::{
+        advance_nonce_account, allocate, allocate_with_seed, assign, assign_with_seed,
+        authorize_nonce_account, create_account, create_account_with_seed, create_nonce_account,
+        transfer, transfer_with_seed, withdraw_nonce_account,
+    },
+    wasm_bindgen::prelude::*,
+};
+
+#[wasm_bindgen]
+struct SystemInstruction;
```
grod220: Does this need to be `pub struct`?

system-wasm-js/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "scripts": {
+    "postinstall": "npm run build",
+    "build": "wasm-pack build --target nodejs --dev --out-dir node_modules/crate --out-name crate"
```
grod220, line 4: This should be `--release` right? or perhaps this is not being used to publish/expose the npm package?

system-wasm-js/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    pub fn transfer(from: &Address, to: &Address, lamports: u64) -> Instruction {
+        transfer(from, to, lamports).into()
+    }
+
+    pub fn transferWithSeed(
+        from: &Address,
+        from_base: &Address,
+        from_seed: String,
```
grod220, line 61: Surprised this is not a `&str`, but saw that `transfer_with_seed` requires an owned string

system-wasm-js/src/lib.rs
joncinque: Ah good point... it certainly can't hurt.

system-wasm-js/src/lib.rs
joncinque: Yeah, this is meant to look like the existing function 🤷

system-wasm-js/package.json
joncinque: This is more of a testing command, since we're not uploading this anywhere.

I tested build times, and release builds take roughly the same amount of time as dev builds. In fact, the initial compilation is faster for release, but then the optimization takes up another 3 seconds.

I'd lean towards sticking with a dev build, but I can be convinced otherwise if you feel better about it.

grod220: Sounds good to me 👍

PR #280, opened 2025-08-12
vote-interface: `VoteStateVersions`: add support for v4

buffalojoec:
Adds support for Vote State v4 to `VoteStateVersions`, by adding a `V4` variant and the necessary conversions.


vote-interface/src/state/mod.rs
```diff
@@ -414,12 +420,30 @@ mod tests {
...
+            .votes
+            .resize(MAX_LOCKOUT_HISTORY, LandedVote::default());
+        vote_state.root_slot = Some(1);
+        let versioned = VoteStateVersions::new_v4(vote_state);
+        assert!(VoteStateV4::serialize(&versioned, &mut buffer[0..4]).is_err());
+        VoteStateV4::serialize(&versioned, &mut buffer).unwrap();
+        assert_eq!(
+            VoteStateV4::deserialize(&buffer, &vote_pubkey).unwrap(),
+            versioned.try_convert_to_v4(&vote_pubkey).unwrap()
+        );
```
jstarry: This test is kinda weird due to the `vote_pubkey` argument actually being ignored for vote state v4. Maybe explicitly pass two unique keys here to make it a little more clear?

vote-interface/src/state/mod.rs
```diff
@@ -889,6 +1113,11 @@ mod tests {
         let vote_state = VoteStateVersions::new_v3(vote_state);
         let size = serialized_size(&vote_state).unwrap();
         assert_eq!(VoteStateV3::size_of() as u64, size);
+
+        let vote_state = VoteStateV4::get_max_sized_vote_state();
```
jstarry, line 1145: nit: prefer to have separate unit tests for v3 and v4

vote-interface/src/state/vote_state_deserialize.rs
```diff
@@ -106,6 +111,127 @@ pub(super) fn deserialize_vote_state_into(
...
+        SourceVersion::V1_14_11 { vote_pubkey } | SourceVersion::V3 { vote_pubkey } => {
+            // V1_14_11 and V3 have commission field here.
+            let commission = read_u8(cursor)?;
+
+            // Set collectors based on SIMD-0185.
+            // Safety: if vote_state is non-null, collectors are guaranteed to be valid too
+            unsafe {
+                // We already read `node_pubkey` earlier, so we can read it
+                // here again.
+                let node_pubkey = (*vote_state).node_pubkey;
```
jstarry: You're right that `node_pubkey` is now initialized but we need to be careful about the syntax here to avoid UB. What you have here will create an intermediate reference to the whole struct which is UB..

https://doc.rust-lang.org/std/ptr/macro.addr_of.html

> Creating a reference with &/&mut is only allowed if the pointer is properly aligned and points to initialized data. For cases where those requirements do not hold, raw pointers should be used instead. However, &expr as *const _ creates a reference before casting it to a raw pointer, and that reference is subject to the same rules as all other references. This macro can create a raw pointer without creating a reference first.

I think we should instead read the node pubkey into a variable above and then reuse it here

vote-interface/src/state/vote_state_deserialize.rs
```diff
@@ -106,6 +111,127 @@ pub(super) fn deserialize_vote_state_into(
...
+                None,
+            )
+        }
+    };
+
+    // For V3 and V4, `has_latency` is always true.
+    let votes = read_votes(
+        cursor,
+        !matches!(source_version, SourceVersion::V1_14_11 { .. }),
+    )?;
```
jstarry: nit: can you create a `has_latency` variable and pass to `read_votes`? I think it's more clear that way

vote-interface/src/state/vote_state_deserialize.rs
```diff
@@ -171,6 +297,36 @@ fn read_prior_voters_into<T: AsRef<[u8]>>(
...
+// Merely updates the cursor to skip over this section.
+fn skip_prior_voters<T: AsRef<[u8]>>(cursor: &mut Cursor<T>) -> Result<(), InstructionError> {
+    for _ in 0..MAX_ITEMS {
+        read_pubkey(cursor)?; // prior_voter
+        read_u64(cursor)?; // from_epoch
+        read_u64(cursor)?; // until_epoch
+    }
+    read_u64(cursor)?; // idx
+    read_bool(cursor)?; // is_empty
+    Ok(())
```
jstarry: `prior_voters` has a fixed size so we can just check if the cursor is long enough to contain that size and throw an error if it doesn't. In `VoteStateView` I did:

```rust
        cursor.consume(PriorVotersFrame::total_size());

        if cursor.position() as usize <= bytes.len() {
            Ok(Self {
                votes_frame,
                root_slot_frame,
                authorized_voters_frame,
                epoch_credits_frame,
            })
        } else {
            Err(VoteStateViewError::AccountDataTooSmall)
        }
```

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -65,3 +69,176 @@ pub struct VoteStateV4 {
...
 }
+
+impl VoteStateV4 {
+    /// Upper limit on the size of the Vote State
+    /// when votes.len() is MAX_LOCKOUT_HISTORY.
+    pub const fn size_of() -> usize {
+        3762 // Same size as V3 to avoid account resizing
+    }
+
+    pub fn new_rand_for_tests(node_pubkey: Pubkey, root_slot: Slot) -> Self {
```
jstarry: I prefer to leave this out.. all the vote state fields are public so callers can implement this if they want

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -65,3 +69,176 @@ pub struct VoteStateV4 {
...
+    pub(crate) fn get_max_sized_vote_state() -> Self {
+        use {
+            super::{MAX_EPOCH_CREDITS_HISTORY, MAX_LOCKOUT_HISTORY},
+            solana_epoch_schedule::MAX_LEADER_SCHEDULE_EPOCH_OFFSET,
+        };
+
+        let mut authorized_voters = AuthorizedVoters::default();
+        for i in 0..=MAX_LEADER_SCHEDULE_EPOCH_OFFSET {
+            authorized_voters.insert(i, Pubkey::new_unique());
+        }
```
jstarry: I'll be honest I don't really understand this logic (I understand it was copied from v3). The check actually looks correct (v3 looks wrong because it stores one less authorized voter than v4) but I think we can be more direct and introduce a local constant saying that there is a max of `4` of authorized voter entries for v4

vote-interface/src/state/vote_state_deserialize.rs
buffalojoec: Ahh good call. I fixed this up.

vote-interface/src/state/vote_state_deserialize.rs
buffalojoec: Ok cool I went with something similar, but `skip_prior_voters` isn't the last read, so I had to adjust the check to instead make sure we didn't overrun the buffer.

vote-interface/src/state/vote_state_deserialize.rs
```diff
@@ -171,6 +292,40 @@ fn read_prior_voters_into<T: AsRef<[u8]>>(
     Ok(())
 }
 
+// Same navigation as `read_prior_voters_into`, but does not perform any writes.
+// Merely updates the cursor to skip over this section.
+fn skip_prior_voters<T: AsRef<[u8]>>(cursor: &mut Cursor<T>) -> Result<(), InstructionError> {
```
jstarry, line 297: can you add a test for this function too?

PR #281, opened 2025-08-12
CI: Run main workflow on PRs targeting maintenance too

joncinque:
#### Problem

PRs against the maintenance branches don't trigger CI, which is an issue for checking backport PRs.

#### Summary of changes

Add the maintenance branches to CI.


yihau (approved):
🫡

PR #285, opened 2025-08-13
Create `InvalidAccountAliasing` error

LucasSte:
**Problem**

Program runtime relies on Rust's `RefCell::try_borrow_mut` function to determine if an instruction is borrowing the same account twice. That might happen when, for instance, two accounts are aliasing each other and the instruction wants them to be different.

The borrow checker adds complexity to the program runtime, so an active approach to detecting aliases with the Pubkey would help us remove the `RefCell`. There is no proper error to replace `AccountBorrowFailed` or `AccountBorrowOutstading` in the repository.

**Solution**

Create error `InvalidAccountAliasing` to throw where the borrow is failing.

### DIFF
--- a/instruction-error/src/lib.rs
+++ b/instruction-error/src/lib.rs
@@ -200,6 +200,9 @@ pub enum InstructionError {
 
     /// Builtin programs must consume compute units
     BuiltinProgramsMustConsumeComputeUnits,
+
+    /// The instruction cannot work with the two aliased accounts received
+    InvalidAccountAliasing,
     // Note: For any new error added here an equivalent ProgramError and its
     // conversions must also be added
 }
@@ -340,6 +343,9 @@ impl fmt::Display for InstructionError {
             }
             InstructionError::BuiltinProgramsMustConsumeComputeUnits => {
                 f.write_str("Builtin programs must consume compute units")
+            }
+            InstructionError::InvalidAccountAliasing => {
+                f.write_str("The instruction does not allow one or more account aliasing")
             }
         }
     }


Lichtso:
Not sure what is different about this error code as opposed to the two existing ones?

LucasSte:
> Not sure what is different about this error code as opposed to the two existing ones?

If there is no more `try_borrow` for accounts, why would we emit `AccountBorrowFailed`? What you actually do not want are aliasing accounts.

LucasSte:
Hmm, maybe we should simply return a `InvalidInstructionData` instead, like zk-token-proof is doing here:

https://github.com/anza-xyz/agave/blob/cd291424d3d71c1a3be0c2c919916dcaa272d162/programs/zk-token-proof/src/lib.rs#L159-L161

Lichtso:
That seems more wrong, it is not about the instruction data after all, but the accounts.

LucasSte:
> That seems more wrong, it is not about the instruction data after all, but the accounts.

What is your suggestion then? What could replace `AccountBorrowFailed` when we don't have RefCells anymore? My idea was `InvalidAccountAliasing`.

LucasSte:
> > That seems more wrong, it is not about the instruction data after all, but the accounts.
> 
> What is your suggestion then? What could replace `AccountBorrowFailed` when we don't have RefCells anymore? My idea was `InvalidAccountAliasing`.

In instructions that rely on the RefCell to prevent the account aliasing, I would check their private keys first, and then return an error.

PR #287, opened 2025-08-15
[bls-signatures] Add signature batch verification functions

samkim-crypto:
#### Problem
There is no function to do batch verification in the bls-signatures crate.

#### Summary of Changes
I added a new function called `par_verify_batch`:
- It first aggregates the signatures and verifies the signature. If the aggregated signature is valid, then it returns `[true, ..., true]` signifying that all the signatures are valid.
- If the aggregate verification fails, then it individually verifies the signatures.

I also added `par_verify_batch_binary_search` function:
- This function also aggregates the signatures and verifies the signatures.
- If the aggregate verification fails, then it recursively applies the aggregate verification in a binary search fashion to identify the false signatures.

According to the benches, if there is more than a single invalid signature in batch, the `par_verify_batch` seems to outperform the `par_verify_batch_binary_search`. For sigverify in alpenglow, let's use the `par_verify_batch` function for now and see if we can really optimize the binary search function later.


bls-signatures/src/signature.rs
```diff
@@ -139,13 +155,55 @@ impl SignatureProjective {
...
+    ///
+    /// This function first attempts to verify all signatures at once via aggregation.
+    /// If the aggregate verification succeeds, it returns a vector of `true`s.
+    /// If it fails, it falls back to verifying each signature individually in parallel
+    /// to identify the valid and invalid ones.
+    #[cfg(feature = "parallel")]
+    pub fn par_verify_batch<P: AsPubkeyProjective + Sync, S: AsSignatureProjective + Sync>(
+        public_keys: &[&P],
+        signatures: &[&S],
+        message: &[u8],
```
wen-coding, line 169: Hmm, if I'm an attacker I would send you different messages to break your batch processing. But of course we can implement rate limiting for votes with messages different from all others.

bls-signatures/src/signature.rs
```diff
@@ -139,13 +155,55 @@ impl SignatureProjective {
...
+        // First, try to verify by aggregating all public keys and signatures.
+        if SignatureProjective::par_aggregate_verify(public_keys, signatures, message)? {
+            Ok(alloc::vec![true; public_keys.len()])
+        } else {
+            // Use Rayon's parallel iterator to verify each signature concurrently.
+            public_keys
+                .par_iter()
+                .zip(signatures.par_iter())
+                .map(|(pubkey, signature)| {
+                    let pubkey_projective = pubkey.try_as_projective()?;
```
wen-coding: We only have 2000 pubkeys per Epoch, if try_as_projective() is not a cheap operation, we can totally pass in PubkeyProjective.

bls-signatures/src/signature.rs
```diff
@@ -139,13 +155,55 @@ impl SignatureProjective {
...
+                    let signature_projective = signature.try_as_projective()?;
```
wen-coding: Could one broken signature break the verification of whole batch for us?

bls-signatures/src/signature.rs
```diff
@@ -154,6 +212,224 @@ impl SignatureProjective {
...
+                    SignatureProjective::identity(),
+                ),
+                |(mut pk_acc, mut sig_acc), (pk, sig)| {
+                    pk_acc.0 += &pk.0;
+                    sig_acc.0 += &sig.0;
+                    (pk_acc, sig_acc)
+                },
+            )
+        } else {
+            // Parallel aggregation
```
wen-coding, line 287: Any performance difference between the two approaches?

Also, because we already have the pubkey and signature aggregation for the whole array, can we do
pubkey_right = pubkey_whole - pubkey_left
To save half of the aggregation?

samkim-crypto: Yeah this depends on the aggregation threshold. If the threshold is about 32, then sequential aggregation seems to perform better on my machine and above the threshold, parallel aggregation seems to perform better. This can be set flexibly with https://github.com/anza-xyz/solana-sdk/pull/287/files#diff-0f77cab3c556ce3179738af6aa815e236128d388be108a6290c3230ba3fd49ddR40, so we should experiment and see what the best threshold is for a typical validator hardware.

samkim-crypto: In terms of `pubkey_right = pubkey_whole - pubkey_left`, we can definitely do that as a form of optimization. To support that, we should create a function that takes in an already aggregated public key instead of a vector/slice of public keys.

But since this is an optimization, I think we should complete sigverify with the current API initially and then incrementally add optimization.

bls-signatures/src/signature.rs
samkim-crypto: Yeah that is an excellent point. The current implementation terminates early if the conversion fails even though a conversion failure means an invalid signature/pubkey. 

After thinking about this, it seems like the best interface for these functions is to just take in a concrete projective type. This way, the function can't terminate early due to conversion error. I updated the code.

bls-signatures/src/signature.rs
samkim-crypto: Yeah I think it is actually possible to just verify on affine types directly without converting the public keys or signatures to projective. I'll address this in a follow-up.

bls-signatures/src/signature.rs
samkim-crypto: Yes, I think that should be taken care of outside the function. This function just takes in a single message for the batch of public keys and signatures.

bls-signatures/src/signature.rs
wen-coding: Sounds good. Do we still need the two `.try_as_projective` above then?

samkim-crypto: Oh sorry I forgot to remove these lines. It is removed now!

bls-signatures/src/signature.rs
```diff
@@ -154,6 +211,216 @@ impl SignatureProjective {
...
+    ) -> Result<Vec<bool>, BlsError> {
+        if public_keys.len() != signatures.len() {
+            return Err(BlsError::InputLengthMismatch);
+        }
+
+        let inputs: Result<Vec<(PubkeyProjective, SignatureProjective)>, BlsError> = public_keys
+            .iter() // Use sequential iterator
+            .zip(signatures.iter())
+            .map(|(pk, sig)| {
+                let pk_proj = pk.try_as_projective()?;
```
wen-coding: Should we do the same for binary search side as well? Pass in Projective and remove `try_as_projective()`.

samkim-crypto: Yeah sorry I actually meant to remove the sequential version of the binary search verification. When I ran the benches, the parallel binary search performed much better. Sorry that I missed removing this earlier 🙏

PR #288, opened 2025-08-19
[signer-store] Implement `thiserror::Error` for `EncodeError` and `DecodeError`

samkim-crypto:
#### Summary of Changes
Implementing `thiserror::Error` for the two error types in signer-store for better ergonomics and display.

### DIFF
--- a/signer-store/src/lib.rs
+++ b/signer-store/src/lib.rs
@@ -58,16 +58,12 @@ pub enum Version {
 /// An error that can occur during the encoding process.
 #[derive(Error, Debug, PartialEq, Eq)]
 pub enum EncodeError {
-    /// In Base3 encoding, the provided bit-vectors have unmatching lengths.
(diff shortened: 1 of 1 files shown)


signer-store/src/lib.rs
```diff
@@ -55,15 +56,19 @@ pub enum Version {
 }
 
 /// An error that can occur during the encoding process.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Error, Debug, PartialEq, Eq)]
 pub enum EncodeError {
     /// In Base3 encoding, the provided bit-vectors have unmatching lengths.
```
wen-coding: nit: do we still need the comments if we have `#[error]` strings?

wen-coding (dismissed):

signer-store/src/lib.rs
samkim-crypto: Good point! I removed the comments.

wen-coding (approved):

PR #290, opened 2025-08-22
[bn254] Refactor out addition, multiplication, and pairing into separate modules

samkim-crypto:
#### Problem
The logic in `lib.rs` is becoming slightly harder to read as it is getting longer and it is organized by submodules inside files.

#### Summary of Changes
I created logical submodules for the `addition`, `multiplication`, and `pairing` operations for elliptic curves. I then moved the relevant logic in `lib.rs` to these relevant submodules. All the functions and constants in `lib.rs` were behind private submodules (`consts` and `target_arch`) and were re-exported as part of `prelude`. I kept this structure the same, so from the perspective of downstream projects, this should not introduce any visible changes.

### DIFF
--- a/bn254/src/addition.rs
+++ b/bn254/src/addition.rs
@@ -2,12 +2,12 @@ use crate::PodG1;
 pub use target_arch::*;
 
 pub(crate) mod consts {
+    use crate::LE_FLAG;
+
     /// Input length for the add operation.
     pub const ALT_BN128_ADDITION_INPUT_LEN: usize = 128;
     /// Output length for the add operation.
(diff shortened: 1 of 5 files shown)


LStan:
What do you think about moving `LE_FLAG` to `lib.rs` as `pub(crate)`? It is repeated in all files

samkim-crypto:
> What do you think about moving `LE_FLAG` to `lib.rs` as `pub(crate)`? It is repeated in all files

Okay, this should be updated.

LStan:
It is also in `compression.rs`

PR #292, opened 2025-08-24
message: re-export types used in public API

kevinheavey:
This doesn't cover all the types but I made some subjective judgments, for example I think nobody cares if we don't re-export solana_sanitize


message/src/lib.rs
```diff
@@ -128,6 +134,6 @@ pub struct MessageHeader {
 /// As used by the `crate::v0` message format.
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct AddressLookupTableAccount {
-    pub key: solana_pubkey::Pubkey,
-    pub addresses: Vec<solana_pubkey::Pubkey>,
+    pub key: Pubkey,
```
apfitzge: now that the `Address` change has been split should this use that instead? I realize it's just an alias, but if we re-export I'd rather export the right one.

apfitzge (commented):
The changes look fine to me, with the exception of the comment on `Pubkey`.

@joncinque could you weigh in on if we've typically re-exported, and if we want to do that in `sdk` crates? To me it seems like it makes it easier to use things

kevinheavey:
For context I have a similar PR open for solana-keypair (#291) and would like to do the same for solana-transaction

message/src/lib.rs
joncinque: I was going to suggest the exact same thing -- let's avoid a breaking change in the future and just use `Address` from the beginning.

apfitzge:
> For context I have a similar PR open for solana-keypair (https://github.com/anza-xyz/solana-sdk/pull/291) and would like to do the same for solana-transaction

👍 yeah, I just want to check with Jon first on what our general practice here is before I start approving. I'm in support since I'm constantly annoyed with pulling in 3-4 smaller base crates to use one that I actually want because it internally uses types that aren't re-exported (rpc-client for example).

joncinque (commented):
This makes sense to me, thanks! Same comment as Andrew

PR #293, opened 2025-08-25
[bls-signatures] Use affine for `Keypair` and signature verification

samkim-crypto:
#### Problem
The current signature verification function takes in the projective form of the public key and signatures. This is a bit inefficient since the public keys and signatures are typically going to be sent over the network in their affine form. In order to invoke the verification function, we have to currently convert them to their projective form just to call the verification function. The verification function would then immediately convert them back to affine internally before performing the pairing operation.

#### Summary of Changes
I made the following changes:
- Updated verification API: The core verification functions, including `verify_signature` and the parallel batch verifiers, now directly accept the more efficient affine types (`Pubkey`, `Signature`). This removes the needless conversions and make the API more intuitive.
- `Keypair`: This is unrelated to the issue above, but I also updated `Keypair` to hold `Pubkey` (affine rep) instead of `PubkeyProjective` to reduce its in-memory footprint.

In the process, I added new conversion traits like `AsPubkey` and `AsSignature` to work with affine types.

### DIFF
--- a/bls-signatures/benches/bls_signatures.rs
+++ b/bls-signatures/benches/bls_signatures.rs
@@ -2,13 +2,19 @@ use {
     criterion::{criterion_group, criterion_main, Criterion},
     solana_bls_signatures::{
         keypair::Keypair,
-        pubkey::{Pubkey, PubkeyProjective, VerifiablePubkey},
-        signature::{Signature, SignatureProjective},
+        pubkey::{PubkeyProjective, VerifiablePubkey},
+        signature::SignatureProjective,
     },
     std::hint::black_box,
 };
 #[cfg(feature = "parallel")]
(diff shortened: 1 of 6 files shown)


deanmlittle:
LGTM. It is entirely reasonable to use Affine coordinates in a keypair. This is a very common convention used across many other digital signature schemes. The now expired draft from Dan Boneh trying to standardize BLS12-381 mentions serialization to [minimal encoding](https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html#section-2.1-2.2.1) which inherently uses Affine coordinates, however since we are not doing compressed keys to enable us to trade twice as many bytes over the wire in order to avoiding slow G2 decompression, this all makes sense to me.

deanmlittle (approved):

PR #296, opened 2025-08-26
transaction: re-export types used in public API

kevinheavey:
and switch to solana-address

### DIFF
--- a/transaction/src/lib.rs
+++ b/transaction/src/lib.rs
@@ -115,18 +115,18 @@ use {
     serde_derive::{Deserialize, Serialize},
     solana_short_vec as short_vec,
 };
+pub use {
+    solana_address::Address,
+    solana_instruction::{AccountMeta, Instruction},
+    solana_instruction_error::InstructionError,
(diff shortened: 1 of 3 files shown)


transaction/src/lib.rs
```diff
@@ -115,22 +115,23 @@ use {
     serde_derive::{Deserialize, Serialize},
     solana_short_vec as short_vec,
 };
+pub use {
+    solana_address::Address,
+    solana_instruction::{AccountMeta, Instruction},
+    solana_instruction_error::InstructionError,
```
joncinque, line 121: Why was this added? I don't see it used explicitly anywhere in the crate

kevinheavey: It's used in a variant of the TransactionError enum. So if you are working with TransactionError you need InstructionError

joncinque: Ah gotcha, makes sense to me, thanks!

joncinque (approved):

apfitzge (approved):
Perfect - thanks!

apfitzge:
Also just want to say I appreciate your clean commits on these, it does make it easier to review!

PR #297, opened 2025-08-27
[bn254] Refactor addition, multiplication, and pairing functions into `validator` and `prelude` modules

samkim-crypto:


bn254/src/multiplication.rs
```diff
@@ -1,118 +1,124 @@
...
-        match endianness {
-            Endianness::BE => {
-                if input.len() > expected_length {
-                    return Err(AltBn128Error::InvalidInputData);
-                }
+pub fn versioned_multiplication(
+    _version: VersionedMultiplication,
+    input: &[u8],
+    endianness: Endianness,
+    expected_length: usize,
```
LStan: Wouldn't it be better to "move" `expected_length` to a version? If `versioned_multiplication` is public, this parameter can create confusion.

bn254/src/multiplication.rs
```diff
@@ -1,118 +1,124 @@
...
+///
+/// This function is intended to be used by the Agave validator client and exists primarily
+/// for validator code. Solana programs or other downstream projects should use
+/// `alt_bn128_multiplication` or `alt_bn128_multiplication_le` instead.
+///
+/// # Warning
+///
+/// Developers should be extremely careful when modifying this function, as a breaking change
+/// can result in a fork in the Solana cluster. Any such change requires an
+/// approved Solana SIMD. Subsequently, a new `VersionedAddition` variant must be added,
```
LStan: `VersionedAddition` -> `VersionedMultiplication`

bn254/src/pairing.rs
```diff
@@ -1,111 +1,112 @@
...
+///
+/// This function is intended to be used by the Agave validator client and exists primarily
+/// for validator code. Solana programs or other downstream projects should use
+/// `alt_bn128_pairing` or `alt_bn128_pairing_le` instead.
+///
+/// # Warning
+///
+/// Developers should be extremely careful when modifying this function, as a breaking change
+/// can result in a fork in the Solana cluster. Any such change requires an
+/// approved Solana SIMD. Subsequently, a new `VersionedAddition` variant must be added,
```
LStan: `VersionedAddition` -> `VersionedPairing`

bn254/src/multiplication.rs
samkim-crypto: Good point! I absorbed the length into the version as you suggested.

bn254/src/multiplication.rs
```diff
@@ -145,3 +160,14 @@ pub(crate) mod target_arch {
...
+
+#[deprecated(
+    since = "3.1.0",
+    note = "Please use `alt_bn128_multiplication` instead"
+)]
+#[cfg(not(target_os = "solana"))]
+#[inline(always)]
+pub fn alt_bn128_multiplication_128(input: &[u8]) -> Result<Vec<u8>, AltBn128Error> {
+    versioned_multiplication(VersionedMultiplication::V0, input, Endianness::BE)
+    // hard-code length; we will remove this function in the future
```
LStan: Why is this comment below the code?

samkim-crypto: I ended up removing this comment actually since we don't hardcode the length as a function parameter any more.

LStan:
Couple of thoughts:
1. When SIMD-0334 is implemented, how `alt_bn128_pairing` will be changed? `target_os = "solana"` part can be easily changed, but `not(solana)` can't be changed because it is used in the validator. Or when updating the dependency in the validator, new `versioned_*` functions will be used? Maybe it is better to deprecate old functions for `not(solana)` and recommend to use `versioned_* ones for everyone?
2. Naming. All functions have `alt_bn128` prefix. Maybe this prefix should also be added to the `versioned_*` functions? Also there is SIMD-0302 that suggests adding g2 addition and multiplication. Maybe add `g1` to the current `versioned_*` functions? So the names will be `alt_bn128_versioned_g1_addition`, `alt_bn128_versioned_g1_multiplication`, `alt_bn128_versioned_pairing`

bn254/src/lib.rs
```diff
@@ -1,14 +1,50 @@
...
 pub mod compression;
-pub mod multiplication;
-pub mod pairing;
+pub(crate) mod multiplication;
+pub(crate) mod pairing;
 
+/// This module contains the versioned syscall implementations and is intended for use
+/// primarily by validator code.
+#[cfg(not(target_os = "solana"))]
+pub mod validator {
```
joncinque: nit: how about just calling this `versioned`? That seems clearer than `validator`

bn254/src/addition.rs
```diff
@@ -1,107 +1,113 @@
...
+pub const ALT_BN128_ADD: u64 = 0;
+pub const ALT_BN128_SUB: u64 = 1;
+pub const ALT_BN128_ADD_LE: u64 = ALT_BN128_ADD | LE_FLAG;
+pub const ALT_BN128_SUB_LE: u64 = ALT_BN128_SUB | LE_FLAG;
+
+/// The version enum used to version changes to the `alt_bn128_addition` syscall.
+#[cfg(not(target_os = "solana"))]
+pub enum VersionedAddition {
+    V0,
 }
```
joncinque: I was reading through the SIMDs, but just so I'm sure I understand -- is the plan to add a V1 for the BE length check?

joncinque (commented):
Looks great! Just the one nit

bn254/src/lib.rs
samkim-crypto: Yep, okay. Updated!

bn254/src/addition.rs
samkim-crypto: Yes, the plan is to add `V1` and then require that the input length is strictly 128 bytes (`ALT_BN128_ADDITION_INPUT_LEN`) for BE addition.

samkim-crypto:
> 1. When SIMD-0334 is implemented, how `alt_bn128_pairing` will be changed? `target_os = "solana"` part can be easily changed, but `not(solana)` can't be changed because it is used in the validator. Or when updating the dependency in the validator, new `versioned_*` functions will be used? Maybe it is better to deprecate old functions for `not(solana)` and recommend to use `versioned_* ones for everyone?

Yes, as you described, my plan was to update agave code so that the new `versioned_*` functions are used for the syscall. Then we can easily feature gate upcoming changes by using the `Versioned*` enum.

I kept the `not(solana)` function so that downstream projects that wishes to invoke the same function as the validator syscall but off-chain, maybe on the client side.

> 2. Naming. All functions have `alt_bn128` prefix. Maybe this prefix should also be added to the `versioned_*` functions? Also there is SIMD-0302 that suggests adding g2 addition and multiplication. Maybe add `g1` to the current `versioned_*` functions? So the names will be `alt_bn128_versioned_g1_addition`, `alt_bn128_versioned_g1_multiplication`, `alt_bn128_versioned_pairing`

These are good ideas. When I was refactoring, I was hoping that the submodule would already scope the functions appropriately, but I ended up re-exporting all the functions as is. I'll add the prefix and also specify the group in the function names.

LStan:
> Yes, as you described, my plan was to update agave code so that the new `versioned_*` functions are used for the syscall. Then we can easily feature gate upcoming changes by using the `Versioned*` enum.
> 
> I kept the `not(solana)` function so that downstream projects that wishes to invoke the same function as the validator syscall but off-chain, maybe on the client side.

But will `alt_bn128_pairing` for `not(solana)` be changed to use `VersionedPairing::V1` when implementing SIMD-0334? Because if yes, then if somebody updates version to 3.1 in the validator without changing to `versioned_` functions, it will break. So this probably should happen in version 4. Or maybe I'm just overcomplicating.

samkim-crypto:
> > Yes, as you described, my plan was to update agave code so that the new `versioned_*` functions are used for the syscall. Then we can easily feature gate upcoming changes by using the `Versioned*` enum.
> > I kept the `not(solana)` function so that downstream projects that wishes to invoke the same function as the validator syscall but off-chain, maybe on the client side.
> 
> But will `alt_bn128_pairing` for `not(solana)` be changed to use `VersionedPairing::V1` when implementing SIMD-0334? Because if yes, then if somebody updates version to 3.1 in the validator without changing to `versioned_` functions, it will break. So this probably should happen in version 4. Or maybe I'm just overcomplicating.

Yeah I think this is fine. As soon as the changes are in v3.1, I can create a PR on the agave side to update the syscall to use the versioned functions.

PR #298, opened 2025-08-27
vote-interface: enable parser on not(solana)

2501babe:
* use the custom `VoteState` parser for all targets
* add `is_uninitialized()` to `VoteState` structs, for agave code that skips `VoteStateVersions`
* fix a bug where `(0, Pubkey::default())` was added to `AuthorizedVoters`


buffalojoec (approved):
Lgtm

PR #300, opened 2025-09-02
[bls-signatures] Add batch signature verification functions on different messages

samkim-crypto:
#### Problem
Currently, there is no public function to a batch of BLS signatures on different messages.

#### Summary of Changes
I added `batch_verify` and `par_batch_verify` functions to verify signatures on different messages.

After I added these functions, I realized that there are many different functions to verify different combination of signatures on messages, which could be quite confusing for developers. I ended up just removing the existing `par_verify_batch` and `par_verify_batch_binary_search` functions.

Now, the exposed API is quite simple. The `aggregate_verify` and `par_aggregate_verify` functions verify BLS signatures on the same message. The `batch_verify` and `par_batch_verify` functions verify BLS signatures on different messages.

### DIFF
--- a/bls-signatures/src/signature.rs
+++ b/bls-signatures/src/signature.rs
@@ -1,16 +1,20 @@
+#[cfg(all(not(target_os = "solana"), feature = "std"))]
+use crate::pubkey::NEG_G1_GENERATOR_AFFINE;
 #[cfg(feature = "bytemuck")]
 use bytemuck::{Pod, PodInOption, Zeroable, ZeroableInOption};
-#[cfg(all(feature = "parallel", not(target_os = "solana")))]
-use {crate::pubkey::Pubkey, alloc::vec::Vec, core::num::NonZeroUsize, rayon::prelude::*};
 #[cfg(not(target_os = "solana"))]
 use {
     crate::{
         error::BlsError,
-        pubkey::{AsPubkeyProjective, PubkeyProjective, VerifiablePubkey},
+        hash::hash_message_to_point,
+        pubkey::{AsPubkeyProjective, Pubkey, PubkeyProjective, VerifiablePubkey},
     },
-    blstrs::{G2Affine, G2Projective},
-    group::Group,
+    blstrs::{Bls12, G1Affine, G2Affine, G2Prepared, G2Projective, Gt},
+    group::{Curve, Group},
+    pairing::{MillerLoopResult, MultiMillerLoop},
 };
+#[cfg(all(feature = "parallel", not(target_os = "solana")))]
+use {alloc::vec::Vec, rayon::prelude::*};
 use {
     base64::{prelude::BASE64_STANDARD, Engine},
     core::fmt,
@@ -32,22 +36,6 @@ pub const BLS_SIGNATURE_AFFINE_SIZE: usize = 192;
 
 /// Size of a BLS signature in an affine point representation in base64
 pub const BLS_SIGNATURE_AFFINE_BASE64_SIZE: usize = 256;
-
-/// Options for configuring advanced verification functions
-#[cfg(all(feature = "parallel", not(target_os = "solana")))]
-pub struct VerificationOptions {
-    /// The number of signatures below which aggregation will be performed sequentially.
-    pub aggregation_threshold: NonZeroUsize,
-}
-
-#[cfg(all(feature = "parallel", not(target_os = "solana")))]
-impl Default for VerificationOptions {
-    fn default() -> Self {
-        Self {
-            aggregation_threshold: NonZeroUsize::new(1).unwrap(),
-        }
-    }
-}
 
 /// A trait for types that can be converted into a `SignatureProjective`.
 #[cfg(not(target_os = "solana"))]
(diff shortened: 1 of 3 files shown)


bls-signatures/src/signature.rs
```diff
@@ -127,6 +117,75 @@ impl SignatureProjective {
...
+        }
+        if public_keys.is_empty() {
+            return Err(BlsError::EmptyAggregation);
+        }
+
+        // TODO: remove `Vec` allocation if possible for efficiency
+        let mut pubkeys_affine = alloc::vec::Vec::with_capacity(public_keys.len());
+        for pubkey in public_keys {
+            let maybe_g1_affine: Option<_> = G1Affine::from_uncompressed(&pubkey.0).into();
+            let g1_affine: G1Affine = maybe_g1_affine.ok_or(BlsError::PointConversion)?;
```
wen-coding, line 155: Since we only have 2000 validators, we could cache this conversion outside the function (e.g. in BLSSigverifier) so you don't have to do it every time?

I'm writing the VAT SIMD, we should do the whole BLS pubkey uncompress and verification during bls_pubkey_to_rank_map generation maybe, so we can kick out all the invalid BLS pubkeys which couldn't be uncompressed correctly later. Which function should I use?

bls-signatures/src/signature.rs
```diff
@@ -127,6 +117,75 @@ impl SignatureProjective {
...
+            pubkeys_affine.push(g1_affine);
+        }
+
+        let mut prepared_hashes = alloc::vec::Vec::with_capacity(messages.len());
+        for message in messages {
+            let hashed_message: G2Affine = hash_message_to_point(message).into();
+            prepared_hashes.push(G2Prepared::from(hashed_message));
```
wen-coding, line 162: I think we talked about one possibility is to aggregate all votes of same message inside BLSSigverifier and then do batch verification on that.

If we implement that, do we still need to pass in an array of messages here?

samkim-crypto: Yes, this is a little confusing, but signatures can still be partially batch verified even if the messages to verify are different. The `batch_verify` function that I added precisely achieves this. When the messages are all the same, then there is the `aggregate_verify` function.

For sigverify, my understanding is that we need to verify signatures on multiple messages because each validator can vote to `Notarize`, `NotarizeFallback`, `Skip`, etc. and all these different kinds of votes will need to be verified in batches. This is why I added the `aggregate_verify` function.

bls-signatures/src/signature.rs
samkim-crypto: The lines
```
let maybe_g1_affine: Option<_> = G1Affine::from_uncompressed(&pubkey.0).into();
let g1_affine: G1Affine = maybe_g1_affine.ok_or(BlsError::PointConversion)?;
```
is actually not decompression, but deserialization. The public key type is defined as simple wrappers around byte arrays `struct Pubkey = Pubkey([u8; 96])` and the core verification function (`multi_miller_loop`) requires us to have a concrete `blstrs::G2Affine` type; so the lines above is just deserializing the byte array to a `G2Affine` type. I think this should be pretty cheap, but we might ultimately want to define another intermediate type that serves as a wrapper of `G2Affine` type directly so that we save on this deserialization.

> I'm writing the VAT SIMD, we should do the whole BLS pubkey uncompress and verification during bls_pubkey_to_rank_map generation maybe, so we can kick out all the invalid BLS pubkeys which couldn't be uncompressed correctly later. Which function should I use?

There is no wrapper type around `G2Affine`, so there is no function exposed to do the pre-deserialization. Can we just use the `Pubkey` type for now? Later, when we really (micro-)optimize sigverify, then we can consider adding an intermediate type that is already deserialized.

bls-signatures/src/signature.rs
wen-coding: Those messages normally don't come at the same time though.

1. most common case: all you get is Notarize, maybe a few Skip from far away slow validators. Then when 60% Notarize land, everyone votes Finalize
3. a bit common case: when the leader is slow or dead, all you get is Skip, maybe a few Notarize from validators close to slow leader

rare case 1: For various reasons (e.g. 50% skip and 50% Notarize), the first round didn't decide, and we go the second round and the people who skipped previously vote NotarizeFallback
rare case 2: For various reasons, the first round didn't decide, we go to the second round and the people who Notarized previously vote SkipFallback
etc etc

So if there is a performance benefit by aggregating similar votes together, we probably should do that. Of course, that depends on how much performance we would gain.

samkim-crypto: Yes, in theory, we would gain quite a lot from restricting batch verification for the same message compared to the different message case. In fact, if we just restrict to signatures on at most two different messages, then this will probably be almost as fast as verifying signatures on a single message. Are there chances that malicious validators submit improper messages? I suppose these validators will be penalized, so I guess we don't need to worry too much.

What I propose is to go with the `batch_verify` function in this PR to complete the initial version of BLS sigverify. I will then add some additional statistics (e.g. average batch size, how frequent do we receive the rare cases outlined above, etc) and also run some benches. With this data, I will make final optimizations in this crate for the release version.

wen-coding: The malicious validators can submit whatever they want, even for ancient slots or future slots. We should rate limit those anomalies, process normal popular traffic first. I'm fine with putting in an initial version. But maybe we should keep the version verifying a batch of messages with same votes, only different pubkeys and signatures?

samkim-crypto: For the case of verifying signatures on the same message, there is `aggregate_verify` and `par_aggregate_verify` to do this. These two functions verify a batch of signatures on the same message and returns either `true` or `false`.

I originally had `par_verify_batch` and `par_verify_batch_binary_search` that were wrappers around `par_aggregate_verify`: if `par_aggregate_verify` fails, then it would individually verify the signatures to identify the invalid signatures.

After experimenting with sigverify in https://github.com/anza-xyz/alpenglow/pull/430 and looking at the current exposed public functions in `solana-bls-signatures`, I thought it would be best to actually clean up `par_verify_batch` and `par_verify_batch_binary_search` from the crate and insert this logic inside BLS sigverify.

The BLS sigverify will optimistically verify the signatures by either calling `par_aggregate_verify` or (the new) `par_verify_batch` and then if this fails, it will (in parallel) verify the batch of signatures to identify the wrong signature.

wen-coding: That sounds good, but do you think we should rename both functions so it's clearer one is for verifying batches with same message and the other is for verifying batches with mixed messages?

samkim-crypto: Okay yep. I renamed `aggregate_verify` to `verify_aggregate` (this is the standard name in cryptography for verifying signatures on a single message) and `batch_verify` to `verify_distinct` (for verifying signatures on different messages). I think this should be more clear than distinguishing by `aggregate` vs `batch`.

PR #308, opened 2025-09-06
vote-interface: vote-state-versions: make converters private

buffalojoec:
Sets the stage for the removal of these converters entirely by removing them from the public interface.

Conversion to v3 is still used in the deserialization for `VoteStateV3`, since the API expects a buffer containing a serialized `VoteStateVersions`. We'll have to do a proper refactor of all the deserialization in order to get rid of these completely. Furthermore, much of the tests in `vote_state_versions` focus on converting to v3 (and v4).

We can make the converters private now before releasing a new vote-interface, and then we can port all of the conversion logic over to the Vote program without breakage.

### DIFF
--- a/vote-interface/src/state/vote_state_versions.rs
+++ b/vote-interface/src/state/vote_state_versions.rs
@@ -4,7 +4,7 @@ use crate::state::{
 };
 #[cfg(test)]
 use arbitrary::{Arbitrary, Unstructured};
-#[cfg(any(test, target_os = "solana", feature = "bincode"))]
+#[cfg(any(test, all(not(target_os = "solana"), feature = "bincode")))]
 use {
     crate::{
         authorized_voters::AuthorizedVoters,
--- a/vote-interface/src/state/vote_state_versions.rs
+++ b/vote-interface/src/state/vote_state_versions.rs
@@ -40,7 +40,7 @@ impl VoteStateVersions {
     ///
     /// NOTE: Does not support conversion from `V4`. Attempting to convert from
     /// v4 to v3 will throw an error.
-    #[cfg(any(test, target_os = "solana", feature = "bincode"))]
+    #[cfg(all(not(target_os = "solana"), feature = "bincode"))]
     pub(crate) fn try_convert_to_v3(self) -> Result<VoteStateV3, InstructionError> {
         match self {
             VoteStateVersions::V0_23_5(state) => {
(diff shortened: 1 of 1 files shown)


vote-interface/src/state/vote_state_versions.rs
```diff
@@ -38,7 +40,8 @@ impl VoteStateVersions {
     ///
     /// NOTE: Does not support conversion from `V4`. Attempting to convert from
     /// v4 to v3 will throw an error.
-    pub fn try_convert_to_v3(self) -> Result<VoteStateV3, InstructionError> {
+    #[cfg(any(test, target_os = "solana", feature = "bincode"))]
```
jstarry: What's the value in the `cfg` modifiers if these methods aren't visible outside the crate anyways. Is it to prune the source tree?

I think they're slightly wrong too.. we only need the conversion methods in this crate for `and(not(target_os = "solana"), feature = "bincode"))

buffalojoec: You get a bunch of clippy/hack "unused" errors if you don't put them behind the `cfg`.

You're right, though. Looks like we just need `#[cfg(all(not(target_os = "solana"), feature = "bincode"))]`.

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -40,7 +40,7 @@ impl VoteStateVersions {
     ///
     /// NOTE: Does not support conversion from `V4`. Attempting to convert from
     /// v4 to v3 will throw an error.
-    #[cfg(any(test, target_os = "solana", feature = "bincode"))]
+    #[cfg(all(not(target_os = "solana"), feature = "bincode"))]
```
jstarry: This assumes tests are either run with the bincode feature or not on the solana target which is probably true but probably better to keep `test` in there?

```
    #[cfg(any(test, all(not(target_os = "solana"), feature = "bincode")))]
```

buffalojoec: Idk it seemed to pass the powerset check and tests no problem. To be safe, I can add it back.

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -1,12 +1,14 @@
+use crate::state::{
+    vote_state_0_23_5::VoteState0_23_5, vote_state_1_14_11::VoteState1_14_11, VoteStateV3,
+    VoteStateV4,
+};
 #[cfg(test)]
 use arbitrary::{Arbitrary, Unstructured};
+#[cfg(any(test, target_os = "solana", feature = "bincode"))]
```
jstarry: forgot this one

jstarry (approved):

PR #309, opened 2025-09-09
ci: publish script: make dependent version a configurable param

buffalojoec:
#### Problem
When publishing the vote-interface, I ran into a hiccup caused by the `cargo-release` argument `--dependent-version`.

Currently, the script hard-codes the argument as `--dependent-version fix`, where `fix` will only update the workspace manifest version of your dependency if it breaks semver.

However, in a situation where you may depend on something additive, such as #280, you might want to publish your lower dependency with a minor or patch but still update the workspace manifest. This functionality is available through the `upgrade` argument instead.

#### Summary of Changes
Add `dependent_version` as a configurable parameter, which defaults to `fix`, in the manually-triggered publish job.

### DIFF
--- a/.github/workflows/publish-rust.yml
+++ b/.github/workflows/publish-rust.yml
@@ -32,7 +32,10 @@ on:
         type: boolean
         default: true
       dependent_version:
-        description: How workspace dependencies should be handled
+        description: |
+          How workspace dependencies should be handled.
+          - "fix": (Default) Only bumps the workspace for semver-breakage - prefer this option
+          - "upgrade": Bumps workspace version regardless - only use if another SDK crate requires new code
         required: true
         default: fix
         type: choice
--- a/.github/workflows/publish-rust.yml
+++ b/.github/workflows/publish-rust.yml
@@ -31,6 +31,14 @@ on:
         required: true
         type: boolean
         default: true
+      dependent_version:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-rust.yml
```diff
@@ -31,6 +31,14 @@ on:
         required: true
         type: boolean
         default: true
+      dependent_version:
+        description: How workspace dependencies should be handled
```
joncinque: To avoid having people misuse this, can we make this description a bit clearer? It could be something like:

```
How workspace dependencies should be handled. Prefer "fix" whenever possible. "upgrade" bumps the workspace version too, so only use it if another sdk crate requires new code.
```

joncinque (commented):
Looks great! Just a nit on the description to help people, let me know what you think

.github/workflows/publish-rust.yml
buffalojoec: Works for me!

buffalojoec: I'm not 100% sure if the multi-line description works, but let's give it a rip.

joncinque (approved):
Beautiful, thanks! And GitHub should tell us right away if there's an issue, so let's just go for it

PR #310, opened 2025-09-10
genesis-config: remove unused native processors field

jstarry:
Agave doesn't actually support adding custom builtin programs from the genesis, this functionality has long been broken. Better to remove it completely since adding a custom builtin program would require forking agave anyways.

### DIFF
--- a/genesis-config/src/lib.rs
+++ b/genesis-config/src/lib.rs
@@ -49,7 +49,7 @@ pub const UNUSED_DEFAULT: u64 = 1024;
 #[cfg_attr(
     feature = "frozen-abi",
     derive(AbiExample),
-    frozen_abi(digest = "3tUUJkZiUUGfuNCXbDuDR6KCQYPsh3m3cPw5vVUSt113")
+    frozen_abi(digest = "2Qkibn4g2baTthkR3o6oP6AHPTw1hCXF96kZscPoBqoo")
 )]
 #[cfg_attr(
     feature = "serde",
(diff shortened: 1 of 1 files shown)


genesis-config/src/lib.rs
```diff
@@ -304,7 +279,11 @@ mod tests {
             solana_pubkey::new_rand(),
             AccountSharedData::new(1, 0, &Pubkey::default()),
         );
-        config.add_native_instruction_processor("hi".to_string(), solana_pubkey::new_rand());
+
+        // Unused fields still impact genesis hash
+        config
+            .__unused_builtin_programs
+            .push(("hi".to_string(), solana_pubkey::new_rand()));
```
buffalojoec, line 286: This kind of stinks. So we can't actually eliminate the field entirely?

genesis-config/src/lib.rs
```diff
@@ -62,7 +62,7 @@ pub struct GenesisConfig {
     /// initial accounts
     pub accounts: BTreeMap<Pubkey, Account>,
     /// built-in programs
-    pub native_instruction_processors: Vec<(String, Pubkey)>,
+    pub __unused_builtin_programs: Vec<(String, Pubkey)>,
```
buffalojoec, line 65: Should this still be `pub`?

genesis-config/src/lib.rs
jstarry: Correct

genesis-config/src/lib.rs
jstarry: I figured it didn't hurt, matched the precedent set by this field below

```rust
    /// this field exists only to ensure that the binary layout of GenesisConfig remains compatible
    /// with the Solana v0.23 release line
    pub __backwards_compat_with_v0_23: u64,
```

jstarry:
How do we track upcoming releases for breaking changes? Or should I just go ahead publish the next major while it's top of mind?

PR #311, opened 2025-09-10
fix: Make `VersionedMessage::Legacy` deserializable via `serde_json`

ShoyuVanilla:
Currently, the following code panics on unwrapping an err: `Error("invalid type: map, expected message prefix byte", line: 1, column: 1)`:

```rust
#[test]
fn legacy_msg_deserialize_with_serde_json() {
    let legacy_msg =
       solana_message::VersionedMessage::Legacy(solana_message::legacy::Message::new(&[], None));
    let serialized = serde_json::to_string(&legacy_msg).unwrap();
    let deserialized: solana_message::VersionedMessage = serde_json::from_str(&serialized).unwrap(); // 💥
}
```

This is because when serializing legacy messages the serializer don't insert version prefix explicitly
https://github.com/anza-xyz/solana-sdk/blob/2cffbca31fe0ad886c340006c0e1d0136f62de02/message/src/versions/mod.rs#L185-L195

but when deserializing, the deserializer implicitly reads the first byte of `LegacyMessage` into a version prefix.
https://github.com/anza-xyz/solana-sdk/blob/2cffbca31fe0ad886c340006c0e1d0136f62de02/message/src/versions/mod.rs#L258-L264

In `bincode` deserialization scenario this causes no problem but with `serde_json`, it causes funny error that the very string that serialized from a `VersionedMessage::Legacy` cannot be deserialized to itself, because `serde_json` expects `Int` but it encounters `{`, the first byte of serialized `LegacyMessage` as a `serde_json::Value::Map`.

### DIFF
--- a/message/src/versions/mod.rs
+++ b/message/src/versions/mod.rs
@@ -246,7 +246,9 @@ impl<'de> serde::Deserialize<'de> for VersionedMessage {
     where
         D: Deserializer<'de>,
     {
-        struct MessageVisitor;
+        struct MessageVisitor {
+            is_human_readable: bool,
+        }
 
         impl<'de> Visitor<'de> for MessageVisitor {
             type Value = VersionedMessage;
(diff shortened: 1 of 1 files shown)


message/src/versions/mod.rs
```diff
@@ -259,10 +261,28 @@ impl<'de> serde::Deserialize<'de> for VersionedMessage {
...
-                let prefix: MessagePrefix = seq
-                    .next_element()?
-                    .ok_or_else(|| de::Error::invalid_length(0, &self))?;
+                let prefix = if self.is_human_readable {
+                    #[derive(Deserialize)]
+                    #[serde(untagged)]
+                    enum PrefixOrLegacy {
+                        Prefix(MessagePrefix),
+                        Legacy(LegacyMessage),
+                    }
```
ShoyuVanilla, line 270: We cannot use this or other branching with `deserialize_any` for `bincode` as it doesn't support `deserialize_any`, so apply this iff the given deserializer is human readable.

ShoyuVanilla:
Though I can workaround this issue but it's a bit irritating in my codebase 😅 Could this be reviewed or merged?

ShoyuVanilla:
@jstarry Sorry for the ping! I was wondering if this PR could be reviewed, or if this change isn’t something the project wants to pursue. I’m currently working on a project that performs serialization/deserialization on transactions, and this issue has required me to implement a custom serializer/deserializer 😅

jstarry:
@ShoyuVanilla apologies on the delay, been backed up on code reviews for awhile. I like this fix but I think we should go further and fix the json string representation of versioned messages as well. Does this fix work for you? https://github.com/anza-xyz/solana-sdk/pull/381

ShoyuVanilla:
> @ShoyuVanilla apologies on the delay, been backed up on code reviews for awhile. I like this fix but I think we should go further and fix the json string representation of versioned messages as well. Does this fix work for you? #381

Yeah, that works fine. Thanks!

ShoyuVanilla:
Closing in favor of #381

PR #312, opened 2025-09-10
Deprecate `InstructionError::NotEnoughAccountKeys`

LucasSte:
**Problem**

`InstructionError::NotEnoughAccountKeys` represent the same error as `InstructionError::MissingAccount`. They are used interchangeably in Agave, and it is becoming difficult to maintain Firedancer compatibility since tracking down where these errors come from is cumbersome.

**Solution**

Deprecate the error.

### DIFF
--- a/instruction-error/src/lib.rs
+++ b/instruction-error/src/lib.rs
@@ -373,7 +373,8 @@ where
             MISSING_REQUIRED_SIGNATURES => Self::MissingRequiredSignature,
             ACCOUNT_ALREADY_INITIALIZED => Self::AccountAlreadyInitialized,
             UNINITIALIZED_ACCOUNT => Self::UninitializedAccount,
-            NOT_ENOUGH_ACCOUNT_KEYS => Self::MissingAccount,
+            #[allow(deprecated)]
+            NOT_ENOUGH_ACCOUNT_KEYS => Self::NotEnoughAccountKeys,
             ACCOUNT_BORROW_FAILED => Self::AccountBorrowFailed,
             MAX_SEED_LENGTH_EXCEEDED => Self::MaxSeedLengthExceeded,
             INVALID_SEEDS => Self::InvalidSeeds,


instruction-error/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![no_std]
+#![allow(deprecated)]
```
LucasSte: This is not ideal, but (I guess) because of the derive attribute above the enum, Rust complains we are using the deprecated type. That is possibly happening in serde.

LucasSte:
I merged https://github.com/anza-xyz/agave/pull/8001 in the Agave repository, and I think this PR here complements that one.

instruction-error/src/lib.rs
```diff
@@ -363,7 +366,7 @@ where
             MISSING_REQUIRED_SIGNATURES => Self::MissingRequiredSignature,
             ACCOUNT_ALREADY_INITIALIZED => Self::AccountAlreadyInitialized,
             UNINITIALIZED_ACCOUNT => Self::UninitializedAccount,
-            NOT_ENOUGH_ACCOUNT_KEYS => Self::NotEnoughAccountKeys,
+            NOT_ENOUGH_ACCOUNT_KEYS => Self::MissingAccount,
```
joncinque: This change means that operations will give a different response.

For example, if we're fetching an old transaction, before we'll receive `NotEnoughAccountKeys`, and with this change, we'll get `MissingAccount`. Are we sure that behavior is ok?

On the other hand, we misuse this int conversion in runtime code, so maybe it isn't such a big deal.

instruction-error/src/lib.rs
joncinque: If you want to get around it, you might be able to declare the enum in a nested mod that has `#[allow(deprecated)]`, and then re-export the enum from here. What do you think?

joncinque (commented):
Just a fly-by

instruction-error/src/lib.rs
LucasSte: Changing something old isn't good, right. I did not want, thought, to explicitly create `NotEnoughAccountKeys` anymore. Do you have any suggestion for that?

It is worth mentioning that `MissingAccount` and `NotEnoughAccountKeys` are very similar in usage and in meaning. Their descriptions are also so close that one cannot probably distinguish them.

https://github.com/anza-xyz/solana-sdk/blob/ca06d99048ea4c2b7a5ae0ef4427486be6b95e5f/instruction-error/src/lib.rs#L94-L95

https://github.com/anza-xyz/solana-sdk/blob/ca06d99048ea4c2b7a5ae0ef4427486be6b95e5f/instruction-error/src/lib.rs#L138-L139

instruction-error/src/lib.rs
LucasSte: Fixed in https://github.com/anza-xyz/solana-sdk/pull/312/commits/c555b5d6acb2dae715463f7e460dda88b525dc24

instruction-error/src/lib.rs
joncinque: If `NotEnoughAccountKeys` is stored as a transaction result in Bigtable, and someone fetches it, they should probably always get `NotEnoughAccountKeys`, so I don't think we can completely avoid creating this type, and that's OK.

You did the most important part in https://github.com/anza-xyz/agave/pull/8001 to stop creating it going forward.

LucasSte: On the other way around, there is no translation from the u64 error code to the correct enum variant for `MissingAccount`. It is returning `Custom`.

Can I create the proper conversion here for `MissingAccount`?

joncinque: It can't hurt, go for it!

LucasSte: I was going to implement that, but I noticed this conversion is from the ProgramError code to the InstructionError enum. Is it supposed to be like that?

We don't have a `MissingAccount` variant for ProgramError, and I don't think one is necessary.

LucasSte: `NotEnoughAccountKeys` is defined here as 11:

https://github.com/anza-xyz/solana-sdk/blob/b6d7418a0d66cf5a4117a58f23e3d8e0b16e8a95/program-error/src/lib.rs#L32

and `ProgramError::NotEnoughAccountKeys` is the eleventh variant of `ProgramError`, but it is not the eleventh variant of `InstructionError`.

joncinque: I think these error conversions are easy to do incorrectly, but yeah the idea is to convert from a number to an instruction error, and the number should be the correct program error.

In this case, it's ok that `NotEnoughAccountKeys` is not the eleventh instruction error, since the `NOT_ENOUGH_ACCOUNT_KEYS` number is converted to `InstructionError::NotEnoughAccountKeys` explicitly.

LucasSte: Given that we are deprecating `InstructionError::NotEnoughAccountKeys`, would it make sense to add a `MissingAccount` variant to `ProgramError` and perform the conversions correctly then?

I'm not sure this is necessary.

joncinque: It depends on how important the deprecation is. If you're ok with `InstructionError::NotEnoughAccountKeys` being created by `ProgramError::NotEnoughAccountKeys`, then we don't need to do anything else.

But if that's a problem, then we need to add `ProgramError::MissingAccount` and also deprecate `ProgramError::NotEnoughAccountKeys`.

Correct me if I'm wrong, but I don't think it's a problem, so this PR looks good to me as it is currently

LucasSte: The goal was to remove the emission of `InstructionError::NotEnoughAccountKeys` in the monorepo and add a deprecation warning if somebody tries to use it.

Having said that and considering your comment, I believe this PR is ready for review and merge.

joncinque (approved):
Looks great to me, thanks!

PR #313, opened 2025-09-10
manifest: bump serialize-utils at workspace level

buffalojoec:
Bumps the version (minor) of `solana-serialize-utils` so we can publish a new `solana-vote-interface` and set the minimum version to 3.1.0.

This step was not automatically done during publish. See #309.


joncinque (approved):
Looks great!

PR #316, opened 2025-09-11
vote-interface: drop all vote-program-specific methods

buffalojoec:
As part of the ongoing work to overhaul the vote state processing in the vote program, we've decided to move all of the functionality that's currently represented in the public API for `VoteStateVersions` and `VoteStateV3` into the vote program itself.

This will be a major breaking change, as it removes all methods from `VoteStateVersions` and `VoteStateV3` that are really only necessary for the Vote program.

I've done each method one at a time, and I've also removed any tests that were only testing that method itself. These tests have been moved to the Vote program.


jstarry:
What are your thoughts on keeping some of the getter methods? I'm wary of breaking downstream users which call some of these read methods directly. Some of them might be useful in ways I can't exactly think of. I think keeping the setter methods would cause more harm than good so I'm in favor of removing those.

jstarry:
> What are your thoughts on keeping some of the getter methods? I'm wary of breaking downstream users which call some of these read methods directly. Some of them might be useful in ways I can't exactly think of. I think keeping the setter methods would cause more harm than good so I'm in favor of removing those.

As you pointed out, many of the getters can be replaced by directly accessing the already public fields. I'd be in favor of removing all methods you've removed except for the following since I think they're simple and convenient and their logic is not super intuitive to end users to implement
- `is_uninitialized`
- `get_rent_exempt_reserve`
- `current_epoch`
- `credits`

buffalojoec:
> > What are your thoughts on keeping some of the getter methods? I'm wary of breaking downstream users which call some of these read methods directly. Some of them might be useful in ways I can't exactly think of. I think keeping the setter methods would cause more harm than good so I'm in favor of removing those.
> 
> As you pointed out, many of the getters can be replaced by directly accessing the already public fields. I'd be in favor of removing all methods you've removed except for the following since I think they're simple and convenient and their logic is not super intuitive to end users to implement
> 
> * `is_uninitialized`
> * `get_rent_exempt_reserve`
> * `current_epoch`
> * `credits`

Sounds good. Sorry, I tried to add clean reverts but it was getting messy, so I just did a rebase and dropped those commits where certain methods were removed. The four you listed should be retained now.

jstarry:
Looks like credits and current epoch aren't retained

buffalojoec:
> Looks like credits and current epoch aren't retained

My bad, now we're good.

jstarry (approved):

PR #320, opened 2025-09-15
Revert "genesis-config: remove unused native processors field"

jstarry:
Unfortunately can't remove this field because the mainnet genesis config has this field set:

```
[genesis-config/src/lib.rs:313:9] genesis_config.native_instruction_processors = [
    (
        "solana_config_program",
        Config1111111111111111111111111111111111111,
    ),
    (
        "solana_stake_program",
        Stake11111111111111111111111111111111111111,
    ),
    (
        "solana_system_program",
        11111111111111111111111111111111,
    ),
    (
        "solana_vote_program",
        Vote111111111111111111111111111111111111111,
    ),
]
```

### DIFF
--- a/genesis-config/src/lib.rs
+++ b/genesis-config/src/lib.rs
@@ -49,7 +49,7 @@ pub const UNUSED_DEFAULT: u64 = 1024;
 #[cfg_attr(
     feature = "frozen-abi",
     derive(AbiExample),
-    frozen_abi(digest = "2Qkibn4g2baTthkR3o6oP6AHPTw1hCXF96kZscPoBqoo")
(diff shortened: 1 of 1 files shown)


buffalojoec (approved):
Ah crud

jstarry:
And just to be clear, the change to rename the field was still compatible with `genesis.bin`'s used for our live clusters but I didn't want the new name of "unused" to imply that it could be reused for something else (like I was intending to do). And it should be possible in the SDK to recreate the `genesis.bin` used for live clusters even if those fields don't have any real effect on the cluster now.

PR #322, opened 2025-09-15
chore: refactor constants in bn254

LStan:
Currently, it is unclear how the values of some of the constants were calculated. In this PR, I tried to make the calculation more obvious.
I also renamed some constants to make them more standardized (`ALT_BN128_POINT_SIZE` -> `ALT_BN128_G1_POINT_SIZE`, `Gx_COMPRESSED` -> `ALT_BN128_Gx_COMPRESSED_SIZE`).

### DIFF
--- a/bn254/src/addition.rs
+++ b/bn254/src/addition.rs
@@ -10,9 +10,9 @@ use {
     ark_serialize::{CanonicalSerialize, Compress},
 };
 
-/// Input length for the add operation.
+/// Input size for the add operation.
 pub const ALT_BN128_ADDITION_INPUT_SIZE: usize = ALT_BN128_G1_POINT_SIZE * 2; // 128
-/// Output length for the add operation.
+/// Output size for the add operation.
 pub const ALT_BN128_ADDITION_OUTPUT_SIZE: usize = ALT_BN128_G1_POINT_SIZE; // 64
 
 #[deprecated(
--- a/bn254/src/multiplication.rs
+++ b/bn254/src/multiplication.rs
@@ -15,10 +15,10 @@ use {
     ark_serialize::{CanonicalDeserialize, CanonicalSerialize, Compress},
 };
 
-/// Input length for the multiplication operation.
+/// Input size for the multiplication operation.
 pub const ALT_BN128_MULTIPLICATION_INPUT_SIZE: usize =
     ALT_BN128_G1_POINT_SIZE + ALT_BN128_FIELD_SIZE; // 96
-/// Output length for the multiplication operation.
+/// Output size for the multiplication operation.
 pub const ALT_BN128_MULTIPLICATION_OUTPUT_SIZE: usize = ALT_BN128_G1_POINT_SIZE; // 64
 
 #[deprecated(
--- a/bn254/src/pairing.rs
+++ b/bn254/src/pairing.rs
@@ -16,9 +16,9 @@ use {
     ark_ff::{BigInteger, BigInteger256, One},
 };
 
-/// Pair element length.
+/// Pair element size.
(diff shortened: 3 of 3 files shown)


LStan:
Also, some constants have suffix `_SIZE`, and some `_LEN`. Should I do something with it?
@samkim-crypto

LStan:
I changed the suffix.

I also want to make two changes in another PRs:
1. Add `_be` suffix to BE variants of functions.
2. Change input for `target(solana)` for LE addition and multiplication from slice to a sized array (like this `input: &[u8; ALT_BN128_ADDITION_INPUT_SIZE]`). This can save some CUs. But for that I will have to split `target(solana)` and `not(solana)` to two functions.

The second change should also be done for compress/decompress functions. Currently, only `alt_bn128_g2_decompress` has a sized array as input, while the others use a slice. However, this is a breaking change.

What do you think?

samkim-crypto:
> Add _be suffix to BE variants of functions.

Yes, I was originally thinking about making this change myself. I am all in favor of this to make the function names consistent.

> Change input for target(solana) for LE addition and multiplication from slice to a sized array (like this input: &[u8; ALT_BN128_ADDITION_INPUT_SIZE]). This can save some CUs. But for that I will have to split target(solana) and not(solana) to two functions.

I am generally in favor of this, but is there a reason to keep the target(solana) and not(solana) functions to have two separate function interfaces? Updating from slice inputs to sized array inputs will be a bit more restrictive in terms of function flexibility, but the point of having the not(solana) function is to have an off-chain function to mirror the on-chain target(solana) function. What do you think about unifying the current `alt_bn128_g2_decompress` functions into a single function that take in a reference to a fixed size array instead?

LStan:
> I am generally in favor of this, but is there a reason to keep the target(solana) and not(solana) functions to have two separate function interfaces? Updating from slice inputs to sized array inputs will be a bit more restrictive in terms of function flexibility, but the point of having the not(solana) function is to have an off-chain function to mirror the on-chain target(solana) function.

I completely forgot that in the validator `versioned_` functions will be used. So, no reason.


> What do you think about unifying the current `alt_bn128_g2_decompress` functions into a single function that take in a reference to a fixed size array instead?
Not sure, I understand you. Do you suggest putting only `alt_bn128_g2_decompress` in a one functions with a sized input for both 
target(solana) and not(solana)? How it will be called from the validator which has &[u8] as input variable?  What about other compress/decompress functions?

samkim-crypto:
>> What do you think about unifying the current alt_bn128_g2_decompress functions into a single function that take in a reference to a fixed size array instead?

> Not sure, I understand you. Do you suggest putting only alt_bn128_g2_decompress in a one functions with a sized input for both
target(solana) and not(solana)? How it will be called from the validator which has &[u8] as input variable? What about other compress/decompress functions?

Yeah sorry I misspoke here. I forgot that we don't have validator `version_` functions for the decompression functions.

There is going to be a little bit of inconsistency in the crate, but what do you think about having the split function syntax as you suggested (slice for non-solana target and fixed array for solana target) just for the compression/decompression submodule, and then use the using unified functions (sized array input for both targets) in the rest of the crate?

An alternative would be to add `version_` functions for the compression/decompression as well though we don't have any necessary changes lined up for these functions (if I remember correctly).

LStan:
> There is going to be a little bit of inconsistency in the crate, but what do you think about having the split function syntax as you suggested (slice for non-solana target and fixed array for solana target) just for the compression/decompression submodule, and then use the using unified functions (sized array input for both targets) in the rest of the crate?

This is exactly what I suggested. 
My point was that changing input to a sized array for LE add and mult is not a breaking change, because it has not yet published. But changing input in compress/decompress is a breaking change.

PR #324, opened 2025-09-16
SIMD-0334: Fix alt_bn128_pairing syscall length check

LStan:
Implemented SDK changes for [SIMD-0334](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0334-fix-alt-bn128-pairing-length-check.md)

### DIFF
--- a/bn254/src/pairing.rs
+++ b/bn254/src/pairing.rs
@@ -121,7 +121,7 @@ pub fn alt_bn128_pairing(input: &[u8]) -> Result<Vec<u8>, AltBn128Error> {
     }
     #[cfg(target_os = "solana")]
     {
-        if !input.len().is_multiple_of(ALT_BN128_PAIRING_ELEMENT_SIZE) {
+        if input.len() % ALT_BN128_PAIRING_ELEMENT_SIZE != 0 {
             return Err(AltBn128Error::InvalidInputData);
         }
         let mut result_buffer = [0u8; 32];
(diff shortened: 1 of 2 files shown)


LStan:
Clippy gives error only for `not(solana)`, this lint was added only in rust 1.89.0. Should I fix it for both targets or only for `not(solana)`?

LStan:
I fixed everywhere. It looks like clippy does not see code inside `target_os = "solana"`.

bn254/src/pairing.rs
```diff
@@ -106,15 +117,11 @@ pub fn alt_bn128_versioned_pairing(
...
+        alt_bn128_versioned_pairing(VersionedPairing::V1, input, Endianness::BE)
     }
     #[cfg(target_os = "solana")]
     {
-        if input
-            .len()
-            .checked_rem(ALT_BN128_PAIRING_ELEMENT_SIZE)
-            .is_none()
-        {
+        if input.len() % ALT_BN128_PAIRING_ELEMENT_SIZE != 0 {
```
samkim-crypto, line 124: Let's change this to the clippy style as was done in `alt_bn128_versioned_pairing`.

bn254/src/pairing.rs
```diff
@@ -138,15 +145,11 @@ pub fn alt_bn128_pairing(input: &[u8]) -> Result<Vec<u8>, AltBn128Error> {
...
+        alt_bn128_versioned_pairing(VersionedPairing::V1, input, Endianness::LE)
     }
     #[cfg(target_os = "solana")]
     {
-        if input
-            .len()
-            .checked_rem(ALT_BN128_PAIRING_ELEMENT_SIZE)
-            .is_none()
-        {
+        if input.len() % ALT_BN128_PAIRING_ELEMENT_SIZE != 0 {
```
samkim-crypto, line 152: Same here.

samkim-crypto (commented):
Yep, all the changes look good to me. Let's just make the length check style more consistent for the Solana target and then this should be good.

bn254/src/pairing.rs
LStan: This is what I did at first, but CI failed because `is_multiple_of` is unstable feature in sbf compiler.

samkim-crypto: Oh I see. Sorry, I was also wondering why you reverted there. I think the changes are good as is then.

samkim-crypto (approved):

PR #325, opened 2025-09-16
address!: Add conditional debug derive

febo:
### Problem

The `Address` type only includes a `Debug` impl when the feature "decode" is enabled, since it includes a base58 formatting. There are cases where it would be useful to have a "default" `Debug` derive without additional dependencies.

### Solution

Add a `Debug` derive for cases where the feature "decode" is not enabled.


joncinque:
I was just thinking about this in the context of #302, and this seems like the easiest path forward.

We'll get some confusion if users are outputting pubkeys without the `std` feature on solana-pubkey, so we should document the change in behavior in the other PR.

What do you think @vadorovsky?

joncinque (approved):
Looks great!

PR #329, opened 2025-09-17
address!: Use uninit memory

febo:
### Problem

When finding or creating program addresses, memory is zeroed before the syscall. This is wasteful since the memory will be overwritten by the syscall.

### Solution

Use uninitialized memory (`MaybeUninit`) instead. This PR also adds `inline` hints to the compiler.


rustopian (approved):
Straightforward, looks good

PR #330, opened 2025-09-18
address: Add `declare_id` / `declare_deprecated_id` from pubkey

joncinque:
#### Problem

We should start using solana-address instead of solana-pubkey in most places, but solana-address doesn't have `declare_id` or `declare_deprecated_id`.

I noticed this while trying to get solana-pubkey properly no-std compatible.

#### Summary of changes

Move them from solana-pubkey, and re-export from solana-pubkey.

### DIFF
--- a/address/src/lib.rs
+++ b/address/src/lib.rs
@@ -356,6 +356,85 @@ macro_rules! address {
     };
 }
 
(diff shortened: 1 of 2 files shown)


febo (commented):
Looks great – thanks! I will just wait CI to pass and then approve it.

PR #331, opened 2025-09-19
feat: `VoteStateV4` goodies

ksn6:


jstarry:
This is a lossy conversion. Can you explain why this would be needed? I don't think we want this

ksn6:
> This is a lossy conversion. Can you explain why this would be needed? I don't think we want this

https://github.com/anza-xyz/agave/blob/3684e5825abf375e9318a0bf9ccca1e8edfc0a7e/programs/vote/src/vote_state/handler.rs#L393

How would we handle the above function for `set_vote_account_state`? Should we panic if the account isn't large enough to store the vote state?

vote-interface/src/state/vote_instruction_data.rs
```diff
@@ -192,13 +194,21 @@ impl TowerSync {
...
 
+#[cfg_attr(feature = "serde", cfg_eval::cfg_eval, serde_as)]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[derive(Default, Debug, PartialEq, Eq, Clone, Copy)]
 pub struct VoteInit {
     pub node_pubkey: Pubkey,
     pub authorized_voter: Pubkey,
     pub authorized_withdrawer: Pubkey,
     pub commission: u8,
+
```
ksn6, line 205: @jstarry @buffalojoec - should we add in fields for the inflation bps and block rewards bps commissions?

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -71,6 +70,21 @@ pub struct VoteStateV4 {
 }
 
 impl VoteStateV4 {
+    pub fn new(vote_init: &VoteInit, clock: &Clock) -> Self {
+        let inflation_rewards_commission_bps = (vote_init.commission as u16)
```
ksn6, line 74: @jstarry @buffalojoec - leaving this as a placeholder, though we probably don't want to do this. My guess is that we want `VoteInit` to have both commission bps fields.

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -214,4 +228,13 @@ impl VoteStateV4 {
             ..Self::default()
         }
     }
+
+    pub fn commission(&self) -> Result<u8, InstructionError> {
+        let commission_bps = self
```
ksn6, line 233: (same here)

jstarry:
We can close this in favor of https://github.com/anza-xyz/agave/pull/8108 / https://github.com/anza-xyz/agave/pull/8120

PR #332, opened 2025-09-20
[bn254] Add _be suffix to BE functions

LStan:
@samkim-crypto I also added `g1` to addition and multiplication as a preparation to SIMD-0302.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -28,6 +28,9 @@ inputs:
   miri:
     description: Install Miri if `true`. Defaults to `false`.
     required: false
+  rust-src:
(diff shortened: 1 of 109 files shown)


LStan:
@samkim-crypto Could you please review this PR? I want to rebase #334 on it.

samkim-crypto (commented):
Sorry for the delay on this. Everything looks good to me!

Seems like a new CI workflow was added in https://github.com/anza-xyz/solana-sdk/pull/340, so there is some issue with the no_std workflow starting. Do you mind rebasing master? I'll start the CI when you do and hopefully that starts the no_std workflow.

samkim-crypto (approved):

PR #333, opened 2025-09-20
address!: Add derive address helpers

febo:
### Problem

Even when a program stores a bump to derive a program address, it is necessary to use the `create_program_address` syscall to validate the derivation. The base cost of the syscall is `1500` CUs.

In most cases, the program has the correct seeds for the derivation, so it would be sufficient to just perform the derivation and compare it against the expected resulting address.

### Solution

This PR adds a new helper `derive_address` that Instead of using the `create_program_address`, it uses the `sol_sha256` syscall to perform the derivation. This significantly reduces the amount of CUs consumed.

To derive a PDA with seeds `&[program_id, &[255]]`:
| function       | CUs |
|------------| --- |
| `create_program_address` | 🟥 1519 (+1363) |
| (*new*) `derive_address`  |  🟩 156 |

There is also compile-time constant version `derive_address_const`.

> [!NOTE]
> The `derive_address` and `derive_address_const` do not validate if the address is a valid PDA (i.e., off-curve address).

### DIFF
--- a/address/Cargo.toml
+++ b/address/Cargo.toml
@@ -16,6 +16,7 @@ borsh = ["dep:borsh", "std"]
 bytemuck = ["dep:bytemuck", "dep:bytemuck_derive"]
 curve25519 = ["dep:curve25519-dalek", "error", "sha2"]
 decode = ["dep:five8", "dep:five8_const", "error"]
+derive = ["sha2-const-stable"]
 dev-context-only-utils = ["dep:arbitrary", "rand"]
 error = ["dep:solana-program-error"]
 frozen-abi = [
@@ -41,6 +42,7 @@ five8_const = { workspace = true, optional = true }
 rand = { workspace = true, optional = true }
 serde = { workspace = true, optional = true }
 serde_derive = { workspace = true, optional = true }
+sha2-const-stable = { version = "0.1.0", optional = true }
 solana-atomic-u64 = { workspace = true, optional = true }
 solana-frozen-abi = { workspace = true, features = ["frozen-abi"], optional = true }
 solana-frozen-abi-macro = { workspace = true, features = ["frozen-abi"], optional = true }
--- a/address/src/derive.rs
+++ b/address/src/derive.rs
@@ -1,0 +1,142 @@
+use crate::{Address, MAX_SEEDS, PDA_MARKER};
+#[cfg(target_os = "solana")]
+use core::mem::MaybeUninit;
+
+/// Derive a [program address][pda] from the given seeds, optional bump and
+/// program id.
+///
+/// [pda]: https://solana.com/docs/core/pda
+///
+/// In general, the derivation uses an optional bump (byte) value to ensure a
+/// valid PDA (off-curve) is generated. Even when a program stores a bump to
+/// derive a program address, it is necessary to use the
+/// [`pinocchio::pubkey::create_program_address`] to validate the derivation. In
+/// most cases, the program has the correct seeds for the derivation, so it would
+/// be sufficient to just perform the derivation and compare it against the
+/// expected resulting address.
+///
+/// This function avoids the cost of the `create_program_address` syscall
+/// (`1500` compute units) by directly computing the derived address
+/// calculating the hash of the seeds, bump and program id using the
+/// `sol_sha256` syscall.
+///
+/// # Important
+///
+/// This function differs from [`Address::create_program_address`] in that
+/// it does not perform a validation to ensure that the derived address is a valid
+/// (off-curve) program derived address. It is intended for use in cases where the
+/// seeds, bump, and program id are known to be valid, and the caller wants to derive
+/// the address without incurring the cost of the `create_program_address` syscall.
+pub fn derive_address<const N: usize>(
+    seeds: &[&[u8]; N],
+    bump: Option<u8>,
+    program_id: &Address,
+) -> Address {
+    #[cfg(target_os = "solana")]
+    {
+        const {
+            assert!(N < MAX_SEEDS, "number of seeds must be less than MAX_SEEDS");
+        }
+
+        const UNINIT: MaybeUninit<&[u8]> = MaybeUninit::<&[u8]>::uninit();
+        let mut data = [UNINIT; MAX_SEEDS + 2];
+        let mut i = 0;
+
+        while i < N {
+            // SAFETY: `data` is guaranteed to have enough space for `N` seeds,
+            // so `i` will always be within bounds.
+            unsafe {
+                data.get_unchecked_mut(i).write(seeds.get_unchecked(i));
+            }
+            i += 1;
+        }
+
+        // TODO: replace this with `as_slice` when the MSRV is upgraded
+        // to `1.84.0+`.
+        let bump_seed = [bump.unwrap_or_default()];
+
+        // SAFETY: `data` is guaranteed to have enough space for `MAX_SEEDS + 2`
+        // elements, and `MAX_SEEDS` is as large as `N`.
+        unsafe {
+            if bump.is_some() {
+                data.get_unchecked_mut(i).write(&bump_seed);
+                i += 1;
+            }
+            data.get_unchecked_mut(i).write(program_id.as_ref());
+            data.get_unchecked_mut(i + 1).write(PDA_MARKER.as_ref());
+        }
+
+        let mut pda = MaybeUninit::<Address>::uninit();
+
+        // SAFETY: `data` has `i + 2` elements initialized.
+        unsafe {
+            sol_sha256(
+                data.as_ptr() as *const u8,
+                (i + 2) as u64,
+                pda.as_mut_ptr() as *mut u8,
+            );
+        }
+
+        // SAFETY: `pda` has been initialized by the syscall.
+        Address::new_from_array(unsafe { pda.assume_init() })
+    }
+
+    #[cfg(not(target_os = "solana"))]
+    derive_address_const(seeds, bump, program_id)
+}
+
+/// Derive a [program address][pda] from the given seeds, optional bump and
+/// program id.
+///
+/// [pda]: https://solana.com/docs/core/pda
+///
+/// In general, the derivation uses an optional bump (byte) value to ensure a
+/// valid PDA (off-curve) is generated.
+///
+/// This function is intended for use in `const` contexts - i.e., the seeds and
+/// bump are known at compile time and the program id is also a constant. It avoids
+/// the cost of the `create_program_address` syscall (`1500` compute units) by
+/// directly computing the derived address using the SHA-256 hash of the seeds,
+/// bump and program id.
+///
+/// # Important
+///
+/// This function differs from [`Address::create_program_address`] in that
+/// it does not perform a validation to ensure that the derived address is a valid
+/// (off-curve) program derived address. It is intended for use in cases where the
+/// seeds, bump, and program id are known to be valid, and the caller wants to derive
+/// the address without incurring the cost of the `create_program_address` syscall.
+///
+/// This function is a compile-time constant version of [`derive_address`].
+pub const fn derive_address_const<const N: usize>(
+    seeds: &[&[u8]; N],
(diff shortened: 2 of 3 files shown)


address/Cargo.toml
```diff
@@ -16,6 +16,7 @@ borsh = ["dep:borsh", "std"]
 bytemuck = ["dep:bytemuck", "dep:bytemuck_derive"]
 curve25519 = ["dep:curve25519-dalek", "error", "sha2"]
 decode = ["dep:five8", "dep:five8_const", "error"]
+derive = ["dep:sha2-const-stable", "dep:solana-define-syscall"]
```
joncinque: I'm wary about adding another feature for this when we already have the sha2 feature, which only controls the `create_with_seed` function, which also hashes things together to produce an address -- how about moving the new functionality under the `sha2` feature?

If we need to remove the solana-sha256-hasher dependency for that to work properly, that's fine by me 😅

address/src/derive.rs
```diff
@@ -0,0 +1,142 @@
...
+/// This function differs from `Address::create_program_address` in that
+/// it does not perform a validation to ensure that the derived address is a valid
+/// (off-curve) program derived address. It is intended for use in cases where the
+/// seeds, bump, and program id are known to be valid, and the caller wants to derive
+/// the address without incurring the cost of the `create_program_address` syscall.
+pub fn derive_address<const N: usize>(
+    seeds: &[&[u8]; N],
+    bump: Option<u8>,
+    program_id: &Address,
+) -> Address {
```
joncinque: `create_program_derived_address` exists in an `impl Address` block, maybe we do the same here?

address/src/derive.rs
```diff
@@ -0,0 +1,142 @@
...
+/// `sol_sha256` syscall.
+///
+/// # Important
+///
+/// This function differs from `Address::create_program_address` in that
+/// it does not perform a validation to ensure that the derived address is a valid
+/// (off-curve) program derived address. It is intended for use in cases where the
+/// seeds, bump, and program id are known to be valid, and the caller wants to derive
+/// the address without incurring the cost of the `create_program_address` syscall.
+pub fn derive_address<const N: usize>(
```
joncinque: nit: We could rename this to `derive_program_address` to be a little clearer. I also thought about `create_program_address_unchecked`, but that might be less clear since it's not actually using the syscall underneath

address/src/derive.rs
```diff
@@ -0,0 +1,142 @@
...
+    seeds: &[&[u8]; N],
+    bump: Option<u8>,
+    program_id: &Address,
+) -> Address {
+    #[cfg(target_os = "solana")]
+    {
+        const {
+            assert!(N < MAX_SEEDS, "number of seeds must be less than MAX_SEEDS");
```
joncinque, line 40: nit: It's technically possible to pass in an array with size `N == MAX_SEEDS` if the bump is not specified, but that isn't possible to check in a const context.

We can relax the const check to `N <= MAX_SEEDS` and then also add a runtime check that `N + 1 <= MAX_SEEDS` if bump is specified. Or we can create two versions of the function to perform the check in a const context properly.

Or we can do nothing and wait for someone to complain about this in the future, since I don't think many people are using `MAX_SEEDS` 😅

address/src/derive.rs
```diff
@@ -0,0 +1,142 @@
...
+            // so `i` will always be within bounds.
+            unsafe {
+                data.get_unchecked_mut(i).write(seeds.get_unchecked(i));
+            }
+            i += 1;
+        }
+
+        // TODO: replace this with `as_slice` when the MSRV is upgraded
+        // to `1.84.0+`.
+        let bump_seed = [bump.unwrap_or_default()];
```
joncinque: It looks like `as_slice()` is stable in non-const contexts as of 1.75: https://doc.rust-lang.org/std/option/enum.Option.html#method.as_slice, so from what I see, this should be safe to change, no?

address/src/derive.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        // SAFETY: `data` is guaranteed to have enough space for `MAX_SEEDS + 2`
+        // elements, and `MAX_SEEDS` is as large as `N`.
```
joncinque: nit
```suggestion
        // elements, and `MAX_SEEDS` is larger than `N`.
```

address/src/derive.rs
```diff
@@ -0,0 +1,142 @@
...
+pub const fn derive_address_const<const N: usize>(
+    seeds: &[&[u8]; N],
+    bump: Option<u8>,
+    program_id: &Address,
+) -> Address {
+    const {
+        assert!(N < MAX_SEEDS, "number of seeds must be less than MAX_SEEDS");
+    }
+
+    let mut hasher = sha2_const_stable::Sha256::new();
```
joncinque: Doesn't need to be done here, but we could consider reimplementing solana-sha256-hasher using `sha2_const_stable`, marking all its functions const, and then using it here instead.

joncinque (commented):
Looks great overall! Mostly small bits

> Note: It was not possible to use solana-sha256-hasher since the helpers use a MaybeUninit array to handle the optional bump – it would be inefficient to repeat the logic of going through the seeds array twice to build the hasher.

Since the `MaybeUninit` code is all gated under `target_os = "solana"`, you could use `solana_sha256_hasher::hashv(...)` safely, since that'll just use the syscall underneath. But that'll require getting hash and solana-sha256-hasher to be properly no-std, which isn't currently the case.

If you need to land this soon, we can add it in as is, and then update it later once the other parts become properly no-std. Just remember to create an issue for it so we don't forget.

address/src/derive.rs
febo: I was considering saving the name `derive_program_address` for another helper that does the curve validation via the "on-curve" syscall and does not take a bump value – e.g.:
```rust
pub fn derive_program_address<const N: usize>(
    seeds: &[&[u8]; N],
    program_id: &Address,
) -> Address;
```
Most of the time the bump is found in about 3 iterations, so this could be cheaper than using the `find_program_address` syscall.

address/src/derive.rs
febo: Tempted to do nothing at this point. 😀 I think is it unlikely to hit that limit, so waiting for someone to complain sounds good.

address/src/derive.rs
febo: Good catch! We only need `1.84+` for the `const` implementation.

febo:
Since this is not urgent, let's improve `solana-sha256-hasher` as you suggested and then we can use it here. I will put this in draft for now.

address/src/derive.rs
febo: Did a small benchmark and `sha2_const_stable` is slightly slower than `sha2`, so perhaps is not great to change `solana-sha256-hasher`. As an alternative, I made `solana-sha256-hasher` only a dependency for non-programs and called the syscall directly instead.

febo: I had to add `solana-hash` as a dependency, so not sure it will be a lot of difference in the end. 😅

febo:
@joncinque This is ready for another review.

rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.93.0"
+channel = "nightly-2026-01-22"
```
joncinque: Oops, can you revert this?

address/src/derive.rs
```diff
@@ -0,0 +1,160 @@
...
+            // SAFETY: This is sound as `sol_sha256` always fills all 32 bytes of our hash.
+            unsafe {
+                solana_define_syscall::definitions::sol_sha256(
+                    data.as_ptr() as *const _ as *const u8,
+                    (i + 2) as u64,
+                    hash_result.as_mut_ptr() as *mut u8,
+                );
+                Address::from(hash_result.assume_init())
+            }
+        }
```
joncinque: nit: Maybe I'm missing something, but this is pretty much the exact same code as `hashv`: https://github.com/anza-xyz/solana-sdk/blob/a467058aabc453c7d749a4993c56df293d1d75c3/sha256-hasher/src/lib.rs#L62-L73

Maybe we can just use that directly and avoid the `#[cfg(...)]`s here?

address/src/derive.rs
```diff
@@ -0,0 +1,160 @@
...
+    ///
+    /// # Important
+    ///
+    /// This function differs from [`Address::create_program_address`] in that
+    /// it does not perform a validation to ensure that the derived address is a valid
+    /// (off-curve) program derived address. It is intended for use in cases where the
+    /// seeds, bump, and program id are known to be valid, and the caller wants to derive
+    /// the address without incurring the cost of the `create_program_address` syscall.
+    ///
+    /// This function is a compile-time constant version of [`Address::derive_address`].
```
joncinque, line 92: nit: how about adding some information like:

"This function has worse performance than `derive_address`, so only use this function in const contexts, where all parameters are known at compile-time."

address/src/derive.rs
```diff
@@ -0,0 +1,160 @@
...
+    #[test]
+    fn test_derive_address() {
+        let program_id = Address::new_from_array([1u8; 32]);
+        let seeds: &[&[u8]; 2] = &[b"seed1", b"seed2"];
+        let bump = Some(255u8);
+
+        let derived_address = Address::derive_address(seeds, bump, &program_id);
+        let derived_address_const = Address::derive_address_const(seeds, bump, &program_id);
+
+        assert_eq!(derived_address, derived_address_const);
```
joncinque: nit: up to you, but how about comparing these derivations to the known / correct result?

address/src/lib.rs
```diff
@@ -287,8 +263,27 @@ impl Address {
...
+            // SAFETY: This is sound as `sol_sha256` always fills all 32 bytes of our hash.
+            unsafe {
+                solana_define_syscall::definitions::sol_sha256(
+                    vals.as_ptr() as *const _ as *const u8,
+                    vals.len() as u64,
+                    hash_result.as_mut_ptr() as *mut u8,
+                );
+                Ok(Address::from(hash_result.assume_init()))
+            }
+        }
```
joncinque: This will also be the same as `solana_sha256_hasher::hashv` -- is the idea to avoid that dependency for solana builds?

joncinque (commented):
Looks great overall! Just some small final things

address/src/lib.rs
febo: Yeah, but maybe we should just use `solana_sha256_hasher`. It does not make a big difference.

rust-toolchain.toml
febo: Oh, sorry, my bad.

address/src/derive.rs
febo: Reverted to use `solana_sha256_hasher`.

address/src/derive.rs
febo: Calculated the  address and bump using `find_program_address` and them used that in the comparisons.

address/src/derive.rs
```diff
@@ -0,0 +1,154 @@
...
+    ///
+    /// # Important
+    ///
+    /// This function differs from [`Address::create_program_address`] in that
+    /// it does not perform a validation to ensure that the derived address is a valid
+    /// (off-curve) program derived address. It is intended for use in cases where the
+    /// seeds, bump, and program id are known to be valid, and the caller wants to derive
+    /// the address without incurring the cost of the `create_program_address` syscall.
+    pub fn derive_address<const N: usize>(
+        seeds: &[&[u8]; N],
```
febo, line 35: @joncinque Just thought of something. We could make this accept an slice to give a bit of flexibility. The only advantage of taking an array is that we can make the length check at compile-time, but it won't allow a program to dynamically build the seeds.

joncinque: I'd say to keep it like this for now -- I don't know of too many use-cases where the seeds are dynamically built like this, especially to add more seeds.

But if we do get requests for it, it'll be easy to add 😄

PR #334, opened 2025-09-21
[bn-254] Change input to a sized array for LE addition, multiplication, and compression

LStan:
Currently, functions take slice as input. However, LE addition, multiplication, and compression require an input of a concrete length. This PR changes the input to a sized array, which is less error-prone.
This change can also be applied to BE compression, but it is a breaking change.


LStan:
Ready for review @samkim-crypto

LStan:
When will be the next major version release? Should I create a PR now?

samkim-crypto:
Let's add all the non-breaking changes in, publish v3.1.0, and then add breaking changes for v4. Did we have other changes that we wanted to include for v3.1 before we publish?

LStan:
> Did we have other changes that we wanted to include for v3.1 before we publish?

I don't have anything in mind

PR #335, opened 2025-09-23
use serde_core where applicable

kevinheavey:
serde_core is a new official serde crate pulled out of serde: https://docs.rs/serde_core/latest/serde_core/

It can be used when none of the serde derive macros are needed, and gives faster compile times. I went through the repo and found where it can be used (ie where we use serde without serde_derive). Turns out there are only three such crates in here.

### DIFF
--- a/transaction/src/lib.rs
+++ b/transaction/src/lib.rs
@@ -177,7 +177,7 @@ const NONCED_TX_MARKER_IX_INDEX: u8 = 0;
 #[cfg_attr(
     feature = "frozen-abi",
     derive(solana_frozen_abi_macro::AbiExample),
-    solana_frozen_abi_macro::frozen_abi(digest = "KSndwV1Ezw3xDX3Mz4Sg2vY22dx9mGTCFzo1RxbwaV8")
+    solana_frozen_abi_macro::frozen_abi(digest = "BLig4G2ysd7dcensK9bhKtnKvCQc1n65XdanyzsdWGXN")
 )]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[derive(Debug, PartialEq, Default, Eq, Clone)]


transaction/src/lib.rs
```diff
@@ -177,7 +177,7 @@ const NONCED_TX_MARKER_IX_INDEX: u8 = 0;
 #[cfg_attr(
     feature = "frozen-abi",
     derive(solana_frozen_abi_macro::AbiExample),
-    solana_frozen_abi_macro::frozen_abi(digest = "KSndwV1Ezw3xDX3Mz4Sg2vY22dx9mGTCFzo1RxbwaV8")
+    solana_frozen_abi_macro::frozen_abi(digest = "BLig4G2ysd7dcensK9bhKtnKvCQc1n65XdanyzsdWGXN")
```
apfitzge, line 180: I'm struggling to think why this would have changed from this.

apfitzge (commented):
added a question

transaction/src/lib.rs
joncinque: All I can see is that it's using short-vec to serialize signatures, and short-vec's serde traits comes from a different crate?

I ran the frozen-abi tests with `RUST_LOG=info` and got these little nuggets:

```
 updating with:             element solana_short_vec::ShortU16
 AbiEnumVisitor for &T: solana_short_vec::ShortU16
 AbiEnumVisitor for T: solana_short_vec::ShortU16
 AbiExample for struct: solana_short_vec::ShortU16
 updating with:                 tuple (elements = 1)
 updating with:                     element u8
 AbiEnumVisitor for &T: u8
 AbiEnumVisitor for T: u8
 updating with:                         primitive u8
 updating with:             element solana_signature::Signature
 AbiEnumVisitor for &T: solana_signature::Signature
 AbiEnumVisitor for T: solana_signature::Signature
 AbiExample for struct: solana_signature::Signature
 updating with:                 struct Signature(solana_signature::_::<impl serde_core::ser::Serialize for solana_signature::Signature>::serialize::__SerializeWith) (newtype) 
```

I imagine that one name change in the last line is enough to change the digest.

joncinque (approved):
This looks correct to me! The only other instance I found was as a dev-dependency of `serialize-utils`, but that's still pulling in serde through other deps anyway, and it's only a dev-dependency

apfitzge (approved):
lgtm. Thanks @joncinque for tracking down the digest difference

PR #338, opened 2025-09-24
feat: add `no_std` support for commitment-config crate

sonicfromnewyoke:
### Problem

`solana-commitment-config` crate is not `no_std` when it's no_std-friendly

### Summary of Changes

- make `solana-commitment-config` as `no_std` crate
- disable default (std) features for `serde` dependency

### DIFF
--- a/commitment-config/Cargo.toml
+++ b/commitment-config/Cargo.toml
@@ -17,5 +17,5 @@ targets = ["x86_64-unknown-linux-gnu"]
 serde = ["dep:serde", "dep:serde_derive"]
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


vadorovsky (commented):
Looks great! Could you add the commitment-config crate to `scripts/check-no-std.sh`, to make sure it's tested in CI?

sonicfromnewyoke:
this one passes `./cargo nightly check -Zbuild-std=core --target=bpfel-unknown-none --no-default-features -p solana-commitment-config` without any warnings/errors

PR #339, opened 2025-09-24
feat: add `no_std` support for compute-budget-interface crate

sonicfromnewyoke:
### Problem

`solana-compute-budget-interface` crate has no `no_std` support

### Summary of Changes
- add `std` feature and make it enabled by default
- use optional `serde` and its std feature in std mode of the `solana-compute-budget-interface` crate
- add `std` sub-feature in the `dev-context-only-utils` feature

### DIFF
--- a/compute-budget-interface/Cargo.toml
+++ b/compute-budget-interface/Cargo.toml
@@ -15,10 +15,12 @@ all-features = true
 rustdoc-args = ["--cfg=docsrs"]
 
 [features]
+default = ["std"]
 borsh = ["dep:borsh"]
-dev-context-only-utils = ["borsh"]
+dev-context-only-utils = ["borsh", "std"]
 frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro", "serde"]
 serde = ["dep:serde", "dep:serde_derive"]
+std = ["serde?/std", "borsh?/std"]
 
 [dependencies]
 borsh = { workspace = true, optional = true }
--- a/compute-budget-interface/Cargo.toml
+++ b/compute-budget-interface/Cargo.toml
@@ -15,8 +15,8 @@ all-features = true
 rustdoc-args = ["--cfg=docsrs"]
 
 [features]
-default = ["std"]
 borsh = ["dep:borsh"]
+default = ["std"]
 dev-context-only-utils = ["borsh", "std"]
 frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro", "serde"]
 serde = ["dep:serde", "dep:serde_derive"]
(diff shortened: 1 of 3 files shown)


vadorovsky (commented):
Amazing stuff! Let's add that crate to `scripts/check-no-std.sh`, to make sure it's tested in CI.

sonicfromnewyoke:
> Amazing stuff! Let's add that crate to `scripts/check-no-std.sh`, to make sure it's tested in CI.

added as suggestion to your MR for introducing this CI step

sonicfromnewyoke:
also fails on CI step with `solana-atomic-u64` with
```sh
error[E0463]: can't find crate for `std`
 --> atomic-u64/src/lib.rs:5:9
  |
5 |     use std::sync::atomic;
  |         ^^^ can't find crate
  |
  = note: the `bpfel-unknown-none` target may not support the standard library
  = help: consider building the standard library from source with `cargo build -Zbuild-std`

For more information about this error, try `rustc --explain E0463`.
error: could not compile `solana-atomic-u64` (lib) due to 2 previous errors
```

compute-budget-interface/Cargo.toml
```diff
@@ -16,20 +16,18 @@ rustdoc-args = ["--cfg=docsrs"]
...
 serde_derive = { workspace = true, optional = true }
-solana-frozen-abi = { workspace = true, features = [
-    "frozen-abi",
-], optional = true }
-solana-frozen-abi-macro = { workspace = true, features = [
-    "frozen-abi",
-], optional = true }
+solana-frozen-abi = { workspace = true, optional = true }
+solana-frozen-abi-macro = { workspace = true, optional = true }
 solana-instruction = { workspace = true, features = ["std"] }
```
raushan728, line 31: This is causing the CI failure on `atomic-u64`. Change this to `default-features = false` and add `solana-instruction/std` to the `std` array on line 23.

compute-budget-interface/src/lib.rs
```diff
@@ -45,7 +48,7 @@ macro_rules! to_instruction {
         Instruction {
             program_id: id(),
             data: data.to_vec(),
-            accounts: vec![],
+            accounts: [].to_vec(),
```
raushan728, line 51: `[].to_vec()` is a bit hacky for `no_std`. Add `extern crate alloc`; at the top of the file and use `alloc::vec::Vec::new()` here instead.

compute-budget-interface/src/lib.rs
```diff
@@ -1,20 +1,23 @@
...
+
+#[cfg(feature = "frozen-abi")]
+use solana_frozen_abi_macro::{AbiEnumVisitor, AbiExample};
 
 #[cfg(feature = "borsh")]
 use borsh::{BorshDeserialize, BorshSerialize};
 use solana_instruction::Instruction;
+
+#[cfg(feature = "std")]
+extern crate std;
```
raushan728, line 14: You can remove this. Rust automatically links `std` when the feature is enabled.

compute-budget-interface/src/lib.rs
```diff
@@ -86,8 +89,11 @@ mod tests {
     #[test]
     fn test_to_instruction() {
         let ix = ComputeBudgetInstruction::set_compute_unit_limit(257);
-        assert_eq!(ix.data, vec![2, 1, 1, 0, 0]);
+        assert_eq!(ix.data, [2, 1, 1, 0, 0].to_vec());
```
raushan728, line 92: Revert this back to `vec![]`. Tests run on the host with `std` enabled by default, so the macro works fine and keeps the code cleaner. (Same for the other .`to_vec()`

PR #341, opened 2025-09-25
[sdk-wasm-js] Add length check on type conversion

samkim-crypto:
#### Problem

The `solana-sdk-wasm-js` crate contain several functions that copy an outside-controlled JS data into WebAssembly memory before validating the input's size. This "copy-before-validate" can lead to excessive memory growth when oversized data are provided as input.

#### Summary of Changes

I went through the crate to added length checks for all the functions that "copy-before-validate".
- `hash.rs`: This was pretty straightforward. I just added length check if the value that is provided to the constructor does not match the expected hash bytes (32 bytes).

- `keypair.rs`: I updated the API of `fromBytes` function to directly take in a `Uint8Array` instead of a bytes slice. 

  When a function exported via #[wasm_bindgen] accepts a byte slice (&[u8]) argument, wasm-bindgen generates JavaScript glue code. This glue code automatically allocates space in Wasm linear memory and copies the entire input Uint8Array before the Rust function body executes. This prevents the Rust code from validating the size before the allocation occurs. 

  Other parts should be straightforward.

- `transaction.rs`: This was straightforward as well. The only nontrvial part is bounding the maximum transaction size length. I set this constant to be the maximum packet size from the `solana-packet` crate.

- `instruction.rs`: I changed the `setData` function to take in a `Uint8Array` instead of a byte slice.

- `address.rs`: This file required the most changes in terms of code lines, but should be straightforward I think.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3713,6 +3713,7 @@ dependencies = [
  "solana-instruction",
  "solana-keypair",
  "solana-message",
+ "solana-packet",
  "solana-signature",
  "solana-signer",
  "solana-transaction",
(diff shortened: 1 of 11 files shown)


joncinque:
Sorry for the late review, looks good on my side too!

PR #343, opened 2025-09-26
address: Add copy feature

### LINKED ISSUE #342: address: Only derive `Copy` if a crate feature is enabled
#### Problem

The `Address` type derives `Copy` by default, which is very useful in a lot of cases. For example, bytemuck's `Pod` trait requires all types within to implement `Copy`.

However, the `Copy` trait makes it easy to create many more copies than expected, making on-chain programs less efficient.

#### Proposed Solution

Only derive the `Copy` trait if a feature is enabled on `solana-address`.

cc @febo

febo:
### Problem

As described in #342 , deriving `Copy` by default makes it easier to create more copies  than expected, in particular when writing on-chain programs.

### Solution

Add a feature `"copy"` so the derive is opt-in.

Closes #342

### DIFF
--- a/address/src/lib.rs
+++ b/address/src/lib.rs
@@ -257,7 +257,7 @@ impl Address {
         Ok(Address::from(hash.to_bytes()))
     }
 
-    pub const fn to_bytes(self) -> [u8; 32] {
+    pub const fn to_bytes(&self) -> [u8; 32] {
         self.0
     }
 


joncinque:
Looks great! We just need to have a firm idea of how to release it. I think it's the perfect time for the semver trick

address/src/lib.rs
```diff
@@ -254,7 +255,7 @@ impl Address {
         Ok(Address::from(hash.to_bytes()))
     }
 
-    pub const fn to_bytes(self) -> [u8; 32] {
+    pub const fn to_bytes(&self) -> [u8; 32] {
```
febo, line 260: Taking a reference avoids breaking code that relied on address being `Copy`.

joncinque (commented):
This one still looks good to me! The changes are consistent with the `Copy` changes from Hash since we *don't* add it as a default feature, and `to_bytes` has been modified to take a reference.

Edit: it just needs a rebase

PR #346, opened 2025-09-29
Use address

joncinque:
#### Problem

The sdk repo still uses Pubkey mostly, which makes it tougher to make more of the SDK no-std.

#### Summary of changes

Use Address in more places, to avoid adding the std feature on solana-pubkey as much as possible.

This will help #302 land with minimal additional changes.


account-info/src/lib.rs
```diff
@@ -20,14 +20,14 @@ pub const MAX_PERMITTED_DATA_INCREASE: usize = 1_024 * 10;
 #[derive(Clone)]
 #[repr(C)]
 pub struct AccountInfo<'a> {
-    /// Public key of the account
-    pub key: &'a Pubkey,
+    /// Address of the account
+    pub key: &'a Address,
```
febo, line 24: Related to this PR, will we continue to use `key` to refer to the address of an account?

sdk-ids/src/lib.rs
```diff
@@ -1,121 +1,121 @@
...
 
 pub mod vote {
-    solana_pubkey::declare_id!("Vote111111111111111111111111111111111111111");
+    solana_address::declare_id!("Vote111111111111111111111111111111111111111");
 }
 
 pub mod sysvar {
     // Owner pubkey for sysvar accounts
-    solana_pubkey::declare_id!("Sysvar1111111111111111111111111111111111111");
+    solana_address::declare_id!("Sysvar1111111111111111111111111111111111111");
```
febo, line 77: The comment above says "Owner pubkey" – wondering if it should be "Owner address".

febo (dismissed):
Looks really good! Noticed that some of the crates are still referencing `solana_pubkey` – I guess they will be eventually updated in a separate PR.

account-info/src/lib.rs
joncinque: I want to avoid the breaking change, especially considering we want people to switch to Pinocchio anyway

sdk-ids/src/lib.rs
joncinque: Ah nice find, thanks!

joncinque:
Yeah exactly, this change is the base of what's required to make pubkey no-std without breaking things. We'll have to make more changes later

febo (approved):
Looks great!

PR #347, opened 2025-09-29
msg: Use dependency without default features

joncinque:
#### Problem

The solana-msg crate is brought in with default features, which makes it impossible to make downstream users of the crate no-std.

#### Summary of changes

Add `default-features = false` for solana-msg at the workspace-level, and add the std feature to solana-program, which is the only place that requires it.

As part of this, I noticed that solana-program-entrypoint re-exports msg, but the re-export is never used, so I removed it.

### DIFF
--- a/program-entrypoint/src/lib.rs
+++ b/program-entrypoint/src/lib.rs
@@ -12,14 +12,12 @@ use {
         slice::{from_raw_parts, from_raw_parts_mut},
     },
 };
-// need to re-export msg for custom_heap_default macro, `AccountInfo` and `Pubkey` for
-// entrypoint_no_alloc macro
+// need to re-export `AccountInfo` and `Pubkey` for entrypoint_no_alloc macro
 pub use {
     solana_account_info::AccountInfo as __AccountInfo,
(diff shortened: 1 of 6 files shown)


program-entrypoint/src/lib.rs
```diff
@@ -12,14 +12,12 @@ use {
...
 };
-// need to re-export msg for custom_heap_default macro, `AccountInfo` and `Pubkey` for
-// entrypoint_no_alloc macro
+// need to re-export `AccountInfo` and `Pubkey` for entrypoint_no_alloc macro
 pub use {
     solana_account_info::AccountInfo as __AccountInfo,
     solana_account_info::MAX_PERMITTED_DATA_INCREASE,
     // Re-exporting for custom_panic
     solana_define_syscall::definitions::{sol_log_ as __log, sol_panic_ as __panic},
-    solana_msg::msg as __msg,
```
febo, line 22: Good catch! I think the `custom_panic_default` used to rely on `msg` to log a panic message, but when it was refactored it is using the `log` syscall directly.

febo (approved):
Looks great!

PR #348, opened 2025-09-29
hash: Only implement Copy on feature

joncinque:
#### Problem

The Hash type, like Address, implements `Copy` by default, which makes it easy for downstream users to accidentally incur a copy when they don't want to.

#### Summary of changes

Gate the `Copy` derivation on a `copy` feature, and make it a default feature. Fix downstream breakage within the repo.


joncinque:
Note that I'm not totally sold on this... most people don't use hashes on-chain, so I'm not sure how much benefit this will bring

blake3-hasher/src/lib.rs
```diff
@@ -76,7 +76,9 @@ mod tests {
         let val_hash = hash(val.as_bytes());
 
         let ext = "lM890t";
-        let ext_hash = hashv(&[&val_hash.to_bytes(), ext.as_bytes()]);
+        // Builds that enable copy on hash will trip this
+        #[allow(clippy::clone_on_copy)]
+        let ext_hash = hashv(&[&val_hash.clone().to_bytes(), ext.as_bytes()]);
```
febo: What do you think if we change the `to_bytes` implementation on the `Hash` type from:
```rust
pub const fn to_bytes(self) -> [u8; HASH_BYTES] {
    self.0
}
```
to:
```rust
pub const fn to_bytes(&self) -> [u8; HASH_BYTES] {
    self.0
}
```
Then we can use `to_bytes` directly without the `clone_on_copy` issue. Since `Hash` is currently copy by default, I think this won't break things.

hash/Cargo.toml
```diff
@@ -17,8 +17,9 @@ rustdoc-args = ["--cfg=docsrs"]
 
 [features]
 borsh = ["dep:borsh", "std"]
-bytemuck = ["dep:bytemuck", "dep:bytemuck_derive"]
-default = ["std"]
+bytemuck = ["copy", "dep:bytemuck", "dep:bytemuck_derive"]
+copy = []
+default = ["copy", "std"]
```
febo: Since this is a breaking change, could we remove the default features and both `"std"` and `"copy"` have to be explicitly enabled? I am thinking that people might want to have a `Hash` field on an account, so it could be useful to have the crate "lighter" by default.

sdk-macro/src/lib.rs
```diff
@@ -239,7 +239,7 @@ pub fn derive_clone_zeroed(input: proc_macro::TokenStream) -> proc_macro::TokenS
                 syn::Fields::Named(ref fields) => fields.named.iter().map(|f| {
                     let name = &f.ident;
                     quote! {
-                        core::ptr::addr_of_mut!((*ptr).#name).write(self.#name);
+                        core::ptr::addr_of_mut!((*ptr).#name).write(self.#name.clone());
```
febo, line 243: Would this trigger the `clone_on_copy` warning since `"copy"` is enabled by default?

apfitzge:
Can you explain the justification a bit more here?
I realize copying 32 bytes all over the place is not good for perf. At the same time, it seems almost to me this is backward? Did we consider opting-out of `Copy`?

For 99% of people this is just going to be annoying and they'll always enable the feature, and for them it probably doesn't affect performance in a concrete way.
It really only matters for performance maximization, which seems a much smaller set of devs.

sdk-macro/src/lib.rs
joncinque: Hm... maybe, I didn't see it in any of my testing, but I could add the clippy allow here if you think that'll be safer

joncinque:
The idea is that it was a mistake to add `Copy` to `Hash` and `Pubkey` in the first place, so we can correct it.

Typically, opt-out crate features are considered an anti-pattern, and crate features should add functionality.

The best possibility I see for downstream devs is to make `copy` a default crate feature, which I'm open to. Maybe @febo can provide reasoning for his side

sdk-macro/src/lib.rs
febo: I think we are ok – apparently clippy does not apply to code generated by proc macros.

febo:
> The idea is that it was a mistake to add `Copy` to `Hash` and `Pubkey` in the first place, so we can correct it.
> 
> Typically, opt-out crate features are considered an anti-pattern, and crate features should add functionality.
> 
> The best possibility I see for downstream devs is to make `copy` a default crate feature, which I'm open to. Maybe @febo can provide reasoning for his side

Yeah, making it a default crate feature is a good compromise. We are doing a similar change to `Address` – mentioning this here in case we want to have a similar behaviour. In the case of `Address`, we probably don't want to have it as a default crate feature.

joncinque:
Ok great, added it as a default feature then!

sdk-wasm-js/src/hash.rs
```diff
@@ -82,6 +82,6 @@ impl Hash {
 
     /// Return the `Uint8Array` representation of the hash
     pub fn toBytes(&self) -> Box<[u8]> {
-        self.inner.to_bytes().into()
+        self.inner.clone().to_bytes().into()
```
febo: With `to_bytes` taking a reference now, we don't need `clone()`.

message/src/legacy.rs
```diff
@@ -286,7 +286,7 @@ impl Message {
             header.num_readonly_signed_accounts,
             header.num_readonly_unsigned_accounts,
             account_keys,
-            *blockhash,
+            Hash::new_from_array(blockhash.to_bytes()),
```
febo, line 289: nit: if `"copy"` is enabled by default, we could keep the original code.

shred-version/src/lib.rs
```diff
@@ -24,7 +24,7 @@ pub fn version_from_hash(hash: &Hash) -> u16 {
 }
 
 pub fn compute_shred_version(genesis_hash: &Hash, hard_forks: Option<&HardForks>) -> u16 {
-    let mut hash = *genesis_hash;
+    let mut hash = Hash::new_from_array(genesis_hash.to_bytes());
```
febo, line 27: nit: same here, if `"copy"` is enabled by default we can keep the original code.

vote-interface/src/instruction.rs
```diff
@@ -219,15 +219,16 @@ impl VoteInstruction {
     /// Only to be used on vote instructions (guard with is_simple_vote), panics otherwise
     pub fn hash(&self) -> Hash {
         assert!(self.is_simple_vote());
-        match self {
-            Self::Vote(v) | Self::VoteSwitch(v, _) => v.hash,
+        let hash = match self {
```
febo, line 222: nit: same note about `"copy"` being enabled by default, keep the original code?

vote-interface/src/state/mod.rs
```diff
@@ -254,7 +254,7 @@ pub mod serde_compact_vote_state_update {
         let compact_vote_state_update = CompactVoteStateUpdate {
             root: vote_state_update.root.unwrap_or(Slot::MAX),
             lockout_offsets: lockout_offsets.collect::<Result<_, _>>()?,
-            hash: vote_state_update.hash,
+            hash: Hash::new_from_array(vote_state_update.hash.to_bytes()),
```
febo, line 257: nit: same note about `"copy"` being enabled by default, keep the original code?

febo (commented):
Looks great - just a couple of nits.

apfitzge:
> n the case of Address, we probably don't want to have it as a default crate feature.

Could you explain why we wouldn't also want it as default? It seems like the same things I mentioned above would apply there too 😄

febo:
> > n the case of Address, we probably don't want to have it as a default crate feature.
> 
> Could you explain why we wouldn't also want it as default? It seems like the same things I mentioned above would apply there too 😄

Well, `Address` is used in every program while `Hash` is probably a small percentage. Having to enable the "copy" feature on `Address` is an explicitly action to indicate "I know that I will be copying 32 bytes on an on-chain program" instead of a default behaviour. 😅

sdk-wasm-js/src/hash.rs
joncinque: Ah right, fixed!

message/src/legacy.rs
joncinque: The workspace uses `default-features = false` when importing solana-hash, so we would need to enable the copy feature on `solana-message` explicitly to get the functionality.

Should we enable the feature for all the other crates that you point out in your comments? I'm leaning towards keeping it as it is, so that other downstream users can choose to enable the feature, and the sdk can work around it.

febo: Yeah, this is better – let's keep it as it is.

nonce/Cargo.toml
```diff
@@ -29,7 +29,7 @@ serde = [
 serde = { workspace = true, optional = true }
 serde_derive = { workspace = true, optional = true }
 solana-fee-calculator = { workspace = true }
-solana-hash = { workspace = true, default-features = false }
+solana-hash = { workspace = true, default-features = false, features = ["copy"] }
```
febo: nit: the workspace already declares `default-features = false`, so not sure whether it is meant to be explicit here or not.

apfitzge:
> Well, Address is used in every program while Hash is probably a small percentage. Having to enable the "copy" feature on Address is an explicitly action to indicate "I know that I will be copying 32 bytes on an on-chain program" instead of a default behaviour. 😅

I think we should have consistent default features here. So if we're not gonna do it for `Address` I think we'd not want it for `Hash`.

I'd make the same arguments for `Address` though. Most people are probably just going to get annoyed at the break and then just enable the feature. Only the people who are optimizing CUs are going to want to not have `Copy` so they don't accidently do it.

joncinque:
@febo and I discussed a bit offline, and we're leaning back towards having it off by default for hash and address. Since we're making breaking changes that are all based around features, we can experiment and see how much grumbling we get from downstream devs. Since there's no need to upgrade, we can allow complaints to pile up.

For example, if it's a huge amount of churn for agave, we can just enable the features by default and release a new version, without any breakage. Going the other way and removing a default feature is a breaking change though, so we should avoid that. What do you think @apfitzge ?

febo (approved):
Looks great!

apfitzge (approved):

PR #351, opened 2025-09-30
Improve `fn download_file` for Gtihub REST API

LucasSte:
**Problem**

Developers are having problem downloading platform tools directly from a Github link, specially in slow connections. That is happening because the JWT token GitHub creates for the download expires in five minutes.

A way around this problem is to use instead the Github REST API for the download. It requires a couple of headers to work, but the function we use for downloads does not support headers.

**Summary of changes**

Add support for headers, without requiring an external `reqwest::header::HeaderMap`.

### DIFF
--- a/file-download/src/lib.rs
+++ b/file-download/src/lib.rs
@@ -7,6 +7,7 @@ use {
         fs::{self, File},
         io::{self, Read},
         path::Path,
+        str::FromStr,
         time::{Duration, Instant},
     },
 };
@@ -61,7 +62,7 @@ pub fn download_file<'a, 'b>(
     use_progress_bar: bool,
     progress_notify_callback: &'a mut DownloadProgressCallbackOption<'b>,
 ) -> Result<(), String> {
-    download_file_with_headers(
+    download_file_with_headers::<&str>(
         url,
         destination_file,
         use_progress_bar,
@@ -71,12 +72,12 @@ pub fn download_file<'a, 'b>(
 }
 
 /// This function works like `download_file`, but accepts HTTP headers.
-pub fn download_file_with_headers<'a, 'b>(
+pub fn download_file_with_headers<'a, 'b, S: AsRef<str>>(
     url: &str,
     destination_file: &Path,
     use_progress_bar: bool,
     progress_notify_callback: &'a mut DownloadProgressCallbackOption<'b>,
-    headers: &[(&'static str, &'static str)],
+    headers: &[(S, S)],
 ) -> Result<(), String> {
     if destination_file.is_file() {
         return Err(format!("{destination_file:?} already exists"));
@@ -103,7 +104,10 @@ pub fn download_file_with_headers<'a, 'b>(
 
     let mut header_map = reqwest::header::HeaderMap::new();
     for (key, value) in headers {
-        header_map.insert(*key, value.parse().unwrap());
+        header_map.insert(
+            reqwest::header::HeaderName::from_str(key.as_ref()).map_err(|e| format!("{e}"))?,
+            value.as_ref().parse().map_err(|e| format!("{e}"))?,
+        );
     }
 
     let response = reqwest::blocking::Client::new()


Lichtso (approved):

LucasSte:
I'll wait for @joncinque's review before merging to ensure this PR does not brake anything.

file-download/src/lib.rs
```diff
@@ -60,6 +60,7 @@ pub fn download_file<'a, 'b>(
     destination_file: &Path,
     use_progress_bar: bool,
     progress_notify_callback: &'a mut DownloadProgressCallbackOption<'b>,
+    headers: &[(&'static str, &'static str)],
```
joncinque: Since these just get parsed anyway, how about taking a more permissive type?
```suggestion
    headers: &[(AsRef<str>, AsRef<str>)],
```

file-download/src/lib.rs
joncinque, line 59: To avoid the breaking change here, how about adding a new function like `download_file_with_headers` with all the logic, and have `download_file` call into `download_file_with_headers` with empty headers?

file-download/src/lib.rs
LucasSte: I'm afraid that may not be possible:

```
error[E0782]: expected a type, found a trait
  --> file-download/src/lib.rs:63:17
   |
63 |     headers: &[(AsRef<&str>, AsRef<&str>)],
   |                 ^^^^^^^^^^^
   |
help: you can add the `dyn` keyword if you want a trait object
   |
63 |     headers: &[(dyn AsRef<&str>, AsRef<&str>)],
   |                 +++
```

And then,

```
error[E0277]: the size for values of type `dyn AsRef<&str>` cannot be known at compilation time
  --> file-download/src/lib.rs:63:16
   |
63 |     headers: &[(dyn AsRef<&str>, dyn AsRef<&str>)],
   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
   |
   = help: the trait `Sized` is not implemented for `dyn AsRef<&str>`
   = note: only the last element of a tuple may have a dynamically sized type
```

file-download/src/lib.rs
LucasSte: Fixed in https://github.com/anza-xyz/solana-sdk/pull/351/commits/fc69a6952296b3c66a1a1643f4bf53f2642763ba

file-download/src/lib.rs
LucasSte: I just noticed I was using `AsRef<&str>`, and not `AsRef<str>`, as the suggestion, but the errors are the same:

```
error[E0782]: expected a type, found a trait
  --> file-download/src/lib.rs:79:17
   |
79 |     headers: &[(AsRef<str>, AsRef<str>)],
   |                 ^^^^^^^^^^
   |
help: you can add the `dyn` keyword if you want a trait object
   |
79 |     headers: &[(dyn AsRef<str>, AsRef<str>)],
   |                 +++
```

And then,

```
error[E0277]: the size for values of type `dyn AsRef<str>` cannot be known at compilation time
  --> file-download/src/lib.rs:79:16
   |
79 |     headers: &[(dyn AsRef<str>, dyn AsRef<str>)],
   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
   |
   = help: the trait `Sized` is not implemented for `dyn AsRef<str>`
   = note: only the last element of a tuple may have a dynamically sized type
```

joncinque: Sorry, I didn't mean for that to be used literally -- you typically need some generic types declared to make it work. I made that change on a branch, and also added error handling in case the headers can't be parsed, let me know what you think! https://github.com/joncinque/solana-sdk/commit/6d36fc2c0a18814af40d8554d0e66e215f2d64d6

LucasSte: Ok, accepted your suggestion in https://github.com/anza-xyz/solana-sdk/pull/351/commits/6bb7c2ce9892a31cddc2d0168354fdcd440241f7

PR #352, opened 2025-10-02
SIMD-0312: add CreateAccountAllowPrefund system instruction

rustopian:
[SIMD](https://github.com/solana-foundation/solana-improvement-documents/pull/312)

Adds the `CreateAccountAllowPrefund` instruction. This may be a breaking change in rare cases, since any exhaustive pattern matching will now have an additional variant of `SystemInstruction` to deal with.

### DIFF
--- a/system-interface/src/instruction.rs
+++ b/system-interface/src/instruction.rs
@@ -1720,16 +1720,16 @@ pub fn upgrade_nonce_account(nonce_address: Address) -> Instruction {
 /// # Required signers
 ///
 /// The `new_account_address` signer must sign the transaction. If present,
-/// the signer for the address in `funding` must also sign the transaction.
+/// the payer in `payer_and_lamports` must also sign the transaction.
 #[cfg(feature = "bincode")]
 pub fn create_account_allow_prefund(
     new_account_address: &Address,
-    funding: Option<(&Address, u64)>,
+    payer_and_lamports: Option<(&Address, u64)>,
     space: u64,
     owner: &Address,
 ) -> Instruction {
     let mut account_metas = vec![AccountMeta::new(*new_account_address, true)];
-    let lamports = match funding {
+    let lamports = match payer_and_lamports {
         None => 0,
         Some((from, lamports)) => {
             account_metas.push(AccountMeta::new(*from, true));


system-interface/src/instruction.rs
```diff
@@ -1669,6 +1713,43 @@ pub fn upgrade_nonce_account(nonce_address: Address) -> Instruction {
...
+/// The `to_address` signer must sign the transaction. If present, the
+/// `from_address` signer must also sign the transaction.
+///
+/// `None` may be passed for `from_address` when `lamports == 0` (no transfer occurs).
+/// When `lamports > 0`, pass `Some(funding_account)` so the transfer can occur, or
+/// the instruction will fail downstream.
+#[cfg(feature = "bincode")]
+pub fn create_account_allow_prefund(
+    from_address: &Option<Address>,
+    to_address: &Address,
```
grod220: I tripped up at first with the from/to language in this case. What do you think of `new_account_address` & `funding_account_address`?

grod220: If `lamports > 0 requires -> Some(address)`, we could prevent an invalid argument in the typing itself with an enum:
```rust
pub enum Funding {
    None,
    Transfer {
        from: Address,
        lamports: u64,
    },
}

pub fn create_account_allow_prefund(
    new_account_address: &Address,
    funding: Funding,
    space: u64,
    owner: &Address,
) -> Instruction {
```

system-interface/src/instruction.rs
```diff
@@ -249,6 +256,43 @@ pub enum SystemInstruction {
...
+    /// account. This instruction does not prevent the new account from having more
+    /// lamports than required for rent exemption, and all lamports will become locked.
+    ///
+    /// # Account references
+    /// If `lamports > 0` (meaning lamports are being transferred):
+    ///   0. `[WRITE, SIGNER]` New account
+    ///   1. `[WRITE, SIGNER]` Funding account
+    ///
+    /// If `lamports == 0` (no lamports to be transferred), you may omit funding account:
+    ///   0. `[WRITE, SIGNER]` New account
```
grod220: Very clear docs!

system-interface/src/instruction.rs
grod220: Also, feels like the optional address should be second to match the instruction ordering.

rustopian: I like this new terminology. However, it does break from what is [currently used](https://github.com/anza-xyz/solana-sdk/blob/1a91efa795650d14169c8b5bb91fea26133c8f2f/system-interface/src/instruction.rs#L408-L410) by `create_account`.

As these are just parameter names, and I agree can be confusing, we can consider updating the names in the other functions in a subsequent PR (even the example functions in the doc comments use `payer` and `new_account` instead).

rustopian:
@grod220 I added your suggested `Funding` type and clearer arg names. Nice little protection.

This file needs a bit more unrelated touch up (to use solana-sdk-ids for example), but I'll handle that in a separate chores PR.

system-interface/src/instruction.rs
```diff
@@ -1669,6 +1725,41 @@ pub fn upgrade_nonce_account(nonce_address: Address) -> Instruction {
     Instruction::new_with_bincode(ID, &SystemInstruction::UpgradeNonceAccount, account_metas)
 }
 
+/// Create a new account without enforcing zero lamports.
```
joncinque: nit for clarity
```suggestion
/// Create a new account without enforcing zero lamports on the destination account.
```

system-interface/src/instruction.rs
```diff
@@ -249,6 +256,55 @@ pub enum SystemInstruction {
...
+/// A funding argument where any amount of `lamports` requires a `from`
+/// address. Currently only for use with `create_account_allow_prefund`.
+#[cfg_attr(
+    feature = "serde",
+    derive(serde_derive::Deserialize, serde_derive::Serialize)
+)]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum Funding {
+    None,
+    Transfer { from: Address, lamports: u64 },
```
joncinque: To be honest, I'm not sure if this is a great idea. It's currently passed in by value, which forces a copy of the pubkey and lamports.

Even if it's passed by reference, users are forced to create an intermediate struct just to create the instruction.

Finally, the resulting code is more verbose than just taking in `lamports` and `Option<&Address>` always.

Is that worth the tradeoff to make it impossible for someone to specify an amount of lamports without also specifying a funder? Maybe! But I'm not convinced.

system-interface/src/instruction.rs
```diff
@@ -14,6 +14,13 @@
 //!
 //! [rent exempt]: https://solana.com/docs/core/accounts#rent-exemption
 //!
+//! The [`create_account`] function requires that the account have zero
+//! lamports. [`create_account_allow_prefund`] allows for the account to have
+//! lamports prefunded; note that without feature activation of SIMD-0312,
```
joncinque: Did you want the link later to reference here?
```suggestion
//! lamports prefunded; note that without feature activation of [SIMD-0312],
```

joncinque (commented):
Looks great! I'm not sure about the extra enum, but maybe I can be convinced

system-interface/src/instruction.rs
rustopian: Cc @grod220 - how strongly do you feel on the `Funding` type protection here? 

I love such things, but since we were already checking the `Option<Address>` in the original anyway...

```
if lamports > 0 {
  if let Some(from) = payer {
    account_metas.push(
      AccountMeta::new(*from, true)
    );
  } // could `else {error}` here
}
```

...maybe a middle ground is to just error here? Better UX than failing downstream, no extra type.

At the cost of returning a `Result`-wrapped `Instruction` tho.

grod220: I don't feel strongly, especially if it breaks expected patterns of the crate. I suppose it's more of a question _where_ we want this case to throw:

- At compile time -> use enum
- At runtime call site -> add error to this helper
- At program execution -> no guards

rustopian: Let's see if `Option<(Address, u64>)` is good. Then if `is_none()`, no transfer, no funding account. No extra type, just a tuple, but still keeps the u64 and Address linked.

joncinque (approved):
Looks great!

grod220 (approved):
💯

PR #355, opened 2025-10-07
Feat: borsh, serde, and address no_std

### LINKED ISSUE #354: Disable borsh default features
Borsh has support for no_std by disabling default features (it defaults to std). This results in an extern crate alloc and would work in no_std contexts. Setting default_features = false in the root cargo toml would enable use of borsh in no_std (with the address crate for example).

stegaBOB:
Disables default features for `serde` and `borsh`, enabing no_std support. That on its own shouldn't be a breaking change since neither dependency is re-exported. Also updated the address crate to support no_std with alloc/borsh/serde enabled.

I enabled no_std for the `atomic-u64` crate, but atomics aren't actually supported on the `bpfel-unknown-none` target (unless I'm configuring something wrong, which very well could be the case). Can definitely remove that change, but still probably beneficial for other no_std targets that do support atomics.

To check that this all works I ran the following in the address crate:
```shell
cargo +nightly build -Zbuild-std=alloc,core --target=bpfel-unknown-none --no-default-features --features "decode, error, sanitize, syscalls, borsh, serde, bytemuck"
```

Closes #354

### DIFF
--- a/scripts/check-no-std.sh
+++ b/scripts/check-no-std.sh
@@ -35,9 +35,9 @@ target="bpfel-unknown-none"
   --no-default-features \
   "${no_std_crates[@]}"
 
-# Check that `solana-address` w/ all features that work with no_std + alloc still works!
+# Check that all crates with features that work with no_std + alloc still work!
 ./cargo nightly check -Zbuild-std=alloc,core \
   "--target=${target}" \
   --no-default-features \
-  --features "decode, error, sanitize, syscalls, borsh, serde, bytemuck" \
-  -p solana-address
+  --features "decode, error, sanitize, syscalls, borsh, serde, bytemuck, alloc" \
+  "${no_std_crates[@]}"


atomic-u64/src/lib.rs
```diff
@@ -1,9 +1,10 @@
+#![no_std]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 pub use implementation::AtomicU64;
 
 #[cfg(target_pointer_width = "64")]
 mod implementation {
-    use std::sync::atomic;
+    use core::sync::atomic;
```
joncinque, line 7: This is fine as it is, so there's nothing to change.

With that said, the crate still won't work for upstream BPF, since atomics aren't available on that target. We'll still need to correctly gate usage of `solana-atomic-u64` in downstream crates

msg/src/lib.rs
```diff
@@ -1,7 +1,15 @@
...
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#[cfg(feature = "alloc")]
+extern crate alloc;
 #[cfg(feature = "std")]
 extern crate std;
+
+/// We re-export the `format!` macro from `alloc` for use in the `msg!` macro
+#[cfg(feature = "alloc")]
+#[doc(hidden)]
+pub use alloc::format;
```
joncinque, line 11: Is this needed? If we reference it as `::alloc::format` in the macro, I think we can get by without this

Cargo.toml
```diff
@@ -201,10 +201,10 @@ quote = "1.0.35"
 rand = "0.8.5"
 rayon = "1.10.0"
 reqwest = { version = "0.11.27", default-features = false }
-serde = "1.0.226" # must match the serde_derive version, see https://github.com/serde-rs/serde/issues/2584#issuecomment-1685252251
+serde = { version = "1.0.226", default-features = false, features = ["alloc"] } # must match the serde_derive version, see https://github.com/serde-rs/serde/issues/2584#issuecomment-1685252251
 serde-big-array = "0.5.1"
 serde_bytes = "0.11.15"
-serde_core = "1.0.226" # must match the serde version, see https://github.com/serde-rs/serde/issues/2584#issuecomment-1685252251
+serde_core = { version = "1.0.226", default-features = false, features = ["alloc"] }  # must match the serde version, see https://github.com/serde-rs/serde/issues/2584#issuecomment-1685252251
```
joncinque: Out of curiosity, I also removed the `alloc` feature, and with a few other tweaks, I got the powerset build worked, so how about we go completely without features on serde?

signature/Cargo.toml
```diff
@@ -22,7 +22,7 @@ bytemuck = ["dep:bytemuck", "dep:bytemuck_derive"]
 frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro", "std"]
 rand = ["dep:rand"]
 serde = ["dep:serde", "dep:serde_derive", "dep:serde-big-array"]
-std = ["alloc"]
+std = ["alloc", "serde/std"]
```
joncinque: This should only enable std on serde if the serde crate is included, right?
```suggestion
std = ["alloc", "serde?/std"]
```

epoch-rewards/Cargo.toml
```diff
@@ -22,7 +22,7 @@ frozen-abi = [
     "std",
 ]
 serde = ["dep:serde", "dep:serde_derive", "solana-hash/serde"]
-std = []
+std = ["serde/std"]
```
joncinque: Same here, we should probably only enable the feature if serde is enabled, right?
```suggestion
std = ["serde?/std"]
```

address/Cargo.toml
```diff
@@ -33,7 +33,8 @@ rand = ["dep:rand", "atomic", "std"]
 sanitize = ["dep:solana-sanitize"]
 serde = ["dep:serde", "dep:serde_derive"]
 sha2 = ["dep:solana-sha256-hasher", "error"]
-std = ["decode"]
+std = ["decode", "borsh?/std", "serde?/std"]
+alloc = []
```
joncinque: I think this is a breaking change, since someone just using `std` currently won't get the `TryFrom<Vec<u8>>` that used to be gated by std and is now gated by `alloc`.

That might be ok though, because we can use the semver trick to always enable the proper features. We'll just need to be very careful in that backport.

With that said, how about having `std` also enable `alloc`, and also adding a comment saying to remove the enabling of `alloc` with `std` enabled? That way, this PR contains no breaking changes, and we can make the breaking change in future work.

address/Cargo.toml
```diff
@@ -17,7 +17,7 @@ rustdoc-args = ["--cfg=docsrs"]
 
 [features]
 atomic = ["dep:solana-atomic-u64"]
-borsh = ["dep:borsh", "std"]
+borsh = ["dep:borsh", "alloc"]
```
joncinque, line 20: Confirming this is correct: borsh just uses `alloc` if std is turned off

hash/src/lib.rs
```diff
@@ -3,14 +3,15 @@
 #![cfg_attr(feature = "frozen-abi", feature(min_specialization))]
 #[cfg(feature = "borsh")]
 use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
-#[cfg(feature = "std")]
-extern crate std;
```
joncinque, line 7: We'll still need this for frozen-abi to work

hash/Cargo.toml
```diff
@@ -16,12 +16,12 @@ all-features = true
 rustdoc-args = ["--cfg=docsrs"]
 
 [features]
-borsh = ["dep:borsh", "std"]
+borsh = ["dep:borsh"]
```
joncinque, line 19: Surprisingly, I don't think this is a breaking change, except potentially for a misconfigured frozen-abi build.

msg/src/lib.rs
stegaBOB: I think it is needed, since alloc might not be an extern crate where this is pulled in. I couldn't get it to compile without this fwiw

Cargo.toml
stegaBOB: Sure! I thought alloc was needed somewhere for some reason, but if you're getting it to build that's great!

signature/Cargo.toml
stegaBOB: Good catch!

address/Cargo.toml
stegaBOB: That should be right!

address/Cargo.toml
stegaBOB: Good catch. I think std enabling alloc is probably the way we'd want this to work anyways. There should be anything that alloc enables that we wouldnt want std to enable too

hash/src/lib.rs
stegaBOB: Oops! Added back

Cargo.toml
stegaBOB: Ah looks like you actually made it work. I saw that a number of crates weren't working without alloc and just gave up haha. Yours is definitely more precise

address/Cargo.toml
```diff
@@ -33,7 +33,8 @@ rand = ["dep:rand", "atomic", "std"]
 sanitize = ["dep:solana-sanitize"]
 serde = ["dep:serde", "dep:serde_derive"]
 sha2 = ["dep:solana-sha256-hasher", "error"]
-std = ["decode"]
+std = ["decode", "borsh?/std", "serde?/std", "alloc"]
```
joncinque, line 36: nit: Can you add a comment above this saying to remove the enabling of `alloc` during the next breaking release?

stegaBOB: I'm not sure we'd want this. I think std should probably enable alloc as well

stegaBOB:
> Separately, I'm wondering how we can test this. The upstream BPF target is no-std, but also doesn't allow for alloc. I looked through different possible targets that are no-std but allow alloc, and found this list: https://github.com/search?q=repo%3Arust-lang%2Frust+path%3A%2F%5Esrc%5C%2Fdoc%5C%2Frustc%5C%2F%2F+%22core+and+alloc%22&type=code

Are you certain this is the case? If you add alloc with `-Zbuild-std=alloc,core`, targeting `bpfel-unknown-none` seems to work fine

```shell
cargo +nightly build -Zbuild-std=alloc,core --target=bpfel-unknown-none --no-default-features --features "decode, error, sanitize, syscalls, borsh, serde, bytemuck"
```

address/Cargo.toml
joncinque: We're going to break up a few of these features, but it's true that it might be odd to need to enable both separately. What do you think @febo ?

joncinque:
> Are you certain this is the case? If you add alloc with -Zbuild-std=alloc,core, targeting bpfel-unknown-none seems to work fine

That's much smarter, let's do that. Can you add another line in `check-no-std.sh` that also runs that command?

stegaBOB:
> That's much smarter, let's do that. Can you add another line in `check-no-std.sh` that also runs that command?

Done @joncinque!

scripts/check-no-std.sh
```diff
@@ -34,3 +34,10 @@ target="bpfel-unknown-none"
   "--target=$target" \
   --no-default-features \
   "${no_std_crates[@]}"
+
+# Check that `solana-address` w/ all features that work with no_std + alloc still works!
+./cargo nightly check -Zbuild-std=alloc,core \
+  "--target=${target}" \
+  --no-default-features \
+  --features "decode, error, sanitize, syscalls, borsh, serde, bytemuck" \
+  -p solana-address
```
joncinque: Let's go wild and test all of them 😈 
```suggestion
  "${no_std_crates[@]}"
```

stegaBOB: Done :smiling_imp: 

also added the alloc feature too!

address/Cargo.toml
febo: Agree, I think we should enable `alloc` when `std` is enabled – feels like it would be the expected behaviour since the std includes alloc.

joncinque:
> curious about the practice of Cargo formatting in this repo - there are a few lines that seem to be changed without reason. Is this not formatted as part of CI somehow? seems like different formatter settings could cause a lot of annoying churn if not.

We run `cargo sort` on the workspace, but that only sorts dependencies, not feature declarations within dependencies. We'd need some separate tool to do that, and I haven't looked into it

stegaBOB:
> curious about the practice of Cargo formatting in this repo - there are a few lines that seem to be changed without reason. Is this not formatted as part of CI somehow? seems like different formatter settings could cause a lot of annoying churn if not.

This was annoying me as well. I'm pretty sure I'm just using the defaults in vscode, so not super sure why it was messing things up

PR #357, opened 2025-10-08
vote-interface: add read api

buffalojoec:
Adds a read-only API that covers `VoteStateVersions`, `VoteStateV4` and `VoteStateV3`.

### DIFF
--- a/vote-interface/src/state/vote_state_read.rs
+++ b/vote-interface/src/state/vote_state_read.rs
@@ -1,0 +1,44 @@
+//! Vote state reader API.
+
+use {
+    super::{AuthorizedVoters, BlockTimestamp, LandedVote},
+    solana_clock::{Epoch, Slot},
+    solana_pubkey::Pubkey,
+    std::collections::VecDeque,
+};
+
+/// Read-only trait for accessing vote state data.
+pub trait VoteStateRead {
+    /// Returns the authorized voters.
+    fn authorized_voters(&self) -> &AuthorizedVoters;
+
+    /// Returns the authorized withdrawer.
+    fn authorized_withdrawer(&self) -> &Pubkey;
+
+    /// Returns the total credits.
+    fn credits(&self) -> u64 {
+        if self.epoch_credits().is_empty() {
+            0
+        } else {
+            self.epoch_credits().last().unwrap().1
+        }
+    }
+
+    /// Returns the epoch credits.
+    fn epoch_credits(&self) -> &Vec<(Epoch, u64, u64)>;
+
+    /// Returns the inflation rewards commission in basis points.
+    fn inflation_rewards_commission_bps(&self) -> u16;
+
+    /// Returns the last timestamp.
+    fn last_timestamp(&self) -> &BlockTimestamp;
+
+    /// Returns the node pubkey.
+    fn node_pubkey(&self) -> &Pubkey;
+
+    /// Returns the root slot.
+    fn root_slot(&self) -> Option<Slot>;
+
+    /// Returns the votes deque.
+    fn votes(&self) -> &VecDeque<LandedVote>;
+}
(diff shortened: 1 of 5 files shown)


vote-interface/src/state/vote_state_read.rs
buffalojoec, line 1: Each method here is used by some component in Agave, such as CLI, RPC, Core, or Stakes, as well as the Stake program.

vote-interface/src/state/vote_state_read.rs
```diff
@@ -0,0 +1,44 @@
...
+        } else {
+            self.epoch_credits().last().unwrap().1
+        }
+    }
+
+    /// Returns the epoch credits.
+    fn epoch_credits(&self) -> &Vec<(Epoch, u64, u64)>;
+
+    /// Returns the inflation rewards commission in basis points.
+    fn inflation_rewards_commission_bps(&self) -> u16;
```
buffalojoec, line 27: I went with "inflation rewards" instead of "commission", since it's a more accurate name for even the legacy field, and basis points rather than percentage to avoid a future breaking change. It's also easier to test with v4 in Agave if we can get the value in basis points, even though the program floors to the nearest percentage anyway.

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -192,6 +192,89 @@ impl VoteStateVersions {
     }
 }
 
+impl VoteStateRead for VoteStateVersions {
+    fn authorized_voters(&self) -> &AuthorizedVoters {
+        match self {
+            VoteStateVersions::V0_23_5(_state) => {
+                // V0_23_5 does not have AuthorizedVoters struct.
+                unimplemented!("V0_23_5 does not support authorized_voters.")
+            }
```
buffalojoec: Obviously this is a bit problematic. It has an `authorized_voter` field, but it doesn't allow us to return a reference to an actual `AuthorizedVoters` type.

We have a few options:
1. Refactor this method and use a trait-based accessor for querying authorized voters. Many callsites in Agave only care about the length, but a few do actually want an iterator over voters or some way to query an authorized voter by epoch.
2. Either panic here or make the function fallible (ie. `Result<&AuthorizedVoters, InstructionError>`.
3. Remove `V0_23_5` from `VoteStateVersions`.

The last one is probably the best call IMO, since it's not used anywhere in Agave and we know there are no v1 vote accounts on-chain anymore.

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -192,6 +192,89 @@ impl VoteStateVersions {
...
+                unimplemented!("V0_23_5 does not support votes.")
+            }
+            VoteStateVersions::V1_14_11(_state) => {
+                // V1_14_11 uses VecDeque<Lockout>, not VecDeque<LandedVote>.
+                unimplemented!("V1_14_11 does not support votes.")
+            }
+            VoteStateVersions::V3(state) => state.votes(),
+            VoteStateVersions::V4(state) => state.votes(),
+        }
+    }
```
buffalojoec: Here we can probably remedy the variance in `votes` types between older vote states and newer ones by making this function return an iterator over `LandedVote` items. However, a generic iterator doesn't guarantee any dequeue behavior. The program would still need access to `&VecDequeue<LandedVote>`, but that can live under the `VoteStateHandle` trait if need be.

vote-interface/src/state/vote_state_versions.rs
jstarry: We will still probably want to keep an enum variant in `VoteStateVersions` for deserializing `Uninitialized` state. Maybe we add a separate enum for vote state versions which excludes v1 and only includes the versions which have authorized voters

vote-interface/src/state/vote_state_versions.rs
jstarry: do we need `VecDeque` in the read api?

buffalojoec:
@jstarry I actually think - if we're going to go this route - the best move is to just only include the ones that are straightforward for now, and see if we can live without `authorized_voters` and `votes` on the Agave side.

Adding new methods is technically breaking, but this change right now isn't.

PR #361, opened 2025-10-09
allow with_seed functions to accept any bytes, not just UTF-8

kevinheavey:
The `with_seed` functions (eg `Address::create_with_seed`) and `WithSeed` instruction variants (eg `SystemInstruction::CreateAccountWithSeed`) all expect `&str` or `String`. But they don't do anything that requires UTF-8, and they just convert these values back to bytes.

Two reasons this is bad:

1. Bincode wastes time on UTF-8 validation
2. It would be nice to use arbitrary bytes (such as an `Address`) as a seed

Breaking changes:

- `Address::create_with_seed` is technically a breaking change. Changing the type to `&(impl AsRef<[u8]> + ?Sized),` means it works as before for most people, but breaks for anyone passing `create_with_seed` itself to a function that expected the `seed` param to be `&str`. It's possible that nobody is doing this.
- system-interface, vote-interface and the wasm helpers just change to `&[u8]` or `Vec<u8>`

In Agave there would be some small changes to system_processor.rs and vote_processor.rs. Firedancer would need to remove the UTF-8 check. Does this need a SIMD?

### DIFF
--- a/system-interface/src/instruction.rs
+++ b/system-interface/src/instruction.rs
@@ -75,7 +75,7 @@ const NONCE_STATE_SIZE: usize = 80;
 /// An instruction to the system program.
 #[cfg_attr(
     feature = "frozen-abi",
-    solana_frozen_abi_macro::frozen_abi(digest = "CBvp4X1gf36kwDqnprAa6MpKckptiAHfXSxFRHFnNRVw"),
+    solana_frozen_abi_macro::frozen_abi(digest = "7Hc7pX4JYnEsBEyixxKP76aDSq3668bLUc1V7L9Vqm68"),
     derive(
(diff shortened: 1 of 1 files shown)


t-nelson:
i don't think there is any rationale for enforcing utf8 here

> Does this need a SIMD?

yes. during adoption of the change, upgraded vs not nodes would have differing execution results for non-utf8 inputs

kevinheavey:
Pre-SIMD discussion here https://github.com/solana-foundation/solana-improvement-documents/discussions/378

kevinheavey:
SIMD: https://github.com/solana-foundation/solana-improvement-documents/pull/390

PR #363, opened 2025-10-09
system-interface: Make it no-std compatible

joncinque:
#### Problem

We're about to publish a breaking change to system-interface, so why not have a little fun and make it no-std beforehand!

#### Summary of changes

Add alloc feature for vec / string and std feature for frozen-abi. There's not a ton you can do with the crate currently since the instructions require bincode / alloc, but we should be able to trim this more in the future with pinocchio-compatible types.

While testing this, I noticed that the frozen-abi tests didn't work for no-std crates, which will likely become a problem in the future, so I fixed the macro to properly emit eprintln rather than through the hacky way it's done now.

### DIFF
--- a/frozen-abi-macro/src/lib.rs
+++ b/frozen-abi-macro/src/lib.rs
@@ -279,8 +279,6 @@ fn quote_for_test(
     type_name: &Ident,
     expected_digest: &str,
 ) -> TokenStream2 {
-    // escape from nits.sh...
-    let p = Ident::new(&("ep".to_owned() + "rintln"), Span::call_site());
     quote! {
         #[cfg(test)]
         mod #test_mod_ident {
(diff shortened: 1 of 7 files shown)


system-interface/src/instruction.rs
```diff
@@ -85,6 +87,7 @@ const NONCE_STATE_SIZE: usize = 80;
     feature = "serde",
     derive(serde_derive::Deserialize, serde_derive::Serialize)
 )]
+#[cfg(feature = "alloc")]
```
joncinque, line 90: This is the breaking change, since `SystemInstruction` isn't available without `alloc` enabled.

frozen-abi-macro/src/lib.rs
```diff
@@ -302,7 +300,7 @@ fn quote_for_test(
                 let actual_digest = ::std::format!("{}", hash);
                 if ::std::env::var("SOLANA_ABI_BULK_UPDATE").is_ok() {
                     if #expected_digest != actual_digest {
-                        #p!("sed -i -e 's/{}/{}/g' $(git grep --files-with-matches frozen_abi)", #expected_digest, hash);
+                        ::std::eprintln!("sed -i -e 's/{}/{}/g' $(git grep --files-with-matches frozen_abi)", #expected_digest, hash);
```
rustopian, line 303: If we want to keep this file evaluated by nits, we could consider directly expanding `eprintln!` to smth like `::std::io::stderr().write_fmt()`

...less hacky than `p` was at least:)

Unless you argue this violates the contract of membership in print_free, that is.

joncinque: There's enough poorly understood code in frozen-abi, so I prefer to keep things simpler with `eprintln`, but I imagine we'll have to use that trick in the future, thanks!

Side note: I spent almost an hour yesterday trying to figure out where `eprintln` was being called from until I stumbled on `p`. `p` was the bane of my evening.

PR #364, opened 2025-10-10
Unify syscall definitions

febo:
### Problem

While most syscall definitions are found in `define-syscall`, there are two that are defined in the crate where they are used to avoid bringing dependencies. This is not ideal when other crates need access to these syscalls.

### Solution

Move the syscall definitions to `define-syscall` crate and use "generic" pointers.

### DIFF
--- a/cpi/src/lib.rs
+++ b/cpi/src/lib.rs
@@ -382,7 +382,7 @@ pub fn get_return_data() -> Option<(Pubkey, Vec<u8>)> {
         let mut program_id = Pubkey::default();
 
         let size = unsafe {
-            crate::syscalls::sol_get_return_data(
+            solana_define_syscall::definitions::sol_get_return_data(
                 buf.as_mut_ptr(),
                 buf.len() as u64,
                 &mut program_id as *mut _ as *mut u8,
--- a/cpi/src/syscalls.rs
+++ b/cpi/src/syscalls.rs
@@ -1,4 +1,13 @@
 /// Syscall definitions used by `solana_cpi`.
 pub use solana_define_syscall::definitions::{
-    sol_get_return_data, sol_invoke_signed_c, sol_invoke_signed_rust, sol_set_return_data,
+    sol_invoke_signed_c, sol_invoke_signed_rust, sol_set_return_data,
 };
+use solana_pubkey::Pubkey;
+
+#[deprecated(
+    since = "3.1.0",
+    note = "Use `solana_define_syscall::definitions::sol_get_return_data` instead"
+)]
+pub unsafe fn sol_get_return_data(data: *mut u8, length: u64, program_id: *mut Pubkey) -> u64 {
+    solana_define_syscall::definitions::sol_get_return_data(data, length, program_id as *mut u8)
+}
--- a/instruction/src/syscalls.rs
+++ b/instruction/src/syscalls.rs
@@ -3,12 +3,32 @@ use crate::Instruction;
 #[cfg(target_os = "solana")]
 pub use {
     crate::{AccountMeta, ProcessedSiblingInstruction},
-    solana_define_syscall::{
-        define_syscall,
-        definitions::{sol_get_processed_sibling_instruction, sol_get_stack_height},
-    },
+    solana_define_syscall::definitions::sol_get_stack_height,
     solana_pubkey::Pubkey,
 };
+
+#[cfg(target_os = "solana")]
+#[deprecated(
+    since = "3.1.0",
+    note = "Use `solana_define_syscall::definitions::get_processed_sibling_instruction` instead"
+)]
+pub unsafe fn sol_get_processed_sibling_instruction(
+    index: u64,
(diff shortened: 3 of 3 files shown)


cpi/src/syscalls.rs
```diff
@@ -1,7 +1,4 @@
 /// Syscall definitions used by `solana_cpi`.
 pub use solana_define_syscall::definitions::{
-    sol_invoke_signed_c, sol_invoke_signed_rust, sol_set_return_data,
+    sol_get_return_data, sol_invoke_signed_c, sol_invoke_signed_rust, sol_set_return_data,
 };
-use {solana_define_syscall::define_syscall, solana_pubkey::Pubkey};
-
-define_syscall!(fn sol_get_return_data(data: *mut u8, length: u64, program_id: *mut Pubkey) -> u64);
```
febo, line 7: To avoid a breaking change, we could consider just deprecating this definition.

instruction/src/syscalls.rs
```diff
@@ -3,13 +3,13 @@ use crate::Instruction;
...
-    solana_define_syscall::{define_syscall, definitions::sol_get_stack_height},
+    solana_define_syscall::{
+        define_syscall,
+        definitions::{sol_get_processed_sibling_instruction, sol_get_stack_height},
+    },
     solana_pubkey::Pubkey,
 };
 
-#[cfg(target_os = "solana")]
-define_syscall!(fn sol_get_processed_sibling_instruction(index: u64, meta: *mut ProcessedSiblingInstruction, program_id: *mut Pubkey, data: *mut u8, accounts: *mut AccountMeta) -> u64);
```
febo, line 11: Same thing here, we could consider just deprecating this definition instead.

febo:
> Looks great to me! We're making a breaking change here anyway with #358, so let's just break the interface for these other syscalls too.
> 
> Also, since it's only function definitions, it shouldn't be too annoying for downstream users to deal with the change.

For the `cpi` crate it is straightforward, there are no types there. For the `instruction` crate, we will need the semver trick since there are types there (e.g., `Instruction`, `AccountMeta`).

Another option is to write a wrapper function with the current signature over the new syscall definition.
```rust
pub fn sol_get_processed_sibling_instruction(
   index: u64,
   meta: *mut ProcessedSiblingInstruction,
   program_id: *mut Pubkey,
   data: *mut u8,
   accounts: *mut AccountMeta
) -> u64 {
   solana_define_syscall::definitions::sol_get_processed_sibling_instruction(
      index,
      meta as *mut u8,
      program_id as *mut u8,
      data,
      account_meta as *mut u8,
   )
}
```

joncinque:
> For the instruction crate, we will need the semver trick since there are types there (e.g., Instruction, AccountMeta). Another option is to write a wrapper function with the current signature over the new syscall definition.

Oh right, I forgot about that redefinition. Let's do a wrapping function now then, and deprecate it. I don't think anyone downstream is using it directly, but better safe than sorry.

cpi/src/lib.rs
```diff
@@ -385,7 +385,7 @@ pub fn get_return_data() -> Option<(Pubkey, Vec<u8>)> {
             crate::syscalls::sol_get_return_data(
                 buf.as_mut_ptr(),
                 buf.len() as u64,
-                &mut program_id,
+                &mut program_id as *mut _ as *mut u8,
```
LucasSte, line 388: Do you actually need two casts?

define-syscall/src/definitions.rs
```diff
@@ -33,6 +33,21 @@ define_syscall!(fn sol_get_sysvar(sysvar_id_addr: *const u8, result: *mut u8, of
...
+// - `data_len` (`u64`): length of the instruction data
+// - `accounts_len` (`u64`): number of `AccountMeta` entries
+//
+// `program_id` points to an `Address` (`[u8; 32]`).
+//
+// `accounts` is a pointer to an array of `AccountMeta`, each containing:
+// - `address` (`[u8; 32]`): account address
+// - `is_signer` (`u8`): `true` if the instruction requires a signature
+// - `is_writable` (`u8`): `true` if the instruction requires the account to be writable
+define_syscall!(fn sol_get_processed_sibling_instruction(index: u64, meta: *mut u8, program_id: *mut u8, data: *mut u8, accounts: *mut u8) -> u64);
```
LucasSte, line 49: I know this is just a redefinition, but it looks unsafe to pass an array without telling the length. There is nothing to do about this, though.

cpi/src/lib.rs
febo: Apparently yes, you can't go from `&Pubkey` to `*mut u8` directly.

define-syscall/src/definitions.rs
febo: Yeah, agree. At least all syscalls are marked `unsafe`. 😅

febo:
Added deprecated wrapping functions and removed the "breaking" label.

PR #373, opened 2025-10-12
refactor: `bls-signatures`: use `ExactSizedIterators` to replace more slices with iterators

akhi3030:
Similar work as #371.

PR #374, opened 2025-10-12
refactor: `bls-signatures`: remove some slices for parallel iterators

akhi3030:
Updates some of the parallel APIs to use iterators where possible.

Work related to #373, #372, and #371

### DIFF
--- a/bls-signatures/src/pubkey.rs
+++ b/bls-signatures/src/pubkey.rs
@@ -148,23 +148,14 @@ impl PubkeyProjective {
     pub fn par_aggregate<'a, P: AsPubkeyProjective + Sync + 'a>(
         pubkeys: impl ParallelIterator<Item = &'a P>,
     ) -> Result<PubkeyProjective, BlsError> {
-        // cnt is used to detect empty iterator which should return an error
-        let (cnt, res) = pubkeys
+        pubkeys
             .into_par_iter()
-            .map(|key| key.try_as_projective().map(|r| (1, r)))
-            .reduce(
-                || Ok((0, PubkeyProjective::identity())),
-                |a, b| {
-                    let (a_cnt, mut a) = a?;
-                    let (b_cnt, b) = b?;
-                    a.0 += &b.0;
-                    Ok((a_cnt + b_cnt, a))
-                },
-            )?;
-        if cnt == 0 {
-            return Err(BlsError::EmptyAggregation);
-        }
-        Ok(res)
+            .map(|key| key.try_as_projective())
+            .try_reduce_with(|mut a, b| {
+                a.0 += b.0;
+                Ok(a)
+            })
+            .ok_or(BlsError::EmptyAggregation)?
     }
 }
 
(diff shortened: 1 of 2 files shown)


bls-signatures/src/pubkey.rs
```diff
@@ -148,24 +145,26 @@ impl PubkeyProjective {
...
-    pub fn par_aggregate<P: AsPubkeyProjective + Sync>(
-        pubkeys: &[&P],
+    pub fn par_aggregate<'a, P: AsPubkeyProjective + Sync + 'a>(
+        pubkeys: impl ParallelIterator<Item = &'a P>,
     ) -> Result<PubkeyProjective, BlsError> {
-        if pubkeys.is_empty() {
-            return Err(BlsError::EmptyAggregation);
-        }
-        pubkeys
+        // cnt is used to detect empty iterator which should return an error
```
akhi3030: this ended up being fairly ugly.  I couldn't come up with a cleaner way to check for empty iterators.  There doesn't seem to be a parallel equivalent of `ExactSizedIterator` which exposes a `len()` function.

akhi3030: A simpler check would be to check if the result is equal to `identity()` and then assume that that could've only happened if the length of the iterator was 0.  Not sure how safe that is though.

Alternative is to not handle error in the API and let the caller worry about the empty case?

bls-signatures/src/signature.rs
```diff
@@ -205,40 +205,42 @@ impl SignatureProjective {
...
-    pub fn par_aggregate<S: AsSignatureProjective + Sync>(
-        signatures: &[&S],
+    pub fn par_aggregate<'a, S: AsSignatureProjective + Sync + 'a>(
+        signatures: impl ParallelIterator<Item = &'a S>,
     ) -> Result<SignatureProjective, BlsError> {
-        if signatures.is_empty() {
-            return Err(BlsError::EmptyAggregation);
-        }
-        signatures
+        // cnt is used to detect empty iterator which should return an error
```
akhi3030: similar issues as discussed above.

bls-signatures/src/pubkey.rs
samkim-crypto: Hm, yeah this is a little tricky. Have you tried using `try_reduce_with`? It seems to return `Result<Option<T>, E>` for handling fallible reductions on potentially empty iterators.

```
pub fn par_aggregate<'a, P: AsPubkeyProjective + Sync + 'a>(
    pubkeys: impl ParallelIterator<Item = &'a P>,
) -> Result<PubkeyProjective, BlsError> {
    pubkeys
        .map(|key| key.try_as_projective())
        .try_reduce_with(|mut a, b| {
            a.0 += &b.0;
            Ok(a)
        })? // This should evaluate to Result<Option<PubkeyProjective>, BlsError>
        .ok_or(BlsError::EmptyAggregation) // Maps None to the desired error
}
```

akhi3030: Did not know about `try_reduce_with`!  Absolutely brilliant.  Thanks!

PR #379, opened 2025-10-14
hash: Make crate no-std with new "atomic" feature

joncinque:
#### Problem

The solana-hash crate should be compatible in a no-std environment, but the `new_unique` function makes that impossible.

#### Summary of changes

Featurize the function on the "atomic" feature, similar to solana-address. Enable the feature as needed.

Closes #323

### DIFF
--- a/transaction/Cargo.toml
+++ b/transaction/Cargo.toml
@@ -22,7 +22,7 @@ bincode = [
     "solana-message/bincode",
 ]
 blake3 = ["bincode", "solana-message/blake3"]
-dev-context-only-utils = ["blake3", "serde", "verify"]
+dev-context-only-utils = ["blake3", "serde", "verify", "solana-hash/atomic"]
 frozen-abi = [
     "dep:solana-frozen-abi",
     "dep:solana-frozen-abi-macro",


apfitzge:
https://github.com/anza-xyz/solana-sdk/pull/355/files#diff-57c9d33183d2544f785f029e6400b88a87cf3e5a79378a847fec72696a010c29R7

This above linked PR uses `core` instead of `std`. Wouldn't that change be sufficient to make this no_std without adding the `atomic` feature?
The cfg for non-64 ptr width seemed to use parking_lot instead of std as well, not sure if that is no_std compatiable.

joncinque:
It doesn't work unfortunately. Here's the error you get:

```
error[E0599]: no method named `fetch_add` found for struct `core::sync::atomic::AtomicU64` in the current scope
  --> atomic-u64/src/lib.rs:16:20
   |
16 |             self.0.fetch_add(v, atomic::Ordering::Relaxed)
   |                    ^^^^^^^^^ method not found in `core::sync::atomic::AtomicU64`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `solana-atomic-u64` (lib) due to 1 previous error
```

You need additional features available on the target in order to use certain operations. I didn't dig in to find out exactly which operations we can and can't use, since using atomics is probably a bad idea for on-chain programs in general

apfitzge:
> You need additional features available on the target in order to use certain operations. I didn't dig in to find out exactly which operations we can and can't use, since using atomics is probably a bad idea for on-chain programs in general

Yeah I think it probably makes sense for on-chain places where atomics do not exist. I think I was probably hung up on the framing in the description as

> should be compatible in a no-std environment

which tmk is less strict than what we actually want/need here.

PR #380, opened 2025-10-14
hash: Make dependencies optional

febo:
### Problem

It is not possible to use the `Hash` type without bringing dependencies.

### Solution

This PR creates two features to make all dependencies optional:
* `decode`: enables `five8`
* `sanitize`: enables `solana-sanitize`

### DIFF
--- a/hash/src/lib.rs
+++ b/hash/src/lib.rs
@@ -43,6 +43,7 @@ pub const MAX_BASE58_LEN: usize = 44;
 #[cfg_attr(feature = "bytemuck", derive(Pod, Zeroable))]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize,))]
 #[cfg_attr(feature = "copy", derive(Copy))]
+#[cfg_attr(not(feature = "decode"), derive(Debug))]
 #[derive(Clone, Default, Eq, PartialEq, Ord, PartialOrd, Hash)]
 #[repr(transparent)]
 pub struct Hash(pub(crate) [u8; HASH_BYTES]);


hash/src/lib.rs
```diff
@@ -46,6 +46,7 @@ pub const MAX_BASE58_LEN: usize = 44;
 #[repr(transparent)]
 pub struct Hash(pub(crate) [u8; HASH_BYTES]);
```
joncinque: Similar to address, can this implement the default `Debug` if `decode` isn't enabled?

febo: Good catch!

febo:
> Looks good to me!

Great – waiting on #348 so we can rebase.

hash/src/lib.rs
```diff
@@ -43,10 +43,12 @@ pub const MAX_BASE58_LEN: usize = 44;
 #[cfg_attr(feature = "bytemuck", derive(Pod, Zeroable))]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize,))]
 #[cfg_attr(feature = "copy", derive(Copy))]
+#[cfg_attr(not(feature = "decode"), derive(Debug))]
 #[derive(Clone, Default, Eq, PartialEq, Ord, PartialOrd, Hash)]
 #[repr(transparent)]
 pub struct Hash(pub(crate) [u8; HASH_BYTES]);
 
+#[cfg(feature = "sanitize")]
 impl Sanitize for Hash {}
```
apfitzge, line 52: it would be great if we could just get rid of `Sanitize` entirely.
This feels so dumb

PR #381, opened 2025-10-15
BREAKING: human readable de/ser for versioned message

jstarry:
Versioned messages are not properly supported for human readable de/ser due to special handling of the version prefix byte. This is a breaking change to how versioned messages (including v0 and legacy) are serialized and deserialized when using a human readable serialization format. More context about the consequences of this lack of support can be found in https://github.com/anza-xyz/solana-sdk/pull/311

Old serialization:

```json
[
  128,
  {
    "header": {
      "numRequiredSignatures": 0,
      "numReadonlySignedAccounts": 0,
      "numReadonlyUnsignedAccounts": 0
    },
    "accountKeys": [..],
    "recentBlockhash": [..],
    [..]
  }
]
```

New serialization:
```json
{                                                                                                                                                                                                          
  "version": "0",                                                                                                                                                                                          
  "message": {                                                                                                                                                                                             
    "header": {                                                                                                                                                                                            
      "numRequiredSignatures": 0,                                                                                                                                                                          
      "numReadonlySignedAccounts": 0,                                                                                                                                                                      
      "numReadonlyUnsignedAccounts": 0                                                                                                                                                                     
    },                                                                                                                                                                                                     
    "accountKeys": [..]              
    "recentBlockhash": [..],
    [..]
  }
}
```

### DIFF
--- a/message/src/versions/mod.rs
+++ b/message/src/versions/mod.rs
@@ -391,11 +391,14 @@ impl<'de> serde::Deserialize<'de> for VersionedMessage {
                 Legacy { message: LegacyMessage },
                 #[serde(rename = "0")]
                 V0 { message: v0::Message },
+                #[serde(rename = "1")]
+                V1 { message: v1::Message },
             }
 
             match HR::deserialize(deserializer)? {
                 HR::Legacy { message } => Ok(VersionedMessage::Legacy(message)),
                 HR::V0 { message } => Ok(VersionedMessage::V0(message)),
+                HR::V1 { message } => Ok(VersionedMessage::V1(message)),
             }
         } else {
             deserializer.deserialize_tuple(2, MessageVisitor)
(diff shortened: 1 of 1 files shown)


apfitzge:
@jstarry can you add a description to the PR?

jstarry:
Thanks for the review @apfitzge! Sorry I was in the middle of updating the description but got side tracked while looking into another human readable serialization issue. ~~I was thinking of marking this as breaking because any v0 messages serialized in the original compact format will not be deserializable. But since legacy messages were not deserializable I tend to think of this as a bug fix rather than a breaking change. Thoughts?~~

EDIT: Just conservatively marked this as breaking because my other change will definitely breaking

joncinque:
Do we want to land this? We're about to publish a breaking change to solana-message for wincode and txv1, so now's a great time!

jstarry:
@joncinque I think it's still a desirable change, I've rebased and fixed the conflicts

message/src/versions/mod.rs
```diff
@@ -364,7 +383,23 @@ impl<'de> serde::Deserialize<'de> for VersionedMessage {
...
 
-        deserializer.deserialize_tuple(2, MessageVisitor)
+        if deserializer.is_human_readable() {
+            #[derive(Deserialize)]
+            #[serde(tag = "version")]
+            enum HR {
+                #[serde(rename = "legacy")]
+                Legacy { message: LegacyMessage },
+                #[serde(rename = "0")]
+                V0 { message: v0::Message },
```
apfitzge, line 393: need a V1 now I think?

jstarry: Ah sorry, will add that and a test

jstarry: Fixed

PR #382, opened 2025-10-16
BREAKING: human readable de/ser for shortvec

jstarry:
Short vec fields are not properly supported for human readable de/ser. This is a breaking change to how short vec fields are serialized and deserialized for human readable formats.

Old serialization of `ShortVec(vec![0, 1, 2])`:

```json
[
  [3],
  0,
  1,
  2
]
```

New serialization:
```json
[
  0,
  1,
  2
]
```


jstarry:
Depends on https://github.com/anza-xyz/solana-sdk/pull/381 (to avoid abi digest conflicts) so wait on reviewing this until that one is in

PR #383, opened 2025-10-16
vote-interface: add fast deserialize for `VoteStateVersions`

buffalojoec:
#### Problem
So far, we've mostly been serializing `VoteStateVersions` with `bincode`, or using the optimized path on either recent vote state (`VoteStateV3::deserialize` or `VoteStateV4::deserialize`). However, the former is not performant enough and the latter will coerce the underlying type into the target version upon deserialization regardless of its actual serialized version.

The coercion of the underlying type is actually a problem and should be eliminated in future breaking changes to the vote interface. However, in the immediate term, we require an optimized path for deserializing `VoteStateVersions` in the Vote program without coercing the underlying type at all.

#### Summary of Changes
Similar to `VoteStateV3` and `VoteStateV4`, adds an optimized deserialization API for `VoteStateVersions`. Simply includes one more crate-private free function to deserialize into a `V1_14_11` vote state pointer and hooks up the deserializers to a new `VoteStateVersions::deserialize` function that will deserialize into `VoteStateVersions` as-is with no coercion.

I made the choice to disallow support for `V0_23_5`, since it's been entirely phased out on all networks, it's not supported by the Vote program, and it has never really had deserialization support in this library. The `V0_23_5` variant should be deprecated as well, which we can do in follow-up work.

### DIFF
--- a/vote-interface/src/state/vote_state_deserialize.rs
+++ b/vote-interface/src/state/vote_state_deserialize.rs
@@ -72,6 +72,56 @@ pub(crate) fn deserialize_into<T: Default>(
     res
 }
 
+pub(crate) fn deserialize_vote_state_into_v1_14_11(
+    cursor: &mut Cursor<&[u8]>,
+    vote_state: *mut crate::state::vote_state_1_14_11::VoteState1_14_11,
+) -> Result<(), InstructionError> {
+    // General safety note: we must use addr_of_mut! to access the `vote_state` fields as the value
+    // is assumed to be _uninitialized_, so creating references to the state or any of its inner
+    // fields is UB.
+
+    read_pubkey_into(
+        cursor,
+        // Safety: if vote_state is non-null, node_pubkey is guaranteed to be valid too
+        unsafe { addr_of_mut!((*vote_state).node_pubkey) },
+    )?;
+    read_pubkey_into(
+        cursor,
+        // Safety: if vote_state is non-null, authorized_withdrawer is guaranteed to be valid too
+        unsafe { addr_of_mut!((*vote_state).authorized_withdrawer) },
+    )?;
+    let commission = read_u8(cursor)?;
+    let votes = read_votes_as_lockouts(cursor)?; // `Vec<Lockout>`
+    let root_slot = read_option_u64(cursor)?;
+    let authorized_voters = read_authorized_voters(cursor)?;
+
+    read_prior_voters_into(
+        cursor,
+        // Safety: if vote_state is non-null, prior_voters is guaranteed to be valid too
+        unsafe { addr_of_mut!((*vote_state).prior_voters) },
+    )?;
+
+    let epoch_credits = read_epoch_credits(cursor)?;
+    let last_timestamp = read_last_timestamp(cursor)?;
+
+    // Safety: if vote_state is non-null, all the fields are guaranteed to be
+    // valid pointers.
+    //
+    // Heap allocated collections - votes, authorized_voters, prior_voters, and epoch_credits -
+    // are guaranteed not to leak after this point as the VoteState1_14_11 is fully
+    // initialized and will be regularly dropped.
+    unsafe {
+        addr_of_mut!((*vote_state).commission).write(commission);
+        addr_of_mut!((*vote_state).votes).write(votes);
+        addr_of_mut!((*vote_state).root_slot).write(root_slot);
+        addr_of_mut!((*vote_state).authorized_voters).write(authorized_voters);
+        addr_of_mut!((*vote_state).epoch_credits).write(epoch_credits);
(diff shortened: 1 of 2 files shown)


vote-interface/src/state/vote_state_versions.rs
```diff
@@ -190,6 +191,66 @@ impl VoteStateVersions {
...
+        use crate::state::vote_state_deserialize::{
+            deserialize_vote_state_into_v1_14_11, deserialize_vote_state_into_v3,
+            deserialize_vote_state_into_v4, SourceVersion,
+        };
+
+        let mut cursor = std::io::Cursor::new(input);
+
+        let variant = solana_serialize_utils::cursor::read_u32(&mut cursor)?;
+        match variant {
+            // V0_23_5 not supported.
```
jstarry, line 213: Let's return `InstructionError::UninitializedAccount` for this instead of `InvalidAccountData`

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -190,6 +191,66 @@ impl VoteStateVersions {
...
+            //
+            // V1_14_11
+            1 => {
+                let mut vote_state = Box::new(std::mem::MaybeUninit::uninit());
+                deserialize_vote_state_into_v1_14_11(&mut cursor, vote_state.as_mut_ptr())?;
+                let vote_state =
+                    unsafe { Box::from_raw(Box::into_raw(vote_state) as *mut VoteState1_14_11) };
```
jstarry, line 220: looks like you can do this without the into/from
```suggestion
               let vote_state = unsafe { vote_state.assume_init() };
```

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -190,6 +191,66 @@ impl VoteStateVersions {
...
+
+        let mut cursor = std::io::Cursor::new(input);
+
+        let variant = solana_serialize_utils::cursor::read_u32(&mut cursor)?;
+        match variant {
+            // V0_23_5 not supported.
+            //
+            // V1_14_11
+            1 => {
+                let mut vote_state = Box::new(std::mem::MaybeUninit::uninit());
```
jstarry: nit: can you import `MaybeUninit` locally as well?

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -190,6 +191,66 @@ impl VoteStateVersions {
...
+            }
+            _ => Err(InstructionError::InvalidAccountData),
+        }
+    }
+
+    /// Serializes the `VoteStateVersions` into the output buffer.
+    ///
+    /// This is a convenience wrapper around `bincode::serialize_into`.
+    #[cfg(feature = "bincode")]
+    pub fn serialize(&self, output: &mut [u8]) -> Result<(), InstructionError> {
```
jstarry: do we need this?

vote-interface/src/state/vote_state_deserialize.rs
```diff
@@ -72,6 +72,56 @@ pub(crate) fn deserialize_into<T: Default>(
...
+        // Safety: if vote_state is non-null, node_pubkey is guaranteed to be valid too
+        unsafe { addr_of_mut!((*vote_state).node_pubkey) },
+    )?;
+    read_pubkey_into(
+        cursor,
+        // Safety: if vote_state is non-null, authorized_withdrawer is guaranteed to be valid too
+        unsafe { addr_of_mut!((*vote_state).authorized_withdrawer) },
+    )?;
+    let commission = read_u8(cursor)?;
+    let votes = read_votes_as_lockouts(cursor)?; // `Vec<Lockout>`
```
jstarry, line 94: nit: remove this comment?

jstarry (commented):
Looks solid, just some small things

vote-interface/src/state/vote_state_versions.rs
buffalojoec: You can, but only for Rust 1.82 and above. I would have to reconfigure the MSRV checks on the workspace in order to allow `vote-interface` to move past everything else, or bump everything to 1.82.

vote-interface/src/state/vote_state_deserialize.rs
buffalojoec: This feels incredibly pedantic.

vote-interface/src/state/vote_state_versions.rs
buffalojoec: We can ride without it for now. If we need it later we can add it.

vote-interface/src/state/vote_state_deserialize.rs
jstarry: haha yeah but it's also not the correct type so it stood out :P

jstarry (approved):

PR #384, opened 2025-10-16
sha256-hasher: Make it no-std compatible

joncinque:
#### Problem

The sha256-hasher doesn't compile in real no-std environments.

#### Summary of changes

This was an easy fix, simply don't pull in default features for the sha2 crate. This means we disable OIDs for hashes, and the ability to finalize hashes into Box'ed values. Since we don't use that, it should be safe to disable.

Also, the crate was a no-op to build without the sha2 feature, so I changed the no-std checking strategy to *exclude* features rather than include certain features, and moved it to `cargo hack` so that each crate and feature is tested individually.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -571,5 +571,10 @@ jobs:
           cargo-cache-key: cargo-nightly-no-std
           cargo-cache-fallback-key: cargo-nightly
 
+      - name: Install cargo-hack
+        uses: taiki-e/cache-cargo-install-action@v2
+        with:
+          tool: cargo-hack
+
       - name: Run check
         run: ./scripts/check-no-std.sh


scripts/check-no-std.sh
```diff
@@ -31,14 +32,22 @@ no_std_crates=(
...
-./cargo nightly check -Zbuild-std=core \
+# These features require alloc
+exclude_features_no_alloc="alloc,borsh,curve25519,serde"
+# These features never work on upstream BPF
+exclude_features="atomic,bincode,default,dev-context-only-utils,frozen-abi,rand,std,verify"
+
+./cargo nightly hack check \
+  -Zbuild-std=core \
   "--target=$target" \
-  --no-default-features \
```
febo, line 36: Should we be building each individual feature with no defaults, so having the `--no-default-features` argument?

febo:
> Also, the crate was a no-op to build without the sha2 feature, so I changed the no-std checking strategy to _exclude_ features rather than include certain features, and moved it to `cargo hack` so that each crate and feature is tested individually.

Do we need to test the build with just `--no-default-features` as well?

scripts/check-no-std.sh
joncinque: Yep, that's included with `--each-feature` on cargo hack:

```
+ exec cargo +nightly-2025-06-29 hack check -Zbuild-std=core --target=bpfel-unknown-none --exclude-features=atomic,bincode,default,dev-context-only-utils,frozen-abi,rand,std,verify,alloc,borsh,curve25519,serde --each-feature -p solana-address -p solana-clock -p solana-commitment-config -p solana-define-syscall -p solana-epoch-info -p solana-fee-calculator -p solana-hash -p solana-msg -p solana-program-error -p solana-program-log -p solana-program-log-macro -p solana-program-memory -p solana-rent -p solana-sanitize -p solana-sdk-ids -p solana-sha256-hasher -p solana-signature -p solana-sysvar-id -p solana-system-interface
info: running `cargo check -Zbuild-std=core --target bpfel-unknown-none --no-default-features` on solana-address (1/32)
...
```

joncinque: And then it enables each feature one by one

febo: Perfect!

PR #386, opened 2025-10-17
Remove `to_account_shared_data` from ReadableAccount

LucasSte:
**Problem**

We are introducing a new account layout for the SVM. We'll also have "view" structs that do not own the fields for when programs borrow an account. Those structs implement `ReadableAccount`, but cannot be converted into `AccountSharedData` without losing references to the underlying data.

**Solution**

Remove the `to_account_shared_data` from `ReadableAccount`.

### DIFF
--- a/account/src/lib.rs
+++ b/account/src/lib.rs
@@ -227,17 +227,6 @@ pub trait ReadableAccount: Sized {
     fn owner(&self) -> &Pubkey;
     fn executable(&self) -> bool;
     fn rent_epoch(&self) -> Epoch;
-    #[deprecated(since = "3.2.0")]
-    fn to_account_shared_data(&self) -> AccountSharedData {
-        #[allow(deprecated)]
-        AccountSharedData::create(
-            self.lamports(),
-            self.data().to_vec(),
-            *self.owner(),
-            self.executable(),
-            self.rent_epoch(),
-        )
-    }
 }
 
 impl ReadableAccount for Account {
@@ -346,10 +335,6 @@ impl ReadableAccount for AccountSharedData {
     fn rent_epoch(&self) -> Epoch {
         self.rent_epoch
     }
-    fn to_account_shared_data(&self) -> AccountSharedData {
-        // avoid data copy here
-        self.clone()
-    }
 }
 
 impl ReadableAccount for Ref<'_, AccountSharedData> {
@@ -367,16 +352,6 @@ impl ReadableAccount for Ref<'_, AccountSharedData> {
     }
     fn rent_epoch(&self) -> Epoch {
         self.rent_epoch
-    }
-    fn to_account_shared_data(&self) -> AccountSharedData {
-        AccountSharedData {
-            lamports: self.lamports(),
-            // avoid data copy here
-            data: Arc::clone(&self.data),
-            owner: *self.owner(),
-            executable: self.executable(),
-            rent_epoch: self.rent_epoch(),
-        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


buffalojoec (commented):
@joncinque can weigh in, but I believe you should do one PR to deprecate it and then another to remove it with the "Breaking" label.

brooksprumo (commented):
If we remove this from the sdk, then agave won't be able to upgrade to this new version until we replace all the call sites.

What is the migration path for agave?

LucasSte:
> If we remove this from the sdk, then agave won't be able to upgrade to this new version until we replace all the call sites.
> 
> What is the migration path for agave?

Since the change is simple, I was thinking of bumping the version and upgrading the call sites in the same PR. I can do the latter first if that is what you prefer.

brooksprumo:
> Since the change is simple

I haven't looked at doing this myself, but I think it would be good to update the call sites first and bump the sdk second. This makes a (hopefully unnecessary) revert less painful (in case there are other PRs that build on the sdk bump).

LucasSte:
> What's the new method, and where will it live?

We don't want to allow the conversion to `AccountSharedData` inside runtime, since we are introducing a new account layout that implements `ReadableAccount`, but does not allow the conversion.

The idea I discussed with @brooksprumo was to have the new method internal to accounts-db.

> I agree that it's better to add the new method and deprecate the old method first, then remove the old method.

Sure, I'll create the new method and remove the old usages first.

brooksprumo:
Yeah, I think in accounts-db is the right place.

account/src/lib.rs
```diff
@@ -226,6 +226,7 @@ pub trait ReadableAccount: Sized {
     fn owner(&self) -> &Pubkey;
     fn executable(&self) -> bool;
     fn rent_epoch(&self) -> Epoch;
+<<<<<<< HEAD
```
Lichtso: unresolved merge conflict

LucasSte: Fixed

LucasSte:
I removed `to_account_shared_data` usages from the monorepo in https://github.com/anza-xyz/agave/pull/8668, deprecated the function in https://github.com/anza-xyz/solana-sdk/pull/407 and bumped the crate on the monorepo in https://github.com/anza-xyz/agave/pull/8754.

I believe this PR is now ready for review :)

joncinque (approved):
Looks good to me!

account/src/lib.rs
```diff
@@ -226,16 +226,6 @@ pub trait ReadableAccount: Sized {
...
-    #[deprecated(since = "3.2.0")]
-    fn to_account_shared_data(&self) -> AccountSharedData {
-        AccountSharedData::create(
-            self.lamports(),
-            self.data().to_vec(),
-            *self.owner(),
-            self.executable(),
-            self.rent_epoch(),
-        )
-    }
```
brooksprumo, line 240: Don't we have to bump the version to v4.0 before we can remove this?

https://github.com/anza-xyz/solana-sdk/blob/ea08ee9cb6c863f6fc498f913f6611fcb2735ec0/account/Cargo.toml#L5

LucasSte: Actually, I believe the Github release job automatically bumps the version when we are running it, so we usually don't update the version here.

@joncinque Did I get that right?

LucasSte:
Let's hold off merging this PR for a moment. We were thinking of removing `fn create` from `WritableAccount` as well. If we can achieve that, then we can make a single major release with all the breaking changes together.

account/src/lib.rs
joncinque: Yep that's correct!

LucasSte:
> Let's hold off merging this PR for a moment. We were thinking of removing `fn create` from `WritableAccount` as well. If we can achieve that, then we can make a single major release with all the breaking changes together.

Both `to_account_shared_data` and `fn create` have been deprecated. I'm thinking of merging this PR now (after resolving the conflicts) and opening a new one to remove `fn create` from `WritableAccount`.

Is everyone all good to merge this?

PR #387, opened 2025-10-17
define-syscall: Add bpf target

febo:
### Problem

### Solution

### DIFF
--- a/define-syscall/Cargo.toml
+++ b/define-syscall/Cargo.toml
@@ -13,6 +13,8 @@ edition = { workspace = true }
 targets = ["x86_64-unknown-linux-gnu"]
 
 [features]
+# static syscalls for eBPF are experimental and
+# not guaranteed to work
 unstable-static-syscalls = []
 
 [lints.rust]


vadorovsky (approved):

define-syscall/src/lib.rs
```diff
@@ -3,7 +3,7 @@
 
 pub mod definitions;
 
-#[cfg(target_feature = "static-syscalls")]
+#[cfg(any(target_feature = "static-syscalls", target_arch = "bpf"))]
```
joncinque: Sorry, just to make sure, but is this correct?

From what I understand, static-syscalls are only supported with sbfv3, which might not be enabled. The SIMD still hasn't been accepted: https://github.com/solana-foundation/solana-improvement-documents/pull/178

In that case, shouldn't we avoid using static syscalls for bpf for now and instead use the old model?

febo: Good point – we had an offline discussion about this. The use of static syscalls for sbpf v0 is unstable, since it was not designed to work with it. While the current implementation works, they have never been tested for v0.

To make this more explicitly, an `"unstable-static-syscalls`" feature was added to the crate, which needs to be enabled on `target_arch = "bpf"`.

define-syscall/Cargo.toml
```diff
@@ -12,6 +12,9 @@ edition = { workspace = true }
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
 
+[features]
+unstable-static-syscalls = []
+
```
LucasSte, line 19: I believe this feature is reasonable, because it conveys the message that it is usable, but we don't guarantee it to work.

To complement, we could write a short comment here saying "static syscalls for eBPF are experimental and not guaranteed to work".

febo: Done.

joncinque (approved):
Looks good to me!

PR #388, opened 2025-10-17
pubkey: Make no-std compatible by enabling atomic with std

joncinque:
#### Problem

The pubkey crate doesn't compile for no-std environments because the atomic feature is always enabled on address.

This is an alternative to #302.

#### Summary of changes

Add the "atomic" feature to the "std" feature in pubkey, and enable "std" as appropriate for certain tests.

We could do the same with "decode", but since it isn't required, I'm inclined to make a minimal change. People can switch to address for more control.


vadorovsky (approved):
Awesome, much cleaner than my solution!

PR #389, opened 2025-10-17
address: Add bpf target

febo:
### Problem

### Solution

### DIFF
--- a/address/Cargo.toml
+++ b/address/Cargo.toml
@@ -54,13 +54,13 @@ solana-frozen-abi-macro = { workspace = true, features = ["frozen-abi"], optiona
 solana-program-error = { workspace = true, optional = true }
 solana-sanitize = { workspace = true, optional = true }
 
-[target.'cfg(not(any(target_os = "solana", target_arch = "bpf")))'.dependencies]
-curve25519-dalek = { workspace = true, optional = true }
-solana-sha256-hasher = { workspace = true, features = ["sha2"], optional = true }
-
 [target.'cfg(any(target_os = "solana", target_arch = "bpf"))'.dependencies]
 solana-define-syscall = { workspace = true, optional = true }
 solana-sha256-hasher = { workspace = true, optional = true }
+
+[target.'cfg(not(any(target_os = "solana", target_arch = "bpf")))'.dependencies]
+curve25519-dalek = { workspace = true, optional = true }
+solana-sha256-hasher = { workspace = true, features = ["sha2"], optional = true }
 
 [dev-dependencies]
 anyhow = { workspace = true }


febo:
Needs #387 to go first.

address/Cargo.toml
```diff
@@ -20,7 +20,7 @@ atomic = ["dep:solana-atomic-u64"]
 borsh = ["dep:borsh", "alloc"]
 bytemuck = ["copy", "dep:bytemuck", "dep:bytemuck_derive"]
 copy = []
-curve25519 = ["dep:curve25519-dalek", "error", "sha2"]
+curve25519 = ["dep:curve25519-dalek", "sha2", "syscalls"]
```
febo, line 23: Enabling feature "curve25519" will enable the `syscalls` module, which will try to use solana-define-syscall dependency when `target_os = "solana"` or `target_arch = "bpf"`, breaking the build. The fix here enables the "syscalls" feature so the dependency is found; it has no effect on other targets, it will only bring the required "error" feature.

Alternatively, we could improve the separation between the "curve25519" and "syscalls" features, but this does not seem necessary at the moment.

address/src/lib.rs
```diff
@@ -57,7 +57,12 @@ pub const MAX_SEEDS: usize = 16;
 #[cfg(feature = "decode")]
 /// Maximum string length of a base58 encoded address.
 const MAX_BASE58_LEN: usize = 44;
+
+/// Marker used to find program derived addresses (PDAs).
+#[cfg(target_arch = "bpf")]
+pub static PDA_MARKER: &[u8; 21] = b"ProgramDerivedAddress";
```
febo, line 63: `target_arch = "bpf"` only supports scalar constants, so we need to switch this to be `static` instead.

address/src/lib.rs
```diff
@@ -411,8 +416,56 @@ macro_rules! declare_id {
...
+///
+/// let my_id = Address::from_str("My11111111111111111111111111111111111111111").unwrap();
+/// assert_eq!(id(), my_id);
+/// ```
+#[cfg(all(feature = "decode", target_arch = "bpf"))]
+#[macro_export]
+macro_rules! declare_id {
+    ($address:expr) => {
+        /// The const program ID.
+        pub static ID: $crate::Address = $crate::Address::from_str_const($address);
```
febo, line 401: Same here, needs to be `static` for `target_arch = "bpf"`.

address/src/lib.rs
joncinque: Makes sense!

address/src/lib.rs
joncinque: The duplication is a bit annoying. I played around with a bit, and we *could* avoid the duplication with something like
```suggestion
        #[cfg(target_arch = "bpf")]
        pub static ID: $crate::Address = $crate::Address::from_str_const($address);
        #[cfg(not(target_arch = "bpf"))]
        pub const ID: $crate::Address = $crate::Address::from_str_const($address);
```

address/src/lib.rs
```diff
@@ -411,8 +416,56 @@ macro_rules! declare_id {
...
+        // TODO make this const once `derive_const` makes it out of nightly
+        // and we can `derive_const(PartialEq)` on `Address`.
+        pub fn check_id(id: &$crate::Address) -> bool {
+            id == &ID
+        }
+
+        /// Returns the ID.
+        pub const fn id() -> $crate::Address {
+            $crate::Address::from_str_const($address)
+        }
```
joncinque, line 418: If we decide to avoid the duplication, this also needs to change

address/Cargo.toml
```diff
@@ -54,10 +54,14 @@ solana-frozen-abi-macro = { workspace = true, features = ["frozen-abi"], optiona
...
-[target.'cfg(not(target_os = "solana"))'.dependencies]
+[target.'cfg(not(any(target_os = "solana", target_arch = "bpf")))'.dependencies]
 curve25519-dalek = { workspace = true, optional = true }
 solana-sha256-hasher = { workspace = true, features = ["sha2"], optional = true }
 
+[target.'cfg(target_arch = "bpf")'.dependencies]
+solana-define-syscall = { workspace = true, optional = true }
+solana-sha256-hasher = { workspace = true, optional = true }
+
 [target.'cfg(target_os = "solana")'.dependencies]
```
joncinque: Rather than copying, can we just do this?
```suggestion
[target.'cfg(any(target_os = "solana", target_arch = "bpf"))'.dependencies]
```

address/Cargo.toml
joncinque: That all makes sense, but how does this work right now?

febo: It does not work – I am not sure why CI has not picked this up.

image

address/src/lib.rs
febo: Good point – it looks better with less duplication.

address/src/lib.rs
febo: Done.

address/Cargo.toml
joncinque: Oh actually this makes sense -- we check `cargo-build-sbf` without enabling features, so we just never caught this

PR #390, opened 2025-10-17
account-view: Add new crate

febo:
### Problem

Currently, Pinocchio uses its own zero-copy representation of `Account` and `AccountInfo`, which saves CUs on entrypoint definitions. By using the same name to represent an `AccountInfo`, it is cumbersome to write code that uses both types at the same time.

These types are useful for other entrypoint definitions and they will have greater visibility if they are added to the SDK.

### Solution

Add a new `solana-account-view` crate with two new types: `Account` and `AccountView`. The `Account` type has the same memory layout than the serialized account data provided by the runtime to a program entrypoint. Therefore, it can be used to represent the account data in a zero-copy fashion. The `AccountView` is the equivalent of an `AccountInfo`, wrapping an `Account` and providing safe borrow methods.

**Note:** This PR moves the code from pinocchio repository maintaining history.

### DIFF
--- a/account-view/src/lib.rs
+++ b/account-view/src/lib.rs
@@ -30,11 +30,11 @@ pub const NOT_BORROWED: u8 = u8::MAX;
 /// This struct is wrapped by [`AccountView`], which provides safe access
 /// to account information. At runtime, the account's data is serialized
 /// directly after the `Account` struct in memory, with its size specified
-/// by [`Account::data_len`].
+/// by [`RuntimeAccount::data_len`].
 #[repr(C)]
 #[cfg_attr(feature = "copy", derive(Copy))]
 #[derive(Clone, Default)]
-pub struct Account {
+pub struct RuntimeAccount {
     /// Borrow state for account data.
     ///
     /// This reuses the memory reserved for the duplicate flag in the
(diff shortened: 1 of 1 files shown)


account-view/src/lib.rs
```diff
@@ -0,0 +1,873 @@
...
+/// Raw account data.
+///
+/// This struct is wrapped by [`AccountView`], which provides safe access
+/// to account information. At runtime, the account's data is serialized
+/// directly after the `Account` struct in memory, with its size specified
+/// by [`Account::data_len`].
+#[repr(C)]
+#[cfg_attr(feature = "copy", derive(Copy))]
+#[derive(Clone, Default)]
+pub struct Account {
```
febo: There is already a public `Account` type defined in solana-account. We could name this one as `RuntimeAccount` or another name to avoid the name clash. It is not really used directly in programs, only the entrypoint uses it.

joncinque: A new name might be better to be safe, especially if it won't really be consumed from the outside. `RuntimeAccount` is pretty good, I also came up with `EntrypointAccount` or `RuntimeSerializedAccount` or `OnchainAccount`

febo:
> Just looked at the crate definition and your comment, but it looks good! Feel free to do the rename or not

Renamed to `RuntimeAccount`, which matches the crate description.

PR #391, opened 2025-10-19
program-memory: Add bpf target

febo:
### Problem

### Solution

**Note:** Needs https://github.com/anza-xyz/solana-sdk/pull/387 to go first.

### DIFF
--- a/program-memory/Cargo.toml
+++ b/program-memory/Cargo.toml
@@ -12,10 +12,7 @@ edition = { workspace = true }
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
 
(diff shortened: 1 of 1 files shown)


program-memory/Cargo.toml
```diff
@@ -12,7 +12,7 @@ edition = { workspace = true }
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
 
-[target.'cfg(target_os = "solana")'.dependencies]
+[target.'cfg(any(target_os = "solana", target_arch = "bpf"))'.dependencies]
```
febo, line 15: @joncinque Updated this to include `any` as suggested on #389.

PR #395, opened 2025-10-21
extend program entry to 128

mgild:
This pull request increases the maximum number of `AccountInfo` objects that can be processed in the `entrypoint_no_alloc` macro within `program-entrypoint/src/lib.rs`. This change allows handling more accounts per transaction, which can be important for complex programs or instructions.

Resource limit adjustment:

* Increased the `MAX_ACCOUNT_INFOS` constant from 64 to 128 in the `entrypoint_no_alloc` macro to support processing up to 128 accounts per transaction.

### DIFF
--- a/program-entrypoint/src/lib.rs
+++ b/program-entrypoint/src/lib.rs
@@ -169,7 +169,7 @@ macro_rules! entrypoint_no_alloc {
             #[allow(clippy::declare_interior_mutable_const)]
             const UNINIT_ACCOUNT_INFO: MaybeUninit<AccountInfo> =
                 MaybeUninit::<AccountInfo>::uninit();
-            const MAX_ACCOUNT_INFOS: usize = 64;
+            const MAX_ACCOUNT_INFOS: usize = 128;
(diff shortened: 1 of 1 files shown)


program-entrypoint/src/lib.rs
```diff
@@ -169,7 +169,7 @@ macro_rules! entrypoint_no_alloc {
             #[allow(clippy::declare_interior_mutable_const)]
             const UNINIT_ACCOUNT_INFO: MaybeUninit<AccountInfo> =
                 MaybeUninit::<AccountInfo>::uninit();
-            const MAX_ACCOUNT_INFOS: usize = 64;
+            const MAX_ACCOUNT_INFOS: usize = 128;
```
febo, line 172: Unfortunately this change is not possible since the stack size is limited to `4kb` and 128 `AccountInfo`s require `6144` bytes (`~6kb`)

joncinque:
Thanks for your contribution! As febo pointed out, we can't include this change currently. If we get dynamic stack frames, then we can add it.

With that said, I'll close this PR.

PR #396, opened 2025-10-21
Add bpf target to hashers

febo:
### Problem

### Solution

**Note:** Needs #389 to go first.

### DIFF
--- a/sha256-hasher/src/lib.rs
+++ b/sha256-hasher/src/lib.rs
@@ -1,11 +1,11 @@
 #![no_std]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
+#[cfg(any(target_os = "solana", target_arch = "bpf"))]
+pub use solana_define_syscall::definitions::sol_sha256;
 use solana_hash::Hash;
 #[cfg(any(target_os = "solana", target_arch = "bpf"))]
-use {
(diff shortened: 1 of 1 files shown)


febo:
Waiting on #408 to rebase.

sha256-hasher/src/lib.rs
```diff
@@ -1,16 +1,24 @@
 #![no_std]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
-#[cfg(all(feature = "sha2", not(target_os = "solana")))]
-use sha2::{Digest, Sha256};
+
 use solana_hash::Hash;
+#[cfg(any(target_os = "solana", target_arch = "bpf"))]
+pub use {
+    core::mem::MaybeUninit, solana_define_syscall::definitions::sol_sha256, solana_hash::HASH_BYTES,
+};
```
joncinque: This shouldn't be `pub use`, right?
```suggestion
use {
    core::mem::MaybeUninit, solana_define_syscall::definitions::sol_sha256, solana_hash::HASH_BYTES,
};
```

febo: Fixed. We still need to re-export the `sol_sha256` syscall – solana-program references it.

febo:
@joncinque Rebased and expanded the PR to include all hasher crates. `secp256k1-recover` still not `no_std` because of the thiserror dependency, I will address that in a separate PR.

PR #398, opened 2025-10-21
feat: add `no_std` support for nonce-account crate

sonicfromnewyoke:
### Problem

`solana-nonce-account` crate has no `no_std` support

### Summary of Changes
- add crate level attribute
- replace import of `std::cell::RefCell` with ` core::cell::RefCell`


febo:
I think we need to make its dependencies `no_std` first. For example:
```
solana-account = { workspace = true, features = ["bincode"] }
```
`bincode` is "std" by default, so probably won't work as it is.

Note: I didn't mean to close the PR.

PR #400, opened 2025-10-22
Remove frozen-abi-macro dependency on solana-logger

steviez:
A couple items are logged, and we can simply use eprintln!() instead.

This is part of a larger initiative to remove `solana-logger` from the SDK as it seemingly belongs with Agave

### DIFF
--- a/scripts/check-nits.sh
+++ b/scripts/check-nits.sh
@@ -19,6 +19,7 @@ declare print_free_tree=(
   ':**.rs'
   ':^address/src/hasher.rs'
   ':^address/src/lib.rs'
+  ':^frozen-abi/src/*.rs'
   ':^frozen-abi-macro/src/lib.rs'
   ':^logger/src/lib.rs'
   ':^msg/src/lib.rs'
@@ -31,13 +32,7 @@ declare print_free_tree=(
   ':^**/build.rs'
 )
 
-# Parts of the tree that may match a print_free_tree pattern but are deemed
-# an exception to the print free hygiene rule
-declare print_allowed_tree=(
-  ':!^frozen-abi/src/*.rs'
-)
-
-if git --no-pager grep -n "${prints[@]/#/-e}" -- "${print_free_tree[@]}" "${print_allowed_tree[@]}"; then
+if git --no-pager grep -n "${prints[@]/#/-e}" -- "${print_free_tree[@]}"; then
     exit 1
 fi
 


steviez:
Hmm ok, CI failed because of this line:
https://github.com/anza-xyz/solana-sdk/blob/1ad7ca9234186b1c53c7f5fac1ab842d70db5b69/frozen-abi-macro/src/lib.rs#L290

Trying to figure out if we need this or not

Edit: I have updated the PR to remove the use of `solana-logger` from `frozen-abi-macro`

apfitzge (dismissed):
seems reasonable to print instead of log. logger should 100% be agave-logger not solana-logger

steviez:
Hmm, I just noticed there are some log invocations in `solana-frozen-abi` like this one:
https://github.com/anza-xyz/solana-sdk/blob/1ad7ca9234186b1c53c7f5fac1ab842d70db5b69/frozen-abi/src/abi_digester.rs#L151

These are basically unprintable now if we do not setup a logger; we could either setup `env_logger` where `solana_logger::setup()` was OR update all those to `println!()` too. Thoughts ?

joncinque:
I'd say go with `println!` everywhere there was an `info!`, and `eprintln!` everywhere there was a `warn!` or `error!`. These logs are just used for debugging frozen-abi changes

scripts/check-nits.sh
```diff
@@ -31,7 +31,13 @@ declare print_free_tree=(
...
 )
 
-if git --no-pager grep -n "${prints[@]/#/-e}" -- "${print_free_tree[@]}"; then
+# Parts of the tree that may match a print_free_tree pattern but are deemed
+# an exception to the print free hygiene rule
+declare print_allowed_tree=(
+  ':!^frozen-abi/src/*.rs'
+)
+
+if git --no-pager grep -n "${prints[@]/#/-e}" -- "${print_free_tree[@]}" "${print_allowed_tree[@]}"; then
```
steviez: I'll call people's attention to this change to make sure we're ok with adding this crate to the newly created exclusion list

frozen-abi/Cargo.toml
apfitzge, line 29: if we want to be sure we're not using log macros, we should probably get rid of this dep?

scripts/check-nits.sh
joncinque: The `^` sign in the `print_free_tree` is actually for exceptions, so you should be able to just add `':^frozen-abi/src/*.rs'` to `print_free_tree`

frozen-abi/Cargo.toml
steviez: I had the same thought, but we're stuck that dep a little longer:
https://github.com/anza-xyz/solana-sdk/blob/ed42870a2858e890e7de996e88500a78c25ce84f/frozen-abi/src/lib.rs#L19-L24

After this PR ships, I plan to deprecate that re-export and then we can kill the dep once we bump major

scripts/check-nits.sh
steviez: Well that's a little embarrassing 😆; I snapped into thinking "start of line" anchor. Thanks for catching / resolved in https://github.com/anza-xyz/solana-sdk/pull/400/commits/29d5885b9dfdaae3d59a96de05b9898f9ebdbf6f (and ran `./scripts/check-nits.sh` locally to ensure the files I changed are still granted the exception

joncinque: I've had to remember this so many times, don't worry :sweat_smile:

joncinque (approved):
Looks great to me!

PR #401, opened 2025-10-23
instruction-view: Add new crate

febo:
### Problem

Currently pinocchio provides a CU efficient way to perform cross-program invocation by using zero-copy structs to represent instruction-related data, which can be used with `sol_invoke_signed_c` syscall.

### Solution

Add a new `solana-instruction-view` crate moving the types from pinocchio – the main type of this crate is the `InstructionView`.

The crate also includes a feature for cross-program invocations ("cpi") syscalls.

### DIFF
--- a/instruction-view/src/cpi.rs
+++ b/instruction-view/src/cpi.rs
@@ -305,7 +305,7 @@ pub fn invoke_with_slice(
 ///
 /// Note that this function is inlined to avoid the overhead of a function call,
 /// but uses stack memory allocation. When a large number of accounts is needed,
-/// it is recommended to use the [`invoke_with_slice_signed`] function instead
+/// it is recommended to use the [`invoke_signed_with_slice`] function instead
 /// to reduce stack memory utilization.
 ///
 /// # Important
@@ -369,7 +369,7 @@ pub fn invoke_signed<const ACCOUNTS: usize>(
 ///
 /// Note that this function is inlined to avoid the overhead of a function call,
 /// but uses stack memory allocation. When a large number of accounts is needed,
-/// it is recommended to use the [`invoke_with_slice_signed`] function instead to reduce
+/// it is recommended to use the [`invoke_signed_with_slice`] function instead to reduce
 /// stack memory utilization.
 ///
 /// # Important


instruction-view/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+/// signer, read or written during the execution of that instruction must be supplied.
+/// Any account that may be mutated by the program during execution, either its
+/// data or metadata such as held lamports, must be writable.
+///
+/// Note that because the Solana runtime schedules parallel transaction
+/// execution around which accounts are writable, care should be taken that only
+/// accounts which actually may be mutated are specified as writable.
+#[repr(C)]
+#[derive(Debug, Clone)]
+pub struct AccountRole<'a> {
```
febo: Not 100% sure about this name, happy to change it to a different name.

febo:
Waiting on #442 to rebase.

febo:
> Waiting on #442 to rebase.

@joncinque Rebase done.

instruction-view/src/cpi.rs
```diff
@@ -0,0 +1,733 @@
...
+use {
+    crate::InstructionView,
+    core::{marker::PhantomData, mem::MaybeUninit, ops::Deref, slice::from_raw_parts},
+    solana_account_view::AccountView,
+    solana_address::Address,
+    solana_program_error::{ProgramError, ProgramResult},
+};
+
+/// Maximum number of accounts that can be passed to a cross-program invocation.
+pub const MAX_CPI_ACCOUNTS: usize = 64;
```
febo: This limit will increase soon, so perhaps we should add an `"alloc"` feature and support a larger value?

instruction-view/src/lib.rs
joncinque: I still don't love it, since I always assume it's just the writable / signer flag.

I would prefer any of the following: `InstructionAccount` or `AccountMeta` or `AccountMetadata` or `InstructionAccountMetadata`. Or honestly anything, since `Role` is too vague, and even inconsistent with types like in Kit: https://github.com/anza-xyz/kit/blob/3779dcb610637f223e24486c6226d3a551f9605e/packages/instructions/README.md?plain=1#L18

instruction-view/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    /// Indicates whether the account is writable or not.
+    pub is_writable: bool,
+
+    /// Indicates whether the account signed the instruction or not.
+    pub is_signer: bool,
+}
+
+impl<'a> AccountRole<'a> {
+    /// Creates a new `AccountPrivilege`.
```
joncinque: nit: These comments will need to be updated with whatever final name we agree on

instruction-view/src/cpi.rs
joncinque: What would that look like exactly? Sorry, I don't totally understand. If we add an alloc feature, then we should also support a `Vec<CpiAccount>`, correct?

Either way, I think we should be able to do it in follow-up work 😄

instruction-view/src/cpi.rs
```diff
@@ -0,0 +1,733 @@
...
+    rent_epoch: u64,
+
+    // Transaction was signed by this account's key?
+    is_signer: bool,
+
+    // Is the account writable?
+    is_writable: bool,
+
+    // This account's data contains a loaded program (and is now read-only).
+    executable: bool,
```
joncinque: nit: up to you, but these could all be doc comments `///` instead

instruction-view/src/cpi.rs
```diff
@@ -0,0 +1,733 @@
...
+///      `Address` matches the `address` in the `AccountView`.
+///   3. The borrow state of the accounts is compatible with the mutability of the
+///      accounts in the instruction.
+///
+/// # Safety
+///
+/// This function is unsafe because it does not check that the stack allocated account
+/// storage `MAX_ACCOUNTS` is sufficient for the number of accounts expected by the
+/// instruction. Using a value of `MAX_ACCOUNTS` that is less than the number of accounts
+/// expected by the instruction will result in undefined behavior.
```
joncinque: nit: To be honest, I don't know if this function is unsafe. This sounds more like a usability error to me. The invoked program should error if it doesn't receive a well-formed instruction, but it won't cause an issue like reading uninitialized memory. Am I understanding this correctly?

instruction-view/src/cpi.rs
```diff
@@ -0,0 +1,733 @@
...
+    }
+}
+
+impl Deref for ReturnData {
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        self.as_slice()
+    }
+}
```
joncinque: Does this have a particular use? Since it omits the `program_id`, I worry that it could cause some issues or confusion

instruction-view/src/cpi.rs
febo: This [line](https://github.com/anza-xyz/solana-sdk/blob/219c2f6dca1f1eb83899c7118baa09a9f04c66e1/instruction-view/src/cpi.rs#L517) is unsafe. The `accounts` array is sized by `MAX_ACCOUNTS`, which coud be smaller than `instruction.accounts.len()`, so the resulting slice overflows.

There is no issue since we validate the value before calling this function – the function is private. I think we talked about this and decided to make it unsafe for maintainability.

instruction-view/src/cpi.rs
febo: We statically allocate an array for the `slice_*` variants:
```rust
const UNINIT: MaybeUninit<CpiAccount> = MaybeUninit::<CpiAccount>::uninit();
let mut accounts = [UNINIT; MAX_ACCOUNTS];
```
We are limited by `64` accounts due to memory requirement (each `CpiAccount` is `56` bytes). If we want to support more accounts, we need to use a `Vec`. Then all `slice_*` variants can be behind an `"alloc"` feature.

Might be cleaner to do it now, otherwise it is a breaking change.

instruction-view/src/lib.rs
febo: `InstructionAccount` seems to fit well here. Keeping `AccountMeta` would be ideal, but it will be confusing to have two types with the same name in the SDK.

instruction-view/src/cpi.rs
joncinque: Ah ok, sorry for rehashing this, all good then!

instruction-view/src/cpi.rs
febo: Or an alternative would be to add the `Vec`-based ones with a different name afterwards.

instruction-view/src/cpi.rs
febo: Not a strong one, just a convenience to get the bytes without having to use `as_slice()`. But I can see how it can be confusing, so let's remove it.

instruction-view/src/cpi.rs
joncinque: Ah ok, sorry, I think I see what you mean now. So essentially, with the `alloc` feature enabled, in the inner function, we dynamically allocate a vec based on the instruction size instead of using an array, and thus support any sized slice. Do I have that right?

The `alloc` feature typically means "add or support types from the alloc crate", which we're not really doing here. Instead, we're using alloc types under the hood and causing worse performance for CPIs using `slice_*`. I think people will accidentally enable the feature more easily since it's a common name.

Instead, we could name the feature `alloc-cpi-accounts` or `alloc-cpi` or `huge-cpi` or `simd339-cpi` with a comment saying that performance will be worse, but you'll be able to support CPIs with more than 64 accounts. What do you think?

joncinque: On the other hand, are dynamic stack frames coming out? If so, maybe we can just punt :sweat_smile:

instruction-view/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+/// data or metadata such as held lamports, must be writable.
+///
+/// Note that because the Solana runtime schedules parallel transaction
+/// execution around which accounts are writable, care should be taken that only
+/// accounts which actually may be mutated are specified as writable.
+#[repr(C)]
+#[derive(Debug, Clone)]
+pub struct InstructionAccount<'a> {
+    /// Address of the account.
+    pub address: &'a Address,
```
febo, line 46: Are we going with `addr` or `address`?

febo: ^ forgot to ask this earlier @joncinque.

joncinque: I'd go with `address` since `addr` is used a lot with pointers in rust: https://doc.rust-lang.org/stable/std/index.html?search=addr

On the other hand, I never had a problem with the name

instruction-view/src/cpi.rs
febo: > Do I have that right?

Yes, exactly!

> On the other hand, are dynamic stack frames coming out? If so, maybe we can just punt 😅

That is a good point, so maybe we can hold off on changing it.

instruction-view/src/lib.rs
febo: I think `address` is fine as well.

febo:
> Looks good to me! It's up to you if you want to do the alloc work here though, whichever you prefer

I had a go at refactoring the invoke to support the actual maximum of CPI accounts (commit [491515b](https://github.com/anza-xyz/solana-sdk/pull/401/commits/491515ba8be71ef8800b21a215150f569bd72e13)). The downside is that it requires a small bump on the minimum Rust version. We could implement with a `Vec`, but using the `Box` seems cleaner. If we don't want to bump the MSRV, then I can change the implementation.

scripts/check-msrv.sh
```diff
@@ -5,4 +5,21 @@ here="$(dirname "$0")"
...
+
+    # If the crate does not specify a rust-version, fall back to
+    # the "program" crate MSRV
+    if [[ -z "$minimum_version" ]]; then
+      minimum_version=$(readCargoVariable rust-version "program/Cargo.toml")
+    fi
+
+    cargo +$minimum_version check
+  fi
+done
```
joncinque: Take it or leave it, but it could be simpler to pull out the toml files directly, ie:
```suggestion
for cargo_toml in $(git ls-files -- '**/Cargo.toml'); do
  # Read the MSRV from the crate
  minimum_version=$(readCargoVariable rust-version "$cargo_toml" 2>/dev/null)

  # If the crate does not specify a rust-version, fall back to
  # the "program" crate MSRV
  if [[ -z "$minimum_version" ]]; then
    minimum_version=$(readCargoVariable rust-version "program/Cargo.toml")
  fi

  cargo +$minimum_version check --manifest-path $cargo_toml
done
```

instruction-view/src/cpi.rs
```diff
@@ -0,0 +1,729 @@
...
+/// stack memory utilization.
+///
+/// # Important
+///
+/// The accounts on the `account_views` slice must be in the same order as the
+/// `accounts` field of the `instruction`. When the instruction has duplicated
+/// accounts, it is necessary to pass a duplicated reference to the same account
+/// to maintain the 1:1 relationship between accounts and instruction accounts.
+#[inline(always)]
+pub fn invoke_with_bounds_signed<const MAX_ACCOUNTS: usize>(
```
joncinque: nit: I preferred the old name of `invoke_signed_with_bounds`, but up to you!

instruction-view/src/cpi.rs
```diff
@@ -0,0 +1,729 @@
...
+/// `MAX_CPI_ACCOUNTS` accounts.
+///
+/// # Important
+///
+/// The accounts on the `account_views` slice must be in the same order as the
+/// `accounts` field of the `instruction`. When the instruction has duplicated
+/// accounts, it is necessary to pass a duplicated reference to the same account
+/// to maintain the 1:1 relationship between accounts and instruction accounts.
+#[inline(always)]
+pub fn invoke_with_slice_signed(
```
joncinque: nit: same here, I preferred `invoke_signed_with_slice`

instruction-view/src/cpi.rs
```diff
@@ -0,0 +1,729 @@
...
+///      instruction accounts.
+///
+/// # Safety
+///
+/// This function is unsafe because it does not check that `accounts` is sufficiently
+/// large for the number of accounts expected by the instruction. Using an `accounts` slice
+/// shorter than the number of accounts expected by the instruction will result in
+/// undefined behavior.
+#[inline(always)]
+unsafe fn _invoke_with_slice_signed<'account, 'cpi>(
```
joncinque: nit: a name starting with an underscore typically means it's unused in Rust, whereas here it means that it's private. Same with the others, I preferred the previous name of prefixing with `inner_`

instruction-view/src/cpi.rs
```diff
@@ -0,0 +1,729 @@
+//! Cross-program invocation helpers.
+
+extern crate alloc;
```
joncinque, line 4: Since `alloc` is only needed for the slice invoke functions, how about a crate feature for the slice functions? That way people can completely avoid `alloc` in their build if they want. With the current code, they'll always get `alloc` if they want to CPI.

We can have `slice-cpi` for those functions, and it enables `cpi`.

scripts/check-msrv.sh
febo: This looks way better. 😀

.github/workflows/main.yml
```diff
@@ -224,6 +224,13 @@ jobs:
           cargo-cache-key: cargo-minimum-check
           cargo-cache-fallback-key: cargo-minimum
 
+      - name: Install additional toolchains
```
febo, line 227: We could add a cache to this step, but it does not seem to take very long since one of the toolchains will usually be installed already.

scripts/read-rust-versions.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
febo, line 1: Not sure if we need a separate script for this – we could just add its contents to the github workflow directly.

.github/workflows/main.yml
joncinque: This looks fine to me! It only takes 8 seconds, and the whole check is less than a minute

scripts/read-rust-versions.sh
joncinque: It can't hurt, looks good!

PR #402, opened 2025-10-23
frozen-abi: Deprecate the log crate re-export

steviez:
The re-export was previously only used to support macros in frozen-abi-macro. That crate has been updated to no longer use anything from the log crate so we can rip this out at the next major version bump


joncinque (approved):
Looks great!

PR #403, opened 2025-10-23
CI: Only run semver checks on crates with a lib target

joncinque:
#### Problem

The semver-checks step during publish doesn't distinguish the current crate targets, and semver-checks only works on crates with lib targets. This means it fails on proc-macro crates, preventing us from publishing.

#### Summary of changes

Similar to what's done in the program repos at [this workflow](https://github.com/solana-program/actions/blob/3823505ada06f9c8362cd935b73a7f22a1490374/.github/workflows/publish-rust.yml#L103), check for a lib target and then run semver-checks.

### DIFF
--- a/.github/workflows/publish-rust.yml
+++ b/.github/workflows/publish-rust.yml
@@ -151,14 +151,28 @@ jobs:
       - name: Install cargo-semver-checks
         uses: taiki-e/install-action@v2
         with:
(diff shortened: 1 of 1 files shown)


joncinque:
Merging before CI finishes, this change is unaffected by CI and will unblock publishing frozen-abi-macro

Issue #394, opened 2025-10-20
Bug report: Breaking change in BorshIoError enum leading to deserialization errors


pmantica11:
BorshIoError enum got changed from `BorshIoError(String)` to only just `BorshIoError` here:
https://github.com/anza-xyz/solana-sdk/blob/master/instruction-error/src/lib.rs#L184C9-L184C21

This is leading to a tricky error in Anza archival where old blocks encode this error as an object: 
```
{
   "BorshIoErrror": "Reason for the error" 
} 
```

New blocks simply encode this error as strings:
```
"BorshIoError"
``` 

Because of the two different ways to deserialize `BorshIoError` there is currently no solana-sdk that can correctly deserialize all blocks. To fix this, we need to implement a customer serde deserializer for the `InstructionError` that handles both cases.

steveluscher:
Related: https://github.com/anza-xyz/solana-sdk/pull/12#discussion_r1999878716

joncinque:
Fixed with #410 and deployed solana-instruction v2.3.1, please use that in your build. The v3 fixes will happen at the agave level

PR #404, opened 2025-10-24
instruction-error: deserialize old format for BorshIoError

fanatid:
Resolves https://github.com/anza-xyz/solana-sdk/issues/394

Custom visitor allows to deserialize both old and new formats:
```
{"BorshIoError": "42"}
"BorshIoError"
```

### DIFF
--- a/instruction-error/src/lib.rs
+++ b/instruction-error/src/lib.rs
@@ -66,7 +66,7 @@ mod instruction_error_module {
                         type Value = InstructionError;
 
                         fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-                            formatter.write_str(concat!("newtype variant"))
+                            formatter.write_str("newtype variant")
                         }
 
                         fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>


joncinque (commented):
Thanks for the contribution! I'm thinking it would be easiest to make a change like this directly to `maintenance/v2.x` and publish that for total coverage. v3 will be fine once all RPCs are running v3 of Agave, so I'm less worried about making the low-level change for that.

However, I'm wondering how you generated this. If I run `cargo expand` on instruction-error, I get a few more visitor functions implemented, like `visit_u64`, `visit_enum`, and `visit_bytes`, which are not included in this hand-implemented deserializer. Are those still needed?

It does also change the order of the instruction errors, which could be an issue for anyone expecting certain integer values for the errors, but maybe that isn't such a big deal.

fanatid:
> v3 will be fine once all RPCs are running v3 of Agave

That's exactly why I have created issues. Problem is that some systems currently on v2 can be not be compatible with other systems updates to v3. It's not possible to update consuming software first to v3 and still use v2 output... also some system can cache old v2 output in s3-like storages and now have problem with it in v3 updates caches.

> However, I'm wondering how you generated this

I checked cargo expand too but implemented manually with some tweaks from claude, in the past I created few visitors so have some experience with it

> It does also change the order of the instruction errors, which could be an issue for anyone expecting certain integer values for the errors, but maybe that isn't such a big deal.

That's a real issue with it, thank you! I really missed that integer values also would be changed and serialization would be different?

I happy to fix this problem if you think this is major blocker for merging

joncinque:
Closing with #410, we'll handle the v3 fix at the agave level with https://github.com/anza-xyz/agave/pull/8625

PR #405, opened 2025-10-24
fee-calculator: Remove solana-logger dev-dependency

steviez:
Remove the dependency and the couple info!() statements from unit tests since they would no longer print without a logger enabled

The `info!` statements seemed non-essential so I decided to remove them; we could leave them if we manually call `env_logger` to initialize a logger. I don't feel too strongly one way or another there.

This is the last use of `solana-logger` in this crate; I will remove `solana-logger `completely after this PR lands

### DIFF
--- a/fee-calculator/src/lib.rs
+++ b/fee-calculator/src/lib.rs
@@ -232,6 +232,7 @@ mod tests {
         let mut count = 0;
         loop {
             let last_lamports_per_signature = f.lamports_per_signature;
+            f = FeeRateGovernor::new_derived(&f, u64::MAX);
 
             // some maximum target reached
             if f.lamports_per_signature == last_lamports_per_signature {


fee-calculator/src/lib.rs
```diff
@@ -237,9 +233,6 @@ mod tests {
         loop {
             let last_lamports_per_signature = f.lamports_per_signature;
 
-            f = FeeRateGovernor::new_derived(&f, u64::MAX);
```
joncinque, line 240: Did you mean to remove this line?

steviez: I did not, corrected in 8f9d1a8f14a7fe2c9cfc9baf4c181784dcd22db0

steviez:
Thanks for catching that issue. Since this was only a dev dependency, I don't believe we need to push a new tag. Do you agree with that or am I overlooking anything ?

joncinque:
> Thanks for catching that issue. Since this was only a dev dependency, I don't believe we need to push a new tag. Do you agree with that or am I overlooking anything ?

Yep that's correct, no need to do anything more

PR #406, opened 2025-10-24
Remove solana-logger

steviez:
This crate probably should have never been brought over to the SDK, and the relevant logic is getting re-added as an Agave crate directly. So, remove the crate

### DIFF
--- a/scripts/check-nits.sh
+++ b/scripts/check-nits.sh
@@ -21,7 +21,6 @@ declare print_free_tree=(
   ':^address/src/lib.rs'
   ':^frozen-abi/src/*.rs'
   ':^frozen-abi-macro/src/lib.rs'
-  ':^logger/src/lib.rs'
   ':^msg/src/lib.rs'
   ':^program-log/src/logger.rs'
   ':^program-option/src/lib.rs'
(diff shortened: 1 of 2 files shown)


joncinque (commented):
Looks great! Can you also remove the references in [check-nits.sh](https://github.com/anza-xyz/solana-sdk/blob/40218b936836bd2cdb8346e45a0676b34172a91b/scripts/check-nits.sh#L24) and [patch-crates-functions.sh](https://github.com/anza-xyz/solana-sdk/blob/40218b936836bd2cdb8346e45a0676b34172a91b/scripts/patch-crates-functions.sh#L52)?

steviez:
> Looks great! Can you also remove the references in [check-nits.sh](https://github.com/anza-xyz/solana-sdk/blob/40218b936836bd2cdb8346e45a0676b34172a91b/scripts/check-nits.sh#L24) and [patch-crates-functions.sh](https://github.com/anza-xyz/solana-sdk/blob/40218b936836bd2cdb8346e45a0676b34172a91b/scripts/patch-crates-functions.sh#L52)?

Of course, apologies for missing these; I searched `solana-logger` initially and when I searched `logger` I saw some hits and disregarded them as being from `env_logger`

steviez:
Just to confirm - nothing further needed on this right ? Ie, we leave the existing crates that we have published alone, but we'll just never publish it again since it is deleted.

joncinque:
yep that's right. It was a non-breaking change to remove the dep, so we're all good. New publishes will just stop using the dep

PR #408, opened 2025-10-28
faster cryptography syscalls

deanmlittle:
This PR contains several of optimizations I have been personally running in production for over a year without incident. The majority of them pertain to using `MaybeUninit<[u8;32]>` in hashing as we have guaranteed soundness due to our syscalls always filling all 32 bytes, thus we can safely skip the existing 32 byte 0 allocation. Inlining the `target_os="solana"` target also saves us several additional CUs, as there is really no reason why you'd ever want to burn more CUs by not inlining a hashing syscall. In the case of `secp256k1_recover`, there are two optimizations, the first being the same `MaybeUninit<[u8;64]>` which is only ever assumed initialized in our pass case, giving us the same soundness guarantees as the hashing example. The second optimization is that we can avoid a `Clone` by directly consuming the recovered `[u8;64]` in constructing a `Secp256k1Pubkey` instead of invoking `Secp256k1Pubkey::new` and making a superfluous clone of it. Resulting CU savings are as follows:

| syscall | Δ CU |
| -------- | ------- |
|sol_sha256 | -20 |
|sol_keccak256 | -20 |
|sol_blake3 | -20 |
| sol_secp256k1_recover | -46 |

### DIFF
--- a/blake3-hasher/src/lib.rs
+++ b/blake3-hasher/src/lib.rs
@@ -50,7 +50,7 @@ pub fn hashv(vals: &[&[u8]]) -> Hash {
     #[cfg(target_os = "solana")]
     {
         let mut hash_result = core::mem::MaybeUninit::<[u8; solana_hash::HASH_BYTES]>::uninit();
-        // Safety: This is sound as sol_blake3 always fills all 32 bytes of our hash
+        // SAFETY: This is sound as sol_blake3 always fills all 32 bytes of our hash
         unsafe {
             solana_define_syscall::definitions::sol_blake3(
                 vals as *const _ as *const u8,
--- a/keccak-hasher/src/lib.rs
+++ b/keccak-hasher/src/lib.rs
@@ -52,7 +52,7 @@ pub fn hashv(vals: &[&[u8]]) -> Hash {
     #[cfg(target_os = "solana")]
     {
         let mut hash_result = core::mem::MaybeUninit::<[u8; solana_hash::HASH_BYTES]>::uninit();
-        // Safety: This is sound as sol_keccak256 always fills all 32 bytes of our hash
+        // SAFETY: This is sound as sol_keccak256 always fills all 32 bytes of our hash
         unsafe {
             solana_define_syscall::definitions::sol_keccak256(
                 vals as *const _ as *const u8,
--- a/secp256k1-recover/src/lib.rs
+++ b/secp256k1-recover/src/lib.rs
@@ -421,7 +421,7 @@ pub fn secp256k1_recover(
             )
         };
 
-        // Safety: This is sound as in our pass case, all 64 bytes of the pubkey are always initialized by sol_secp256k1_recover
+        // SAFETY: This is sound as in our pass case, all 64 bytes of the pubkey are always initialized by sol_secp256k1_recover
         match result {
             0 => Ok(Secp256k1Pubkey(unsafe { pubkey_buffer.assume_init() })),
             error => Err(Secp256k1RecoverError::from(error)),


deanmlittle:
@steviez @joncinque plz merge ty 🫶

blake3-hasher/src/lib.rs
```diff
@@ -28,6 +28,7 @@ impl Hasher {
 }
 
 /// Return a Blake3 hash for the given data.
+#[cfg_attr(target_os = "solana", inline(always))]
```
febo, line 31: Since we are updating the SDK to also work with upstream BPF, we could update this to:
```suggestion
#[cfg_attr(any(target_os = "solana", target_arch = "bpf"), inline(always))]
```

blake3-hasher/src/lib.rs
```diff
@@ -48,19 +49,21 @@ pub fn hashv(vals: &[&[u8]]) -> Hash {
     // Call via a system call to perform the calculation
     #[cfg(target_os = "solana")]
     {
-        let mut hash_result = [0; HASH_BYTES];
+        let mut hash_result = core::mem::MaybeUninit::<[u8; solana_hash::HASH_BYTES]>::uninit();
+        // Safety: This is sound as sol_blake3 always fills all 32 bytes of our hash
```
febo: nit: We tend to use `SAFETY` in all caps:
```suggestion
        // SAFETY: This is sound as sol_blake3 always fills all 32 bytes of our hash
```

febo:
Looks great – just left 2 small comments that apply in multiple places.

sha256-hasher/src/lib.rs
```diff
@@ -50,19 +51,21 @@ pub fn hashv(vals: &[&[u8]]) -> Hash {
...
     {
-        let mut hash_result = [0; solana_hash::HASH_BYTES];
+        let mut hash_result = core::mem::MaybeUninit::<[u8; solana_hash::HASH_BYTES]>::uninit();
+        // Safety: This is sound as sol_sha256 always fills all 32 bytes of our hash
         unsafe {
             sol_sha256(
                 vals as *const _ as *const u8,
                 vals.len() as u64,
-                &mut hash_result as *mut _ as *mut u8,
+                hash_result.assume_init_mut() as *mut u8,
```
febo: Memory is not initialized at this point:
```suggestion
                hash_result.as_mut_ptr() as *mut u8,
```

blake3-hasher/src/lib.rs
deanmlittle: Fixed `SAFETY` here: https://github.com/anza-xyz/solana-sdk/commit/687729927c4fd14d40dd95eca4191eeea5cfc150

sha256-hasher/src/lib.rs
deanmlittle: autocomplete definitely won this round 😅

fix here: https://github.com/anza-xyz/solana-sdk/commit/8a9ddcf9d8cc461db730475c9332b67331d4cc92

deanmlittle:
> Looks good to me with febo's points addressed!

@joncinque typos fixed, CI passed, @febo will rebase off this version for bpf arch flags once merged. Send it ser 🫡

keccak-hasher/src/lib.rs
```diff
@@ -50,19 +51,21 @@ pub fn hashv(vals: &[&[u8]]) -> Hash {
     // Call via a system call to perform the calculation
     #[cfg(target_os = "solana")]
     {
-        let mut hash_result = [0; HASH_BYTES];
+        let mut hash_result = core::mem::MaybeUninit::<[u8; solana_hash::HASH_BYTES]>::uninit();
+        // Safety: This is sound as sol_keccak256 always fills all 32 bytes of our hash
```
febo:
```suggestion
        // SAFETY: This is sound as sol_keccak256 always fills all 32 bytes of our hash
```

secp256k1-recover/src/lib.rs
```diff
@@ -402,25 +402,28 @@ pub use solana_define_syscall::definitions::sol_secp256k1_recover;
...
             sol_secp256k1_recover(
                 hash.as_ptr(),
                 recovery_id as u64,
                 signature.as_ptr(),
-                pubkey_buffer.as_mut_ptr(),
+                pubkey_buffer.as_mut_ptr() as *mut u8,
             )
         };
 
+        // Safety: This is sound as in our pass case, all 64 bytes of the pubkey are always initialized by sol_secp256k1_recover
```
febo:
```suggestion
        // SAFETY: This is sound as in our pass case, all 64 bytes of the pubkey are always initialized by sol_secp256k1_recover
```

blake3-hasher/src/lib.rs
febo: There were multiple occurrences, so this one still not fixed.

blake3-hasher/src/lib.rs
febo: We can fix these as part of #396.

deanmlittle:
@febo mind checking my skill issues one last time?

kwdila:
awesome work! Now I finally don't need to have the non-std deps

PR #409, opened 2025-10-28
add src pubkey meta field to packet meta

dachen0:
necessary for https://github.com/anza-xyz/agave/issues/8420


joncinque:
Is this the same as #436? cc @lijunwangs

lijunwangs:
> Is this the same as #436? cc @lijunwangs

Yes. I think it is for the same goal.

joncinque:
Closing since this was done with https://github.com/anza-xyz/solana-sdk/pull/436

PR #410, opened 2025-10-29
instruction: Forward compatibility for BorshIoError serde

joncinque:
#### Problem

As noted in #404, InstructionError in v2 can't deserialize the new BorshIoError variant in v3 because we removed the string component.

#### Summary of changes

After trying many different things, this seems like the least bad option, and actually recommended by serde's maintainer for this kind of situation.

Agave should not be updated to use this, but all clients can pull in the new crate for their v2 builds to work with v3 RPC nodes.

cc @fanatid

### DIFF
--- a/instruction/src/error.rs
+++ b/instruction/src/error.rs
@@ -247,7 +247,7 @@ pub enum InstructionError {
 // variant or a newtype, since there's no way for an enum variant deserializer
 // to work with both a newtype and nothing:
 // https://github.com/serde-rs/serde/issues/1174#issuecomment-372411280
-#[cfg(feature = "serde")]
+#[cfg(all(feature = "std", feature = "serde"))]
 impl<'de> serde::de::Deserialize<'de> for InstructionError {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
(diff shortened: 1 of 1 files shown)


instruction/src/error.rs
```diff
@@ -242,6 +243,35 @@ pub enum InstructionError {
...
+// https://github.com/serde-rs/serde/issues/1174#issuecomment-372411280
+#[cfg(all(feature = "std", feature = "serde"))]
+impl<'de> serde::de::Deserialize<'de> for InstructionError {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::de::Deserializer<'de>,
+    {
+        let s = serde_json::Value::deserialize(deserializer)?;
+        if s.as_str().is_some_and(|v| v == "BorshIoError") {
+            Ok(Self::BorshIoError(String::new()))
```
fanatid, line 258: I had seen that default value is `Unknown`, should it be from default? but probably empty is valid :confused: 

https://github.com/anza-xyz/solana-sdk/blob/transaction-error%40v2.2.1/instruction/src/error.rs#L410

fanatid (approved):
lgtm

instruction/src/error.rs
joncinque: I thought about that, but decided to go with the simplest option, since I don't even think RPC returns a value for these errors when coming from blocks.

PR #412, opened 2025-10-29
v3: Use `doc_cfg` instead of `doc_auto_cfg`

joncinque:
#### Problem

This is the v3 version of #411.

#### Summary of changes

Same thing, replace `doc_auto_cfg` with `doc_cfg`.

### DIFF
--- a/account-info/src/lib.rs
+++ b/account-info/src/lib.rs
@@ -1,5 +1,5 @@
 //! Account information.
-#![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
 use {
     solana_address::Address,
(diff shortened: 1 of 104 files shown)


apfitzge:
why not backport?

joncinque:
Great point! I thought we had different crates between the two branches, but it's not the case

joncinque:
Welp, never mind, the bot told me to do it by hand, so we're back!

joncinque:
https://github.com/anza-xyz/solana-sdk/pull/411#issuecomment-3464890516

febo:
> Welp, never mind, the bot told me to do it by hand, so we're back!

The master branch has `account-view` crate that is not on the maintenance branch.

febo (approved):

PR #413, opened 2025-10-30
ci: Add docs.rs build to publish

febo:
### Problem

As described on #411 , docs.rs build failed after publishing the crate since it uses a nightly toolchain where `doc_auto_cfg` feature has been removed.

### Solution

This PR adds a job to the publish workflow to build the documentation using docs.rs image and a nightly toolchain.

A successful run of the job can be seen here: https://github.com/febo/solana-sdk/actions/runs/18941243715/job/54080342807

### DIFF
--- a/.github/workflows/publish-rust.yml
+++ b/.github/workflows/publish-rust.yml
@@ -113,20 +113,12 @@ jobs:
       - name: Install 'nightly' toolchain
         uses: dtolnay/rust-toolchain@nightly
 
-      - name: Run docs.rs container build
+      - name: Run docs.rs build
         run: |
-          docker run --rm \
-            -v ~/.cargo:/cargo \
-            -v ~/.rustup:/rustup \
-            -v "$PWD":/work \
-            -w /work \
-            -e CARGO_HOME=/cargo \
-            -e RUSTUP_HOME=/rustup \
-            -e PATH=/cargo/bin:$PATH \
-            -e DOCS_RS=1 \
-            -e RUSTDOCFLAGS="--cfg docsrs" \
-            ghcr.io/rust-lang/crates-build-env/linux:latest \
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-rust.yml
```diff
@@ -102,6 +102,32 @@ jobs:
...
+          docker run --rm \
+            -v ~/.cargo:/cargo \
+            -v ~/.rustup:/rustup \
+            -v "$PWD":/work \
+            -w /work \
+            -e CARGO_HOME=/cargo \
+            -e RUSTUP_HOME=/rustup \
+            -e PATH=/cargo/bin:$PATH \
+            -e DOCS_RS=1 \
+            -e RUSTDOCFLAGS="--cfg docsrs" \
```
joncinque: I'm not sure if you ran this command successfully on your machine, but on mine, I need to quote all of the env declarations
```suggestion
            -e "CARGO_HOME=/cargo" \
            -e "RUSTUP_HOME=/rustup" \
            -e "PATH=/cargo/bin:$PATH" \
            -e "DOCS_RS=1" \
            -e "RUSTDOCFLAGS=\"--cfg docsrs\"" \
```

febo:
> We could potentially eliminate the whole docker setup if you want to simplify a bit more and just call `cargo +nightly ...` directly, that worked equally well for me.

Agree , I think  it is sufficient to run `cargo +nightly doc`. Here is an example run: https://github.com/febo/solana-sdk/actions/runs/18959537456/job/54143821569

> We'll need to update the proc-macro2 dependency in our lockfile, because the docs build fails locally for me.

Hmmm, I ran `cargo +nightly doc --no-deps --all-features` locally from the root and it succeeds.

joncinque (approved):
> Hmmm, I ran cargo +nightly doc --no-deps --all-features locally from the root and it succeeds.

Alright, let's roll the dice and see what happens!

PR #414, opened 2025-10-31
Faster bn254 syscalls

cryptopapi997:
Inspired by https://github.com/anza-xyz/solana-sdk/pull/408 , we can ~steal~ adapt these ideas to optimize the bn254 syscalls as well, since those all overwrite the full memory buffer as well. For the compression syscalls that return a fixed size array, we can utilize the same `MaybeUninit` trick. For the addition, multiplication, and pairing syscalls we can alloc a vec once at the beginning instead of allocing an array and then converting that to a vec. This results in the following CU diffs:

| operation      | Δ CU |
| ----------- | ----------- |
| add | -27 |
| mul | -27 |
| pairing | -13 |
| g1_compress | -4 |
| g2_compress | -8 |
| g1_decompress | -8 |
| g2_decompress | -22 |

Note: I didn't apply this to the deprecated bn254 syscalls, but can do as well.
Note 2: Technically these optimizations could be applied inside the actual syscall implementation too instead of the wrapper functions interfacing with them for a couple extra CUs. However, this seemed more low risk, and idk if a SIMD would apply in that case so I went with this for now.

### DIFF
--- a/bn254/src/addition.rs
+++ b/bn254/src/addition.rs
@@ -126,7 +126,7 @@ pub fn alt_bn128_g1_addition_be(input: &[u8]) -> Result<Vec<u8>, AltBn128Error> 
                 ALT_BN128_G1_ADD_BE,
                 input as *const _ as *const u8,
                 input.len() as u64,
-                result_buffer.as_mut_ptr() as *mut u8,
+                result_buffer.as_mut_ptr(),
             );
             match result {
                 0 => {
@@ -165,7 +165,7 @@ pub fn alt_bn128_g1_addition_le(
                 ALT_BN128_G1_ADD_LE,
                 input as *const _ as *const u8,
                 input.len() as u64,
-                result_buffer.as_mut_ptr() as *mut u8,
+                result_buffer.as_mut_ptr(),
             );
             match result {
                 0 => {
--- a/bn254/src/multiplication.rs
+++ b/bn254/src/multiplication.rs
@@ -140,7 +140,7 @@ pub fn alt_bn128_g1_multiplication_be(input: &[u8]) -> Result<Vec<u8>, AltBn128E
                 ALT_BN128_G1_MUL_BE,
                 input as *const _ as *const u8,
                 input.len() as u64,
-                result_buffer.as_mut_ptr() as *mut u8,
+                result_buffer.as_mut_ptr(),
             );
             match result {
                 0 => {
@@ -179,7 +179,7 @@ pub fn alt_bn128_g1_multiplication_le(
                 ALT_BN128_G1_MUL_LE,
                 input as *const _ as *const u8,
                 input.len() as u64,
-                result_buffer.as_mut_ptr() as *mut u8,
+                result_buffer.as_mut_ptr(),
             );
             match result {
                 0 => {
--- a/bn254/src/pairing.rs
+++ b/bn254/src/pairing.rs
@@ -135,7 +135,7 @@ pub fn alt_bn128_pairing_be(input: &[u8]) -> Result<Vec<u8>, AltBn128Error> {
                 ALT_BN128_PAIRING_BE,
                 input as *const _ as *const u8,
                 input.len() as u64,
-                result_buffer.as_mut_ptr() as *mut u8,
+                result_buffer.as_mut_ptr(),
             );
             match result {
                 0 => {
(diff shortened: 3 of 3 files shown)


bn254/src/compression.rs
```diff
@@ -345,152 +345,156 @@ mod target_arch {
...
                 &mut result_buffer as *mut _ as *mut u8,
-            )
-        };
-
-        match result {
-            0 => Ok(result_buffer),
-            _ => Err(AltBn128CompressionError::UnexpectedError),
+            );
+            match result {
+                0 => Ok(result_buffer),
```
joncinque: Does this work? We should probably have an `assume_init()` here
```suggestion
                0 => Ok(result_buffer.assume_init()),
```

joncinque (commented):
Thanks for your contribution, looks great to me! Just the one comment, @samkim-crypto can you also take a look?

joncinque:
Noting for anyone watching, that the `Vec::with_capacity` -> `as_mut_ptr()` -> `set_len()` flow is very well demonstrated here: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_mut_ptr

bn254/src/compression.rs
```diff
@@ -345,152 +345,156 @@ mod target_arch {
...
                 input as *const _ as *const u8,
                 input.len() as u64,
-                &mut result_buffer as *mut _ as *mut u8,
-            )
-        };
-
-        match result {
-            0 => Ok(result_buffer),
-            _ => Err(AltBn128CompressionError::UnexpectedError),
+                result_buffer.as_mut_ptr() as *mut u8,
```
samkim-crypto, line 450: I could be missing something, but for the addition, multiplication, and pairing syscalls, it seems like `result_buffer.as_mut_ptr()` already returns `*mut u8`. Is the cast `as *mut u8` redundant?

samkim-crypto (commented):
Just one small clarification below, but everything looks fine to me!

bn254/src/compression.rs
cryptopapi997: No, it returns `*mut [u8; OUTPUT_LENGTH]` bc `as_mut_ptr` just returns `*mut T` and result_buffer is a fixed size byte array. If we remove the cast, rust complains since it's a different type than the expected `*mut u8` (even though we know these are the same here)

samkim-crypto: Oh sorry, I think I put the comment in the wrong place. For `compression.rs` where we use `MaybeUninit<[u8; N]>`, you are right. The `as_mut_ptr()` returns `*mut T`, which in that case is `*mut [u8; N]`, so we need to cast. For `addition.rs`, `multiplication.rs`, and `pairing.rs`, where we use `Vec<T>`, then it seems like we can remove the cast? (again, correct me if I am wrong 🙏 )

cryptopapi997: ah, i forgot about those, good catch! Fixed now :+1:

LStan:
@cryptopapi997 Why did you change `&mut result_buffer as *mut _ as *mut u8` to `result_buffer.as_mut_ptr() as *mut u8`? Is there any difference?
Also, you didn't make this change in `alt_bn128_g1_compress_be`, `alt_bn128_g1_decompress_be`, `alt_bn128_g2_compress_be`, `alt_bn128_g2_compress_le`, `alt_bn128_pairing`

cryptopapi997:
I changed `&mut result_buffer as *mut _ as *mut u8` to `as_mut_ptr` bc we introduced dealing with the raw bytes of `vec`s in some of the code. For these, those two operations are very different, the former will give a pointer to the beginning of the struct (i.e. the length field) whereas the latter will give a pointer to the first element (what we want). For the raw `MaybeUninit` arrays you're right this switch is technically not needed, but imo lowers the possibility of someone seeing the pointer cast being done differently, trying to unify them, and messing up the vec stuff & just felt a bit cleaner overall.

> Also, you didn't make this change in alt_bn128_g1_compress_be, alt_bn128_g1_decompress_be, alt_bn128_g2_compress_be, alt_bn128_g2_compress_le, alt_bn128_pairing

I mentioned this in my PR description, `alt_bn128_pairing` is deprecated, so I left it as is. For the other compression syscalls I did [make the change](https://github.com/anza-xyz/solana-sdk/blob/master/bn254/src/compression.rs#L423-L424) using `core::mem::MaybeUninit`, unless you're thinking of a different optimization I missed.

LStan:
> I mentioned this in my PR description, `alt_bn128_pairing` is deprecated, so I left it as is.

But you changed other deprecated functions in `compression`. The confusion probably appeared after you merged my code.

> For the other compression syscalls I did [make the change](https://github.com/anza-xyz/solana-sdk/blob/master/bn254/src/compression.rs#L423-L424) using `core::mem::MaybeUninit`, unless you're thinking of a different optimization I missed

I meant the change from  `&mut result_buffer as *mut _ as *mut u8` to `esult_buffer.as_mut_ptr() as *mut u8` in those functions.

PR #415, opened 2025-10-31
[bn254] Add _be suffix to BE compress/decompress functions

LStan:
This is a follow-up to #332, where I forgot to add the `_be` suffix to the BE compress/decompress functions.
I also changed the input to a sized array for the new functions, as in #334.

### DIFF
--- a/bn254/src/compression.rs
+++ b/bn254/src/compression.rs
@@ -126,8 +126,19 @@ mod target_arch {
         LE,
     }
 
+    #[deprecated(
+        since = "3.2.0",
+        note = "Please use `alt_bn128_g1_decompress_be` instead"
+    )]
     #[inline(always)]
     pub fn alt_bn128_g1_decompress(
+        g1_bytes: &[u8],
+    ) -> Result<[u8; ALT_BN128_G1_POINT_SIZE], AltBn128CompressionError> {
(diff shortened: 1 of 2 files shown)


LStan:
@samkim-crypto I set the deprecated since version to `3.2.0`. But maybe it can be released as `3.1.3`. If so, I'll change the version.

bn254/src/compression.rs
```diff
@@ -361,6 +415,25 @@ mod target_arch {
         }
     }
 
+    pub fn alt_bn128_g1_compress_be(
```
samkim-crypto, line 418: `alt_bn128_g1_compress` is not invoking `alt_bn128_g1_compress_be` to save on CUs here, correct?

samkim-crypto (commented):
Looks good to me. Thanks!

I don't think we should introduce deprecations on a patch release, so v3.2 seems fine to me.

bn254/src/compression.rs
LStan: If it was invoking `alt_bn128_g1_compress_be` here, then the length of the input would be checked, which is not the case in the original logic.

samkim-crypto: Ah yes makes sense to me. Thanks!

PR #416, opened 2025-10-31
Document security hazards of `system_interface::instruction::create_account`

tobias-nd:


system-interface/src/instruction.rs
```diff
@@ -312,6 +312,13 @@ pub enum SystemInstruction {
 /// [`transfer`] lamports for rent, [`assign`] to its owning program. The
 /// [`create_account`] function does all three at once.
 ///
+/// # Security issues
+///
+/// Using this function is a security issue if the `to_address` is predictable
+/// by an attacker. The attacker can prefund the address with lamports and
+/// thereby prevent the successful execution of the `create_account` call. This
+/// can DoS an on-chain program.
```
joncinque, line 320: This makes sense to me! Should we add a call-out to use `create_account_allow_prefund` to avoid the error? cc @rustopian

joncinque (commented):
Thanks for your contribution! Also asking for a recommendation about `create_account_allow_prefund`, but maybe we wait to include that until it's been enabled

system-interface/src/instruction.rs
rustopian: Excellent. Yes, we should recommend `CreateAccountAllowPrefund` here in any case where `to_address` could be predictable ... once the feature is active.

PR #418, opened 2025-10-31
v3: Re-export v4 hash

febo:
### Problem

There was a major version of `solana-hash` since features were created so dependencies are optional and the crate was made `no_std` compatible. Although the types between the version have not changed, the major bump makes them to be considered different.

### Solution

Follow the semver trick and re-export the v4 type on v3, so they will be considered the same type.

### DIFF
--- a/scripts/order-crates-for-publishing.py
+++ b/scripts/order-crates-for-publishing.py
@@ -63,6 +63,20 @@ def is_self_dev_dep_with_dev_context_only_utils(package, dependency, wrong_self_
 
     return is_special_cased
 
+# When using the semver trick to make version X compatible with version X + 1,
+# we add a self-dependency on a different major version.
+# This is totally fine from a publishing perspective.
+def is_self_dep_with_different_major(package, dependency):
+    no_explicit_version = '*'
+
+    is_special_cased = False
+    if (dependency['name'] == package['name'] and
+        dependency['req'] != no_explicit_version and
+        dependency['req'].lstrip('^~>=<')[0] != package['version'][0]):
+        is_special_cased = True
+
+    return is_special_cased
+
 
 # `cargo publish` is fine with circular dev-dependencies if
 # they are path deps.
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -342,3 +342,11 @@ opt-level = 1
 # Enable optimizations for procmacros for faster recompile
 [profile.dev.build-override]
 opt-level = 1
+
+[patch.crates-io]
```
febo, line 346: These are all `solana-*` dependencies from the hash crate so they will always use the repository version instead of crates.io ones.

scripts/test-miri.sh
```diff
@@ -5,4 +5,8 @@ here="$(dirname "$0")"
 src_root="$(readlink -f "${here}/..")"
 cd "${src_root}"
 # miri is very slow; so only run very few of selective tests!
-./cargo nightly miri test -p solana-hash -p solana-account-info
+#
+# Note: fixed version of solana-hash is needed here to avoid
+#       dependency resolution issues since v3 just re-exports
+#       v4 of the `solana-hash` crate.
+./cargo nightly miri test -p solana-hash@3.0.0 -p solana-account-info
```
febo: This was needed since miri was complaining that there were 2 `solana-hash` crates available.

Cargo.lock
```diff
@@ -3149,11 +3149,33 @@ dependencies = [
...
+ "im",
+ "log",
+ "memmap2",
+ "serde",
+ "serde_derive",
+ "serde_with",
+ "sha2",
+ "solana-frozen-abi-macro 3.0.1",
+ "thiserror 2.0.12",
+]
```
joncinque: The new entries for frozen-abi and frozen-abi-macro need to be removed to avoid bringing in two versions of frozen-abi and frozen-abi-macro. That's causing the build issues

joncinque (commented):
Just a fly-by

scripts/test-miri.sh
febo: I think we can actually remove `solana-hash` from this script, since there are no tests in the crate.

joncinque (approved):
Looks good to me! (assuming CI passes this time)

PR #423, opened 2025-11-03
v2: instruction: Handle string in all deserializers

joncinque:
#### Problem

The fallback logic to optionally deserialize a string with `BorshIoError` only works with JSON, and doesn't work with binary formats like bincode.

#### Summary of changes

I tried so many different things, including a custom visitor, introducing an intermediate type, deserializing just the bytes first, and a few others not worth mentioning.

Nothing worked in the end, so I've inlined the macro expansion for `Deserialize`, and added custom code for handling the `BorshIoError` variant, where we fallback to an empty string if we fail to deserialize the string.

Since this code is only in the v2 branch, I think we can live with the ugliness, but I'm certainly open to any other suggestions.

### DIFF
--- a/instruction/src/error.rs
+++ b/instruction/src/error.rs
@@ -46,6 +46,7 @@ pub const INCORRECT_AUTHORITY: u64 = to_builtin!(26);
 // - Added as an equivalent to ProgramError and InstructionError
 // - Be featurized in the BPF loader to return `InstructionError::InvalidError`
 //   until the feature is activated
+// - Added to the `Deserialize` implementation, which is done by hand
 
 /// Reasons the runtime might have rejected an instruction.
 ///


instruction/src/lib.rs
```diff
@@ -196,7 +196,7 @@ impl Instruction {
     /// ```
     /// # use solana_pubkey::Pubkey;
     /// # use solana_instruction::{AccountMeta, Instruction};
-    /// # use serde::{Serialize, Deserialize};
+    /// # use serde_derive::{Serialize, Deserialize};
```
joncinque, line 199: Not strictly necessary, but the doc tests were failing to build once I enabled `bincode` in dev-dependencies

instruction/src/error.rs
```diff
@@ -243,34 +239,979 @@ pub enum InstructionError {
...
+                            _serde::de::VariantAccess::unit_variant(__variant)?;
+                            _serde::__private::Ok(InstructionError::IncorrectAuthority)
+                        }
+                        (__Field::__field44, __variant) => {
+                            // START CUSTOM CODE: FALL BACK TO AN EMPTY STRING
+                            _serde::__private::Ok(InstructionError::BorshIoError(
+                                _serde::de::VariantAccess::newtype_variant::<String>(__variant)
+                                    .unwrap_or_default(),
+                            ))
+                            // END CUSTOM CODE
```
joncinque, line 1100: This is the only change required. Here's the original macro expansion:

```
_serde::__private::Result::map(
    serde::de::VariantAccess::newtype_variant::<String,>(
        __variant
    ),
    InstructionError::BorshIoError,
)
```

instruction/src/error.rs
```diff
@@ -56,11 +56,7 @@ pub const INCORRECT_AUTHORITY: u64 = to_builtin!(26);
 /// change at any time and changes to them are difficult to detect.
 #[cfg(feature = "std")]
 #[cfg_attr(feature = "frozen-abi", derive(AbiExample, AbiEnumVisitor))]
-#[cfg_attr(
-    feature = "serde",
-    derive(serde_derive::Serialize, serde_derive::Deserialize)
-)]
-#[cfg_attr(feature = "serde", serde(remote = "InstructionError"))]
+#[cfg_attr(feature = "serde", derive(serde_derive::Serialize))]
```
febo, line 60: Wondering if it would be useful to document somewhere that `InstructionError` has a "manual" `Deserialize` implementation. Unlikely that it will change, but if it does, it needs to be updated.

joncinque: Yeah sure, what would be the best way to document that? Add a comment after this line?

febo: It could be simple comment (`//`) after the rustdocs for the type on line 56. There is a "Warning" comment for the error codes just above (line 44), so maybe a similar one?

joncinque: Sure, done!

febo (approved):
Looks great!

PR #424, opened 2025-11-04
Add VoteInitV2 and new variant in VoteAuthorize to add BLS Pubkey to vote program.

wen-coding:
Per [SIMD 387](https://github.com/solana-foundation/solana-improvement-documents/pull/387), do the following to add BLS Pubkey to vote program:
- Add `VoteInitV2`, and `InitializeAccountV2` to create vote account
- Add new variant `VoterWithBLS` in `VoteAuthorize` enum

### DIFF
--- a/vote-interface/src/state/vote_instruction_data.rs
+++ b/vote-interface/src/state/vote_instruction_data.rs
@@ -1,8 +1,3 @@
-#[cfg(feature = "serde")]
-use {
-    serde_derive::{Deserialize, Serialize},
-    serde_with::serde_as,
-};
 #[cfg(feature = "frozen-abi")]
 use solana_frozen_abi_macro::{frozen_abi, AbiExample};
 use {
@@ -13,6 +8,11 @@ use {
     solana_hash::Hash,
     solana_pubkey::Pubkey,
     std::{collections::VecDeque, fmt::Debug},
+};
+#[cfg(feature = "serde")]
+use {
+    serde_derive::{Deserialize, Serialize},
+    serde_with::serde_as,
 };
 
 #[cfg_attr(
@@ -212,7 +212,10 @@ pub struct VoteInit {
 pub struct VoteInitV2 {
     pub node_pubkey: Pubkey,
     pub authorized_voter: Pubkey,
-    #[cfg_attr(feature = "serde", serde_as(as = "[_; BLS_PUBLIC_KEY_COMPRESSED_SIZE]"))]
+    #[cfg_attr(
+        feature = "serde",
+        serde_as(as = "[_; BLS_PUBLIC_KEY_COMPRESSED_SIZE]")
+    )]
     pub authorized_voter_bls_pubkey: [u8; BLS_PUBLIC_KEY_COMPRESSED_SIZE],
     #[cfg_attr(feature = "serde", serde_as(as = "[_; BLS_SIGNATURE_COMPRESSED_SIZE]"))]
     pub authorized_voter_bls_proof_of_possession: [u8; BLS_SIGNATURE_COMPRESSED_SIZE],
@@ -243,7 +246,10 @@ impl Default for VoteInitV2 {
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub struct VoterWithBLSArgs {
-    #[cfg_attr(feature = "serde", serde_as(as = "[_; BLS_PUBLIC_KEY_COMPRESSED_SIZE]"))]
+    #[cfg_attr(
+        feature = "serde",
+        serde_as(as = "[_; BLS_PUBLIC_KEY_COMPRESSED_SIZE]")
+    )]
     pub bls_pub_key: [u8; BLS_PUBLIC_KEY_COMPRESSED_SIZE],
     #[cfg_attr(feature = "serde", serde_as(as = "[_; BLS_SIGNATURE_COMPRESSED_SIZE]"))]
     pub bls_proof_of_possession: [u8; BLS_SIGNATURE_COMPRESSED_SIZE],
--- a/vote-interface/src/state/vote_state_v4.rs
+++ b/vote-interface/src/state/vote_state_v4.rs
@@ -96,7 +96,7 @@ impl VoteStateV4 {
         Self {
             node_pubkey: vote_init.node_pubkey,
             authorized_voters: AuthorizedVoters::new(clock.epoch, vote_init.authorized_voter),
-            bls_pubkey_compressed: Some(vote_init.authorized_voter_bls_pubkey.into()),
+            bls_pubkey_compressed: Some(vote_init.authorized_voter_bls_pubkey),
             authorized_withdrawer: vote_init.authorized_withdrawer,
             inflation_rewards_commission_bps: vote_init.inflation_rewards_commission_bps,
             inflation_rewards_collector: vote_init.inflation_rewards_collector,


vote-interface/src/instruction.rs
```diff
@@ -34,6 +34,15 @@ pub enum VoteInstruction {
...
 
+    // Initialize a vote account using VoteInitV2
+    ///
+    /// # Account references
+    ///   0. `[WRITE]` Uninitialized vote account
+    ///   1. `[]` Rent sysvar
+    ///   2. `[]` Clock sysvar
+    ///   3. `[SIGNER]` New validator identity (node_pubkey)
+    InitializeAccountV2(VoteInitV2),
+
```
buffalojoec: This new variant has to go at the end, otherwise you'll change the discriminator for all of the remaining instructions below this.

vote-interface/src/instruction.rs
```diff
@@ -311,6 +336,37 @@ pub fn create_account_with_config(
     vec![create_ix, init_ix]
 }
 
+#[cfg(feature = "bincode")]
+pub fn create_account_v2(
```
buffalojoec: nit: why not this name to match the other one?

```suggestion
pub fn create_account_with_config_v2(
```

vote-interface/src/instruction.rs
```diff
@@ -34,6 +34,15 @@ pub enum VoteInstruction {
     ///   3. `[SIGNER]` New validator identity (node_pubkey)
     InitializeAccount(VoteInit),
 
+    // Initialize a vote account using VoteInitV2
+    ///
+    /// # Account references
+    ///   0. `[WRITE]` Uninitialized vote account
+    ///   1. `[]` Rent sysvar
+    ///   2. `[]` Clock sysvar
```
buffalojoec: I believe @jstarry was interested in moving the Vote program away from requiring sysvar accounts and instead just querying the information from the sysvar cache, and the syscalls if this program becomes BPF.

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -92,6 +92,20 @@ impl VoteStateV4 {
         }
     }
 
+    pub fn new_with_vote_init_v2(vote_init: &VoteInitV2, clock: &Clock) -> Self {
```
buffalojoec: This is fine, but kinda makes me wish I named `VoteStateV4::new` as `VoteStateV4::new_with_defaults` or something. Maybe we can deprecate `new` in favor of a `new_with_defaults`?

In fact, adding a new enum to the instructions is a breaking change in Rust, so we can probably just do the rename without deprecating, since we're onto a major release here anyway.

vote-interface/src/state/vote_instruction_data.rs
```diff
@@ -201,11 +206,53 @@ pub struct VoteInit {
...
     Voter,
     Withdrawer,
+    VoterWithBLS(
+        // BLS public key
+        #[cfg_attr(feature = "serde", serde(with = "serde_bytes"))]
+        [u8; BLS_PUBLIC_KEY_COMPRESSED_SIZE],
+        // BLS Proof of Possession
+        #[cfg_attr(feature = "serde", serde(with = "serde_bytes"))]
+        [u8; BLS_SIGNATURE_COMPRESSED_SIZE],
+    ),
```
buffalojoec: I'm thinking that these comments here are enough to signal that we should instead use a struct for the input data. That way, the fields are explicitly named.

vote-interface/src/state/vote_instruction_data.rs
```diff
@@ -201,11 +206,53 @@ pub struct VoteInit {
...
+    pub authorized_voter_bls_proof_of_possession: [u8; BLS_SIGNATURE_COMPRESSED_SIZE],
+    pub authorized_withdrawer: Pubkey,
+    pub inflation_rewards_commission_bps: u16,
+    pub inflation_rewards_collector: Pubkey,
+    pub block_revenue_commission_bps: u16,
+    pub block_revenue_collector: Pubkey,
+}
+
+impl Default for VoteInitV2 {
+    fn default() -> Self {
```
buffalojoec, line 230: Is this necessary to implement by hand? Looks like we'd have the same result with it derived.

wen-coding: Tried that and I got this:
```
error[E0277]: the trait bound `[u8; 48]: Default` is not satisfied
   --> vote-interface/src/state/vote_instruction_data.rs:216:5
    |
211 | #[derive(Debug, Default, PartialEq, Eq, Clone, Copy)]
    |                 ------- in this derive macro expansion
...
216 |     pub authorized_voter_bls_pubkey: [u8; BLS_PUBLIC_KEY_COMPRESSED_SIZE],
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Default` is not implemented for `[u8; 48]`
    |

```
This means deriving default only works for arrays smaller than 32 bytes?

Am I doing anything wrong?

vote-interface/src/state/vote_instruction_data.rs
wen-coding: Done.

vote-interface/src/state/vote_state_v4.rs
wen-coding: Sounds good, changed.

vote-interface/src/instruction.rs
wen-coding: Changed.

vote-interface/src/instruction.rs
wen-coding: Oops, thanks for pointing it out, done.

vote-interface/src/state/vote_instruction_data.rs
jstarry: Add a wrapper type for the types that cannot have `Default` derived and manually implement `Default` for them. Then you can derive `Default` for `VoteInitV2`.

vote-interface/src/state/vote_instruction_data.rs
```diff
@@ -1,14 +1,19 @@
...
+    },
     solana_clock::{Slot, UnixTimestamp},
     solana_hash::Hash,
     solana_pubkey::Pubkey,
     std::{collections::VecDeque, fmt::Debug},
 };
+#[cfg(feature = "serde")]
+use {
+    serde_derive::{Deserialize, Serialize},
+    serde_with::serde_as,
```
jstarry, line 15: None of the `serde_as` annotations are needed, can you please remove them and remove this import?

vote-interface/src/state/vote_instruction_data.rs
wen-coding: That's a good point, done.

vote-interface/src/state/vote_instruction_data.rs
wen-coding: Oops, this was leftover from a removed attempt, deleted.

vote-interface/src/state/vote_instruction_data.rs
```diff
@@ -201,11 +203,64 @@ pub struct VoteInit {
...
+    pub block_revenue_commission_bps: u16,
+    pub block_revenue_collector: Pubkey,
+}
+
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+#[derive(Debug, Default, PartialEq, Eq, Clone, Copy)]
+pub struct VoterWithBLSArgs {
+    bls_pub_key: BLSPubkeyCompressed,
+    bls_proof_of_possession: BLSSignatureCompressed,
+}
```
buffalojoec, line 265: Can you make these fields public?

vote-interface/src/state/vote_instruction_data.rs
```diff
@@ -201,11 +203,64 @@ pub struct VoteInit {
     pub commission: u8,
 }
 
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub struct BLSPubkeyCompressed(
+    #[cfg_attr(feature = "serde", serde(with = "serde_bytes"))]
+    [u8; BLS_PUBLIC_KEY_COMPRESSED_SIZE],
+);
```
buffalojoec: Do you need the wrapper structs for these types? I was able to just use `[u8; BLS_PUBLIC_KEY_COMPRESSED_SIZE]` directly in `VoteStateV4`.

https://github.com/anza-xyz/solana-sdk/blob/5390fa973897e969c5adea858079c7de1fa67d07/vote-interface/src/state/vote_state_v4.rs#L51-L56

vote-interface/Cargo.toml
```diff
@@ -54,6 +55,7 @@ cfg_eval = { workspace = true, optional = true }
 num-derive = { workspace = true }
 num-traits = { workspace = true }
 serde = { workspace = true, optional = true }
+serde_bytes = { workspace = true, optional = true }
```
buffalojoec: If we can just use `serde_as` without `serde_bytes`, we might be able to remove this additional dep!

vote-interface/src/instruction.rs
buffalojoec, line 1: Can we also add a very brief line to each `Authorize` instruction describing the constraints imposed by SIMD-0387? Maybe something like:

```rust
/// When SIMD-0387 is enabled, the `VoteAuthorize::Voter` variant is
/// disallowed for any vote accounts whose BLS pubkey is set to `Some`.
```

vote-interface/src/state/vote_instruction_data.rs
wen-coding: Done.

vote-interface/src/state/vote_instruction_data.rs
wen-coding: We can do that as well, I was just using wrapper structs because I used it in multiple places.

vote-interface/Cargo.toml
wen-coding: Removed

vote-interface/src/instruction.rs
wen-coding: Done.

buffalojoec (approved):
Lgtm!

wen-coding:
> Lgtm!

SIMD 387 is finally merged, I'm merging this now. Is the next step doing a release so we can continue developing in agave repo?

vote-interface/src/state/vote_instruction_data.rs
joncinque: Is this meant to be done in follow-up work?

vote-interface/src/state/vote_instruction_data.rs
joncinque: Was this meant to be done? I still see `serde_as` annotations

vote-interface/src/state/vote_instruction_data.rs
```diff
@@ -201,11 +206,70 @@ pub struct VoteInit {
...
+}
+
+impl Default for VoterWithBLSArgs {
+    fn default() -> Self {
+        Self {
+            bls_pub_key: [0u8; BLS_PUBLIC_KEY_COMPRESSED_SIZE],
+            bls_proof_of_possession: [0u8; BLS_SIGNATURE_COMPRESSED_SIZE],
+        }
+    }
+}
```
joncinque, line 265: It looks like this could also get similar `Default` treatment as the other struct, by deriving it from newtypes instead of doing it by hand

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -92,6 +92,20 @@ impl VoteStateV4 {
         }
     }
 
+    pub fn new(vote_init: &VoteInitV2, clock: &Clock) -> Self {
```
joncinque, line 95: I'm not sure how this is being used downstream already or how the transition is being planned, but was this breaking change intended?

wen-coding: https://github.com/anza-xyz/solana-sdk/pull/424#discussion_r2490198188

vote-interface/src/state/vote_instruction_data.rs
wen-coding: It was done in https://github.com/anza-xyz/solana-sdk/pull/424/commits/7b86969ba6bca1e4ce73ef17d6902a348407d597

But I then got comments that wrapper is not needed:
https://github.com/anza-xyz/solana-sdk/pull/424#discussion_r2537818677

So I removed the wrapper. Which one is the preferred approach?

vote-interface/src/state/vote_instruction_data.rs
wen-coding: See comment above.

vote-interface/src/state/vote_instruction_data.rs
wen-coding: Do you mean I should use a wrapper here?

vote-interface/src/state/vote_instruction_data.rs
joncinque: Ah ok, I missed that part, sorry. I would have gone with a wrapper to make the serde code simpler, but this works

vote-interface/src/state/vote_instruction_data.rs
joncinque: Yeah that's what I was getting at, but it looks like the final decision was to not use wrapper structs, so no need to do anything here

vote-interface/src/state/vote_state_v4.rs
joncinque: Ok great, thanks! Sorry for the noise then

vote-interface/src/state/vote_instruction_data.rs
buffalojoec: Maybe I'm missing something, but why is a wrapper simpler than a 4-line macro attribute?

joncinque: To be clear, this is totally fine! It's really a preference, but I tend towards default behavior where possible, since it's predictable and well understood. I find serde annotations harder to understand than newtypes. Newtypes are also more expressive.

PR #426, opened 2025-11-04
Fix hasher performance degredation caused by rust upgrade

roryharr:
#### Problem Statement:
When upgrading Rust from 1.86 to 1.87, it was found the performance of the PoH hashing degraded. Since then, all rust upgrades have been blocked.

#### Solution
After analysis it was found that inlining differences during the LLVM compilation were causing the performance difference. By force inlining the performance improved in current nightly to better than 1.87

Results were verified with proof of hash bench:
1.86 without force inline
```
test bench_poh_hash                   ... bench:   1,397,297.50 ns/iter (+/- 4,870.21)
```

1.86 with force inline
```
test bench_poh_hash                   ... bench:   1,391,815.20 ns/iter (+/- 2,922.94)
```

nightly without force inline
```
test bench_poh_hash                   ... bench:   1,474,467.60 ns/iter (+/- 6,121.96)
```

nightly with force inline
```
test bench_poh_hash                   ... bench:   1,348,648.70 ns/iter (+/- 2,663.92)
```

And on chain:

Below shows performance with nightly
Before the restart, nightly without the fix is running. 
After the restart, nightly with the fix is running.
It is also better than 1.86 (which is not shown on the graph)
Image

https://github.com/anza-xyz/agave/issues/8869

### DIFF
--- a/sha256-hasher/src/lib.rs
+++ b/sha256-hasher/src/lib.rs
@@ -20,14 +20,19 @@ pub struct Hasher {
 
 #[cfg(all(feature = "sha2", not(any(target_os = "solana", target_arch = "bpf"))))]
 impl Hasher {
+    #[inline(always)]
     pub fn hash(&mut self, val: &[u8]) {
         self.hasher.update(val);
     }
+
+    #[inline(always)]
     pub fn hashv(&mut self, vals: &[&[u8]]) {
         for val in vals {
             self.hash(val);
         }
     }
+
+    #[inline(always)]
     pub fn result(self) -> Hash {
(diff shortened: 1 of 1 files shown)


steviez:
> https://github.com/anza-xyz/agave/issues/8869

I'm trying to reproduce Rory's results in this issue, I think it would be good for us to figure that out before merging this PR. Discussion still happening over there

joncinque (approved):
This is great, thanks for digging in and finding a fix!

joncinque:
We're in the process of publishing breaking changes on the sdk, so I'll backport to v3 so Agave can get the fix quicker

PR #428, opened 2025-11-04
v3: CI: Pin cargo-release and cargo-semver-checks

joncinque:
#### Problem

cargo-semver-checks now requires Rust 1.89, but the v3 branch is still on 1.88.

#### Summary of changes

Pin cargo-semver-checks to the last known working version. Do the same with cargo-release, since that just upgraded to Rust 1.89.

### DIFF
--- a/.github/workflows/publish-rust.yml
+++ b/.github/workflows/publish-rust.yml
@@ -151,7 +151,7 @@ jobs:
       - name: Install cargo-semver-checks
         uses: taiki-e/install-action@v2
         with:
-          tool: cargo-semver-checks,cargo-release
+          tool: cargo-semver-checks@0.44.0,cargo-release@0.25.20
 
       - name: Set Git Author (required for cargo-release)
         run: |


steviez (commented):
Ha nice, I was just looking into this / about to create a PR when I saw yours.

Should we also remove this statement so that we get the semver check to run in normal CI ? I didn't `git blame` my way to see why it was disabled in the first place, but the comment there would suggest we could re-enable it now
https://github.com/anza-xyz/solana-sdk/blob/f544553db09f462cb94f9dca6b07a9824723d817/.github/workflows/main.yml#L186-L204

For example, I can see that `Check semver` was skipped for this PR:
https://github.com/anza-xyz/solana-sdk/actions/runs/19083195426/job/54517055088?pr=428

joncinque:
Running semver-checks on every PR can be very noisy, especially when the repo is in a broken (EDIT: meaning, contains breaking changes) state, so we've decided to only run it on publish. I'll remove that check.

Feel free to approve this PR if you want though :wink:

febo (approved):
Looks great!

PR #433, opened 2025-11-05
hash: Add semver-checks lints

febo:
### Problem

The `hash` crate now re-exports the types from `hash@v4` to maintain compatibility between the different major versions. But it is still not possible to publish a patch version from the maintenance/v3.x branch since semver-checks  does not support checking cross-crate elements and reports the re-exports as missing/breaking changes.

This is known limitation of semver-cheks:
* https://github.com/obi1kenobi/cargo-semver-checks/issues/638
* https://rust-lang.github.io/rust-project-goals/2024h2/cargo-semver-checks.html

### Solution

Disable the semver-checks lints for the `hash` crate so the re-exported elements are not flagged as missing.

Unsuccessful semver-checks (current): https://github.com/anza-xyz/solana-sdk/actions/runs/19088345538/job/54533431155

Successful semver-checks (this PR): https://github.com/febo/solana-sdk/actions/runs/19100305658/job/54570359447

### DIFF
--- a/hash/Cargo.toml
+++ b/hash/Cargo.toml
@@ -10,6 +10,14 @@ homepage = { workspace = true }
 license = { workspace = true }
 edition = { workspace = true }
 
+[package.metadata.cargo-semver-checks.lints]
+# cargo-semver-checks performs checks on the exported symbol graph, so re-exports
+# are considered different elements and they will be flagged as breaking changes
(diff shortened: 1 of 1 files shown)


joncinque (approved):
Thanks for digging into this, looks great!

PR #434, opened 2025-11-05
v2: instruction: Don't use serde::__private

joncinque:
#### Problem

Serde breaks the `__private` module in patch releases, which makes the current code imcompatible with newer versions of serde.

#### Summary of changes

Don't use the `__private` module, and import from std directly. The only code change is using `String::from_utf8_lossy` directly, everything else is just a global removal of `_serde::__private::` everywhere.


buffalojoec (approved):
Easy enough! Thanks!

PR #435, opened 2025-11-05
v3/address: Re-export v2

febo:
### Problem

There was a major version of `solana-address` since a `"copy"` feature was created. Although the type between the versions has not changed, the major bump makes `Address` to be considered different.

### Solution

Follow the semver trick and re-export the v2 type on v1, so they will be considered the same type.


febo:
> Just a fly-by, but I looked through and everything looks good to me. Looks like there's just a small fix required for the no-std build script

Yeah, we need to specify a version for duplicated crates. Fixed here: 3e33e4fd0264019e69c6cc52357a18c3f78d2325

address/Cargo.toml
```diff
@@ -10,79 +10,41 @@ homepage = { workspace = true }
...
-]
-rand = ["dep:rand", "atomic", "std"]
-sanitize = ["dep:solana-sanitize"]
-serde = ["dep:serde", "dep:serde_derive"]
-sha2 = ["dep:solana-sha256-hasher", "error"]
-std = ["decode", "borsh?/std", "serde?/std", "alloc"]
-alloc = []
-syscalls = ["dep:solana-define-syscall", "error"]
+atomic = ["solana-address-v2/atomic"]
+borsh = ["solana-address-v2/borsh", "alloc"]
```
febo, line 31: @joncinque Added reference to other features so the powerset generated is smaller – looks redundant but reduces the number of combinations from `33k+` to `~3k`. Without this, CI was running for more than 1 hour evaluating feature combinations and still had more combinations to test.

PR #436, opened 2025-11-06
Add remote_pubkey to Packet Meta

lijunwangs:
This is related to https://github.com/anza-xyz/agave/pull/8836 for https://github.com/issues/assigned?issue=anza-xyz%7Cagave%7C8420

Alpenglow requires sending the pubkey of the client for QUIC connection to the consumer to support censoring a client using the key if the consumer determines the client is misbehaving.

We considered alternatives of adding the information to PacketBatch, BytesPacket, and we think having it in Meta is cleanest.

### DIFF
--- a/packet/Cargo.toml
+++ b/packet/Cargo.toml
@@ -39,8 +39,8 @@ solana-frozen-abi-macro = { workspace = true, optional = true, features = ["froz
 solana-pubkey = { workspace = true }
 
 [dev-dependencies]
-solana-pubkey = { workspace = true, features = ["std"] }
 solana-packet = { path = ".", features = ["dev-context-only-utils"] }
+solana-pubkey = { workspace = true, features = ["std"] }
 static_assertions = { workspace = true }
 
 [lints]


alexpyattaev:
Overall LGTM.
I wonder if we should save a bit of memory here and store None as array of zeros (since no valid Pubkey can be made of all zeros).

lijunwangs:
> Overall LGTM. I wonder if we should save a bit of memory here and store None as array of zeros (since no valid Pubkey can be made of all zeros).

You meant instead of Option<Pubkey>, just store Pubkey?

alexpyattaev:
> You meant instead of Option, just store Pubkey?

Yes, and set it to all zeros when no Pubkey is provided. Have a getter and setter that operate on Option<Pubkey> of course. Saves 8 bytes per packet - not sure it is worth the effort though.

tao-stones:
> > You meant instead of Option, just store Pubkey?
> 
> Yes, and set it to all zeros when no Pubkey is provided. Have a getter and setter that operate on Option of course. Saves 8 bytes per packet - not sure it is worth the effort though.

imo, worth it.

lijunwangs:
> > > You meant instead of Option, just store Pubkey?
> > 
> > 
> > Yes, and set it to all zeros when no Pubkey is provided. Have a getter and setter that operate on Option of course. Saves 8 bytes per packet - not sure it is worth the effort though.
> 
> imo, worth it.

Done

packet/src/lib.rs
```diff
@@ -300,6 +302,20 @@ impl Meta {
     pub fn is_from_staked_node(&self) -> bool {
         self.flags.contains(PacketFlags::FROM_STAKED_NODE)
     }
+
+    #[inline]
+    pub fn remote_pubkey(&self) -> Option<Pubkey> {
+        if self.remote_pubkey == Pubkey::default() {
```
alexpyattaev, line 308: In such cases it may be preferable to explicitly compare with array of zeros rather than relying on the Default impl staying stable. Unlikely to be a problem for us of course.

alexpyattaev (dismissed):
LGTM (please wait for SDK SME approval)

lijunwangs:
@joncinque can you help review from sdk side?

packet/Cargo.toml
```diff
@@ -35,8 +35,10 @@ serde_derive = { workspace = true, optional = true }
 serde_with = { workspace = true, optional = true, features = ["macros"] }
 solana-frozen-abi = { workspace = true, optional = true, features = ["frozen-abi"] }
 solana-frozen-abi-macro = { workspace = true, optional = true, features = ["frozen-abi"] }
+solana-pubkey = { workspace = true, features = ["serde"] }
```
joncinque: Rather than always enabling the serde feature, it should only be enabled if the serde feature on this crate is enabled, ie back up around line 20, adding:

```
serde = ["solana-pubkey/serde"... ]
```

packet/Cargo.toml
```diff
@@ -35,8 +35,10 @@ serde_derive = { workspace = true, optional = true }
...
 
 [dev-dependencies]
+solana-address = { workspace = true, features = ["atomic"] }
```
joncinque: nit: it's a bit confusing to mix address and pubkey, so probably clearer to use the `std` feature on `solana-pubkey`
```suggestion
solana-pubkey = { workspace = true, features = ["std"] }
```

packet/src/lib.rs
```diff
@@ -72,6 +73,7 @@ pub struct Meta {
     pub addr: IpAddr,
     pub port: u16,
     pub flags: PacketFlags,
+    remote_pubkey: Pubkey,
```
joncinque, line 76: Flagging that this is technically a breaking change for anyone who builds a `Meta` without using `Default` in some way, ie:

```
let m = Meta { size, addr, port, flags };
```

It's not a problem, but probably requires a breaking release

packet/src/lib.rs
```diff
@@ -300,6 +302,20 @@ impl Meta {
...
+        if self.remote_pubkey == Pubkey::default() {
+            None
+        } else {
+            Some(self.remote_pubkey)
+        }
+    }
+
+    #[inline]
+    pub fn set_remote_pubkey(&mut self, pubkey: Pubkey) {
+        self.remote_pubkey = pubkey;
```
joncinque, line 318: nit: Up to you if this is the API you want to use, but it might be good flagging that passing `Pubkey::default()` results in nulling the remote pubkey in a doc comment

packet/src/lib.rs
lijunwangs: I think they can still use struct Meta, by putting the new flag with default
Met {
    size,
    add,
    port,
    flags,
    ..Meta::default()
}

packet/Cargo.toml
lijunwangs: done

packet/src/lib.rs
lijunwangs: added comment

packet/Cargo.toml
lijunwangs: done

joncinque (approved):
Looks good to me!

packet/src/lib.rs
joncinque: Correct, but that does still constitute a breaking change since old code will no longer compile.

lijunwangs: True, whenever we introduce a field it will break as we lack a constructor.

PR #437, opened 2025-11-07
fix(transaction): apply features to `solana-transaction-error` dep

ifiokjr:
When the `serde` feature is applied to the `solana-transaction` it should also be applied to `solana-transaction-error`. 

Also do the same for the `frozen-abi` feature.


joncinque (dismissed):
This makes sense, considering solana-transaction re-exports solana-transaction-error. Thanks for your contribution!

joncinque:
Can you just `./cargo nightly sort --workspace` and commit the results? Then we can get this in

ifiokjr:
All done! I had to install `cargo-sort` which wasn't obvious at first.

PR #439, opened 2025-11-09
program-log: Add cfg for 128-bits types

febo:
### Problem

While program-log "claims" to be upstream BPF compatible, it uses `u128`/`i128` types which are not supported.

### Solution

Any code that references `u128/`i128` is now behind a `#[cfg(not(target_arch = "bpf"))]`.

### DIFF
--- a/program-log/src/lib.rs
+++ b/program-log/src/lib.rs
@@ -473,14 +473,9 @@ mod tests {
             };
         }
 
-        unsigned_test_case!(
-            (u8, 3),
-            (u16, 5),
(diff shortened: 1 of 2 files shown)


joncinque (approved):
Looks great! We'll probably be able to revert this once eBPF properly supports 128-bit integers, correct?

febo:
> Looks great! We'll probably be able to revert this once eBPF properly supports 128-bit integers, correct?

Yes, it would probably only require a bump on the minimum version of Rust for the crate.

PR #440, opened 2025-11-12
ci: improve test-frozen-abi script

yihau:
#### Problem

I noticed that some crates' frozen-abi features can't be run individually. The reason is that our tests run them all together, which allows them to rely on tainted packages and hide potential errors. As you can see in https://github.com/anza-xyz/solana-sdk/actions/runs/19302341784/job/55200264398, it caught a missing `solana_short_vec/frozen-abi` in the message.

#### Summary of Changes

test frozen-abi crate by crate

### DIFF
--- a/scripts/test-frozen-abi.sh
+++ b/scripts/test-frozen-abi.sh
@@ -5,9 +5,4 @@ here="$(dirname "$0")"
 src_root="$(readlink -f "${here}/..")"
 cd "${src_root}"
 
-packages=$(./cargo nightly metadata --no-deps --format-version=1 | jq -r '.packages[] | select(.features | has("frozen-abi")) | .name')
-for package in $packages; do
-  echo "::group::./cargo nightly test -p $package --features frozen-abi --lib -- test_abi_ --nocapture"
-  ./cargo nightly test -p "$package" --features frozen-abi --lib -- test_abi_ --nocapture
-  echo "::endgroup::"
-done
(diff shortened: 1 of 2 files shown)


tao-stones (dismissed):
looks good for metadata

scripts/test-frozen-abi.sh
```diff
@@ -5,4 +5,9 @@ here="$(dirname "$0")"
 src_root="$(readlink -f "${here}/..")"
 cd "${src_root}"
 
-./cargo nightly test --features frozen-abi --lib -- test_abi_ --nocapture
+packages=$(./cargo nightly metadata --no-deps --format-version=1 | jq -r '.packages[] | select(.features | has("frozen-abi")) | .name')
+for package in $packages; do
+  echo "::group::./cargo nightly test -p $package --features frozen-abi --lib -- test_abi_ --nocapture"
+  ./cargo nightly test -p "$package" --features frozen-abi --lib -- test_abi_ --nocapture
+  echo "::endgroup::"
+done
```
joncinque: If you want to use cargo hack, we can simplify the change to simply:

```
./cargo nightly hack --features frozen-abi --ignore-unknown-features test --lib -- test_abi_ --nocapture
```

What do you think? We'll also need to install cargo-hack for this step in CI if we go this route

yihau: sure! let's do it: https://github.com/anza-xyz/solana-sdk/pull/440/commits/c0869000e462d10bb8474a4d4fa18df1bd388cbd

tao-stones (approved):

PR #443, opened 2025-11-14
frozen-abi: check actual ABI compatibility ( agave issue #8091 )

puhtaytow:
#### Problem

> Frozen ABI does not actually check ABI and checks API compatibility instead

[agave#8091](https://github.com/anza-xyz/agave/issues/8091)

#### Summary of Changes
Frozen ABI now supports both API compatibility checks and ABI stability checks.
- Introduced the new `StableAbi` trait.
- Updated macros to generate ABI tests conditionally.
- Added a `test_abi_` test that compares the provided abi_digest with a deterministically computed digest based on hashing multiple randomized instances of the tested type.

Its backward compatible drop-in, with optional abi compatibility check.
[example of use](https://github.com/anza-xyz/solana-sdk/pull/459/files#diff-6468a960878ea059c99ec897f59740910b7206e270eab369c147a2688830eacc)

### DIFF
--- a/frozen-abi-macro/src/lib.rs
+++ b/frozen-abi-macro/src/lib.rs
@@ -23,6 +23,42 @@ pub fn derive_abi_enum_visitor(_item: TokenStream) -> TokenStream {
     "".parse().unwrap()
 }
 
+#[cfg(not(feature = "frozen-abi"))]
+#[proc_macro_derive(StableAbi)]
+pub fn derive_stable_abi(_item: TokenStream) -> TokenStream {
+    "".parse().unwrap()
+}
+
+#[cfg(feature = "frozen-abi")]
+#[proc_macro_derive(StableAbi)]
+pub fn derive_stable_abi(item: TokenStream) -> TokenStream {
+    use {
+        quote::quote,
+        syn::{parse_macro_input, Error, Item},
+    };
+
+    let item = parse_macro_input!(item as Item);
+    let ident = match item {
+        Item::Struct(ref s) => &s.ident,
+        Item::Enum(ref e) => &e.ident,
+        Item::Type(ref t) => &t.ident,
+        _ => {
+            return Error::new_spanned(
+                item,
+                "StableAbi can only be derived for struct, enum, or type alias",
+            )
+            .to_compile_error()
+            .into();
+        }
+    };
+
+    let expanded = quote! {
+        #[automatically_derived]
+        impl ::solana_frozen_abi::stable_abi::StableAbi for #ident {}
+    };
+    expanded.into()
+}
+
 #[cfg(feature = "frozen-abi")]
 use proc_macro2::{Span, TokenStream as TokenStream2, TokenTree};
 #[cfg(feature = "frozen-abi")]
@@ -156,8 +192,8 @@ fn derive_abi_sample_enum_type(input: ItemEnum) -> TokenStream {
 #[cfg(feature = "frozen-abi")]
 fn derive_abi_sample_struct_type(input: ItemStruct) -> TokenStream {
     let type_name = &input.ident;
-    let mut sample_fields = quote! {};
     let fields = &input.fields;
+    let mut sample_fields = quote! {};
 
     match fields {
         Fields::Named(_) => {
@@ -167,9 +203,7 @@ fn derive_abi_sample_struct_type(input: ItemStruct) -> TokenStream {
                     #field_name: AbiExample::example(),
                 });
             }
-            sample_fields = quote! {
-                { #sample_fields }
-            }
+            sample_fields = quote! {{ #sample_fields }};
         }
         Fields::Unnamed(_) => {
             for _ in fields {
@@ -177,9 +211,7 @@ fn derive_abi_sample_struct_type(input: ItemStruct) -> TokenStream {
                     AbiExample::example(),
                 });
             }
-            sample_fields = quote! {
-                ( #sample_fields )
-            }
+            sample_fields = quote! {( #sample_fields )};
         }
         _ => unimplemented!("fields: {:?}", fields),
     }
@@ -195,7 +227,7 @@ fn derive_abi_sample_struct_type(input: ItemStruct) -> TokenStream {
         impl #impl_generics ::solana_frozen_abi::abi_example::AbiExample for #type_name #ty_generics #where_clause {
             fn example() -> Self {
                 ::std::println!(
-                    "AbiExample for struct: {}",
+                    "AbiExample::example for struct: {}",
                     std::any::type_name::<#type_name #ty_generics>()
                 );
                 use ::solana_frozen_abi::abi_example::AbiExample;
@@ -277,16 +309,14 @@ pub fn derive_abi_enum_visitor(item: TokenStream) -> TokenStream {
 fn quote_for_test(
     test_mod_ident: &Ident,
     type_name: &Ident,
-    expected_digest: &str,
+    expected_api_digest: &str,
+    expected_abi_digest: Option<&str>,
 ) -> TokenStream2 {
-    quote! {
-        #[cfg(test)]
-        mod #test_mod_ident {
-            use super::*;
-            use ::solana_frozen_abi::abi_example::{AbiExample, AbiEnumVisitor};
-
+    let test_api = quote! {
             #[test]
-            fn test_abi_digest() {
+            fn test_api_digest() {
+                use ::solana_frozen_abi::abi_example::{AbiExample, AbiEnumVisitor};
+
                 let mut digester = ::solana_frozen_abi::abi_digester::AbiDigester::create();
                 let example = <#type_name>::example();
                 let result = <_>::visit_for_abi(&&example, &mut digester);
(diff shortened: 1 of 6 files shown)


frozen-abi/Cargo.toml
```diff
@@ -27,6 +27,7 @@ bv = { workspace = true, features = ["serde"] }
 bytes = { workspace = true }
 dashmap = { workspace = true }
 log = { workspace = true, features = ["std"] }
+rand = { workspace = true }
```
joncinque: If we move this down into the `[target.'cfg(not(target_os = "solana"))'.dependencies]` block, and gate the new module on `[cfg(not(target_os = "solana"))]`, we can avoid the breaking change on Solana builds

scripts/build-sbf.sh
```diff
@@ -23,6 +23,8 @@ build_sbf_excludes=(
   --exclude solana-system-wasm-js
   --exclude solana-transaction
   --exclude solana-sdk
+  --exclude solana-frozen-abi
+  --exclude solana-frozen-abi-macro
```
joncinque: We'll also be able to remove this change

frozen-abi/src/lib.rs
```diff
@@ -10,7 +10,9 @@ pub mod abi_digester;
 #[cfg(feature = "frozen-abi")]
 pub mod abi_example;
 #[cfg(feature = "frozen-abi")]
-mod hash;
+pub mod hash;
+#[cfg(feature = "frozen-abi")]
+pub mod stable_abi;
```
joncinque, line 16: As mentioned elsewhere, we can gate this to keep sbf builds working:
```suggestion
#[cfg(feature = "frozen-abi")]
#[cfg(not(target_os = "solana"))]
pub mod stable_abi;
```

frozen-abi-macro/src/lib.rs
```diff
@@ -400,31 +482,39 @@ fn frozen_abi_enum_type(input: ItemEnum, expected_digest: &str) -> TokenStream {
 #[cfg(feature = "frozen-abi")]
 #[proc_macro_attribute]
 pub fn frozen_abi(attrs: TokenStream, item: TokenStream) -> TokenStream {
-    let mut expected_digest: Option<String> = None;
+    let mut api_expected_digest: Option<String> = None;
+    let mut abi_expected_digest: Option<String> = None;
+
     let attrs_parser = syn::meta::parser(|meta| {
         if meta.path.is_ident("digest") {
```
joncinque: How about deprecating `digest` as an attribute and having `api_digest` instead? It'll be clearer to downstream users what's really going on.

frozen-abi-macro/src/lib.rs
```diff
@@ -400,31 +482,39 @@ fn frozen_abi_enum_type(input: ItemEnum, expected_digest: &str) -> TokenStream {
...
     parse_macro_input!(attrs with attrs_parser);
 
-    let Some(expected_digest) = expected_digest else {
-        return Error::new_spanned(
-            TokenStream2::from(item),
-            "the required \"digest\" = ... attribute is missing.",
-        )
-        .to_compile_error()
-        .into();
-    };
```
joncinque, line 421: What's the motivation for removing these checks? With this PR as it is, someone can use `frozen_abi` without specifying either an `abi_digest` or a `digest`, which would be dangerous if they're expecting something to be tested.

If that makes sense to you, let's keep the existing check that `api_expected_digest` is set, and we can eventually do a breaking change that also requires `abi_expected_digest` to be set.

joncinque (commented):
Great job overall! A couple of questions and suggestions on my side

frozen-abi-macro/src/lib.rs
alexpyattaev: We initially wanted this to be more of a drop-in update, but making things clear is probably preferable indeed.

frozen-abi-macro/src/lib.rs
alexpyattaev: Right now you can have a struct deriving AbiExample and not specifying digests and it would work. This is often used for nested structs.

puhtaytow: Exactly as Alex wrote. I recently have seen a bunch of structs with missing `digest` attribute.

Removed the checks as it seemed redundant. Brought it back. Thank you for pointing out.

brooksprumo:
What does it look like when there is an ABI change? The current code lets you dump the output to diff. Is that still the case? Is the new version deterministic too? (I assume it has to be, but I also see some random shuffling/distribution. And lastly, does the random sampling affect runtime?

frozen-abi-macro/src/lib.rs
joncinque: I might be misunderstanding something, but the `AbiExample` and `frozen_abi` macros are not the same -- this is changing the `frozen_abi` macro behavior by not forcing someone to specify `digest`.

So with this change, someone could do:

```
#[derive(AbiExample)]
#[frozen_abi]
```

Which would be useless.

frozen-abi-macro/src/lib.rs
```diff
@@ -400,31 +482,39 @@ fn frozen_abi_enum_type(input: ItemEnum, expected_digest: &str) -> TokenStream {
...
 #[proc_macro_attribute]
 pub fn frozen_abi(attrs: TokenStream, item: TokenStream) -> TokenStream {
-    let mut expected_digest: Option<String> = None;
+    let mut api_expected_digest: Option<String> = None;
+    let mut abi_expected_digest: Option<String> = None;
+
     let attrs_parser = syn::meta::parser(|meta| {
-        if meta.path.is_ident("digest") {
-            expected_digest = Some(meta.value()?.parse::<LitStr>()?.value());
+        if meta.path.is_ident("api_digest") {
```
joncinque: This becomes a breaking change -- up to you if you want a break, but to avoid the break, it would be better to support both `"digest"` and `"api_digest"`

joncinque (commented):
Just two last bits, then this should be able to go in!

alexpyattaev:
Thank you for review @joncinque ! I wonder though what could poasibly be causing the CI failures here. Do you have any idea!

puhtaytow:
@brooksprumo 

> What does it look like when there is an ABI change?

```
thread 'epoch_slots::EpochSlots_frozen_abi::test_abi_digest' panicked at gossip/src/epoch_slots.rs:254:5:
assertion `left == right` failed: ABI layout has changed!
  left: "outdatedhash"
 right: "EUSmnof24f7q8258wPBLiW7VAdtZXfRa1mN8j6aVwDAB"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test epoch_slots::EpochSlots_frozen_abi::test_abi_digest ... FAILED

failures:

failures:
    epoch_slots::EpochSlots_frozen_abi::test_abi_digest
```

For the API change it remains as it was before.


> The current code lets you dump the output to diff. Is that still the case?

The current code is still there as it was. This PR is focused on bringing the separate functionality, `StableAbi`. It doesn't interfere with the already existing `AbiExample`.

> Is the new version deterministic too? 

Yes, for the `AbiExample` nothing has changed, and the `StableAbi` RNG always starts from the same seed.

> does the random sampling affect runtime?

It does affect the tests runtime, but only for types which are `StableAbi` enabled.  It would generate thousands of instances of the tested type to hash these together into the `abi_digest`.

frozen-abi-macro/src/lib.rs
puhtaytow: Yeah, after thinking about it for a while.. it would indeed be better to support both. Thank you :pray:

frozen-abi-macro/src/lib.rs
puhtaytow: > I might be misunderstanding something

More likely, im missing something. This is the example from top of my head https://github.com/anza-xyz/agave/blob/master/gossip/src/epoch_slots.rs#L260

joncinque: There are many different macros in this crate. Specifically, your change affects the `frozen_abi` macro, ie https://github.com/anza-xyz/solana-sdk/blob/0da89f297a7450c1129c40c064ced83a04903f74/vote-interface/src/state/vote_state_v4.rs#L23

With the code as it is, someone can just specify `#[frozen_abi]` without a digest, which would be a problem. Let's ensure that someone does specify the api_digest at the very least

puhtaytow: Oki. I have added the check in `#[frozen_abi]`.
Now, when it can't see `digest` or `api_digest` attr it would result in 

```
error: missing required attribute: #[frozen_abi(api_digest = "...")]
  --> account/src/lib.rs:70:60
   |
70 |     #[cfg_attr(feature = "frozen-abi", derive(AbiExample), frozen_abi())]
```

Thank you for pointing it out :pray:

frozen-abi-macro/src/lib.rs
```diff
@@ -400,31 +491,48 @@ fn frozen_abi_enum_type(input: ItemEnum, expected_digest: &str) -> TokenStream {
...
-    let Some(expected_digest) = expected_digest else {
+    if api_expected_digest.is_none() {
         return Error::new_spanned(
             TokenStream2::from(item),
-            "the required \"digest\" = ... attribute is missing.",
+            "missing required attribute: #[frozen_abi(api_digest = \"...\")]",
         )
         .to_compile_error()
         .into();
-    };
```
joncinque, line 421: Since we're requiring `api_digest` to be specified, we can keep this old let...else expression and have `frozen_abi_struct_type` / `frozen_abi_enum_type` / `frozen_abi_type_alias` all take `api_expected_digest` as a `&str` instead of `Option<&str>`

puhtaytow: Makes sense. Thank you :1st_place_medal:

puhtaytow:
> Looks great to me, thanks for integrating all the feedback!

Absolutely. Thank you for the feedback! :)

alexpyattaev (approved):
Excellent work than you!

puhtaytow:
> Excellent work than you!

Thank you! 🙏 🙏 🙏

PR #444, opened 2025-11-15
packet: add constructor to Meta

lijunwangs:
We added a private field remote_pubkey, the "Meta {"  based construction is broken. Adding a constructor to simplify creating the object.


lijunwangs:
@joncinque can you please take a look at this?

PR #446, opened 2025-11-17
Remove too granular Copy implementation

lvboudre:
Since an Address is a wrapper type around a Copy type ([u8;32]) it uselessly break older user-code during crates upgrade which didn't required `copy` in older version of the crate. Moreover the feature-flag `copy` is not documented in the documentation

### DIFF
--- a/address/src/lib.rs
+++ b/address/src/lib.rs
@@ -91,8 +91,7 @@ pub const PDA_MARKER: &[u8; 21] = b"ProgramDerivedAddress";
 #[cfg_attr(feature = "bytemuck", derive(Pod, Zeroable))]
 #[cfg_attr(feature = "dev-context-only-utils", derive(Arbitrary))]
 #[cfg_attr(not(feature = "decode"), derive(Debug))]
-#[cfg_attr(feature = "copy", derive(Copy))]
-#[derive(Clone, Default, Eq, Ord, PartialEq, PartialOrd)]
+#[derive(Clone, Default, Eq, Ord, PartialEq, PartialOrd, Copy)]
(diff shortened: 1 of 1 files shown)


febo:
> Since an Address is a wrapper type around a Copy type ([u8;32]) it uselessly break older user-code during crates upgrade which didn't required `copy` in older version of the crate.

The rationale is to avoid unintentional copies when writing programs, as discussed [here](https://github.com/anza-xyz/solana-sdk/issues/342). Versions `v1.1` and `v2.0` are compatible, which should help with upgrades.

> Moreover the feature-flag `copy` is not documented in the documentation

This is a good point – not clear why the documentation does not show the `copy` feature requirement for the `impl Copy`. I will have a look at it.

lvboudre:
I guess its subjective, since Pubkey is constant size implements Clone, PartialEq, Eq, Hash it makes sense to always implement Copy for such type. I never had issue with unintentional Copy seems a no-issue to me.

febo:
> I guess its subjective, since Pubkey is constant size implements Clone, PartialEq, Eq, Hash it makes sense to always implement Copy for such type. I never had issue with unintentional Copy seems a no-issue to me.

Yeah, it is definitely subjective. The `Copy` derive is still available, but having it behind a feature means that this behaviour is intentional. The `"copy"` feature is also being added to other crates for the same reason – programs should avoid unnecessary memory copies, since it is more expensive than passing a reference.

joncinque:
Thanks for your PR, but we won't integrate this change for the moment. We'll keep gathering feedback and see if we should add `Copy` back by default.

PR #448, opened 2025-11-19
shred-version: Move test over from Agave

steviez:
This test currently lives in `solana-ledger` crate; however, the test is for logic in the `solana-shred-version` crate in this repo so the test should live here. The test will be deleted from Agave after this PR merges.

Agave permalink:
https://github.com/anza-xyz/agave/blob/6545467a8981b758101eb7cc36aef2064e431ecc/ledger/src/shred.rs#L1067-L1088

### DIFF
--- a/shred-version/Cargo.toml
+++ b/shred-version/Cargo.toml
@@ -17,5 +17,8 @@ solana-hard-forks = { workspace = true }
 solana-hash = { workspace = true }
 solana-sha256-hasher = { workspace = true }
 
+[dev-dependencies]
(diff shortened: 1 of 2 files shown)


steviez:
@buffalojoec - GH suggested you review, seemingly since you authored https://github.com/anza-xyz/solana-sdk/commit/9e61f39bdfd6bdac402ae7a60653d78750d06828. If you don't feel comfortable reviewing, lemme know and I can find someone else. I'm not too hopeful on getting a review from the original author that has most of the code in `git blame` 😅

PR #450, opened 2025-11-21
ci: Pin cargo-release version

febo:
### Problem

Currently the publish workflow is failing since the latest `cargo-release` version requires Rust `1.90.0` to build it.

### Solution

Switch the action to `taiki-e/install-action@v2` so the binary is used.


joncinque:
We should probably bump the rust version soon, but no particular rush

PR #451, opened 2025-11-22
sysvar: Fix UB in sysvar loading

jamie-osec:
https://github.com/anza-xyz/solana-sdk/pull/257#discussion_r2553138460

Slices must be initialized, but the implementation of `Sysvar::get` creates a `&mut [u8]` from an uninitialized buffer. This causes UB and miscompiles.
Resolve this by introducing a new private function which takes a mutable _pointer_ (may be uninitialized) and use that instead.

### DIFF
--- a/sysvar/src/lib.rs
+++ b/sysvar/src/lib.rs
@@ -182,23 +182,27 @@ macro_rules! impl_sysvar_get {
             }
         }
     };
-    ($sysvar_id:expr) => {
+    // This version requires the `bincode` feature
+    ($sysvar_id:expr, $serialized_size:expr) => {
         fn get() -> Result<Self, $crate::__private::ProgramError> {
             // Allocate uninitialized memory for the sysvar struct
-            let mut uninit = ::core::mem::MaybeUninit::<Self>::uninit();
-            let size: u64 = size_of::<Self>() as u64;
+            let mut uninit = ::core::mem::MaybeUninit::<[u8; $serialized_size]>::uninit();
             let sysvar_id = $sysvar_id;
+            // Attempt to load the sysvar data using the provided sysvar id.
+            // SAFETY: `size` is exactly the size of the buffer passed
             unsafe {
-                // Attempt to load the sysvar data using the provided sysvar id.
-                // SAFETY: `size` is exactly the size of the buffer passed
                 $crate::get_sysvar_unchecked(
                     uninit.as_mut_ptr() as *mut u8,
                     ::core::ptr::addr_of!(sysvar_id) as *const u8,
                     0,
-                    size,
+                    $serialized_size as u64,
                 )?;
-                // SAFETY: If `get_sysvar_unchecked` succeeded, the buffer will have been initialized
-                Ok(uninit.assume_init())
+            }
+            // SAFETY: If `get_sysvar_unchecked` succeeded, the buffer will have been initialized
+            let buffer: [u8; $serialized_size] = unsafe { uninit.assume_init() };
+            match ::bincode::deserialize::<Self>(&buffer) {
+                Ok(var) => Ok(var),
+                Err(_) => Err($crate::__private::ProgramError::UnsupportedSysvar),
             }
         }
     };
@@ -226,8 +230,9 @@ pub fn get_sysvar(
 
 /// Same as [`get_sysvar`], but takes a raw pointer and does not check destination length.
 ///
-/// # SAFETY:
-/// - `dst` must point to `length` mutable bytes
+/// # Safety
+///
+/// The caller must ensure that `dst` is valid for writes of `length` bytes.
 unsafe fn get_sysvar_unchecked(
     dst: *mut u8,
     sysvar_id: *const u8,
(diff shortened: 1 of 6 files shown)


rustopian:
Good find, and thanks for moving on it quickly @jamie-osec! This is similar to an auditor-noted issue in p-token (since remedied), and it should be fixed.

Your proposed fix isn't quite there yet. I'm away from my machine, but for now I've approved the CI so that you can easily see its objections.

sysvar/src/lib.rs
```diff
@@ -185,19 +185,21 @@ macro_rules! impl_sysvar_get {
...
+                // SAFETY: `size` is exactly the size of the buffer passed
+                $crate::get_sysvar_unchecked(
+                    uninit.as_mut_ptr() as *mut u8,
+                    ::core::ptr::addr_of!(sysvar_id) as *const u8,
+                    0,
+                    size,
+                )?;
+                // SAFETY: If `get_sysvar_unchecked` succeeded, the buffer will have been initialized
+                Ok(uninit.assume_init())
+            }
```
febo, line 206: nit: Since we have two different "safety" comments, we could split the `unsafe` block into two different ones. Also, `addr_of!` is *soft-deprecated* – the suggestion is to use `&raw const` instead.

```suggestion
            // Attempt to load the sysvar data using the provided sysvar id.
            //
            // SAFETY: `size` is exactly the size of the uninitialized buffer passed.
            unsafe {
                $crate::get_sysvar_unchecked(
                    uninit.as_mut_ptr() as *mut u8,
                    &raw const sysvar_id as *const u8,
                    0,
                    size,
                )?;
            }                
            // SAFETY: If `get_sysvar_unchecked` succeeded, the buffer will have been initialized.
            Ok(unsafe { uninit.assume_init() })
```

sysvar/src/lib.rs
```diff
@@ -219,13 +221,26 @@ pub fn get_sysvar(
     let sysvar_id = sysvar_id as *const _ as *const u8;
     let var_addr = dst as *mut _ as *mut u8;
 
+    unsafe { get_sysvar_unchecked(var_addr, sysvar_id, offset, length) }
+}
+
+/// Same as [`get_sysvar`], but takes a raw pointer and does not check destination length.
+///
+/// # SAFETY:
+/// - `dst` must point to `length` mutable bytes
```
febo:
```suggestion
/// # Safety
///
/// The caller must ensure that `dst` is valid for writes of `length` bytes.
```

jamie-osec:
I've been investigating this further to confirm the issue and I've realized the problem is _actually_ that `get_sysvar` returns the raw sysvar buffer data. In the case of `Rent` (and other sysvars), this data is stored as `bincode`-serialized data, whereas the old implementation deserialized the data with the desired type.
This means that:
- There is a mismatch between the length param the program passes (i.e. 24 bytes) and the sysvar buffer data (17 bytes)
- This causes the runtime to return `OFFSET_LENGTH_EXCEEDS_SYSVAR`-
- Which `get_sysvar` maps to `UnsupportedSysvar`
- Which manifests as the error `The arguments provided to a program instruction were invalid.`

While the original issue is library-level UB according to the documentation, it does _not_ appear to be currently considered UB by Rust, and Miri does not report an issue.

jamie-osec:
Addressed the UB nits, and pushed a fix for the root cause.
- Added a `serialized_size` param to the macro, and test that the size is indeed correct for the serialized data
- Adjusted the `to_bytes` function to serialize rather than using raw bytes to match the runtime
- Fallback to the original sysvars when `bincode` is not enabled

I'm not familiar enough with the context to know if this breakage was intended - it seems unlikely given that the signature is still returning `Self` which in many cases is not possible without deserialization. If it is, I can revert back to just the UB fixes.

(Relevant: #87)

rustopian:
> While the original issue is library-level UB according to the documentation, it does _not_ appear to be currently considered UB by Rust, and Miri does not report

Indeed, this is why it didn't raise red flags earlier, but it's still a good idea to fix it, as it violates the contract of `from_raw_parts_mut` and could in the future cause issues.

rustopian:
> Fallback to the original sysvars when bincode is not enabled

Thanks for tracking down the issue!

FYI, we likely want to use the general syscall when possible due to https://github.com/anza-xyz/solana-sdk/issues/235.

febo:
> > Fallback to the original sysvars when bincode is not enabled
> 
> Thanks for tracking down the issue!
> 
> FYI, we likely want to use the general syscall when possible due to #235.

Using bincode on-chain will most likely offset any gains from `sol_get_sysvar`. An alternative is to create 1-byte aligned sysvar representations so we can zero-copy the data; or keep using `sol_get_<NAME>_sysvar`.

rustopian:
> Using bincode on-chain will most likely offset any gains from `sol_get_sysvar`. An alternative is to create 1-byte aligned sysvar representations so we can zero-copy the data; or keep using `sol_get_<NAME>_sysvar`.

Yes, I'm definitely not suggesting we require/encourage bincode.

rustopian:
@jamie-osec, thanks much for your work here. Due to the need to fix the 3.1.0 crate in the interim, I'm going to close this PR and just revert to the old (3.0.0) sysvar behavior for now. Then, we'll ensure everything works properly and cleanly implement an update. There are clearly some test coverage improvements needed.

Your UB fix will be incorporated and I'll make sure you receive attribution for it. I'll tag you on the new PR once open.

jamie-osec:
Great, thanks for the update

PR #452, opened 2025-11-23
[bls-signatures] Add support for custom payloads for bls PoP

samkim-crypto:
#### Problem

Currently, the PoP generation API in the crate doesn't support custom payload. To support https://github.com/solana-foundation/solana-improvement-documents/pull/387 where custom [message](https://github.com/solana-foundation/solana-improvement-documents/pull/387/files#diff-45cf6d94dd49af7b43169bd0433976c0a0596b8d18659267a7cabe37de639ad7R124) is used for domain separation, we need the interface for the caller to generate and verify PoP on custom payload.

#### Summary of Changes

I added a field `payload: Option<&[u8]>` to the PoP generation and verification functions. If this field is `None`, then the functions just use the BLS public key as bytes for the payload. If this field is `Some(...)`, then it uses the provided bytes to generate the message for PoP.

### DIFF
--- a/bls-signatures/src/hash.rs
+++ b/bls-signatures/src/hash.rs
@@ -15,7 +15,14 @@ pub fn hash_message_to_point(message: &[u8]) -> G2Projective {
 }
 
 /// Hash a pubkey to a G2 point
-pub(crate) fn hash_pubkey_to_g2(public_key: &PubkeyProjective) -> G2Projective {
-    let pubkey_bytes = public_key.0.to_compressed();
-    G2Projective::hash_to_curve(&pubkey_bytes, POP_DST, &[])
+pub(crate) fn hash_pubkey_to_g2(
+    public_key: &PubkeyProjective,
+    payload: Option<&[u8]>,
+) -> G2Projective {
+    if let Some(bytes) = payload {
+        G2Projective::hash_to_curve(bytes, POP_DST, &[])
+    } else {
+        let public_key_bytes = public_key.0.to_compressed();
+        G2Projective::hash_to_curve(&public_key_bytes, POP_DST, &[])
+    }
 }
--- a/bls-signatures/src/proof_of_possession.rs
+++ b/bls-signatures/src/proof_of_possession.rs
@@ -18,7 +18,7 @@ use {
 /// Domain separation tag used when hashing public keys to G2 in the proof of
 /// possession signing and verification functions. See the
 /// [standard](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#section-4.2.3).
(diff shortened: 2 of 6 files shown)


bls-signatures/src/proof_of_possession.rs
```diff
@@ -18,7 +18,7 @@ use {
 /// Domain separation tag used when hashing public keys to G2 in the proof of
 /// possession signing and verification functions. See the
 /// [standard](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#section-4.2.3).
-pub const POP_DST: &[u8] = b"BLS_SIG_BLS12381G1_XMD:SHA-256_SSWU_RO_POP_";
+pub const POP_DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_";
```
samkim-crypto, line 21: I also realized that this domain separation tag was written for G1 while we hash the message on G2 for PoP, so I fixed this.

bls-signatures/src/hash.rs
```diff
@@ -15,7 +15,14 @@ pub fn hash_message_to_point(message: &[u8]) -> G2Projective {
...
 /// Hash a pubkey to a G2 point
-pub(crate) fn hash_pubkey_to_g2(public_key: &PubkeyProjective) -> G2Projective {
-    let pubkey_bytes = public_key.0.to_compressed();
-    G2Projective::hash_to_curve(&pubkey_bytes, POP_DST, &[])
+pub(crate) fn hash_pubkey_to_g2(
+    public_key: &PubkeyProjective,
+    payload: Option<&[u8]>,
+) -> G2Projective {
+    if let Some(bytes) = payload {
+        G2Projective::hash_to_curve(bytes, POP_DST, &[])
```
zz-sol, line 23: nits: this means `bytes = Some(&[])` and `bytes=None` have different outputs. Is this behavior acceptable? Or should `bytes=None` also hash the public key?

samkim-crypto: Yeah, I was thinking that this is acceptable.

I originally made the function take in `prefix: Option<&[u8]>` instead. It would then concatenate `prefix || pubkey_bytes` and then use this as the payload. This is probably the safer API because the public key will always be hashed. But if someone wants to add a suffix instead of prefix `pubkey_bytes || suffix`, for example, they cannot use this API. This is why I made the function to take in `paylod: Option<&[u8]>` instead.

I think hashing on `bytes = Some(&[])` is not any safer than hashing any other string that does not have the public key bytes in it anyway, so there doesn't seem to be a justification to specialize for this case. Of course, if you think specializing addressing the case `Some(&[])` to hash the public key bytes instead, then let me know 😄

zz-sol: sounds good to me

PR #453, opened 2025-11-24
[vote-interface] Rename `BLS_SIGNATURE_COMPRESSED_SIZE` to `BLS_PROOF_OF_POSSESSION_COMPRESSED_SIZE`

samkim-crypto:
#### Summary of Changes

Sorry, this is a small follow-up to https://github.com/anza-xyz/solana-sdk/pull/424. I just wanted to rename `BLS_SIGNATURE_COMPRESSED_SIZE` to `BLS_PROOF_OF_POSSESSION_COMPRESSED_SIZE`. BLS PoP and signatures are the same size, but it would be good to name the constants more precisely. Completely my fault for not commenting on this in https://github.com/anza-xyz/solana-sdk/pull/424.

### DIFF
--- a/vote-interface/src/state/mod.rs
+++ b/vote-interface/src/state/mod.rs
@@ -33,7 +33,7 @@ pub(crate) mod vote_state_deserialize;
 /// Size of a BLS public key in a compressed point representation
 pub const BLS_PUBLIC_KEY_COMPRESSED_SIZE: usize = 48;
 
-/// Size of a BLS proof of possession in a compressed point representation
+/// Size of a BLS proof of possession in a compressed point representation; matches BLS signature size
 pub const BLS_PROOF_OF_POSSESSION_COMPRESSED_SIZE: usize = 96;
 
 // Maximum number of votes to keep around, tightly coupled with epoch_schedule::MINIMUM_SLOTS_PER_EPOCH


vote-interface/src/state/mod.rs
```diff
@@ -33,8 +33,8 @@ pub(crate) mod vote_state_deserialize;
 /// Size of a BLS public key in a compressed point representation
 pub const BLS_PUBLIC_KEY_COMPRESSED_SIZE: usize = 48;
 
-/// Size of a BLS signature in a compressed point representation
-pub const BLS_SIGNATURE_COMPRESSED_SIZE: usize = 96;
+/// Size of a BLS proof of possession in a compressed point representation
```
wen-coding: nit: do we need to mention it is the same size as a signature?

vote-interface/src/state/mod.rs
```diff
@@ -33,8 +33,8 @@ pub(crate) mod vote_state_deserialize;
...
+pub const BLS_PROOF_OF_POSSESSION_COMPRESSED_SIZE: usize = 96;
```
wen-coding, line 37: This looks good to me, but I don't know what is normally the styling requirement in this repo, is this constant a bit too long? @buffalojoec

samkim-crypto: Yeah I can definitely update to `BLS_POP_COMPRESSED_SIZE`. I have been going with mvines style so far, which is to avoid abbreviations as much as possible. Will wait for other's opinion.

buffalojoec: Actually I kinda like the long, descriptive constant. 😂

samkim-crypto: Okay great! I'll go with the current constant name.

PR #457, opened 2025-11-24
[WIP] Use sol_get_sysvar instead of sol_get_<NAME>_sysvar, now with deserialization

rustopian:
[WIP]

This addresses Issue https://github.com/anza-xyz/solana-sdk/issues/235, fixing two issues with the [original](https://github.com/anza-xyz/solana-sdk/pull/257) (reverted) PR:

1. The contract of `from_raw_parts_mut` was being violated, in a way that miri considered safe, but which could have theoretically caused trouble in the future.
2. The implementation assumed that the number of bytes returned by the runtime was exactly
`mem::size_of::<Self>()`, which was not always true. `Rent`, for example, contains a `u64`, an `f64` and a `u8`. In memory the compiler pads this struct to 24 bytes, but the runtime stores its contents as a 17‑byte `bincode` serialization. The `get_sysvar` call thus returns only 17 bytes. Passing a 24‑byte buffer results in an `OFFSET_LENGTH_EXCEEDS_SYSVAR` error and undefined behavior when uninitialized bytes are interpreted as f64.

In order to address this second issue, this PR provides 1‑byte‑aligned representations of the sysvars that match the on‑chain encoding. Then, we can zero-copy the bytes directly into those representations and convert them to the canonical structs without using `bincode`.

This is lightweight but bespoke deserialization to avoid requiring any given dependency.

### DIFF
--- a/define-syscall/src/definitions.rs
+++ b/define-syscall/src/definitions.rs
@@ -49,12 +49,12 @@ define_syscall!(fn sol_get_return_data(data: *mut u8, length: u64, program_id: *
 // - `is_writable` (`u8`): `true` if the instruction requires the account to be writable
 define_syscall!(fn sol_get_processed_sibling_instruction(index: u64, meta: *mut u8, program_id: *mut u8, data: *mut u8, accounts: *mut u8) -> u64);
 
-// these are to be deprecated once they are superceded by sol_get_sysvar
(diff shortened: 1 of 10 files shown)


rustopian:
Replacing for sake of cleaner PR history

PR #458, opened 2025-11-25
Use sol_get_sysvar instead of sol_get_<NAME>_sysvar, fixed

rustopian:
[WIP]

This addresses Issue https://github.com/anza-xyz/solana-sdk/issues/235, fixing two issues with the [original](https://github.com/anza-xyz/solana-sdk/pull/257) (reverted) PR:

1. The contract of `from_raw_parts_mut` was being violated, in a way that miri considered safe, but which could have theoretically caused trouble in the future.
2. The implementation assumed that the number of bytes returned by the runtime was exactly
`mem::size_of::<Self>()`, which was not always true. `Rent`, for example, contains a `u64`, an `f64` and a `u8`. In memory the compiler pads this struct to 24 bytes, but the runtime stores its contents as a 17‑byte `bincode` serialization. The `get_sysvar` call thus returns only 17 bytes. Passing a 24‑byte buffer results in an `OFFSET_LENGTH_EXCEEDS_SYSVAR` error and undefined behavior when uninitialized bytes are interpreted as f64.

Note: this PR only updates the three sysvars `Rent`, `EpochSchedule`, and `EpochRewards`, which require two different approaches as commented below. `Clock` and `LastRestartSlot` do not require special handling.

Thanks to [jamie-osec](https://github.com/jamie-osec) for investigating the cause here and noticing the `from_raw_parts_mut` violation. I'll add you as contributor to the final.

### DIFF
--- a/sysvar/src/epoch_rewards.rs
+++ b/sysvar/src/epoch_rewards.rs
@@ -193,6 +193,7 @@ mod tests {
 
         let data = bincode::serialize(&expected).unwrap();
         assert_eq!(data.len(), 81);
+        assert_eq!(data.len() + 15, core::mem::size_of::<EpochRewards>());
 
         crate::tests::mock_get_sysvar_syscall(&data);
         let got = EpochRewards::get().unwrap();
--- a/sysvar/src/lib.rs
+++ b/sysvar/src/lib.rs
@@ -191,20 +191,17 @@ macro_rules! impl_sysvar_get {
             let var_addr = var.as_mut_ptr() as *mut u8;
             let length = core::mem::size_of::<Self>().saturating_sub($padding);
             let sysvar_id_ptr = (&$sysvar_id) as *const _ as *const u8;
-            // SAFETY: The allocation is valid for `size_of::<Self>()`. We load
-            // `(size - padding)` bytes from the syscall, which matches bincode
-            // serialization. The remaining `padding` bytes are then zeroed.
-            let result =
-                unsafe { $crate::get_sysvar_unchecked(var_addr, sysvar_id_ptr, 0, length as u64) };
+            // SAFETY: The allocation is valid for `size_of::<Self>()`. We zero
+            // the padding bytes first, then load `(size - padding)` bytes from
+            // the syscall, which matches bincode serialization.
+            let result = unsafe {
+                var_addr.add(length).write_bytes(0, $padding);
+                $crate::get_sysvar_unchecked(var_addr, sysvar_id_ptr, 0, length as u64)
+            };
             match result {
-                Ok(()) => {
-                    // SAFETY: All bytes now initialized: syscall filled data
-                    // bytes and we zeroed padding.
-                    unsafe {
-                        var_addr.add(length).write_bytes(0, $padding);
-                        Ok(var.assume_init())
-                    }
-                }
+                // SAFETY: All bytes initialized: padding was zeroed above,
+                // syscall filled the data bytes.
+                Ok(()) => Ok(unsafe { var.assume_init() }),
                 // Unexpected errors are folded into `UnsupportedSysvar`.
                 Err(_) => Err($crate::__private::ProgramError::UnsupportedSysvar),
             }
--- a/sysvar/src/rent.rs
+++ b/sysvar/src/rent.rs
@@ -152,6 +152,7 @@ mod tests {
         };
         let data = bincode::serialize(&expected).unwrap();
         assert_eq!(data.len(), 17);
+        assert_eq!(data.len() + 7, core::mem::size_of::<Rent>());
 
         crate::tests::mock_get_sysvar_syscall(&data);
         let got = Rent::get().unwrap();


sysvar/src/epoch_rewards.rs
```diff
@@ -156,15 +156,112 @@
...
+    total_points: u128,
+    total_rewards: u64,
+    distributed_rewards: u64,
+    active: u8, // bool as u8
+}
+
+const _: () = assert!(core::mem::size_of::<EpochRewardsPacked>() == 81);
+
+impl From<EpochRewardsPacked> for EpochRewards {
+    fn from(p: EpochRewardsPacked) -> Self {
```
rustopian: @febo please lmk what you think of this approach.

Originally I did a whole macro setup, but it was overkill given that we're only worried about 3 padded sysvars here.

Compile-time checks just below ensure 1) field parity between the two structs (repr(C) and packed) and 2) size

define-syscall/src/lib.rs
```diff
@@ -9,7 +9,8 @@ pub mod definitions;
 ))]
 #[macro_export]
 macro_rules! define_syscall {
-    (fn $name:ident($($arg:ident: $typ:ty),*) -> $ret:ty) => {
+    ($(#[$attr:meta])* fn $name:ident($($arg:ident: $typ:ty),*) -> $ret:ty) => {
```
rustopian, line 12: Enabling `deprecated` to be passed through (as in the original sol_get_sysvar PR)

sysvar/src/epoch_schedule.rs
```diff
@@ -121,15 +121,68 @@
...
 
+#[repr(C, packed)]
+#[derive(Clone, Copy)]
+struct EpochSchedulePacked {
+    slots_per_epoch: u64,
+    leader_schedule_slot_offset: u64,
+    warmup: u8, // bool as u8
+    first_normal_epoch: u64,
+    first_normal_slot: u64,
+}
```
febo: Using `packed` is problematic when there are unaligned fields – both `first_normal_epoch` and `first_normal_slot` are unaligned so can't be read directly.

What do you think if we change the definition of all sysvars to be like?
```rust
pub struct EpochSchedule {
    /// The maximum number of slots in each epoch.
    slots_per_epoch: [u8; 8],

    /// A number of slots before beginning of an epoch to calculate
    /// a leader schedule for that epoch.
    leader_schedule_slot_offset: [u8; 8],

    /// Whether epochs start short and grow.
    warmup: u8,

    /// The first epoch after the warmup period.
    ///
    /// Basically: `log2(slots_per_epoch) - log2(MINIMUM_SLOTS_PER_EPOCH)`.
    first_normal_epoch: [u8; 8],

    /// The first slot after the warmup period.
    ///
    /// Basically: `MINIMUM_SLOTS_PER_EPOCH * (2.pow(first_normal_epoch) - 1)`.
    first_normal_slot: [u8; 8],
}

impl EpochSchedule {
    pub fn slots_per_epoch(&self) -> u64 {
        u64::from_le_bytes(self.slots_per_epoch)
    }

    pub fn leader_schedule_slot_offset(&self) -> u64 {
        u64::from_le_bytes(self.leader_schedule_slot_offset)
    }

    // Could also return the `u8` value directly.
    pub fn warmup(&self) -> Result<bool, ProgramError> {
        match self.warmup {
            0 => Ok(false),
            1 => Ok(true),
            _ => Err(ProgramError::InvalidAccountData),
        }
    }

     // other fields...

    pub fn first_normal_slot(&self) -> u64 {
        u64::from_le_bytes(self.first_normal_slot)
    }
}
```
There is no "penalty" for using `u64::from_le_bytes`, although the implementation is a bit more verbose.

sysvar/src/epoch_rewards.rs
febo: Left a comment below about the use of `packed`.

sysvar/src/epoch_schedule.rs
rustopian: Ah, of course. The packed struct works currently since it’s just used for `From`, but that of course copies everything out.

I’ll implement your approach, thanks.

(Despite appearances here, I did in fact attend your optimization talk which mentioned this. exact. pattern. Ahem.)

jamie-osec: Doesn't this constitute a breaking change (public fields removed)? I believe the internal-only packed version is better here - an unaligned read from a packed field produces [the same IR/code](https://godbolt.org/z/Wo4vfezhr) as reading it as an array.

rustopian: @febo I've implemented your suggestion in this PR, and it's much nicer looking (and zero copy). However, as @jamie-osec mentions, the accessors are a breaking change. What formerly was e.g. `.slots_per_epoch` becomes `.slots_per_epoch()`

Note: I've only implemented it for the problematic `Rent`, `EpochRewards`, and `EpochSchedule` for now, pending our decision here.

febo: > Doesn't this constitute a breaking change (public fields removed)?

Yes, that is a drawback of the approach. We should be probably ok since crates are versioned individually.

> I believe the internal-only packed version is better here - an unaligned read from a packed field produces [the same IR/code](https://godbolt.org/z/Wo4vfezhr) as reading it as an array.

The issue is that we do an extra copy to return the "non-packed" version, which is what we would like to avoid. Returning the "packed" version is problematic – easy to end up in UB by taking a reference to an unaligned field.

rustopian: cc @joncinque

joncinque: I think this break would be annoying for Agave and some downstream programs that directly access any of these fields. We've introduced `Pod*` types for other sysvars, so I would recommend doing the same thing here

LStan:
Why did you deprecate the syscalls since `3.0.0` if the current version of `solana-define-syscall` is already `4.0.1`?

rustopian:
> Why did you deprecate the syscalls since `3.0.0` if the current version of `solana-define-syscall` is already `4.0.1`?

Thanks for noticing this! The original work here was done back on 3.x, will update

rustopian:
The compromise we've currently ended up with is this:
- @febo's custom `Rent` suggestion gives us all the benefits of our approach without breaking, and works fine for `EpochRewards` too;
- @joncinque's `Pod` suggestion implemented, but only on `PodEpochSchedule` (due to its mid-struct padding), which:
    - can convert to `EpochSchedule` via `From` (so `get` behavior remains exactly the same)
    - or also can be fetched directly (`PodEpochSchedule::fetch()`) to use zero-copy accessors instead of copying/padding to re-align it.

In this way, there is some benefit for people using this sysvar without them needing to make any changes. However, if they also want to take advantage of the zero-copy accessors, they're available.

`Rent` and `EpochRewards` users do not need to make any updates to get full benefit.

LStan:
There is a `bool` in `EpochRewards`. Is it safe to load it without `Pod`?

rustopian:
> There is a `bool` in `EpochRewards`. Is it safe to load it without `Pod`?

`EpochRewards` is 16-aligned:

```
#[repr(C, align(16))]
...
pub struct EpochRewards {
    /// The starting block height of the rewards distribution in the current
    /// epoch
    pub distribution_starting_block_height: u64,

...

    /// Whether the rewards period (including calculation and distribution) is
    /// active
    pub active: bool,
}
```

So when we call `impl_sysvar_get` for it, we specify 15 zeroes of padding:
```
impl Sysvar for EpochRewards {
    impl_sysvar_get!(id(), 15);
}
```

...just as we do for `Rent` (but with `7`). This zero-fills in the resulting `get()` function:
```
macro_rules! impl_sysvar_get {
    ...
    ($sysvar_id:expr, $padding:literal) => {
        fn get() -> Result<Self, $crate::__private::ProgramError> {
            ...
            let result =
                unsafe { $crate::get_sysvar_unchecked(var_addr, sysvar_id_ptr, 0, length as u64) };
            match result {
                Ok(()) => {
                    // SAFETY: All bytes now initialized: syscall filled data
                    // bytes and we zeroed padding.
                    unsafe {
                        var_addr.add(length).write_bytes(0, $padding);
                        Ok(var.assume_init())
                    }
                }
                // Unexpected errors are folded into `UnsupportedSysvar`.
                Err(_) => Err($crate::__private::ProgramError::UnsupportedSysvar),
            }
        }
    };
    // Variant for sysvars without padding (struct size matches bincode size).
    ($sysvar_id:expr) => {
        $crate::impl_sysvar_get!($sysvar_id, 0);
    };
}
```

We can do this easily (as in `Rent`) since the `bool` is at the end of the struct. The only reason `EpochSchedule` needs special handling (which this PR now does with `PodEpochSchedule`) is that it has a `bool` right in the middle:

```
pub struct EpochSchedule {
    pub slots_per_epoch: u64,
    pub leader_schedule_slot_offset: u64,
    pub warmup: bool,
    pub first_normal_epoch: u64,
    pub first_normal_slot: u64,
}
```

LStan:
I meant the casting of `u8` to `bool`. What if that byte for some reason contains not 0 or 1?

sysvar/src/epoch_schedule.rs
```diff
@@ -119,17 +119,124 @@
...
+        u64::from_le_bytes(self.slots_per_epoch)
+    }
+
+    pub fn leader_schedule_slot_offset(&self) -> u64 {
+        u64::from_le_bytes(self.leader_schedule_slot_offset)
+    }
+
+    pub fn warmup(&self) -> bool {
+        self.warmup != 0
+    }
```
febo, line 177: This works, but I am not sure whether we need to be more specific and really test the only valid values.
```suggestion
    pub fn warmup(&self) -> Result<bool, ProgramError> {
        match self.warmup {
            0 => Ok(false),
            1 => Ok(true),
            _ => Err(ProgramError::InvalidAccountData),
        }
    }
```
The other alternative is to return the `u8` value.
```suggestion
    pub fn warmup(&self) -> u8 {
        self.warmup
    }
```

febo:
> I meant the casting of `u8` to `bool`. What if that byte for some reason contains not 0 or 1?

@rustopian This is a good point, we can't cast a value directly to `bool` unless we validate it is a `0` or `1`. Left a comment about this [here](https://github.com/anza-xyz/solana-sdk/pull/458#discussion_r2584828248).

rustopian:
> > I meant the casting of `u8` to `bool`. What if that byte for some reason contains not 0 or 1?
> 
> @rustopian This is a good point, we can't cast a value directly to `bool` unless we validate it is a `0` or `1`. Left a comment about this [here](https://github.com/anza-xyz/solana-sdk/pull/458#discussion_r2584828248).

I implemented your suggestion, but the question still remains for `EpochRewards`, which uses the same `impl_sysvar_get` approach as `Rent` (pad the end) but with a final item `bool`.

But I think we could make a case for allowing this (otherwise, I'll have to do `PodEpochRewards` too). Because of an upstream invariant: sysvar data is created by the Solana runtime, and of course serializes `bool` as exactly `0` for false, `1` for true.

e.g.
```
// SAFETY: upstream invariant: the sysvar data is created exclusively
// by the Solana runtime and serializes bool as 0x00 or 0x01.
```

Incidentally, this *also* can apply to `PodEpochSchedule::warmup()`. Yes, we get UB on values other than `0` or `1`, but Solana runtime guarantees this is a serialized `bool` and thus cannot be anything else in `solana` target.

 I can simply validate when target is not `solana`.

Wdyt @febo @LStan

LStan:
I meant exactly what @febo said. You could use `bool` for `warmup` in `PodEpochSchedule`, but you didn't because it's UB.

sysvar/src/lib.rs
```diff
@@ -206,6 +241,54 @@ pub fn get_sysvar(
     #[cfg(not(target_os = "solana"))]
     let result = crate::program_stubs::sol_get_sysvar(sysvar_id, var_addr, offset, length);
 
+    match result {
+        solana_program_entrypoint::SUCCESS => Ok(()),
+        OFFSET_LENGTH_EXCEEDS_SYSVAR => Err(solana_program_error::ProgramError::InvalidArgument),
+        _ => Err(solana_program_error::ProgramError::UnsupportedSysvar),
+    }
```
joncinque, line 245: nit: This return was changed to not explicitly check for `SYSVAR_NOT_FOUND`. It doesn't make a difference in the end, but is there a reason for this change?

sysvar/src/lib.rs
```diff
@@ -174,11 +177,43 @@ macro_rules! impl_sysvar_get {
...
+            let result =
+                unsafe { $crate::get_sysvar_unchecked(var_addr, sysvar_id_ptr, 0, length as u64) };
+            match result {
+                Ok(()) => {
+                    // SAFETY: All bytes now initialized: syscall filled data
+                    // bytes and we zeroed padding.
+                    unsafe {
+                        var_addr.add(length).write_bytes(0, $padding);
+                        Ok(var.assume_init())
+                    }
```
joncinque: nit: The Pinocchio version does the zero-ing before the syscall, should we do the same in this implementation?

sysvar/src/epoch_rewards.rs
```diff
@@ -163,8 +163,35 @@ pub use {
...
+            num_partitions: 7,
+            parent_blockhash: solana_hash::Hash::new_unique(),
+            total_points: 1234567890,
+            total_rewards: 100,
+            distributed_rewards: 10,
+            active: true,
+        };
+
+        let data = bincode::serialize(&expected).unwrap();
+        assert_eq!(data.len(), 81);
```
joncinque, line 195: How about adding a check that `data.len() + 15 == mem::size_of::<EpochRewards>()` to make sure padding doesn't change?

sysvar/src/rent.rs
```diff
@@ -128,9 +128,32 @@ pub use {
...
+    #[serial]
+    #[cfg(feature = "bincode")]
+    fn test_rent_get() {
+        let expected = Rent {
+            lamports_per_byte_year: 123,
+            exemption_threshold: 2.5,
+            burn_percent: 7,
+        };
+        let data = bincode::serialize(&expected).unwrap();
+        assert_eq!(data.len(), 17);
```
joncinque, line 154: Same here, let's check that the padding is correct with `data.len() + 7 == size_of::<Rent>()`

febo:
> > > I meant the casting of `u8` to `bool`. What if that byte for some reason contains not 0 or 1?
> > 
> > 
> > @rustopian This is a good point, we can't cast a value directly to `bool` unless we validate it is a `0` or `1`. Left a comment about this [here](https://github.com/anza-xyz/solana-sdk/pull/458#discussion_r2584828248).
> 
> I implemented your suggestion, but the question still remains for `EpochRewards`, which uses the same `impl_sysvar_get` approach as `Rent` (pad the end) but with a final item `bool`.
> 
> But I think we could make a case for allowing this (otherwise, I'll have to do `PodEpochRewards` too). Because of an upstream invariant: sysvar data is created by the Solana runtime, and of course serializes `bool` as exactly `0` for false, `1` for true.
> 
> e.g.
> 
> ```
> // SAFETY: upstream invariant: the sysvar data is created exclusively
> // by the Solana runtime and serializes bool as 0x00 or 0x01.
> ```
> 
> Incidentally, this _also_ can apply to `PodEpochSchedule::warmup()`. Yes, we get UB on values other than `0` or `1`, but Solana runtime guarantees this is a serialized `bool` and thus cannot be anything else in `solana` target.
> 
> I can simply validate when target is not `solana`.
> 
> Wdyt @febo @LStan

I double checked this and you are right, we can rely on the runtime always storing valid values. The sysvar account is always serialized with bincode and bincode encodes `bool` values as `0` or `1`. We could revert the change to `warmup` as well.

rustopian:
@febo great, I rolled back `warmup()` change

sysvar/src/epoch_schedule.rs
rustopian: Subsequent discussion in the PR landed on a safety comment instead of this validation

sysvar/src/epoch_schedule.rs
```diff
@@ -119,17 +119,126 @@
...
 
+/// Pod (Plain Old Data) representation of [`EpochSchedule`] with no padding.
+///
+/// This type can be safely loaded via `sol_get_sysvar` without undefined behavior.
+/// Provides performant zero-copy accessors as an alternative to the `EpochSchedule` type.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct PodEpochSchedule {
+    slots_per_epoch: [u8; 8],
+    leader_schedule_slot_offset: [u8; 8],
```
joncinque, line 138: nit: we *could* have these two as `u64`, which would ensure that the type aligns to 8 bytes and make accesses of these two fields cheaper in non-VM contexts. What do you think? Too much micro-optimization?

joncinque (commented):
Sorry for the lateness, I had a pending review that I apparently never submitted. Mostly small things on my side, looks great overall!

sysvar/src/lib.rs
rustopian: `SYSVAR_NOT_FOUND` already fell through to ` UnsupportedSysvar`.

I didn't keep the explicit arm here because, besides being redundant, it made `SYSVAR_NOT_FOUND` look special when it's not; it's folded into `UnsupportedSysvar` just like any other error.

I don't see any reason for the original form:
```
        SYSVAR_NOT_FOUND => Err(solana_program_error::ProgramError::UnsupportedSysvar),
        // Unexpected errors are folded into `UnsupportedSysvar`.
        _ => Err(solana_program_error::ProgramError::UnsupportedSysvar),
```
... unless maybe it's a very soft reminder to consider handling `SYSVAR_NOT_FOUND` differently in some future version.

sysvar/src/epoch_schedule.rs
rustopian: Well, aligning this type to 8 bytes defeats its purpose (zero-copy accessing the packed format sysvar data, which is not aligned).

Besides, AFAIK `u64::from_le_bytes()` should be optimized to a direct load in contexts that matter?

sysvar/src/rent.rs
rustopian: Done in 93fc197

sysvar/src/epoch_rewards.rs
rustopian: Makes sense to me. Done in 93fc197

sysvar/src/lib.rs
rustopian: Sure, that's a bit cleaner. Done in 93fc197

sysvar/src/epoch_schedule.rs
joncinque: Ah right, I was thinking just in the context of the syscall, where we allocate the stack memory ourselves, and so have control over the alignment. Probably better to keep it usable in any zero-copy context

sysvar/src/lib.rs
joncinque: > ... unless maybe it's a very soft reminder to consider handling SYSVAR_NOT_FOUND differently in some future version

I saw it as a very soft reminder to handle unexpected errors differently in some future version :grin: 

Probably fine to keep this as you've done it

joncinque (commented):
Looks good to me! Please get approval from the others before merging though

sysvar/src/lib.rs
rustopian: Right, that. I'll take a look at this tomorrow and see if maybe some errors need to be better handled here.

PR #461, opened 2025-11-26
Add wincode support for Entry components

cpubot:
`agave`'s `Entry` is comprised of a few core types exposed by the `solana-sdk` crate. [Agave currently has its own wincode-derive implementations for those types](https://github.com/anza-xyz/agave/blob/4890ca53fda1da169a75679cafc1212ade8ced1b/entry/src/wincode.rs). This PR migrates those implementations here to make this crate the source of truth for those implementations, as well as making them available to users of `solana-sdk` more broadly.

This isn't meant to provide comprehensive support for `wincode` to `solana-sdk`, but rather to begin introducing `wincode` as a feature to a few subcrates, laying the groundwork for future implementations.

Implemented types:
- `MessageHeader`
- `CompiledInstruction`
- `Address`
- `Hash`
- `Signature`
- `legacy::Message`
- `MessageAddressTableLookup`
- `v0::Message`
- `VersionedTransaction`

I can also break each into a separate PR if that's preferable, though there will be some PR stacking as many of these types have dependencies in this list.

### DIFF
--- a/message/src/versions/mod.rs
+++ b/message/src/versions/mod.rs
@@ -346,6 +346,7 @@ impl SchemaWrite for VersionedMessage {
         match src {
             VersionedMessage::Legacy(message) => LegacyMessage::size_of(message),
             // +1 for message version prefix
+            #[expect(clippy::arithmetic_side_effects)]
             VersionedMessage::V0(message) => Ok(1 + v0::Message::size_of(message)?),
         }
     }


message/src/versions/v0/loaded.rs
```diff
@@ -23,6 +25,7 @@ pub struct LoadedMessage<'a> {
 /// by readonly and writable.
 #[derive(Clone, Default, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+#[cfg_attr(feature = "wincode", derive(SchemaWrite, SchemaRead))]
```
jstarry: nit: doesn't look this is necessary for `Entry` support

message/src/compiled_instruction.rs
```diff
@@ -17,15 +19,18 @@ use {solana_address::Address, solana_sanitize::Sanitize};
     derive(Deserialize, Serialize),
     serde(rename_all = "camelCase")
 )]
+#[cfg_attr(feature = "wincode", derive(SchemaWrite, SchemaRead))]
```
jstarry, line 22: Can you explain why other structs need the `struct_extensions` attribute but this doesn't have it? I get that it's related to providing helper mut methods to the fields but I don't have a mental model for when it's nice to add or not yet

Cargo.toml
```diff
@@ -195,11 +195,11 @@ openssl = "0.10.72"
...
 parking_lot = "0.12"
 pbkdf2 = { version = "0.11.0", default-features = false }
-proc-macro2 = "1.0.93"
+proc-macro2 = "1.0.101"
 proptest = "1.6"
 qstring = "0.7.2"
 qualifier_attr = { version = "0.2.2", default-features = false }
-quote = "1.0.35"
+quote = "1.0.41"
 rand = "0.8.5"
```
jstarry: Looks like one of these dependencies is causing `rand` to be resolved to `0.9.x` in the `Cargo.lock` file. Do you mind reviewing the lock file changes to ensure they're all ok? I think the 0.9 upgrade is fine for `rand` since we're [working on upgrading `rand` in agave](https://github.com/anza-xyz/agave/issues/4738) to that version as well. But it should probably be reflected here in the `Cargo.toml` too if it's the new min?

message/src/versions/v0/loaded.rs
cpubot: Yep -- extraneous. Will remove

message/src/compiled_instruction.rs
cpubot: Sure. [`struct_extensions`](https://docs.rs/wincode/latest/wincode/#struct_extensions) adds additional methods onto the type that take care of the boilerplate when manually initializing the fields of a `MaybeUninit` struct via `Reader` reads or direct in-memory writes. 

Specifically, it adds `uninit_<field>_mut`, `read_<field>`, and `write_uninit_<field>`.
- `uninit_<field_name>_mut`: Gets a mutable `MaybeUninit` projection to the `<field_name>` slot.
- `read_<field_name>`: Reads into a `MaybeUninit`’s `<field_name>` slot from the given `Reader`.
- `write_uninit_<field_name>`: Writes a `MaybeUninit`’s <field_name> slot with the given value.

This is generally only needed when manually implementing `SchemaRead` where there is some specific logic that precludes reading the type directly from a source. `LegacyMessage` is the prototypical case here. Because the first byte of `VersionedMessage` may actually represent `num_required_signatures` of `LegacyMessage`'s `MessageHeader`, we must read the first byte of the source payload in isolation and check whether it represents a versioned message prefix. If the payload is a `LegacyMessage`, we now have 1 less byte in the source, and simply calling `LegacyMessage::read` without that byte in the source would be incorrect. `struct_extensions` gives us an ergonomic way to handle this, where we call `write_uninit_num_required_signatures` with the byte we already read, and then read the subsequent fields as normal.

Without these helpers, we'd need to project each field of the struct manually before writing or reading (the boilerplate for which `struct_extensions` is used to avoid):
```rs
unsafe {&mut *(&raw mut (*dst.as_mut_ptr()).<field>).cast::<MaybeUninit<field_type>>()}
```

`struct_extensions` is needed on both `MessageHeader` and `LegacyMessage` because `MessageHeader` is part of the `LegacyMessage` payload, and they need to be initialized together in this case.

```rs
let mut msg = MaybeUninit::<LegacyMessage>::uninit();
// We've already read the variant byte which, in the legacy case, represents
// the `num_required_signatures` field.
// As such, we need to write the remaining fields into the message manually,
// as calling `LegacyMessage::read` will miss the first field.
let header_uninit = LegacyMessage::uninit_header_mut(&mut msg);

MessageHeader::write_uninit_num_required_signatures(variant, header_uninit);
MessageHeader::read_num_readonly_signed_accounts(reader, header_uninit)?;
MessageHeader::read_num_readonly_unsigned_accounts(reader, header_uninit)?;

LegacyMessage::read_account_keys(reader, &mut msg)?;
LegacyMessage::read_recent_blockhash(reader, &mut msg)?;
LegacyMessage::read_instructions(reader, &mut msg)?;
```

Cargo.toml
cpubot: It looks like the newer `rand` dependency is only pulled in as a result of adding `proptest` to `transaction`, which is only a `dev-dependency`. 
image

Crates that depend on an older version of rand appear to still point to their original version 
image

jstarry: Ah ok, that's good, thanks for checking

buffalojoec (approved):
Super cool! The library looks awesome.

transaction/Cargo.toml
joncinque: I might have missed something, but since the `wincode` feature isn't always enabled for test builds, we won't actually run the test that's gated on the `wincode` feature -- can you add the `wincode` feature here to make sure we do run the test?

Alternatively, we can add `wincode` to `dev-context-only-utils`, but I'd prefer adding the feature here since there's nothing in `dev-context-only-utils` that requires `wincode` at the moment.

joncinque (changes requested):
Sorry for the fly-by, but the code looks great to me! Just a comment on enabling deps for tests

transaction/Cargo.toml
cpubot: Done! https://github.com/anza-xyz/solana-sdk/pull/461/commits/07c53aea0fe9faf131b8b4235fede4cb45d99488

Cargo.lock
```diff
@@ -3816,6 +3915,15 @@ dependencies = [
...
 ]
 
+[[package]]
+name = "solana-short-vec"
+version = "3.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "79fb1809a32cfcf7d9c47b7070a92fa17cdb620ab5829e9a8a9ff9d138a7a175"
+dependencies = [
+ "serde_core",
+]
```
joncinque, line 3925: Sorry I'm only noticing this now, but we might have some issues with two versions of the same crate in our build at some point. At worst, we'll need to add a patch in Cargo.toml to force using the local solana-short-vec. We could potentially bring solana-short-vec down with the wincode crate if needed too.

Either way, it looks like we're ok for now, so nothing to do yet

cpubot: I could also move the wincode implementations into the sdk as well if it becomes problematic!

apfitzge:
> Tagging @apfitzge -- need someone from tx-metadata to approve

@tao-stones or @t-nelson could you look at this? I briefly looked when the work began, should help replay

tao-stones (approved):
lgtm for tx-metadata

PR #463, opened 2025-11-27
[bls-signatures] Refactor pubkey, signature, and proof of possession modules

samkim-crypto:
#### Problem

As I was tackling https://github.com/anza-xyz/solana-sdk/issues/474, I realized that each of the `pubkey.rs`, `signature.rs`, and `proof_of_possession.rs` files are becoming quite long and harder to read.

#### Summary of Changes

In preparation for adding the new `PubkeyAffine`, `SignatureAffine`, and `ProofOfPossesionAffine` types, I first did a refactor of the `pubkey.rs`, `signature.rs`, and `proof_of_possession.rs` types.

Each each `pubkey`, `signature`, and `proof_of_possession` modules are divided into 
- `points.rs`: contains the `Projective` type and will soon host the `Affine` type
- `bytes.rs`: contains the serialized types like `Pubkey` (soon to be `PubkeyUncompressed`) and `PubkeyCompressed`.
- `conversion.rs`: contains any conversion logic between points


bls-signatures/src/signature/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+    BLS_SIGNATURE_COMPRESSED_BASE64_SIZE, BLS_SIGNATURE_COMPRESSED_SIZE,
+};
+#[cfg(not(target_os = "solana"))]
+pub use {
+    bytes::AsSignature,
+    points::{AsSignatureProjective, SignatureProjective, VerifiableSignature},
+};
+
+#[cfg(test)]
+mod tests {
```
zz-sol, line 16: nits: should test be refactored into test.rs?

zz-sol: and similarly for a few other mod.rs?

samkim-crypto: Yeah I think that is a good idea for a cleaner structure. I'll address that in a follow-up and merge this since this can be done in parallel with the renaming.

PR #464, opened 2025-11-27
[bls12-381-core] Add bls12-381 syscall (SIMD-388)

samkim-crypto:


samkim-crypto:
Replaced with https://github.com/anza-xyz/agave/pull/9759

PR #467, opened 2025-11-29
frozen-abi, script:  (bugfix) some frozen-abi tests are not performed

puhtaytow:
#### Problem

After merging https://github.com/anza-xyz/solana-sdk/pull/440 some frozen-abi tests are skipped which results in, not checked digest.

I discovered this issue accidentially on `solana-transaction::Transaction`, where after changing digest, frozen-abi script still pass green.

#### Summary of changes
Brought back `cargo test` instead of `cargo hack`.

### DIFF
--- a/scripts/test-frozen-abi.sh
+++ b/scripts/test-frozen-abi.sh
@@ -5,5 +5,5 @@ here="$(dirname "$0")"
 src_root="$(readlink -f "${here}/..")"
 cd "${src_root}"
 
-./cargo nightly hack --features frozen-abi --ignore-unknown-features test --lib -- test_abi_digest --nocapture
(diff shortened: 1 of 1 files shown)


puhtaytow:
Closing this one as Alex found the root of the issue https://github.com/anza-xyz/solana-sdk/pull/477

alexpyattaev:
I'm not actually convinced my solution is "better" in any measurable way.

puhtaytow:
> I'm not actually convinced my solution is "better" in any measurable way.

Right. Lets see what Jon would say about your PR then. To me it seem more appropriate fix, but he might think differently, then I could always re-open this one.

PR #468, opened 2025-11-29
transaction: enable frozen-abi StableAbi

puhtaytow:
#### Description

This PR enables `StableAbi`, functionality for frozen-abi types in the crate.

#### Summary of changes
All frozen-abi types got enabled `StableAbi`.

### DIFF
--- a/transaction/src/lib.rs
+++ b/transaction/src/lib.rs
@@ -110,6 +110,8 @@
 //! # Ok::<(), anyhow::Error>(())
 //! ```
 
+#[cfg(feature = "frozen-abi")]
+use solana_frozen_abi_macro::{frozen_abi, AbiExample, StableAbi};
 #[cfg(feature = "serde")]
 use {
     serde_derive::{Deserialize, Serialize},
@@ -177,8 +179,11 @@ const NONCED_TX_MARKER_IX_INDEX: u8 = 0;
 /// redundantly specifying the fee-payer is not strictly required.
 #[cfg_attr(
     feature = "frozen-abi",
-    derive(solana_frozen_abi_macro::AbiExample),
-    solana_frozen_abi_macro::frozen_abi(digest = "ADDDuk3dAZJ5hDxue8v4btH7nhEyngxUpXaC7A4k8gyQ")
+    derive(AbiExample, StableAbi),
+    frozen_abi(
+        api_digest = "ADDDuk3dAZJ5hDxue8v4btH7nhEyngxUpXaC7A4k8gyQ",
+        abi_digest = "B9cGfEfYQtYccbPKXWBRoWhdfLRjaUBimic11Vp5Bjzx"
+    )
 )]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[cfg_attr(feature = "wincode", derive(SchemaWrite, SchemaRead))]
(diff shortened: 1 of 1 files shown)


alexpyattaev (dismissed):
LGTM!

apfitzge:
Think this got broken with TXV1 - and never saw this review request since I was out at the time.

puhtaytow:
> never saw this review request since I was out at the time.

I think i didnt do good job at advertising the feature. If you feel for there are as well https://github.com/anza-xyz/solana-sdk/pull/473 and https://github.com/anza-xyz/solana-sdk/pull/472 both are small :pray:

apfitzge:
Probably a dumb question - but why do we care the type has a stable abi? We only care if the serialization/deserialization is stable, no?

alexpyattaev:
> Probably a dumb question - but why do we care the type has a stable abi? We only care if the serialization/deserialization is stable, no?

This is what StableAbi does (and yes, StableSerialization might have been a better name). It allows us to avoid implementing the code to generate examples of Transaction everywhere we may want to confirm that a packet containing a transaction has consistent wire format. But also this prevents someone from accidentally breaking the Transaction type itself. 

We might want to add checks on the in-memory representation too for things like Geyser though...

puhtaytow:
In addition to what Alex wrote (thank you), i would add [docs](https://github.com/anza-xyz/solana-sdk/blob/master/frozen-abi/src/lib.rs) and the [test](https://github.com/anza-xyz/solana-sdk/blob/master/frozen-abi-macro/src/lib.rs#L344-L362) which ensure the stability.

> We might want to add checks on the in-memory representation too for things like Geyser though...

++

apfitzge:
Hmm I've seen the `test_abi_digest` code now. FYI it uses bincode though, which does not actually serialize transactions/messages correctly anymore (since txv1 has been added).

alexpyattaev:
> Hmm I've seen the `test_abi_digest` code now. FYI it uses bincode though, which does not actually serialize transactions/messages correctly anymore (since txv1 has been added).

oh. guess we need to switch it to wincode then. should be a drop-in replacement.

apfitzge:
> oh. guess we need to switch it to wincode then. should be a drop-in replacement.

hopefully, but only if all the types calling it actually implement the traits I think.

puhtaytow:
> > Hmm I've seen the `test_abi_digest` code now. FYI it uses bincode though, which does not actually serialize transactions/messages correctly anymore (since txv1 has been added).
> 
> oh. guess we need to switch it to wincode then. should be a drop-in replacement.

ill have a look at it shortly.

transaction/src/lib.rs
```diff
@@ -201,6 +207,39 @@ pub struct Transaction {
...
+            .map(|_| Signature::from(std::array::from_fn(|_| rng.random::<u8>())))
+            .collect();
+        let accounts: Vec<AccountMeta> = (0..rng.random_range(1..100))
+            .map(|_| AccountMeta {
+                pubkey: Address::new_from_array(rng.random()),
+                is_signer: rng.random(),
+                is_writable: rng.random(),
+            })
+            .collect();
+        let data: Vec<u8> = (0..rng.random_range(1..1000))
```
alexpyattaev: we probably do not want this to grow too large. Else we'd get a transaction that is many KB in size which is not realistic..

puhtaytow: good catch, thanks! :)

apfitzge (approved):

alexpyattaev (approved):

PR #472, opened 2025-11-29
vote-interface: enable StableAbi

puhtaytow:
#### Description

This PR enables `StableAbi`, functionality for vote interface types.

#### Summary of changes
- all types that used frozen-abi got enabled `StableAbi` thru derive StableAbiSample

### DIFF
--- a/vote-interface/src/state/mod.rs
+++ b/vote-interface/src/state/mod.rs
@@ -5,7 +5,7 @@ use arbitrary::Arbitrary;
 #[cfg(feature = "serde")]
 use serde_derive::{Deserialize, Serialize};
 #[cfg(feature = "frozen-abi")]
-use solana_frozen_abi_macro::AbiExample;
+use solana_frozen_abi_macro::{AbiExample, StableAbi, StableAbiSample};
 use {
     crate::authorized_voters::AuthorizedVoters,
     solana_clock::{Epoch, Slot, UnixTimestamp},
@@ -49,7 +49,7 @@ pub const VOTE_CREDITS_GRACE_SLOTS: u8 = 2;
 // Maximum number of credits to award for a vote; this number of credits is awarded to votes on slots that land within the grace period. After that grace period, vote credits are reduced.
 pub const VOTE_CREDITS_MAXIMUM_PER_SLOT: u8 = 16;
 
-#[cfg_attr(feature = "frozen-abi", derive(AbiExample))]
+#[cfg_attr(feature = "frozen-abi", derive(AbiExample, StableAbi, StableAbiSample))]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[derive(Default, Debug, PartialEq, Eq, Copy, Clone)]
 #[cfg_attr(feature = "dev-context-only-utils", derive(Arbitrary))]
@@ -102,7 +102,7 @@ impl Lockout {
     }
 }
 
-#[cfg_attr(feature = "frozen-abi", derive(AbiExample))]
+#[cfg_attr(feature = "frozen-abi", derive(AbiExample, StableAbi, StableAbiSample))]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[derive(Default, Debug, PartialEq, Eq, Copy, Clone)]
 #[cfg_attr(feature = "dev-context-only-utils", derive(Arbitrary))]
@@ -139,7 +139,7 @@ impl From<Lockout> for LandedVote {
     }
 }
 
-#[cfg_attr(feature = "frozen-abi", derive(AbiExample))]
+#[cfg_attr(feature = "frozen-abi", derive(AbiExample, StableAbi, StableAbiSample))]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[derive(Debug, Default, PartialEq, Eq, Clone)]
 #[cfg_attr(feature = "dev-context-only-utils", derive(Arbitrary))]
@@ -152,7 +152,7 @@ pub struct BlockTimestamp {
 const MAX_ITEMS: usize = 32;
 
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
-#[cfg_attr(feature = "frozen-abi", derive(AbiExample))]
+#[cfg_attr(feature = "frozen-abi", derive(AbiExample, StableAbi, StableAbiSample))]
 #[derive(Debug, PartialEq, Eq, Clone)]
 #[cfg_attr(feature = "dev-context-only-utils", derive(Arbitrary))]
 pub struct CircBuf<I> {
@@ -210,7 +210,7 @@ pub mod serde_compact_vote_state_update {
         solana_serde_varint as serde_varint, solana_short_vec as short_vec,
     };
 
-    #[cfg_attr(feature = "frozen-abi", derive(AbiExample))]
+    #[cfg_attr(feature = "frozen-abi", derive(AbiExample, StableAbi, StableAbiSample))]
     #[derive(serde_derive::Deserialize, serde_derive::Serialize)]
     struct LockoutOffset {
         #[serde(with = "serde_varint")]
@@ -306,7 +306,7 @@ pub mod serde_tower_sync {
         solana_serde_varint as serde_varint, solana_short_vec as short_vec,
     };
 
-    #[cfg_attr(feature = "frozen-abi", derive(AbiExample))]
+    #[cfg_attr(feature = "frozen-abi", derive(AbiExample, StableAbi, StableAbiSample))]
     #[derive(serde_derive::Deserialize, serde_derive::Serialize)]
     struct LockoutOffset {
         #[serde(with = "serde_varint")]
--- a/vote-interface/src/state/vote_instruction_data.rs
+++ b/vote-interface/src/state/vote_instruction_data.rs
@@ -1,5 +1,5 @@
 #[cfg(feature = "frozen-abi")]
-use solana_frozen_abi_macro::{frozen_abi, AbiExample};
+use solana_frozen_abi_macro::{frozen_abi, AbiExample, StableAbi, StableAbiSample};
 use {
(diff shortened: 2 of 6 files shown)


alexpyattaev (dismissed):
LGTM, but I am wondering if we could ever derive those impls instead.

puhtaytow:
> LGTM, but I am wondering if we could ever derive those impls instead.

Yup, good question.. i'm open for any solution that would work.

From my pov,  if we want to avoid raw distribution impl for frozen-abi / stableabi enabled types and derive it thru macro/visitor, then it would mean, every custom type (even those non frozen-abi enabled) used in the api would need to have the implementation. 

It might turn out that we have more raw impls than if we did it only for the already frozen-abi enabled types.

alexpyattaev:
@ilya-bobyr couod you recommend a course of action here? Should we invest into a macro to derive these?

vote-interface/src/state/vote_instruction_data.rs
```diff
@@ -32,6 +35,23 @@ pub struct Vote {
     pub timestamp: Option<UnixTimestamp>,
 }
 
+#[cfg(feature = "frozen-abi")]
+impl solana_frozen_abi::rand::prelude::Distribution<Vote>
+    for solana_frozen_abi::rand::distr::StandardUniform
+{
+    fn sample<R: solana_frozen_abi::rand::Rng + ?Sized>(&self, rng: &mut R) -> Vote {
+        let slots: Vec<Slot> = (0..rng.random_range(0..1000))
```
alexpyattaev: tower votes fit into 1 MTU. 1000 u64's would not =)

vote-interface/src/state/vote_instruction_data.rs
```diff
@@ -64,6 +87,23 @@ pub struct VoteStateUpdate {
     pub timestamp: Option<UnixTimestamp>,
 }
 
+#[cfg(feature = "frozen-abi")]
+impl solana_frozen_abi::rand::prelude::Distribution<VoteStateUpdate>
+    for solana_frozen_abi::rand::distr::StandardUniform
+{
+    fn sample<R: solana_frozen_abi::rand::Rng + ?Sized>(&self, rng: &mut R) -> VoteStateUpdate {
+        let lockouts: VecDeque<_> = (0..rng.random_range(0..1000))
```
alexpyattaev: same here, establish true max length.

vote-interface/src/state/vote_state_1_14_11.rs
```diff
@@ -45,6 +50,47 @@ pub struct VoteState1_14_11 {
     pub last_timestamp: BlockTimestamp,
 }
 
+#[cfg(feature = "frozen-abi")]
+impl solana_frozen_abi::rand::prelude::Distribution<VoteState1_14_11>
+    for solana_frozen_abi::rand::distr::StandardUniform
+{
+    fn sample<R: solana_frozen_abi::rand::Rng + ?Sized>(&self, rng: &mut R) -> VoteState1_14_11 {
+        let votes: VecDeque<_> = (0..rng.random_range(0..1000))
```
alexpyattaev: and here...

alexpyattaev (commented):
Thank you! The direction looks good but we need to figure out how long these vecs can be in practice. For example, for votes 1000 elements is clearly too much (as that would not fit into MTU).

vote-interface/src/state/vote_instruction_data.rs
puhtaytow: ah sorry for triggering you by rebase, i plan to remove all these manual implementations, we just have to finish #730

alexpyattaev: Manual or otherwise, need to establish correct length for those. No problem about the ping =)

kskalski: Technically the length of the collection doesn't really make much difference in terms of validating the serialization format, at least beyond some small values like 3-5.
If we keep adding collections in range of 1000 or even close to real-life sizes for given fields, but nest them in other collections with 1000 elements, the froze-abi tests might start hitting slow execution.

I think most of frozen_abi collections should still stay within 10 elements.

alexpyattaev: Agreed, around 10-30 elements max is probably where the reasonable values are, unless a tighter upper bound is known.

puhtaytow:
to be merged after https://github.com/anza-xyz/solana-sdk/pull/739

vote-interface/src/state/vote_instruction_data.rs
puhtaytow: the defaults are 5 elements, should i change for 10?

vote-interface/src/state/vote_state_1_14_11.rs
```diff
@@ -35,6 +40,12 @@ pub struct VoteState1_14_11 {
     /// history of prior authorized voters and the epochs for which
     /// they were set, the bottom end of the range is inclusive,
     /// the top of the range is exclusive
+    #[cfg_attr(
+        feature = "frozen-abi",
+        stable_abi_sample(
+            with = "CircBuf { buf: rng.random(), idx: usize::from(rng.random::<u8>()) % MAX_ITEMS, is_empty: rng.random() }"
```
puhtaytow: if CircBuf isn't something from the past (i cant see it present in V4 anymore) then we might think about covering this case without manual override in future pr
```
error[E0277]: the trait bound `I: solana_frozen_abi::stable_abi::StableAbi` is not satisfied
   --> vote-interface/src/state/mod.rs:159:10
    |
159 |     buf: [I; MAX_ITEMS],
    |          ^^^^^^^^^^^^^^ the trait `solana_frozen_abi::stable_abi::StableAbi` is not implemented for `I`
    |
    = note: required for `[I; 32]` to implement `solana_frozen_abi::stable_abi::StableAbi`
help: consider restricting type parameter `I` with trait `StableAbi`
    |
158 | pub struct CircBuf<I: solana_frozen_abi::stable_abi::StableAbi> {
    |                     ++++++++++++++++++++++++++++++++++++++++++
```

vote-interface/src/state/vote_instruction_data.rs
alexpyattaev: 5 is fine, thank you!

vote-interface/src/state/vote_state_1_14_11.rs
kskalski: with can use a shared function to be called in both use cases using `with = "generate_stable_abi_sample_circ_buf(rng)"`

vote-interface/src/state/mod.rs
```diff
@@ -174,6 +174,24 @@ impl<I: Default + Copy> Default for CircBuf<I> {
...
+        solana_frozen_abi::rand::distr::Distribution<[I; MAX_ITEMS]>,
+{
+    fn random_with_context(
+        rng: &mut (impl solana_frozen_abi::rand::RngCore + ?Sized),
+        _ctx: (),
+    ) -> Self {
+        Self {
+            buf: solana_frozen_abi::rand::Rng::random::<[I; MAX_ITEMS]>(rng),
+            idx: usize::from(solana_frozen_abi::rand::Rng::random::<u8>(rng)) % MAX_ITEMS,
+            is_empty: solana_frozen_abi::rand::Rng::random::<bool>(rng),
```
kskalski: not sure if we should care, but maybe we could maintain internal consistency here making `is_empty` false for `idx < MAX_ITEMS - 1`

is our support for generic arrays no enough to just derive StableAbis for CircBuf or is there different reason?

puhtaytow: should be good now :+1: and yes this case isnt supported by just derive

kskalski: what is the error? can we work-around it by using with= on one of the fields?

I see https://github.com/anza-xyz/solana-sdk/blob/14a725d6e9180e6cfbd98054473d61ef3aabde57/frozen-abi/src/stable_abi/impls.rs#L99-L106
which should support array properly even when used with generic type

puhtaytow:
```
error[E0277]: the trait bound `I: solana_frozen_abi::stable_abi::StableAbi` is not satisfied
   --> vote-interface/src/state/mod.rs:159:10
    |
159 |     buf: [I; MAX_ITEMS],
    |          ^^^^^^^^^^^^^^ the trait `solana_frozen_abi::stable_abi::StableAbi` is not implemented for `I`
    |
    = note: required for `[I; 32]` to implement `solana_frozen_abi::stable_abi::StableAbi`
help: consider restricting type parameter `I` with trait `StableAbi`
```

puhtaytow: i think it makes sense to keep the manual impl as you rised above it might benefit from real scenario, with derive we wont have even if we figure out the above

kskalski: I suppose macro generated impl should apply StableAbi constraint to generic params of the struct (here https://github.com/anza-xyz/solana-sdk/blob/14a725d6e9180e6cfbd98054473d61ef3aabde57/frozen-abi-macro/src/lib.rs#L36-L53)
something like wincode does here https://github.com/anza-xyz/wincode/blob/a738c8a139629fb5f26e8473a1aad23eb8f0b6cc/wincode-derive/src/schema_write.rs#L259

kskalski: Generating valid instances isn't really that important for checking serialization format, at least in this case (where serialization isn't dependent on that, e.g. creating optimized format covering subset of possible memory bytes).

I think this raised a real and important missing feature in the derive macro, so let's treat this PR as discover and test for the fix. :)

puhtaytow: Oki. I'll see what could be done :)

puhtaytow: for reference https://github.com/anza-xyz/solana-sdk/pull/750

kskalski:
LGTM

alexpyattaev (approved):
Thank you!

kskalski (approved):

PR #478, opened 2025-12-01
frozen-abi: remove deprecated log re-export

puhtaytow:
#### Problem

There was deprecated since 3.0.1 re-export for `log` crate.

#### Summary of changes
Removed `log` re-export and dependency.

### DIFF
--- a/frozen-abi/src/lib.rs
+++ b/frozen-abi/src/lib.rs
@@ -28,13 +28,3 @@ extern crate solana_frozen_abi_macro;
 
 #[cfg(all(feature = "frozen-abi", not(target_os = "solana")))]
 pub use {bincode, rand, rand_chacha};
-
-// Not public API. Previously referenced by macro-generated code. Remove the
-// `log` dependency from Cargo.toml when this is cleaned up in the next major
-// version bump
-#[deprecated(since = "3.0.1", note = "Please use the `log` crate directly instead")]
-#[doc(hidden)]
-pub mod __private {
-    #[doc(hidden)]
-    pub use log;
-}


joncinque:
The change looks good to me, especially because we're planning on making some breaking changes soon.

Keep in mind that a breaking release on solana-frozen-abi will require a lot of crate publishes, since every single crate that uses solana-frozen-abi will need to be updated and published.

If we want to do a breaking release on solana-frozen-abi, we should do a semver trick backport to v3 of solana-frozen-abi, so that we don't have to do so many crate publishes and updates.

puhtaytow:
> The change looks good to me, especially because we're planning on making some breaking changes soon.
> 
> Keep in mind that a breaking release on solana-frozen-abi will require a lot of crate publishes, since every single crate that uses solana-frozen-abi will need to be updated and published.
> 
> If we want to do a breaking release on solana-frozen-abi, we should do a semver trick backport to v3 of solana-frozen-abi, so that we don't have to do so many crate publishes and updates.

Soon sounds good. I didn't know about the symver trick, thank you :pray:

joncinque:
I'm thinking we can just land this -- we should be able to remove a module called `__private` in a non-major release. `serde` does something similar with its macros implementations. Any thoughts on that?

puhtaytow:
> I'm thinking we can just land this -- we should be able to remove a module called `__private` in a non-major release. `serde` does something similar with its macros implementations. Any thoughts on that?

The comment above says explicitly that its not public api `// Not public API. Previously referenced by macro-generated code.` and its hidden from docs.. to me looks ok to proceed tbh

PR #479, opened 2025-12-01
vote-interface: remove v0_23_5

buffalojoec:
Removes the old vote state version `V0_23_5` from the vote-interface crate. For `VoteStateVersions`, replaces `V0_23_5` with an `Uninitialized` variant and treats it as such.

### DIFF
--- a/vote-interface/src/state/vote_state_v3.rs
+++ b/vote-interface/src/state/vote_state_v3.rs
@@ -155,8 +155,8 @@ impl VoteStateV3 {
 
         let variant = solana_serialize_utils::cursor::read_u32(&mut cursor)?;
         match variant {
-            // Variant 0 is uninitialized.
-            0 => Err(InstructionError::UninitializedAccount),
+            // Variant 0 is not a valid vote state.
+            0 => Err(InstructionError::InvalidAccountData),
             // V1_14_11
             1 => deserialize_vote_state_into_v3(&mut cursor, vote_state, false),
             // V3. the only difference from V1_14_11 is the addition of a slot-latency to each vote
--- a/vote-interface/src/state/vote_state_v4.rs
+++ b/vote-interface/src/state/vote_state_v4.rs
@@ -164,8 +164,8 @@ impl VoteStateV4 {
 
         let variant = solana_serialize_utils::cursor::read_u32(&mut cursor)?;
         match variant {
-            // Variant 0 is uninitialized.
-            0 => Err(InstructionError::UninitializedAccount),
+            // Variant 0 is not a valid vote state.
+            0 => Err(InstructionError::InvalidAccountData),
             // V1_14_11
             1 => deserialize_vote_state_into_v4(
                 &mut cursor,


vote-interface/src/state/vote_state_versions.rs
```diff
@@ -37,40 +31,14 @@ impl VoteStateVersions {
...
     }
 
-    /// Convert from vote state `V0_23_5`, `V1_14_11`, or `V3` to `V3`.
+    /// Convert from vote state `V1_14_11` or `V3` to `V3`.
     ///
     /// NOTE: Does not support conversion from `V4`. Attempting to convert from
     /// v4 to v3 will throw an error.
-    #[cfg(any(test, all(not(target_os = "solana"), feature = "bincode")))]
+    #[cfg(test)]
     pub(crate) fn try_convert_to_v3(self) -> Result<VoteStateV3, InstructionError> {
```
buffalojoec, line 39: @jstarry since these are now test-only, we can probable remove them completely here.

vote-interface/src/state/vote_state_v3.rs
```diff
@@ -155,29 +155,9 @@ impl VoteStateV3 {
...
-                        );
-                    }
-                    Ok(())
-                }
-                #[cfg(target_os = "solana")]
-                Err(InstructionError::InvalidAccountData)
-            }
-            // V1_14_11. substantially different layout and data from V0_23_5
+            // Variant 0 is uninitialized.
+            0 => Err(InstructionError::UninitializedAccount),
```
jstarry: Noting that this error change will change instruction errors downstream so we should make sure Firedancer is aware when we pick up this version update.

1. If we try to deserialize an account with `[0, 0, 0, 0]` prefix but it's not big enough to hold `V0_23_5`, we used to return `InvalidAccountData` but now we always return `UninitializedAccount`

2. In vote program, if we successfully deserialized and converted from `V0_23_5` to `V3`, some instructions don't check for initialization (but later fail with some ix error) but now we will always return `UninitializedAccount`

vote-interface/src/state/vote_state_v4.rs
```diff
@@ -164,29 +164,8 @@ impl VoteStateV4 {
...
-                                .and_then(|versioned| versioned.try_convert_to_v4(vote_pubkey))?,
-                        );
-                    }
-                    Ok(())
-                }
-                #[cfg(target_os = "solana")]
-                Err(InstructionError::InvalidAccountData)
-            }
+            // Variant 0 is uninitialized.
+            0 => Err(InstructionError::UninitializedAccount),
```
jstarry: Same as above, this error change will change instruction errors downstream so we should make sure Firedancer is aware when we pick up this version update.

1. If we try to deserialize an account with `[0, 0, 0, 0]` prefix but it's not big enough to hold `V0_23_5`, we used to return `InvalidAccountData` but now we always return `UninitializedAccount`

vote-interface/src/state/vote_state_versions.rs
jstarry: This change doesn't look related to the rest of the PR. Can we remove it in a follow up?

vote-interface/src/state/vote_state_versions.rs
```diff
@@ -97,42 +65,15 @@ impl VoteStateVersions {
         }
     }
 
-    // Currently, all versions can be converted to v4 without data loss, so
-    // this function returns `Ok(..)`. However, future versions may not be
-    // convertible to v4 without data loss, so this function returns a `Result`
-    // for forward compatibility.
-    #[cfg(any(test, all(not(target_os = "solana"), feature = "bincode")))]
```
jstarry, line 104: Same here, can we change this in a follow up PR?

vote-interface/src/state/vote_state_versions.rs
buffalojoec: Yeah sure, we can. That's why I left that commit out for now.

vote-interface/src/state/vote_state_v3.rs
buffalojoec: To me, `UninitializedAccount` feels like the correct error, but we could always leave it as `InvalidAccountData` if it's too painful to change. Tagging @mjain-jump to weigh in.

vote-interface/src/state/vote_state_versions.rs
buffalojoec: This change in particular is to satisfy clippy/hack checks, since the function is only used for deserialization (non-os-solana) of a v0_23_5 underlying vote state. Removal of that variant and deserialization logic leaves this unused outside of test context.

vote-interface/src/state/vote_state_versions.rs
jstarry: I mean this vis change

vote-interface/src/state/vote_state_v3.rs
jstarry: I prefer uninitialized too, just noting that we need to let them know that it's going to change

vote-interface/src/state/vote_state_versions.rs
jstarry: But it's public isn't it? I thought clippy wouldn't complain about unused public methods

jstarry: Sorry the last comment was from my phone.. Makes sense to make this change given that the vis is `pub(crate)` not `pub` as I was thinking.

vote-interface/src/state/vote_state_v3.rs
mjain-jump: `UninitializedAccount` conceptually makes less sense to me because, the account is technically "initialized" as a v0_23_5 account, its just not valid anymore (so invalid account data makes more sense IMO). I have the same proposal here: https://github.com/anza-xyz/solana-sdk/pull/447

jstarry: Yes, right now the account is technically "initialized" as v0 but part of the reason for this PR is to change that notion entirely and instead handle such accounts as technically being "uninitialized" since they can only exist as a fully zeroed accounts

vote-interface/src/state/vote_state_versions.rs
buffalojoec: No problem. Yeah it's a clippy error if we leave the original `cfg` visibility in tact. In follow up, they'll just come out completely.

vote-interface/src/state/vote_state_v3.rs
buffalojoec: Haha fair enough @mjain-jump - you got me. I did suggest either one in my [review comment](https://github.com/anza-xyz/solana-sdk/pull/447#pullrequestreview-3480333175), though.

We know none of these accounts exist on mb, so as @jstarry said, reading an account with `[0, 0, 0, 0]` would be all-zeroes, but I guess the SDK could more agnostically handle serialized state.

> the account is technically "initialized" as a v0_23_5 account, its just not valid anymore 

At this point I could go wither way, but I think this is a solid point.

buffalojoec: I've updated the error to `InvalidAccountData`.

jstarry:
We are going to be setting vote state to all zeroes when closing a v4 vote account. This will now be seen as "invalid" rather than "uninitialized." This still seems strange to me. Typically such variants are called uninitialized but this is all semantic so I'll approve :)

PR #480, opened 2025-12-01
deps: Bump to rand v0.9 where possible

joncinque:
#### Problem

The repo is still on rand v0.8 in a lot of places, but that version isn't very compatible with rust 2024 due to its usage of `gen` as a function name, which is becoming a keyword.

#### Summary of changes

Follow https://rust-random.github.io/book/update-0.9.html to perform the upgrade to rand v0.9 where possible.

Most crypto libraries are still on rand v0.8 / rand_core v0.6, and the blstrs and ecdsa implementations of `random` aren't as simple as just taking 32 or 64 random bytes, so I left those back.

cc @puhtaytow

### DIFF
--- a/serialize-utils/src/cursor.rs
+++ b/serialize-utils/src/cursor.rs
@@ -133,11 +133,7 @@ mod test {
     #[test]
     fn test_read_option_u64() {
         for _ in 0..100 {
-            let test_value = if rand::random_bool(0.5) {
-                Some(rand::random::<u64>())
-            } else {
-                None
-            };
+            let test_value = rand::random_bool(0.5).then(rand::random::<u64>);
             test_read(read_option_u64, test_value);
         }
     }
--- a/vote-interface/src/state/mod.rs
+++ b/vote-interface/src/state/mod.rs
@@ -923,13 +923,13 @@ mod tests {
         .take(32)
         .sorted_by_key(|lockout| lockout.slot())
         .collect();
-        let root = rng.random_ratio(1, 2).then(|| {
+        let root = rng.random_bool(0.5).then(|| {
             lockouts[0]
                 .slot()
                 .checked_sub(rng.random_range(0..1_000))
                 .expect("All slots should be greater than 1_000")
         });
-        let timestamp = rng.random_ratio(1, 2).then(|| rng.random());
+        let timestamp = rng.random_bool(0.5).then(|| rng.random());
         let hash = Hash::from(rng.random::<[u8; 32]>());
         let vote_state_update = VoteStateUpdate {
             lockouts,


frozen-abi/Cargo.toml
```diff
@@ -42,8 +42,8 @@ im = { workspace = true, features = ["rayon", "serde"] }
 memmap2 = { workspace = true }
 # These dependencies are used only to back `frozen-abi` StableAbi API,
 # to avoid version skew and keep digest computation stable regardless of consumer dependencies.
-rand = "0.8.5"
-rand_chacha = "0.3.1"
+rand = { workspace = true }
```
alexpyattaev, line 45: Should we pin the version of rand here to avoid ABI hashes from breaking accidentally? None of this runs in prod, so perf is a non-issue.

keypair/src/lib.rs
```diff
@@ -34,8 +33,8 @@ impl Keypair {
     /// Constructs a new, random `Keypair` using `OsRng`
     #[allow(clippy::new_without_default)]
     pub fn new() -> Self {
-        let mut rng = OsRng;
-        Self(ed25519_dalek::SigningKey::generate(&mut rng))
+        let secret_bytes = rand::random::<[u8; Self::SECRET_KEY_LENGTH]>();
```
alexpyattaev, line 36: Technically, this changes from explicitly using OsRng to using ThreadRng. Not sure it matters, just pointing it out. It is probably for the best.

alexpyattaev (commented):
Checked the crates I know:

- address
- frozen-abi
- keypair
- serde-varint

Overall the change looks good, but I'd prefer if we pinned some version for frozen-abi. Technically, changing the rand version for frozen-abi at this point would require a major version bump (assuming someone has already started using the version with ABI checks).

frozen-abi/Cargo.toml
joncinque: The frozen-abi tests in this repo would catch that, no?

joncinque:
> Technically, changing the rand version for frozen-abi at this point would require a major version bump (assuming someone has already started using the version with ABI checks).

Right, it's a little annoying, but probably for the best

frozen-abi/Cargo.toml
alexpyattaev: We need to add those tests. But yes, they should catch this.

joncinque: How different would they be from these? https://github.com/anza-xyz/solana-sdk/blob/master/scripts/test-frozen-abi.sh

puhtaytow:
> > Technically, changing the rand version for frozen-abi at this point would require a major version bump (assuming someone has already started using the version with ABI checks).
> 
> Right, it's a little annoying, but probably for the best

Unfortunately. 

I have a branch switched to new chacha and rand versions. Tested it with sdk/agave examples, so if we want to bump major version i could PR it now.  Alternatively, we could implement StableAbi for all frozen-abi enabled types and let it settle for a bit then migrate and bump major.

puhtaytow:
> > Technically, changing the rand version for frozen-abi at this point would require a major version bump (assuming someone has already started using the version with ABI checks).
> 
> Right, it's a little annoying, but probably for the best

Unfortunately.

I have spoken with Alex and we agreed that perhaps it would be better if we switch `frozen-abi` to rand/chacha v0.9 using this ocassion. It would be breaking change, but performed now, would reflect in less work later.

If you think this is the good move then please leave `frozen-abi` crate deps as it was and here i have prepared the small PR https://github.com/anza-xyz/solana-sdk/pull/482 to be merged after yours.

joncinque:
> If you think this is the good move then please leave frozen-abi crate deps as it was and here i have prepared the small PR https://github.com/anza-xyz/solana-sdk/pull/482 to be merged after yours.

I don't understand, is the separate PR mainly to do the contribution in a separate PR? The changes are here and passing CI, why force me to spend extra time to revert parts of this PR?

frozen-abi/Cargo.toml
puhtaytow: I have this draft here https://github.com/anza-xyz/solana-sdk/pull/484 this should early catch eventual changes. Somehow it has the same issue as described here https://github.com/anza-xyz/solana-sdk/pull/477 - Works with cargo test, but with cargo hack those are skipped.

puhtaytow:
> > If you think this is the good move then please leave frozen-abi crate deps as it was and here i have prepared the small PR #482 to be merged after yours.
> 
> I don't understand, is the separate PR mainly to do the contribution in a separate PR? The changes are here and passing CI, why force me to spend extra time to revert parts of this PR?

Sorry! I didn't notice that you adjusted the interface definition as well. All good! :pray:

joncinque:
@samkim-crypto can you review the other crates not covered by @alexpyattaev ?

serialize-utils/src/cursor.rs
```diff
@@ -133,7 +133,11 @@ mod test {
     #[test]
     fn test_read_option_u64() {
         for _ in 0..100 {
-            let test_value = rand::random::<Option<u64>>();
+            let test_value = if rand::random_bool(0.5) {
+                Some(rand::random::<u64>())
+            } else {
+                None
+            };
```
samkim-crypto:
```suggestion
            let test_value = rand::rng()
                .random_bool(0.5)
                .then(|| rand::random::<u64>());
```

Very minor nit: I guess this could be written slightly more idiomatically.

vote-interface/src/state/mod.rs
```diff
@@ -908,29 +908,29 @@ mod tests {
...
-            let confirmation_count = rng.gen_range(0..33);
+            let slot = 149_303_885_u64.saturating_add(rng.random_range(0..10_000));
+            let confirmation_count = rng.random_range(0..33);
             Lockout::new_with_confirmation_count(slot, confirmation_count)
         })
         .take(32)
         .sorted_by_key(|lockout| lockout.slot())
         .collect();
-        let root = rng.gen_ratio(1, 2).then(|| {
+        let root = rng.random_ratio(1, 2).then(|| {
```
samkim-crypto: Nit: Seems like `random_ratio` should be used for fractions like `1/3` or `5/7`. For this particular case, it seems like `rng.random_bool(0.5)` is probably more idiomatic.

vote-interface/src/state/mod.rs
```diff
@@ -908,29 +908,29 @@ mod tests {
...
             lockouts[0]
                 .slot()
-                .checked_sub(rng.gen_range(0..1_000))
+                .checked_sub(rng.random_range(0..1_000))
                 .expect("All slots should be greater than 1_000")
         });
-        let timestamp = rng.gen_ratio(1, 2).then(|| rng.gen());
-        let hash = Hash::from(rng.gen::<[u8; 32]>());
+        let timestamp = rng.random_ratio(1, 2).then(|| rng.random());
```
samkim-crypto: Nit: same here

joncinque:
Addressed the nits, let me know what you think!

PR #481, opened 2025-12-01
file-download: Optimize file download with buffered writer

dachen0:
Specify a buffer of 8MB instead of the default 8kb. This increases download speed to be similar to wget.


t-nelson (dismissed):
nice find! didn't even require a grant or rube goldberg machine

joncinque:
Do we want to merge this? The change looks good to me!

PR #482, opened 2025-12-02
frozen-abi: migrate StableAbi to rand v0.9

puhtaytow:
#### Problem
The `frozen-abi` crate has in place `rand` dependency v0.8 while **agave** and **solana-sdk** are moving towards v0.9.

#### Summary of changes
Trait `StableAbi` has been updated to use v0.9 API


puhtaytow:
This PR was open by mistake. I didn't notice that changes here are applied in #480.

PR #486, opened 2025-12-04
rent: Deprecate the concept of paying rent, prep for SIMD-0194

joncinque:
#### Problem

The concept of rent no longer exists in Solana clusters, but we still have the types and functions related to it.

Additionally, with SIMD-0194, we're eliminating the concept of `exemption_threshold`, but the field still exists.

#### Summary of changes

Prepare solana-rent for SIMD-0194, by deprecating
`lamports_per_byte_year` (will be renamed to `lamports_per_byte`), `exemption_threshold` (will just become a `[u8; 8]`), and `burn_percent` (will just go unused). Also deprecate functions and types related to paying rent.

Once this lands, we can do a breaking release that removes types and changes variable names.

This is a part of #275

### DIFF
--- a/rent/src/lib.rs
+++ b/rent/src/lib.rs
@@ -77,7 +77,7 @@ pub const DEFAULT_LAMPORTS_PER_BYTE_YEAR: u64 = 1_000_000_000 / 100 * 365 / (102
 /// - $1 per SOL
 /// - $0.01 per megabyte day
 /// - $7.30 per megabyte
-pub const DEFAULT_LAMPORTS_PER_BYTE: u64 = 2 * 1_000_000_000 / 100 * 365 / (1024 * 1024);
+pub const DEFAULT_LAMPORTS_PER_BYTE: u64 = 6_960;
 
 /// Default amount of time (in years) the balance has to include rent for the
 /// account to be rent exempt.


joncinque:
cc @brooksprumo just in case there's anything I'm missing. The deprecated functions aren't being used by Agave at all, and we can just upgrade Agave to use the new functions to prep for the breaking change.

febo (approved):
Looks great!

brooksprumo:
> cc @brooksprumo just in case there's anything I'm missing. The deprecated functions aren't being used by Agave at all, and we can just upgrade Agave to use the new functions to prep for the breaking change.

Love it. Looks good to me.

febo:
@joncinque I was just going through pinocchio's implementation of `Rent`. Maybe we should add a deprecation note to [`DEFAULT_LAMPORTS_PER_BYTE_YEAR`](https://github.com/anza-xyz/solana-sdk/blob/ea4634f4d90f222e31fe984676561c77bd721149/rent/src/lib.rs#L67) and [`DEFAULT_BURN_PERCENT`](https://github.com/anza-xyz/solana-sdk/blob/ea4634f4d90f222e31fe984676561c77bd721149/rent/src/lib.rs#L77) constants.

And we might want to change the type of [`DEFAULT_EXEMPTION_THRESHOLD`](https://github.com/anza-xyz/solana-sdk/blob/ea4634f4d90f222e31fe984676561c77bd721149/rent/src/lib.rs#L71), so a deprecation note here as well.

joncinque:
> @joncinque I was just going through pinocchio's implementation of Rent. Maybe we should add a deprecation note to [DEFAULT_LAMPORTS_PER_BYTE_YEAR](https://github.com/anza-xyz/solana-sdk/blob/ea4634f4d90f222e31fe984676561c77bd721149/rent/src/lib.rs#L67) and [DEFAULT_BURN_PERCENT](https://github.com/anza-xyz/solana-sdk/blob/ea4634f4d90f222e31fe984676561c77bd721149/rent/src/lib.rs#L77) constants.

> And we might want to change the type of [DEFAULT_EXEMPTION_THRESHOLD](https://github.com/anza-xyz/solana-sdk/blob/ea4634f4d90f222e31fe984676561c77bd721149/rent/src/lib.rs#L71), so a deprecation note here as well.

I just deprecated them entirely and added `DEFAULT_LAMPORTS_PER_BYTE`. While I was at it, I realized that the new constructor isn't correct -- we need to set the exemption threshold to 1.

rent/src/lib.rs
```diff
@@ -52,16 +64,37 @@ pub struct Rent {
...
 pub const DEFAULT_LAMPORTS_PER_BYTE_YEAR: u64 = 1_000_000_000 / 100 * 365 / (1024 * 1024);
 
+/// Default rental rate in lamports/byte.
+///
+/// This calculation is based on:
+/// - 10^9 lamports per SOL
+/// - $1 per SOL
+/// - $0.01 per megabyte day
+/// - $7.30 per megabyte
+pub const DEFAULT_LAMPORTS_PER_BYTE: u64 = 2 * 1_000_000_000 / 100 * 365 / (1024 * 1024);
```
febo: SIMD-0194 specifies this value as `6960`, which is equivalent to doubling the current `3480` value. Because of the precision, `DEFAULT_LAMPORTS_PER_BYTE` is evaluating to `6961`. To match the SIMD value, we could do the following:
```suggestion
pub const DEFAULT_LAMPORTS_PER_BYTE: u64 = 2 * (1_000_000_000 / 100 * 365 / (1024 * 1024) as u64);
```

joncinque: Nice catch, I'll just hard-code it to 6960

PR #487, opened 2025-12-05
rent!: Update default values, remove deprecated code

joncinque:
#### Problem

Rent is gone, and SIMD-0194 will deprecate the exemption threshold, but the field still exists in the actual sysvar on-chain.

#### Summary of changes

Remove all deprecated code related to collecting rent, and use the new default values for Rent after SIMD-0194.

To go with this, avoid floating point math in `minimum_balance()`, which was one of the of motivations for the change in the first place.

Add a test to make sure the old and new calcs are the same.

Closes #275

### DIFF
--- a/rent/src/lib.rs
+++ b/rent/src/lib.rs
@@ -173,9 +173,6 @@ impl Rent {
     /// Returns `ProgramError::InvalidArgument` if `data_len` exceeds the maximum
     /// permitted data length or if the `lamports_per_byte` is too large based on
     /// the `exemption_threshold`, which would cause an overflow.
-    //
-    // Note: Clippy suggests collapsing the `if` statements, but they are kept
-    // separate since it is more CU-efficient this way.
     #[inline(always)]
     pub fn try_minimum_balance(&self, data_len: usize) -> Option<u64> {
         if data_len as u64 > MAX_PERMITTED_DATA_LENGTH {


Cargo.toml
```diff
@@ -196,7 +196,7 @@ pairing = "0.23.0"
 parking_lot = "0.12"
 pbkdf2 = { version = "0.11.0", default-features = false }
 proc-macro2 = "1.0.93"
-proptest = "1.6"
+proptest = "1.9"
```
joncinque, line 202: This upgrade is just to make things easier once #461 lands

rent/src/lib.rs
```diff
@@ -104,220 +55,82 @@ pub const DEFAULT_BURN_PERCENT: u8 = 50;
...
-        (burned_portion, rent_collected - burned_portion)
-    }
-
     /// Minimum balance due for rent-exemption of a given account data size.
-    #[allow(deprecated)]
     pub fn minimum_balance(&self, data_len: usize) -> u64 {
         let bytes = data_len as u64;
-        (((ACCOUNT_STORAGE_OVERHEAD + bytes) * self.lamports_per_byte_year) as f64
-            * self.exemption_threshold) as u64
+        if self.unused_exemption_threshold == DEFAULT_EXEMPTION_THRESHOLD {
```
febo: nit: Currently this is only avoiding the `f64` operation when the threshold is `1.0`.  Wonder if it would be better to also include the current case (threshold equal to `2.0`), so we get the benefit before SIMD-0194 lands.

febo:
We might need to update the ABI diggest on `GenesisConfig` since it includes a `Rent` field.

rent/src/lib.rs
joncinque: My initial implementation actually had that, and then I reverted it :sweat_smile: I'll put it back in, and we can remove it once the SIMD is live, it's a good point. It also lets us (more) safely remove all float logic

joncinque:
Ok, this should be good for another look!

febo (dismissed):
Looks great!

joncinque:
I'm going to hold off on merging this for roughly 10 days, then we can merge, publish, and update agave accordingly

CASABECI (commented):
[Add Security Team to GHSAs](https://github.com/anza-xyz/solana-sdk/actions/workflows/add-team-to-ghsa.yml)

CASABECI:
> [Add Security Team to GHSAs](https://github.com/anza-xyz/solana-sdk/actions/workflows/add-team-to-ghsa.yml)

joncinque:
solana-rent v3.1.0 has been published, so this should be ready now!

I ended up copying the code from pinocchio, with the only change that there aren't any `unlikely` checks, and `try_minimum_balance` returns `Option<u64>` to avoid a dependency on `solana-program-error`. Let me know what you think!

rent/src/lib.rs
```diff
@@ -104,220 +75,190 @@ pub const DEFAULT_BURN_PERCENT: u8 = 50;
...
+    ///
+    /// # Errors
+    ///
+    /// Returns `ProgramError::InvalidArgument` if `data_len` exceeds the maximum
+    /// permitted data length or if the `lamports_per_byte` is too large based on
+    /// the `exemption_threshold`, which would cause an overflow.
+    //
+    // Note: Clippy suggests collapsing the `if` statements, but they are kept
+    // separate since it is more CU-efficient this way.
+    #[allow(clippy::collapsible_if)]
```
febo: nit: Probably this can be removed since the if statement is collapsed.

rent/src/lib.rs
```diff
@@ -104,220 +75,190 @@ pub const DEFAULT_BURN_PERCENT: u8 = 50;
...
         assert_eq!(cloned_rent, rent);
     }
+
+    #[test]
+    fn test_exemption_threshold() {
+        assert_eq!(1f64.to_le_bytes(), SIMD0194_EXEMPTION_THRESHOLD);
+        assert_eq!(2f64.to_le_bytes(), CURRENT_EXEMPTION_THRESHOLD);
+    }
+
+    proptest! {
```
febo, line 242: This is cool!

joncinque:
Sorry for the slowness, finally updated this PR

rent/src/lib.rs
```diff
@@ -104,220 +75,189 @@ pub const DEFAULT_BURN_PERCENT: u8 = 50;
...
+    /// The minimum balance in lamports for rent exemption.
+    ///
+    /// # Errors
+    ///
+    /// Returns `ProgramError::InvalidArgument` if `data_len` exceeds the maximum
+    /// permitted data length or if the `lamports_per_byte` is too large based on
+    /// the `exemption_threshold`, which would cause an overflow.
+    //
+    // Note: Clippy suggests collapsing the `if` statements, but they are kept
+    // separate since it is more CU-efficient this way.
```
febo: nit: I think this comment can be removed.

joncinque: Bah, sorry about that, done

febo (approved):
Looks great!

PR #488, opened 2025-12-05
Deprecate `WritableAccount::create`

LucasSte:
**Problem**

WritableAccount create is not suitable for structs that do not own fields, like the new SVM account format we adopted in runtime for ABIv2.

I have removed all usages from the monorepo in https://github.com/anza-xyz/agave/pull/9279.

**Solution**

Deprecate function and remove usages.


LucasSte:
> Does agave still build when using this branch of the SDK?
> 
> Mainly wondering about the removal of the various constructor helper functions, and if agave uses any of them.
> 
> If agave builds, then :shipit:

Yes, Agave still builds with this patch. I had removed all the use cases in https://github.com/anza-xyz/agave/pull/9279.

PR #489, opened 2025-12-06
program-log: Add log wrappers

febo:
### Problem

There are a few more log syscalls available on the runtime, but in order to use them it is necessary to pass raw pointers.

### Solution

Add wrapper functions to simplify their use.

### DIFF
--- a/program-log/src/wrapper.rs
+++ b/program-log/src/wrapper.rs
@@ -3,11 +3,13 @@
 #[cfg(not(any(target_os = "solana", target_arch = "bpf")))]
 use core::hint::black_box;
 #[cfg(any(target_os = "solana", target_arch = "bpf"))]
-use solana_define_syscall::definitions::{sol_log_, sol_log_64_, sol_log_compute_units_};
+use solana_define_syscall::definitions::{
+    sol_log_, sol_log_64_, sol_log_compute_units_, sol_log_data,
+};
 
 /// Print a string to the log.
 #[inline(always)]
-pub fn sol_log(message: &str) {
+pub fn log(message: &str) {
     #[cfg(any(target_os = "solana", target_arch = "bpf"))]
     unsafe {
         sol_log_(message.as_ptr(), message.len() as u64);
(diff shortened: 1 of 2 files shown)


program-log/src/wrapper.rs
```diff
@@ -0,0 +1,62 @@
+//! Logging wrapper for Solana syscalls.
+
+#[cfg(not(any(target_os = "solana", target_arch = "bpf")))]
+use core::hint::black_box;
+#[cfg(any(target_os = "solana", target_arch = "bpf"))]
+use solana_define_syscall::definitions::{sol_log_, sol_log_64_, sol_log_compute_units_};
+
+/// Print a string to the log.
+#[inline(always)]
+pub fn sol_log(message: &str) {
```
febo: We could consider dropping the `sol_*` prefix.

program-log/src/wrapper.rs
```diff
@@ -0,0 +1,62 @@
...
+    black_box(data);
+}
+
+/// Print the hexadecimal representation of a slice.
+#[inline(always)]
+pub fn sol_log_slice(slice: &[u8]) {
+    for (i, s) in slice.iter().enumerate() {
+        sol_log_64(0, 0, 0, i as u64, *s as u64);
+    }
+}
```
joncinque: Honestly I don't find this one too useful, I'd go with not including it at all

program-log/src/wrapper.rs
joncinque: That's a good point, since these are new functions, let's name them however we want!

I looked around at other rust logging libraries for some standards -- most of them integrate with the standard `log` library, which uses the `info!` / `error!` / `debug!` macros.

I haven't seen too many functions called `log`, but it might be ok in our case for two reasons:
* we don't have log levels, so just `log` makes sense
* we don't want to encourage string formatting, so it shouldn't be a macro

so maybe we have `log`, `log_u64`, log_data`, `log_compute_units`. What do you think?

joncinque (commented):
Looks great! Just some small comments

program-log/src/wrapper.rs
febo: Sounds good – updated.

PR #490, opened 2025-12-09
wip: introduce vote-state-v5

akhi3030:
Proposal for how to update vote state so that it supports both tower and alpenglow vote rewards.

Compared to `VoteStateV4`:
- tower related stuff is moved into a separate struct
- a new struct for alpenglow is introduced

If it helps with migration, we do not have to introduce new structs and just add the new field needed for alpenglow at the bottom of `VoteStateV5`.  

Assuming that after migration to alpenglow, we still need to finish paying tower rewards so both states are accessible at the same time.

### DIFF
--- a/vote-interface/src/state/mod.rs
+++ b/vote-interface/src/state/mod.rs
@@ -26,7 +26,8 @@ pub use vote_state_v3::VoteStateV3;
 pub mod vote_state_v4;
 pub use vote_state_v4::VoteStateV4;
 mod vote_instruction_data;
-pub use vote_instruction_data::*;
+pub mod vote_state_v5;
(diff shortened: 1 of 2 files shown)


puhtaytow:
Would this code be part of another PR?

akhi3030:
this PR was part of a larger discussion to figure out how to store credits for vote rewards in alpenglow.  For V1, we decided that we will "abuse" v4 state.  Longer term, we haven't decided if we want to do PDA or V5.

puhtaytow:
> this PR was part of a larger discussion to figure out how to store credits for vote rewards in alpenglow. For V1, we decided that we will "abuse" v4 state. Longer term, we haven't decided if we want to do PDA or V5.

Oki. Thank you :pray:

PR #493, opened 2025-12-11
[bls-signatures] Add support verifying signatures and pop directly as byte arrays

samkim-crypto:
#### Problem

For places like the vote program https://github.com/anza-xyz/agave/pull/9310, public keys are stored as raw bytes rather than as `Pubkey` or `PubkeyCompressed` types provided from the `solana-bls-signatures` crate. It would be nice if the crate public interface supports verifying signatures or PoPs by directly interpreting the input from raw bytes.

#### Summary of Changes

Updated the verification functions so that it can interpret public keys, signatures, or PoPs directly from byte arrays. It was pretty simple to do this since we can just implement the `AsPubkey`, `AsSignature`, and `AsProofOfPossession` traits for the fixed size byte arrays.

### DIFF
--- a/bls-signatures/src/pubkey/mod.rs
+++ b/bls-signatures/src/pubkey/mod.rs
@@ -14,6 +14,8 @@ pub use {
 
 #[cfg(test)]
 mod tests {
+    #[cfg(feature = "parallel")]
+    use rayon::prelude::*;
     use {
         super::*,
         crate::{
@@ -25,8 +27,6 @@ mod tests {
         core::str::FromStr,
         std::string::ToString,
     };
-    #[cfg(feature = "parallel")]
-    use {crate::error::BlsError, rayon::prelude::*};
 
     #[test]
     fn test_pubkey_verify_signature() {


bls-signatures/src/proof_of_possession/mod.rs
```diff
@@ -140,4 +140,23 @@ mod tests {
...
+    fn test_verify_proof_of_possession_with_raw_bytes() {
+        let keypair = Keypair::new();
+        let pop_projective = keypair.proof_of_possession(None);
+
+        let pubkey_bytes: [u8; 48] = PubkeyCompressed::try_from(keypair.public).unwrap().0;
+
+        let pop_affine = ProofOfPossession::from(pop_projective);
+        let pop_bytes: [u8; 96] = ProofOfPossessionCompressed::try_from(pop_affine).unwrap().0;
+
+        assert!(pubkey_bytes
```
zz-sol, line 154: may be add a test that malleable_pubkey_bytes will return errors?
```
malleable_pubkey_bytes: [u8; 50ish] = pubkey_bytes || random data
```

bls-signatures/src/signature/mod.rs
```diff
@@ -407,4 +407,27 @@ mod tests {
...
+        let signature_bytes: [u8; 96] = SignatureCompressed::try_from(signature_affine).unwrap().0;
+
+        assert!(pubkey_bytes
+            .verify_signature(&signature_bytes, message)
+            .unwrap());
+        assert!(keypair
+            .public
+            .verify_signature(&signature_bytes, message)
+            .unwrap());
+        assert!(pubkey_bytes
```
zz-sol, line 429: same.

bls-signatures/src/proof_of_possession/mod.rs
samkim-crypto: Yep, good idea. I added test cases for bit flipped public keys, signatures, and PoPs to make sure that they error. For the length check, this should be compiler enforced, so I don't think I can pass `[u8; 50]` to a function expecting `[u8; 48]` as input. I still added a check on the underlying public key parsing logic to make sure that wrong size inputs are rejected.

PR #494, opened 2025-12-11
[bn254] Add G2 addition and multiplication syscalls (SIMD-0302)

LStan:


bn254/src/lib.rs
```diff
@@ -342,6 +352,13 @@ mod target_arch {
             .collect::<Vec<u8>>()
     }
 
+    pub(crate) fn convert_endianness_128(bytes: &[u8]) -> Vec<u8> {
```
samkim-crypto: What do you think about naming the `convert_endianness_64` and `convert_endianness_128` to `convert_g1_endianness` and `convert_g2_endianness`? The function `convert_endianness_128` takes chunks of 64 and then reverse each chunks, which might feel magical without proper context.

Also, let's additionally give some docs to these functions as they can be confusing.

For g1, what do you think about
```
/// Reverses byte order within 32-byte chunks to convert G1 point coordinates.
///
/// This is used to convert the `x` and `y` coordinates of G1 points between
/// Big-Endian and Little-Endian formats.
```

For g2:
```
/// Reverses byte order within 64-byte chunks to convert G2 point coordinates.
///
/// This handles G2 field extension elements where the EIP-197 encoding is
/// `c1 || c0` (Big-Endian) and Arkworks expects `c0 || c1` (Little-Endian).
///
/// Reversing the full 64-byte chunk serves a dual purpose:
/// 1. It swaps the coefficient order from (`c1`, `c0`) to (`c0`, `c1`).
/// 2. It reverses the endianness of the individual coefficients simultaneously.
```

bn254/src/multiplication.rs
```diff
@@ -120,11 +141,17 @@ pub fn alt_bn128_versioned_g1_multiplication(
 
     result_point
         .x
-        .serialize_with_mode(&mut result_point_data[..32], Compress::No)
+        .serialize_with_mode(
+            &mut result_point_data[..ALT_BN128_G1_MULTIPLICATION_OUTPUT_SIZE / 2],
```
samkim-crypto: I don't quite remember why we had a separate constant for the output size, but at least in this context, I think it makes sense to just write `ALT_BN128_FIELD_SIZE`. What do you think?

bn254/src/multiplication.rs
```diff
@@ -212,3 +239,131 @@ pub fn alt_bn128_g1_multiplication_le(
...
+        .map_err(|_| AltBn128Error::InvalidInputData)?;
+
+    let result_point: G2 = p.mul_bigint(fr).into();
+
+    let mut result_point_data = [0u8; ALT_BN128_G2_MULTIPLICATION_OUTPUT_SIZE];
+
+    result_point
+        .x
+        .serialize_with_mode(
+            &mut result_point_data[..ALT_BN128_G2_MULTIPLICATION_OUTPUT_SIZE / 2],
```
samkim-crypto: For this, I think it would be cleaner to define the following constants in `lib.rs`:

```
    /// Size of the EC point field, in bytes.
    pub const ALT_BN128_FIELD_SIZE: usize = 32;

    /// Size of the extension field element (Fq2), in bytes.
    pub const ALT_BN128_FQ2_SIZE: usize = ALT_BN128_FIELD_SIZE * 2; <-- added

    /// Size of the EC point. `alt_bn128` point contains
    /// the consistently united x and y fields as 64 bytes.
    pub const ALT_BN128_G1_POINT_SIZE: usize = ALT_BN128_FIELD_SIZE * 2;

    #[deprecated(since = "3.1.0", note = "Please use `ALT_BN128_G1_POINT_SIZE` instead")]
    pub const ALT_BN128_POINT_SIZE: usize = ALT_BN128_G1_POINT_SIZE;

    /// Elements in G2 is represented by 2 field-extension elements `(x, y)`.
    pub const ALT_BN128_G2_POINT_SIZE: usize = ALT_BN128_FQ2_SIZE * 2; <-- updated
```
and then use `ALT_BN128_FQ2_SIZE` here.

bn254/src/lib.rs
LStan: There is `convert_endianness<const CHUNK_SIZE: usize, const ARRAY_SIZE: usize>` function in `compression.lib`. It could be moved to `lib.rs` and used everywhere instead of the current functions. I planned to do it in the next PR. What do you think?

bn254/src/multiplication.rs
LStan: Maybe it is better to use `ALT_BN128_G1_POINT_SIZE` instead of `ALT_BN128_G1_MULTIPLICATION_OUTPUT_SIZE` everywhere. Also, using `ALT_BN128_FIELD_SIZE` seams vague for me, because the array is declared with `ALT_BN128_G1_MULTIPLICATION_OUTPUT_SIZE` size, and it is not obvious that `ALT_BN128_FIELD_SIZE` is a half of it. Isn't it better to do `[..ALT_BN128_G1_POINT_SIZE / 2]` here?
Or maybe after changing `ALT_BN128_G1_MULTIPLICATION_OUTPUT_SIZE`  to  `ALT_BN128_G1_POINT_SIZE` using `ALT_BN128_FIELD_SIZE` will not be vague.

`ALT_BN128_G1_MULTIPLICATION_OUTPUT_SIZE` can be deprecated.

The same for `ALT_BN128_G1_ADDITION_OUTPUT_SIZE`.

bn254/src/multiplication.rs
LStan: I suggest using `ALT_BN128_G2_POINT_SIZE` instead of `ALT_BN128_G2_MULTIPLICATION_OUTPUT_SIZE` everywhere.
The same for `ALT_BN128_G2_ADDITION_OUTPUT_SIZE.

bn254/src/multiplication.rs
samkim-crypto: A group point consists of two field elements `(x, y)`. Here, we are extracting each field element `x` and `y` and serializing them, so I thought it made more intuitive sense to use the `ALT_BN128_FIELD_SIZE` directly. But I don't have strong opinions, so if you find `ALT_BN128_G1_POINT_SIZE / 2` more intuitive, then it is fine with me. Let's at least use the `POINT_SIZE` instead of `OUTPUT_SIZE`.

bn254/src/multiplication.rs
samkim-crypto: Agreed, let's use `POINT_SIZE` instead of `OUTPUT_SIZE`.

bn254/src/lib.rs
samkim-crypto: Yeah I saw that too and using this function directly seems like it will clean up code slightly as it takes in a byte array instead of a slice. 

I think refactoring it to `lib.rs` is a good change. Let's just do it in this PR since I was hoping to cut `v3.2` after this PR is merged.

bn254/src/multiplication.rs
LStan: These consts are used [here](https://github.com/anza-xyz/agave/blob/master/syscalls/src/lib.rs#L1590-L1613). If we change `ALT_BN128_ADDITION_OUTPUT_SIZE` and `ALT_BN128_MULTIPLICATION_OUTPUT_SIZE` to `ALT_BN128_G1_POINT_SIZE` then `ALT_BN128_PAIRING_OUTPUT_SIZE` will look inconsistent. What do you think?

samkim-crypto: Oh that is unfortunate... I think it is fine to have `OUTPUT_SIZE` just for pairing and then use `POINT_SIZE` for addition and multiplication for now. Let's try to address the pairing output size for v4.0.

bn254/src/addition.rs
```diff
@@ -188,3 +220,129 @@ pub fn alt_bn128_g1_addition_le(
...
+/// approved Solana SIMD. Subsequently, a new `VersionedG2Addition` variant must be added,
+/// and the new logic must be scoped to that variant.
+#[cfg(not(target_os = "solana"))]
+pub fn alt_bn128_versioned_g2_addition(
+    _version: VersionedG2Addition,
+    input: &[u8],
+    endianness: Endianness,
+) -> Result<Vec<u8>, AltBn128Error> {
+    if input.len() != ALT_BN128_G2_ADDITION_INPUT_SIZE {
+        return Err(AltBn128Error::InvalidInputData);
```
0x0ece, line 231: Just pointing out that this behavior is different than G1.
I oc like this better.
Maybe let's collect these little differences and then amend the SIMD to highlight them?

bn254/src/multiplication.rs
LStan: Done in [50462e7](https://github.com/anza-xyz/solana-sdk/pull/494/commits/50462e715ac28afb4d714d2ca5f8a62a92fd0f8e). These constants were not published, so I just removed them and changed deprecation notes.

bn254/src/multiplication.rs
```diff
@@ -212,3 +226,129 @@ pub fn alt_bn128_g1_multiplication_le(
...
+pub fn alt_bn128_versioned_g2_multiplication(
+    _version: VersionedG2Multiplication,
+    input: &[u8],
+    endianness: Endianness,
+) -> Result<Vec<u8>, AltBn128Error> {
+    if input.len() != ALT_BN128_G2_MULTIPLICATION_INPUT_SIZE {
+        return Err(AltBn128Error::InvalidInputData);
+    }
+
+    let p: G2 = match endianness {
```
0x0ece, line 252: We need to add subgroup membership check for p, right?
(or if it's done implicitly, we need to remove it from add...)

bn254/src/lib.rs
LStan: Done in [a0b29dc](https://github.com/anza-xyz/solana-sdk/pull/494/commits/a0b29dc8ce5791dacef902fd42d0df5cfb059504). Also removed `reverse_copy` as it can be changed to `convert_endianness`.
Please check the doc for `convert_endianness` to see if it is understandable.
There are also magic numbers for `convert_endianness` params in `compression.rs`. Should I fix them here or later?

bn254/src/addition.rs
LStan: If you mean strict vs non-strict length comparison, then I guess it is [planned](https://github.com/anza-xyz/solana-sdk/pull/297#discussion_r2317538910)

bn254/src/multiplication.rs
LStan: Sorry, I don't understand what you mean

LStan:
@samkim-crypto There is a discussion in discord about return types for the function. Should I change it from `Vec` to `[u8;N]` (for `sdk` functions, not for `versioned`) as in `compression.rs` or leave it for later?

bn254/src/addition.rs
0x0ece: Unless I misunderstood, even with the new version, for BE we're still doing extension with 0s, no?
(As we need to be compatible with Ethereum...)

bn254/src/multiplication.rs
0x0ece: G2 is not a prime order group, unlike G1.
So we need to do 2 checks to verify that a point is valid:
1) validate the curve equation
2) validate that the point is in the correct group (usually a bunch of mul / endomorphisms -- pretty expensive)

We discussed with @samkim-crypto and we're going to do:
- addition: only check 1
- multiplication (and pairing): both checks 1+2

I'm unclear what checks `PodG2::from_??_bytes` does, but there should be a different fn in add vs mul.
Similar to [BLS12-381](https://github.com/anza-xyz/agave/pull/9759/files), where we have:
```
pub fn bls12_381_g1_addition(
...
    // skip subgroup check for efficiency
    let p1_affine = p1.to_affine_subgroup_unchecked(endianness)?;

pub fn bls12_381_g1_multiplication(
...
    let p1_affine = point.to_affine(endianness)?;
```

0x0ece: I guess this is the same fn used in pairing, so it's prob doing the check. So my comment should actually go into the add fn...

bn254/src/addition.rs
LStan: Not sure what you call the "new version", but extension with 0s is currently only for BE G1 multiplication and addition. And you wanted to remove it [here](https://github.com/anza-xyz/solana-sdk/pull/267#issuecomment-3193903458)

bn254/src/multiplication.rs
LStan: Validation happens in `try_from` [here](https://github.com/anza-xyz/solana-sdk/blob/d633129c628eeccecf01521fe9a7a6f865a51162/bn254/src/lib.rs#L291) and [here](https://github.com/anza-xyz/solana-sdk/blob/d633129c628eeccecf01521fe9a7a6f865a51162/bn254/src/lib.rs#L317). I don't know what `ark-bn254` validates

bn254/src/addition.rs
0x0ece: All I'm saying is that:
- G1 behaves in a way (BE does 0s extension)
- G2 behaves in a different way (BE does NOT do 0s extension)
- the SIMD doesn't mention this difference
- I'm personally ok with it
--> we should amend the SIMD

LStan: Oh, I see. I can add 0s extension to BE G2 addition and multiplication for consistency.

0x0ece: My personal preference is to leave the impl as is, and update the SIMD.

The only reason for the 0s extension on G1 add, G1 mul and pairing is Ethereum compatibility. Ethereum doesn't have G2. Also, in adding BLS, Ethereum removed this "feature" of extending with 0s.

The easier the code, the less things we have that can go wrong in multiple implementations. But I let others chime in as well.

bn254/src/lib.rs
samkim-crypto: Looks good to me! I don't think you need to bother with the magic numbers for this PR. Once we publish v3.2 and bump bn254 on the agave side, I was planning on a major refactoring of splitting this crate into two `agave-bn254` and `solana-bn254` for the v4.0 release. We can probably address a lot of these issues then.

bn254/src/addition.rs
samkim-crypto: Yeah I am happy either way: 
- We can make it consistent with G1 and then add a SIMD to amend both G1 and G2
- Or we can leave the current implementation as is, amend the G2 SIMD, and add another SIMD to update G1.

Maybe we can do whatever requires least amount of work (since agave v4.0 is going to be cut next Monday). So my vote would be to leave the current implement in the PR as is.

bn254/src/multiplication.rs
0x0ece: Talked to @samkim-crypto - please disregard my comment for this PR. We will change the behavior in a separated SIMD.

samkim-crypto:
> @samkim-crypto There is a discussion in discord about return types for the function. Should I change it from `Vec` to `[u8;N]` (for `sdk` functions, not for `versioned`) as in `compression.rs` or leave it for later?

Let's leave this later for now and target v4.0 where we can make breaking changes for the G1 components as well. I know it is kind of a mess in terms of consistency between G1 and G2 at this point, but for something like function signatures, I think it would be nice to have them consistent between G1 and G2.

I think the important bit right now is to add the versioned functions and then add it to agave before the cut. For the developer facing functions, the sdk has a flexible publishing schedule, so we can tackle after we make updates to agave.

samkim-crypto (approved):
The current version looks fine to me. @LStan if you are happy with the current changes, then I can go ahead and merge it in.

bn254/src/addition.rs
deanmlittle: +1 @0x0ece's suggestion

deanmlittle:
Correct me if I'm wrong @LStan @samkim-crypto but I believe us changing the SDK types will have no bearing upon the agave implementation, so there shouldn't be a good reason why we can't change from Vec<u8> in the SDK types. I still approve the changes as is, but would love to see us make those changes before 4.0 in another PR if possible.

samkim-crypto:
Since the agave v4.0 is cut on Monday, it would be ideal to cut the v3.2 release ASAP. We still need to assign CU costs and also update the `solana-syscall` crates to actually bind these syscall functions.

I don't think it will take us that long to release a `solana-bn254` v4.0. Maybe by the end of the month or early next month. I'd propose that we merge and cut immediately (esp. since this will introduce some inconsistencies in the function signatures), but if you think you really need it by v3.2, then we should make the changes in this PR.

LStan:
> The current version looks fine to me. @LStan if you are happy with the current changes, then I can go ahead and merge it in.

I guess the consensus now is basically to leave everything as it is. The only question is the return type. It can be changed for G2 functions, but I don't know how to do it for others without breaking changes.

samkim-crypto:
Yeah if we do decide to change the return type, it would just be for G2 only, which I think can wait until v4.0. But let me double check with @deanmlittle one last time.

bn254/src/multiplication.rs
0x0ece: Eventually added: https://github.com/anza-xyz/solana-sdk/pull/530

Issue #474, opened 2025-12-01
Re-organize types for `solana-bls-signatures`


samkim-crypto:
#### Problem

Currently, the `bls-signatures` crate mixes mathematical types with storage types. For example, `PubkeyProjective` wraps a curve point `(blstrs::G1Projective`), but `Pubkey` (affine) wraps a raw byte array (`[u8; 96]`).

This causes performance inefficiencies due to redundant validation. Every time we use a `Pubkey` for verification, we must deserialize and subgroup-check the bytes into a `G1Affine`. Because we don't have a wrapper for `G1Affine`, we throw this work away after the operation, forcing us to pay the validation cost again next time.

#### Proposal

We should split the types into mathematical points (Projective/Affine) and serialize bytes (uncompressed/compressed).

Using public keys as an example, we would transition to four types:

Points (math & arithmetic):
- `PubkeyProjective`: Wraps `G1Projective`. Best for aggregation and signing.
- `PubkeyAffine`: Wraps `G1Affine`. Best for verification (pairing inputs)

Bytes (storage & I/O):
- `Pubkey`: Wraps `[u8; 96]`
- `PubkeyCompressed`: Wraps `[u8; 48]`

This allows us to validate bytes once into a `PubkeyAffine` and reuse the efficient representation for multiple verifications.

joncinque:
Can this be closed?

PR #496, opened 2025-12-12
[bls-signatures] Add public key, signature, and proof of possession affine intermediate types

samkim-crypto:
#### Problem

https://github.com/anza-xyz/solana-sdk/issues/474

#### Summary of Changes

- added three affine types: `PubkeyAffine`, `SignatureAffine`, `ProofOfPossessionAffine`
- added traits `AsPubkeyAffine`, `AsSignatureAffine`, `AsProofOfPossessionAffine`
- updated verification functions to take in a generic that implements `As*Affine`
- Cleaned up and simplified the hashing functions for mapping messages and public keys into `G2`

### DIFF
--- a/bls-signatures/src/keypair.rs
+++ b/bls-signatures/src/keypair.rs
@@ -85,6 +85,11 @@ impl TryFrom<&[u8]> for Keypair {
             .map_err(|_| BlsError::ParseFromBytes)?;
         let public = PubkeyAffine::try_from(pubkey_bytes)?;
 
+        let expected_public: PubkeyAffine = PubkeyProjective::from_secret(&secret).into();
+        if expected_public != public {
+            return Err(BlsError::ParseFromBytes);
+        }
+
         Ok(Self { secret, public })
     }
 }


bls-signatures/src/hash.rs
```diff
@@ -1,28 +1,17 @@
...
 
 /// Domain separation tag used for hashing messages to curve points to prevent
 /// potential conflicts between different BLS implementations. This is defined
 /// as the ciphersuite ID string as recommended in the
 /// [standard](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#section-4.2.1).
 pub const HASH_TO_POINT_DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_NUL_";
 
-/// Hash a message to a G2 point
-pub fn hash_message_to_point(message: &[u8]) -> G2Projective {
+/// Hash a message to a G2 point for signature generation and verification
```
zz-sol, line 9: nit: add in the doc that if the caller is meant to hash the POP, use `hash_pop_payload_to_point ` API.

bls-signatures/src/keypair.rs
```diff
@@ -63,7 +63,11 @@ impl Keypair {
     }
 
     /// Verify a signature against a message and a public key
-    pub fn verify<S: AsSignature>(&self, signature: &S, message: &[u8]) -> Result<bool, BlsError> {
+    pub fn verify<S: AsSignatureAffine>(
+        &self,
+        signature: &S,
+        message: &[u8],
+    ) -> Result<bool, BlsError> {
```
zz-sol, line 70: I remember you wanted to change the return values. Is this going to happen on a follow-up PR

bls-signatures/src/hash.rs
samkim-crypto: Yep, done!

bls-signatures/src/keypair.rs
samkim-crypto: Yeah I wanted to do this in a follow-up PR since it is orthogonal to the introduction of the affine types.

bls-signatures/src/keypair.rs
```diff
@@ -74,14 +78,14 @@ impl TryFrom<&[u8]> for Keypair {
...
-                    .try_into()
-                    .map_err(|_| BlsError::ParseFromBytes)?,
-            ),
-        })
+
+        let secret = SecretKey::try_from(&bytes[..BLS_SECRET_KEY_SIZE])?;
+        let pubkey_bytes: &[u8; BLS_PUBLIC_KEY_AFFINE_SIZE] = bytes[BLS_SECRET_KEY_SIZE..]
+            .try_into()
+            .map_err(|_| BlsError::ParseFromBytes)?;
+        let public = PubkeyAffine::try_from(pubkey_bytes)?;
```
zz-sol, line 86: do we want to check that the SK/PK are indeed a pair? or we should add a comment that the sanity check on the keypair is not performed? I think either is fine.

zz-sol (dismissed):

bls-signatures/src/keypair.rs
samkim-crypto: Good idea! I ended up adding the test to check sk/pk validity.

PR #498, opened 2025-12-17
message: make `is_writable_index` pub

buffalojoec:
#### Problem
In testing tools like [Mollusk](https://github.com/anza-xyz/mollusk), we currently reimplement lots of the message compilation strategy in order to merge privileges and lookup account indices - primarily to compile instruction and transaction accounts.

https://github.com/anza-xyz/mollusk/blob/main/keys/src/keys.rs
https://github.com/anza-xyz/mollusk/blob/main/keys/src/accounts.rs

However, it would be super nice to just use `Message` directly from the SDK and delete all of the above linked code! Instead, we could just do something like this:

```rust
let message = Message::new(&[instruction], None);
let accounts: Vec<_> = message.account_keys.iter().enumerate()
  .map(|(i, key)| {
      InstructionAccount::new(i as IndexOfAccount, message.is_signer(i), message.is_writable_index(i))
  })
  .collect();
```

However, we need public access to `Message::is_writable_index`.

#### Summary of Changes
Make `Message::is_writable_index` public.

### DIFF
--- a/message/src/legacy.rs
+++ b/message/src/legacy.rs
@@ -520,7 +520,7 @@ impl Message {
 
     /// Returns true if the account at the specified index was requested to be
     /// writable. This method should not be used directly.
-    pub(super) fn is_writable_index(&self, i: usize) -> bool {
+    pub fn is_writable_index(&self, i: usize) -> bool {
         i < (self.header.num_required_signatures as usize)
             .saturating_sub(self.header.num_readonly_signed_accounts as usize)
             || (i >= self.header.num_required_signatures as usize


message/src/legacy.rs
```diff
@@ -520,7 +520,7 @@ impl Message {
 
     /// Returns true if the account at the specified index was requested to be
     /// writable. This method should not be used directly.
```
joncinque, line 522: We should also update this comment

joncinque (commented):
This makes sense to me -- the function is named clearly enough to me, so that users shouldn't be confused about whether the account is truly writable, but rather just in the writable portion of account keys.

However, the comment does say it shouldn't be used directly, so we should have @jstarry weigh in to make sure this is OK.

For reference, the function was added in https://github.com/solana-labs/solana/pull/35340/

message/src/legacy.rs
buffalojoec: Will update the comment once I hear from @jstarry! I'm assuming the rationale has to do with the reserved account keys, but maybe there's a way we can safely document it and make it available.

jstarry:
I agree that the current API for writable checks is not ideal at all and I think the main issue is that this API was used by both the core validator and end users. The reason the visibility was conservative is to prevent the core validator from using the wrong API. This is still important as long as these message types are used in agave but it looks like it isn't used anymore? Could you double check? If it's not, then I wouldn't mind a vis change here and I think we could actually just clean up the whole API because ideally end users shouldn't have to worry about write lock demotion at all. It typically means they're doing something wrong and we can document that in the API to help them avoid issues.

buffalojoec:
Cool thanks. Let me look into this some more. I may actually be able to just use Message directly and not even use these accessors. I'll circle back here and either revisit the changes or close the PR.

https://github.com/anza-xyz/mollusk/pull/189

PR #500, opened 2025-12-19
reward-info: change `commission` to `commission_bps`

buffalojoec:
#### Problem
As part of the implementation of [SIMD-0291](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0291-commission-rate-in-basis-points.md), we'll need `RewardInfo` to track inflation rewards commission in basis points, since the runtime will be switching over to this accounting.

Draft implementation of SIMD-0291: https://github.com/anza-xyz/agave/pull/9668

#### Summary of Changes
Change the field on `RewardInfo` from `commission: Option<u8>` to `commission_bps: Option<u16>`.

### DIFF
--- a/reward-info/src/lib.rs
+++ b/reward-info/src/lib.rs
@@ -40,6 +40,7 @@ pub struct RewardInfo {
     pub lamports: i64,
     /// Account balance in lamports after `lamports` was applied
     pub post_balance: u64,
-    /// Vote account commission when the reward was credited, only present for voting and staking rewards
-    pub commission: Option<u8>,
+    /// Vote account commission in basis points when the reward was credited,
+    /// only present for voting and staking rewards.
+    pub commission_bps: Option<u16>,
 }


jstarry:
Unfortunately this struct is exposed in the block meta geyser interface so this is a breaking change there. I think it's ok to make a breaking change there though since the next agave version will be a major version.

jstarry (approved):

jstarry:
Btw this crate could be migrated to the agave repo if you feel like it, it's on used for the runtime for the most part. `RewardType` is used a bit in RPC though.

buffalojoec:
> Btw this crate could be migrated to the agave repo if you feel like it, it's on used for the runtime for the most part. `RewardType` is used a bit in RPC though.

I'm not opposed to it, but we have to consider if anyone uses it on client-side around the ecosystem. @joncinque - what do you think?

joncinque:
I'm fine with moving it to Agave, but note that we would need to move it back out if we split out the clients from the Agave repo, which has been a lingering consideration.

PR #501, opened 2025-12-21
Remove dead code in alt_bn128_pairing

ripatel-fd:
Removes unreachable code. This is not a consensus breaking change

See [usize::checked_rem](https://doc.rust-lang.org/std/primitive.usize.html#method.checked_rem):

> Checked integer remainder. Computes self % rhs, returning None if rhs == 0.

And rhs is a non-zero constant.

### DIFF
--- a/bn254/src/pairing.rs
+++ b/bn254/src/pairing.rs
@@ -158,13 +158,6 @@ pub fn alt_bn128_pairing(input: &[u8]) -> Result<Vec<u8>, AltBn128Error> {
     }
     #[cfg(target_os = "solana")]
     {
-        if input
-            .len()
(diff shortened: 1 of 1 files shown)


joncinque (commented):
Makes sense to me! Let's get @samkim-crypto 's approval though

samkim-crypto (approved):

samkim-crypto:
@ripatel-fd Sorry it seems like the CI is stuck. Can you rebase 🙏 to get the CI restarted? I'll go ahead and merge after CI is completed.

ripatel-fd:
@samkim-crypto Of course, rebased

PR #473, opened 2025-11-29
account: enable frozen-abi StableAbi

puhtaytow:
#### Description

This PR enables `StableAbi`, functionality for frozen-abi types in the crate.

#### Summary of changes
All frozen-abi types got enabled `StableAbi`.

PR #502, opened 2025-12-21
frozen-abi: swap StandardUniform with Arbitrary in StableAbi trait

puhtaytow:
#### Problem

It's tedious to implement manually, the `rand::distr::StandardUniform` trait for each `StableAbi` enabled type.

The suggested solution to mitigate amount of extra code [here](https://discord.com/channels/428295358100013066/476811830145318912/1450967916735500359)

#### Summary of chaneges

- `StableAbi` trait has now constraint `arbitrary::Arbitrary` in default impl of  `random()`
- Added docs
- Added arbitrary re-export (just in case)

### DIFF
--- a/frozen-abi/src/stable_abi.rs
+++ b/frozen-abi/src/stable_abi.rs
@@ -17,15 +17,16 @@ pub trait StableAbi: Sized {
     where
         Self: for<'a> Arbitrary<'a>,
     {
-        // NOTE: As bytes being consumed, it might lead to buffer exhaust.
-        // In such case increase the `MAX_TYPE_SIZE` accordingly
-        // keeping in mind that increasing this value has impact on frozen-abi tests runtime.
-        const MAX_TYPE_SIZE: usize = 4096;
+        const FALLBACK_SIZE: usize = 4096;
+        const DEPTH: usize = 0;
 
-        let mut buffer = vec![0u8; MAX_TYPE_SIZE];
+        let (min, max) = Self::size_hint(DEPTH);
+        let buffer_size = max.unwrap_or(FALLBACK_SIZE).max(min);
+
+        let mut buffer = vec![0u8; buffer_size];
         rng.fill_bytes(&mut buffer);
         let mut unstructured = Unstructured::new(&buffer);
 
-        Self::arbitrary(&mut unstructured).expect("failed to fill - try increasing MAX_TYPE_SIZE")
+        Self::arbitrary(&mut unstructured).expect("failed to fill")
     }
 }


frozen-abi/src/stable_abi.rs
```diff
@@ -1,10 +1,31 @@
...
+///
+/// Meant to be used with `#[derive(StableAbi)]` from `frozen-abi-macro`.
 pub trait StableAbi: Sized {
     fn random(rng: &mut impl RngCore) -> Self
     where
-        StandardUniform: rand::distr::Distribution<Self>,
+        Self: for<'a> Arbitrary<'a>,
     {
-        rng.random::<Self>()
+        // NOTE: As bytes being consumed, it might lead to buffer exhaust.
```
alexpyattaev: pls make sure spelling & grammar are ok in the comments, rustc does not spellcheck (yet)

 Besides that, this will only ever work for Sized types, so we know their size statically at compile time. We can choose the buffer size to match the type byte size. This way we never overrun the buffer, and also do not need to produce RNG data for regions we can never use since our type is 4 bytes and we make 4K RNG bytes.

frozen-abi/src/stable_abi.rs
```diff
@@ -1,10 +1,31 @@
...
+///
+/// Types that implement this trait can be tested via `test_abi_digest` (generated by
+/// the `#[frozen_abi(abi_digest = "...")]` attribute)
+/// which is used to detect unintended changes in serialized representations.
+///
+/// Meant to be used with `#[derive(StableAbi)]` from `frozen-abi-macro`.
 pub trait StableAbi: Sized {
     fn random(rng: &mut impl RngCore) -> Self
     where
-        StandardUniform: rand::distr::Distribution<Self>,
```
alexpyattaev, line 6: I think we can keep this blanket impl alongside with the impl based on Arbitrary. Would require some type system shenanigans along the lines of https://www.greyblake.com/blog/alternative-blanket-implementations-for-single-rust-trait/ to make it work though. Not sure if worth the trouble, but might allow us to add support for Arbitrary as an add-on, rather than replacement for Distribution.

frozen-abi/src/stable_abi.rs
puhtaytow: I had there before `size_of::<Self>()`, but it was causing problems with collections, thats why decided to make static big size buffer.. but it turns out that i should do instead RTFM as there is solution in [Arbitrary](https://docs.rs/arbitrary/latest/arbitrary/trait.Arbitrary.html#method.size_hint).

I still have const with 4096 as fallback coz, returned max from `size_hint()` is type `Option` which might turn to be `None`.

Please have a look and let me know.. and ofc, big thank you for pointing it out :pray:

frozen-abi/src/stable_abi.rs
puhtaytow: I have spent some time on trying to crack on with this idea and despite the fact thats cool trick, i dont see it helping in our case.

Keeping Arbitrary and Distribution would bring unnecesary ambiguity without any clear win (correct me if im wrong) and would demand changes not only in `StableAbi` signature, but as well in makros.

If I had to make decision I would avoid going this way, but if we decide otherwise then ill apply it.

As always thank you for the suggestion :pray: and the link :pray: i didnt know that is even possible.

alexpyattaev: well it was a stretch anyway. type system gymnastics are not the point here.

puhtaytow:
We had discussion with @alexpyattaev and agreed that it would be appropriate to write down pros and cons coming from this PR so we could decide either it is worth moving to `Arbitrary` or keeping the current solution.

Currently` StableAbi` uses `Distribution` from rand crate in the default implementation of [`random()`](https://github.com/anza-xyz/solana-sdk/blob/master/frozen-abi/src/stable_abi.rs#L4-L10) method, which requires us to manually implement it for every `StableAbi` enabled type in order to use the default implementation.  While this is tedious, it also comes with good control over the content of sampled instances, as we have to explicitly construct it. Whether this is a benefit or pain might depend on the use case.

While we have observed that in some cases `Arbitrary` reduces the boilerplate code since we can simply derive, it also rising questions about whether using derive would reflect in a desired type sample for every derivable case, or if it would be only appropriate for a small subset of types, while others would still demand a manual implementation.

Why does it matter? because it might turn out that we would introduce a breaking change / yank wuthout real or substantial benefits.

---

#### Pros

- Possibility of simply using derive instead of manual implementation for types where it doesnt matter (from realistic usecase perspective) whats inside the filled type, since what only matter is the binary layout after all, so as long as its not empty we should be good.

#### Cons

- Breaking change / yank
- Commonly used types would need have derived `Arbitrary` as well to unlock derivation on structs including these e.g. [address](https://github.com/anza-xyz/solana-sdk/pull/503/files), hash and few others.
- Static buffer size for random bytes in the default implementation of `random()` used to fill a type. Might either lead to too much random bytes to fill a type or too little. 
  - **Too small:** Would prevent using the default impllementation of `random()` method, requiring a manual implementation with a appropriately sized buffer to cover the edge case.
  - **Too large:** Would increase test runtime which we ideally should avoid.
  
---
Below examples using both solutions

#### Distribution

- [solana-sdk/account](https://github.com/anza-xyz/solana-sdk/pull/473/files)
- [solana-sdk/vote-interface](https://github.com/anza-xyz/solana-sdk/pull/472/files)

- [agave/accounts-db](https://github.com/anza-xyz/agave/pull/9338/files)
- [agave/votor-messages](https://github.com/anza-xyz/agave/pull/9336/files)

#### Arbitrary
- [solana-sdk/vote-interface/account](https://github.com/puhtaytow/solana-sdk/compare/master...frozen-abi-stable-abi-arbitrary-filler)

- [agave/accounts-db/votor-messages](https://github.com/anza-xyz/agave/compare/master...puhtaytow:agave:frozen-abi-stable-abi-arbitrary?expand=1)

As you can see, the difference is clearly visible in the vote-interface / account case, but its questionable whether we care about them being filled realistically or not, and ultimately whether it's worth a breaking change to save maybe a few hundred lines of boilerplate code across both repos.

Thank you for reaching to this point :pray: and please tell me what do you think :pray: :pray: 

cc @joncinque :pray:

frozen-abi/src/stable_abi.rs
```diff
@@ -1,10 +1,32 @@
...
+        const DEPTH: usize = 0;
+
+        let (min, max) = Self::size_hint(DEPTH);
+        let buffer_size = max.unwrap_or(FALLBACK_SIZE).max(min);
+
+        let mut buffer = vec![0u8; buffer_size];
+        rng.fill_bytes(&mut buffer);
+        let mut unstructured = Unstructured::new(&buffer);
+
+        Self::arbitrary(&mut unstructured).expect("failed to fill")
```
nagisa, line 30: As written on Discord, I believe the `size_hint` use here to ensure infallibility somewhat suspicious. `Arbitrary` trait/its implementation does not guarantee that generating a type won't use more than maximum number of bytes in the hint. It could also return `None` for the max bound and try to use more than `FALLBACK_SIZE` bytes.

puhtaytow: >  It could also return None for the max bound and try to use more than FALLBACK_SIZE bytes.

Thats good catch indeed. You are absolutely right on that one, i just assumed that the default implementation of `random()` would suffice in many cases and if not then user would have to provide manual implementation with appropriate buffer size.

To be honest with you im more convinced that we should stay with `Distribution` coz despite it requiring manual implementation for every `StableAbi` enabled type (cant be derived as simply as `Arbitrary`) it would result in generating exact amount of data to fill the tested type and its down to the implementator how complex or simple it would be.

Thank you for your engagement :pray:

joncinque:
Thanks for the feedback @nagisa -- let's close this PR then and stick with the `rand` trait.

PR #503, opened 2025-12-21
address: enable Arbitrary for frozen-abi feature

puhtaytow:
#### Description

Address is commonly used among frozen-abi enabled types. Without an implementation of `arbitrary::Arbitrary`, they block other types from easy derivation.

#### Summary of changes

- `arbitrary::Arbitrary` being derived now not only in case of `DCOU` but `frozen-abi` as well.


puhtaytow:
closed as there is no intentions to merge #502

PR #504, opened 2025-12-22
short-vec: Make serde an optional (default) feature

cpubot:
Crates like wincode may pull in `solana-short-vec` if the feature is enabled, but will not use any of the `serde` features -- it implements its own crate-specific traits for the `ShortU16` type.

This PR makes `serde` a feature, rather than a hard dependency. The `serde` feature is enabled by default to avoid downstream breakage.

### DIFF
--- a/message/Cargo.toml
+++ b/message/Cargo.toml
@@ -34,6 +34,7 @@ serde = [
     "serde/alloc",
     "solana-address/serde",
     "solana-hash/serde",
+    "solana-short-vec/serde",
 ]
 wincode = ["dep:wincode", "solana-hash/wincode", "solana-address/wincode"]
 
--- a/serde-varint/Cargo.toml
+++ b/serde-varint/Cargo.toml
@@ -19,4 +19,4 @@ serde = { workspace = true }
 bincode = { workspace = true }
 rand = { workspace = true }
 serde_derive = { workspace = true }
-solana-short-vec = { workspace = true }
+solana-short-vec = { workspace = true, features = ["serde"] }
--- a/signature/Cargo.toml
+++ b/signature/Cargo.toml
@@ -48,7 +48,7 @@ ed25519-dalek = { workspace = true }
 serde_derive = { workspace = true }
 serde_json = { workspace = true }
 solana-pubkey = { workspace = true, features = ["std"] }
-solana-short-vec = { workspace = true }
+solana-short-vec = { workspace = true, features = ["serde"] }
 solana-signature = { path = ".", features = ["serde"] }
 
 [lints]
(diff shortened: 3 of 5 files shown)


short-vec/src/lib.rs
```diff
@@ -52,6 +55,7 @@ enum VisitStatus {
 }
 
 #[derive(Debug)]
+#[cfg_attr(not(feature = "serde"), allow(dead_code))]
```
cpubot, line 58: This enum is still used by `decode_shortu16_len` internally when the `serde` feature is disabled, but the `TooShort` variant is not used and `TooLong` and `Overflow` values are never extracted.

short-vec/Cargo.toml
```diff
@@ -13,10 +13,12 @@ edition = { workspace = true }
 targets = ["x86_64-unknown-linux-gnu"]
 
 [features]
+default = ["serde"]
```
joncinque, line 16: This is still *technically* a breaking change, since someone could be pulling in the crate with `default-features = false`, even though the default feature doesn't exist.

Practically speaking though, I think it'll be fine.

short-vec/Cargo.toml
joncinque, line 32: To make sure the tests actually run, can you add `solana-short-vec` as a dep with `serde` enabled?

short-vec/src/lib.rs
joncinque: That's fine by me.

Alternatively, the error variants only used for serde could have a `#[cfg(feature = "serde")]` annotation, but since it's an internal type, it doesn't make a big difference.

short-vec/Cargo.toml
cpubot: Done!

short-vec/src/lib.rs
cpubot: I considered that as an option as well but felt like it might be a bit cumbersome to deal with development-wise.

short-vec/Cargo.toml
cpubot: Actually, this triggers a circular dependency error. I presume the tests should be run regardless since the feature is enabled by default?

cpubot: Looks like it was fixable by using `path = "."` rather than `workspace = true`

short-vec/src/lib.rs
joncinque: Makes sense to me!

short-vec/Cargo.toml
joncinque: Ah sorry I didn't explain -- correct, you have to use `path = "."`

joncinque:
Looks good to me, but we'll need to explicitly enable the serde feature on downstream crates that do require serde from short-vec

PR #505, opened 2025-12-22
ci: Partition powerset job

joncinque:
#### Problem

The powerset job takes a long time in the 30-40 minutes range. This is annoying for development.

#### Summary of changes

We could eventually look into only running on `--each-feature`, but for now, let's partition the powerset job into 6, which should make things much faster for devs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -548,3 +548,32 @@ jobs:
 
       - name: Run check
         run: ./scripts/check-no-std.sh
+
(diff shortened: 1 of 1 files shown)


joncinque:
@yihau I've added a new step at the end of the workflow which just checks that everything else ran, so that the branch protection rule only needs to have that final job as "required", which makes it easier if we want to change the number of partitions in the future.

Let me know what you think!

PR #506, opened 2025-12-27
[bls-signatures] Add support for mixed additions and update verification syntax

samkim-crypto:
### Summary of Changes

This PR bundles three updates for the `bls-signatures` crate for v2.0.0 release targeting agave v4. It would be better to break this PR into three separate PR, but since we are in a tight release schedule for agave v4 (mid Jan), I combined these changed into a single PR 🙏.

1. I enabled mixed addition (adding an Affine point directly to a Projective point), which`blst` seems to support. Previously, we were converting everything to Projective before adding, which was slower.

For this, I added `AddToSignatureProjective` and `AddToPubkeyProjective`. If you pass a Projective point, it does standard addition. If you pass an Affine point (or bytes that convert to Affine), it uses the optimized mixed addition.

The benchmarks show that this cuts public key and signature aggregation time by about 20-25% when inputs are in Affine representation (which would be the case for alpenglow).

```
Benchmarking aggregate_verify/64 signature aggregation: Collecting 100 samples in estimated 5.0554 s (91k iteratiaggregate_verify/64 signature aggregation
                        time:   [55.449 µs 55.466 µs 55.487 µs]
                        change: [-26.774% -26.612% -26.468%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 16 outliers among 100 measurements (16.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild
  10 (10.00%) high severe
Benchmarking aggregate_verify/64 pubkey aggregation: Collecting 100 samples in estimated 5.0431 s (227k iterationaggregate_verify/64 pubkey aggregation
                        time:   [22.153 µs 22.169 µs 22.193 µs]
                        change: [-21.574% -21.360% -21.169%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  2 (2.00%) high mild
  11 (11.00%) high severe
```

2. I noticed that I forgot to update some of the verification functions in https://github.com/anza-xyz/solana-sdk/pull/496. The verification functions were expecting serialized bytes, forcing us to serialize bytes, forcing us to serialize keys just to verify them. I updated these functions to be generic so that they can now verify points directly without the extra serialization step.

3. I updated the verification function return type:
- We previously returned `Result<bool, BlsError>`
- Now we return `Result<(), BlsError>`

These should be all the changes we need to include for the v2.0.0 release!

### DIFF
--- a/bls-signatures/src/signature/points.rs
+++ b/bls-signatures/src/signature/points.rs
@@ -1,5 +1,10 @@
 #[cfg(all(not(target_os = "solana"), feature = "std"))]
 use crate::pubkey::points::NEG_G1_GENERATOR_AFFINE;
+#[cfg(all(
+    not(target_os = "solana"),
+    any(feature = "parallel", not(feature = "std"))
+))]
+use blstrs::G1Affine;
 #[cfg(not(feature = "std"))]
 use blstrs::G1Projective;
 #[cfg(not(target_os = "solana"))]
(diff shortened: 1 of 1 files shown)


pddiffer053-alt:
_[]()_

zz-sol (approved):
LGTM. Nice optimizations!

PR #507, opened 2025-12-29
system-interface: fix docs

LStan:
`to_pubkey` input of `create_account` was renamed to `to_address`, but it is still called `to_pubkey` in the docs.

### DIFF
--- a/system-interface/src/instruction.rs
+++ b/system-interface/src/instruction.rs
@@ -264,7 +264,7 @@ pub enum SystemInstruction {
     /// current lamports must be 0.
     ///
     /// This constructor is identical to [`create_account`] with the exception that it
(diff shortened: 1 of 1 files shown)


system-interface/src/instruction.rs
```diff
@@ -264,7 +264,7 @@ pub enum SystemInstruction {
     /// current lamports must be 0.
     ///
     /// This constructor is identical to [`create_account`] with the exception that it
-    /// **does not** check that the destination account (`to_pubkey`) has a zero
+    /// **does not** check that the destination account (`to_address`) has a zero
```
joncinque, line 267: There's a bit of inconsistency, since `create_account_allow_prefund` calls the `to_address` parameter `new_account_address`, but I think this is clearer.

PR #512, opened 2026-01-02
Bump thiserror to 2.0.17

samkim-crypto:
#### Summary of Changes

Bumping `thiserror` to v2.0.17. Seems like `Cargo.lock` already uses v2.0.17, so not need to update `Cargo.lock`.


samkim-crypto:
Okay makes sense!

PR #515, opened 2026-01-08
SIMD-0430: Loader V3: Relax Program Buffer Constraints

buffalojoec:
#### Problem
[SIMD-0430](https://github.com/solana-foundation/solana-improvement-documents/pull/430) introduces an optional `close_buffer: bool` which tells Loader V3 whether or not to close the buffer account being used for deploy (`DeployWithMaxDataLen`) or upgrade (`Upgrade`).

The interface must be updated to support this change.

#### Summary of Changes
Add an API for deserializing the optional parameter from a slice of bytes.

Unfortunately we can't modify the enum variants directly, since the program still needs to treat legacy payloads - where the boolean argument is not provided - as valid, and default `close_buffer` to `true`.

### DIFF
--- a/loader-v3-interface/src/instruction.rs
+++ b/loader-v3-interface/src/instruction.rs
@@ -89,12 +89,22 @@ pub enum UpgradeableLoaderInstruction {
     ///   1. `[writable]` The uninitialized ProgramData account.
     ///   2. `[writable]` The uninitialized Program account.
     ///   3. `[writable]` The Buffer account where the program data has been
-    ///      written.  The buffer account's authority must match the program's
-    ///      authority
+    ///      written.
     ///   4. `[]` Rent sysvar.
     ///   5. `[]` Clock sysvar.
     ///   6. `[]` System program (`solana_sdk_ids::system_program::id()`).
     ///   7. `[signer]` The program's authority
+    ///
+    /// An additional parameter can be serialized at the end of this
+    /// instruction. Consider this `close_buffer: bool`.
+    /// * If `true`:
+    ///   * The buffer account's authority must match the program's authority.
+    ///   * The buffer account will be closed and its lamports transferred to
+    ///     the payer.
+    /// * If `false`:
+    ///   * The buffer account's authority *does not* have to match the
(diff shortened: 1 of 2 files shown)


joncinque (commented):
This code works, but it seems silly to do all this when we can just add two enum variants.

I've submitted my thoughts to the will of the masses: https://github.com/solana-foundation/solana-improvement-documents/pull/430#issuecomment-4185123268

buffalojoec:
> This code works, but it seems silly to do all this when we can just add two enum variants.
> 
> I've submitted my thoughts to the will of the masses: [solana-foundation/solana-improvement-documents#430 (comment)](https://github.com/solana-foundation/solana-improvement-documents/pull/430#issuecomment-4185123268)

Personally I don't care too much, but the SIMD with an extra byte workflow is approved & merged from both validator teams, and I've got the implementation waiting here. Just let me know who I need to request review from if you are abstaining.

loader-v3-interface/src/instruction.rs
```diff
@@ -89,12 +89,22 @@ pub enum UpgradeableLoaderInstruction {
     ///   1. `[writable]` The uninitialized ProgramData account.
     ///   2. `[writable]` The uninitialized Program account.
     ///   3. `[writable]` The Buffer account where the program data has been
```
Lichtso, line 91: The buffer is not necessarily `[writable]` depending on the `close_buffer` parameter. But that might be hard to convey.

joncinque:
> the SIMD with an extra byte workflow is approved & merged from both validator teams, and I've got the implementation waiting here. Just let me know who I need to request review from if you are abstaining.

In the SIMD it was brought up that it doesn't consider such specific implementation details. I'm not abstaining, I'm actively opposing, there's a difference :grinning:

loader-v3-interface/src/instruction.rs
buffalojoec: Yeah, I thought it was easier to just leave the accounts requirement docs alone. If you think it's necessary I can try to document it, but IMO it's fine.

PR #516, opened 2026-01-08
SIMD-0431: Loader V3: Minimum Extend Program Size

buffalojoec:
#### Problem
[SIMD-0431](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0431-minimum-extend-program-size.md) has been approved and merged. It needs to be implemented in the SDK's loader-v3-interface.

#### Summary of Changes
Implements SIMD-0431 by adding a constant `MINIMUM_EXTEND_PROGRAM_BYTES` and documentation about the new minimum length requirement to the instruction's documentation.

### DIFF
--- a/loader-v3-interface/src/instruction.rs
+++ b/loader-v3-interface/src/instruction.rs
@@ -9,6 +9,14 @@ use {
     solana_system_interface::instruction as system_instruction,
 };
 
+/// Minimum number of bytes for an `ExtendProgram` instruction.
+///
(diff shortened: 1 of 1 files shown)


buffalojoec:
> I do wonder if it's worth creating another instruction creator that returns an error if the amount is too small, since there's no real way to figure that out without reading the code.

Yeah, I thought about this, too, but IMO I left it up to the client. Until the feature is enabled, no reason you can't use this crate still as-is.

PR #517, opened 2026-01-08
delete reward-info crate

buffalojoec:
This crate has been moved to Agave. We can remove it here in the SDK.
https://github.com/anza-xyz/agave/pull/9857


buffalojoec:
Just kidding we re-export `RewardType` (not `RewardInfo`) in RPC client types.

PR #520, opened 2026-01-08
SIMD-406: Restrict to 255 the number of accounts in instructions for transactions v0 and legacy

LucasSte:
In [SIMD-406](https://github.com/solana-foundation/solana-improvement-documents/pull/406), we proposed that v0 and legacy transaction sanitization must now have an extra check to limit the number of accounts per instruction to 255.

This PR implements this check behind a boolean flag that will be coupled with a feature gate in Agave.

### DIFF
--- a/sanitize/src/lib.rs
+++ b/sanitize/src/lib.rs
@@ -34,15 +34,15 @@ impl fmt::Display for SanitizeError {
 /// - All index values are in range.
 /// - All values are within their static max/min bounds.
 pub trait Sanitize {
-    fn sanitize(&self) -> Result<(), SanitizeError> {
+    fn sanitize(&self, _limit_ix_accounts_simd_406: bool) -> Result<(), SanitizeError> {
         Ok(())
     }
 }
 
 impl<T: Sanitize> Sanitize for [T] {
-    fn sanitize(&self) -> Result<(), SanitizeError> {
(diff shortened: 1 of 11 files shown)


sanitize/src/lib.rs
```diff
@@ -34,15 +34,15 @@ impl fmt::Display for SanitizeError {
 /// - All index values are in range.
 /// - All values are within their static max/min bounds.
 pub trait Sanitize {
-    fn sanitize(&self) -> Result<(), SanitizeError> {
+    fn sanitize(&self, _limit_ix_accounts_simd_406: bool) -> Result<(), SanitizeError> {
```
jstarry, line 37: There are a lot of things in the agave repo that implement this trait so I don't think it's a good idea to change this trait method. Check out https://github.com/anza-xyz/agave/pull/8182 for another approach that doesn't involve changing the `Sanitize` trait. After these features are activated we can update the `Sanitize` implementation in the SDK without a boolean for the feature.

jstarry (changes requested):

sanitize/src/lib.rs
LucasSte: Thanks for the hint. I'll follow an approach similar to the PR you pointed.

LucasSte:
Superseded by https://github.com/anza-xyz/agave/pull/10223.

PR #522, opened 2026-01-08
Remove `fn create` from `WritableAccount`

LucasSte:
`fn create` has been deprecated in https://github.com/anza-xyz/solana-sdk/pull/488, its usages have been removed from Agave in https://github.com/anza-xyz/agave/pull/9279 and the monorepo dependency has been bumped in https://github.com/anza-xyz/agave/pull/9706, so it is now safe to remove the function from the crate.

This is a follow up to https://github.com/anza-xyz/solana-sdk/pull/386.


joncinque (approved):
Looks great!

brooksprumo (approved):
:shipit:

PR #526, opened 2026-01-10
address: Implement on-curve check for programs

febo:
### Problem

Currently `bytes_are_curve_point` is not implemented for on-chain programs, although there is a syscall available.

### Solution

 Add the implementation using the `sol_curve_validate_point` syscall.

### DIFF
--- a/address/src/lib.rs
+++ b/address/src/lib.rs
@@ -198,31 +198,22 @@ pub fn bytes_are_curve_point<T: AsRef<[u8]>>(bytes: T) -> bool {
 
     #[cfg(any(target_os = "solana", target_arch = "bpf"))]
     {
-        #[cfg(feature = "syscalls")]
-        {
-            // ID for the Ed25519 as defined in `solana-curve25519::CURVE25519_EDWARDS`.
-            const CURVE25519_EDWARDS: u64 = 0;
+        // ID for the Ed25519 as defined in `solana-curve25519::CURVE25519_EDWARDS`.
+        const CURVE25519_EDWARDS: u64 = 0;
 
-            // The syscall return `0` when the point is valid – i.e., on the curve;
-            // otherwise it returns `1`
-            //
-            // SAFETY: The syscall validates the input.
-            let result = unsafe {
-                syscalls::sol_curve_validate_point(
-                    CURVE25519_EDWARDS,
-                    bytes.as_ref() as *const _ as *const u8,
-                    core::ptr::null_mut(),
-                )
-            };
+        // The syscall return `0` when the point is valid – i.e., on the curve;
+        // otherwise it returns `1`
+        //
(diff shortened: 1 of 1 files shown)


address/src/lib.rs
```diff
@@ -172,23 +173,57 @@ impl TryFrom<&str> for Address {
...
 }
 
-// If target_os = "solana" or target_arch = "bpf", then this panics so there
-// are no dependencies; otherwise this should be opt-in so users don't need the
-// curve25519 dependency.
-#[cfg(any(target_os = "solana", target_arch = "bpf", feature = "curve25519"))]
-#[allow(clippy::used_underscore_binding)]
-pub fn bytes_are_curve_point<T: AsRef<[u8]>>(_bytes: T) -> bool {
+/// Check whether the given bytes represent a valid curve point.
+#[cfg(any(feature = "curve25519", feature = "syscalls"))]
```
febo, line 177: This was needed otherwise the method is not visible when the feature `"curve25519"` is not enabled.

address/src/lib.rs
```diff
@@ -275,10 +310,13 @@ impl Address {
...
-    // If target_os = "solana" or target_arch = "bpf", then this panics so there
-    // are no dependencies; otherwise, this should be opt-in so users don't need
-    // the curve25519 dependency.
-    #[cfg(any(target_os = "solana", target_arch = "bpf", feature = "curve25519"))]
+    /// Checks whether the given address lies on the Ed25519 curve.
+    ///
+    /// On-curve addresses correspond to valid Ed25519 public keys (and therefore
+    /// can have associated private keys). Off-curve addresses are typically used
+    /// for program-derived addresses (PDAs).
+    #[cfg(any(feature = "curve25519", feature = "syscalls"))]
```
febo, line 309: Same here: this was needed otherwise the method is not visible when the feature `"curve25519"` is not enabled.

address/src/lib.rs
```diff
@@ -172,23 +173,57 @@ impl TryFrom<&str> for Address {
...
+            };
+
+            result == 0
+        }
+
+        #[cfg(not(feature = "syscalls"))]
+        {
+            core::hint::black_box(bytes);
+            panic!("bytes_are_curve_point is only available with the `syscalls` feature enabled on this crate")
+        }
```
joncinque: Since the `curve25519` feature enables `syscalls`, I don't think we need this branch at all -- if you're in this function, then `syscalls` is definitely enabled

febo: Good point, removed in https://github.com/anza-xyz/solana-sdk/pull/526/commits/0e2a3316783c19f196f2fd298178cbd5044257f0.

joncinque (approved):
Looks good to me!

PR #529, opened 2026-01-11
docs: add StableAbi docs to frozen-abi

puhtaytow:
### Description

As per request [here](https://github.com/anza-xyz/agave/pull/9338#discussion_r2679683480) 
Last piece of https://github.com/anza-xyz/agave/issues/8091

### Summary of changes
- Added StableAbi docs to `frozen-abi/src/lib.rs`

### DIFF
--- a/frozen-abi/src/lib.rs
+++ b/frozen-abi/src/lib.rs
@@ -1,7 +1,8 @@
 //! # StableAbi
 //!
-//! The `StableAbi` is an optional extension to `frozen-abi` that provides functionality for detecting unintended encoding changes.
-//! It is designed to be used in conjuction with the mandatory `AbiExample`.
+//! The `StableAbi` is an optional extension to `frozen-abi` that provides functionality for
+//! detecting unintended encoding changes. It is designed to be used in conjunction with the
+//! mandatory `AbiExample`.
 //!
 //! ## How it works?
 //!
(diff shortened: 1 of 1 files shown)


puhtaytow:
@brooksprumo :pray:

frozen-abi/src/lib.rs
```diff
@@ -1,3 +1,67 @@
...
+//!
+//! ```rust,ignore
+//! impl ::solana_frozen_abi::stable_abi::StableAbi for MyType {}
+//! ```
+//!
+//! ## Edge Cases
+//!
+//! 1. It will not detect field name or order changes, nor same size type swaps (e.g., `i64` and `u64`). These cases are still covered by `AbiExample`
+//! 2. The implementor must ensure a consistent order of `rng.random()` calls, as any change will result in different hash
+//! 3. For collection types with non deterministic ordering (e.g., `HashMap`), it is recommended to insert only one item to avoid false positives caused by iteration order differences
```
brooksprumo: Probably good to re-format the whole thing now, before merging. Let's not have lines exceeding the 100-char limit.

Rust 2024 edition formatting context: https://github.com/anza-xyz/agave/issues/6203#issuecomment-3709875936

puhtaytow: Now should be good. Thank you :pray:

brooksprumo (approved):
:shipit:

PR #530, opened 2026-01-14
[bn254] Remove subgroup check on G2 addition

samkim-crypto:
#### Problem

I benchmarked the G2 addition and multiplication to assign their compute units. However, it seems like G2 addition performs about 20x that of G1 addition. We can go ahead and assign appropriate CUs, but 20x seems a little too much.

The reason why G2 addition is so much slower than G1 addition is due to the subgroup (coset) check. The subgroup validity check is almost trivial for G1 (has cofactor of 1), but computationally very expensive for G2 since the cofactor is quite large.

We can assign the higher CUs for Agave 4.0 and relax the check in 4.1 (and lower CU then), but given the demand for this feature in Agave 4.0, I think it is preferable to ship with optimal CUs now for 4.0.

After the subgroup check is removed, the G2 addition costs about x2.15 that of G1 addition.

#### Summary of Changes

I removed the subgroup check for G2. Changing the subgroup check for G1 will be a breaking change and will require a SIMD, so I left it the same.

Let's relax G2 addition and merge it to agave 4.0. Then let's add a SIMD to relax G1 subgroup check for agave 4.1.

Related discussion: https://github.com/anza-xyz/solana-sdk/pull/494#discussion_r2683751718

cc: @0x0ece

### DIFF
--- a/bn254/src/addition.rs
+++ b/bn254/src/addition.rs
@@ -67,10 +67,6 @@ pub enum VersionedG2Addition {
 /// for validator code. Solana programs or other downstream projects should use
 /// `alt_bn128_g1_addition_be` or `alt_bn128_g1_addition_le` instead.
 ///
-/// # Security Note: Unlike G2 addition, G1 addition validates both the curve equation
-/// and the subgroup (coset). Future versions may relax this to check only the
-/// curve equation.
-///
 /// # Warning
 ///
 /// Developers should be extremely careful when modifying this function, as a breaking change
@@ -219,8 +215,9 @@ pub fn alt_bn128_g1_addition_le(
 /// for validator code. Solana programs or other downstream projects should use
 /// `alt_bn128_g2_addition_be` or `alt_bn128_g2_addition_le` instead.
 ///
-/// # Security Note: Unlike G1 addition, G2 addition validates only the curve equation;
-/// it does not perform a subgroup (coset) check.
+/// # Security Note: Unlike G1, which has cofactor 1, the group G2 has a high cofactor.
+/// This G2 addition function validates only the curve equation; it does not perform
+/// a subgroup (coset) check.
 ///
 /// # Warning
 ///
--- a/bn254/src/lib.rs
+++ b/bn254/src/lib.rs
@@ -237,8 +237,7 @@ mod target_arch {
     }
 
     impl PodG2 {
-        /// Takes in an EIP-197 (big-endian) byte encoding of a group element in G2
-        /// and constructs a `PodG2` struct that encodes the same bytes in little-endian.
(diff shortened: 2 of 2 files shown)


0x0ece:
> Changing the subgroup check for G1 will be a breaking change and will require a SIMD, so I left it the same.

Thankfully, for BN curves, G1 has prime order. So nothing to do there 🎉

bn254/src/lib.rs
```diff
@@ -237,6 +237,30 @@ mod target_arch {
     }
 
     impl PodG2 {
+        /// Takes in an EIP-197 (big-endian) byte encoding of a group element in G2
+        /// and constructs a `PodG2` struct that encodes the same bytes in little-endian.
```
LStan:
```suggestion
        /// Deserializes to an affine point in G2
```

bn254/src/addition.rs
```diff
@@ -67,6 +67,10 @@ pub enum VersionedG2Addition {
 /// for validator code. Solana programs or other downstream projects should use
 /// `alt_bn128_g1_addition_be` or `alt_bn128_g1_addition_le` instead.
 ///
+/// # Security Note: Unlike G2 addition, G1 addition validates both the curve equation
```
zz-sol: Nits: i would suggest to improve wording -- G1's cofactor = 1

```
/// # Security Note: G1 addition validates both the curve equation. 
/// G1 has cofactor 1, so any point on the curve is automatically
/// in the correct subgroup—no separate subgroup check is needed. 
```

zz-sol (dismissed):
I think this is fine. 

IIRC Geth/Reth follows a same approach

bn254/src/addition.rs
samkim-crypto: Okay great! I just removed the note on G1 because the cofactor check does not actually matter for G1. I reworded the note on top of G2 addition to make clear that unlike G1, the group G2 has a non-trivial cofactor and we skip the coset check.

bn254/src/lib.rs
samkim-crypto: Thanks! Fixed.

samkim-crypto:
> > Changing the subgroup check for G1 will be a breaking change and will require a SIMD, so I left it the same.
> 
> Thankfully, for BN curves, G1 has prime order. So nothing to do there 🎉

Oh yeah I missed that since we have cofactor of 1, we shouldn't even need the coset check.

I still did the benchmarks and if we use the unchecked deserialization with manual `is_on_curve` check, then we get about 5~6% improvement in the benches. At this point, I think we should merge and cut 3.2.1 right away, but for `solana-bn254` v4.0, let's address this.

PR #531, opened 2026-01-15
Add support for V1 transactions

grod220:
Reference: [SIMD-0296](https://github.com/solana-foundation/solana-improvement-documents/pull/296) & [SIMD-0385](https://github.com/solana-foundation/solana-improvement-documents/pull/385)

### Add V1 transaction format

V1 transactions increase the maximum transaction size from 1232 bytes to 4096 bytes and embed compute budget configuration directly in the message header, eliminating the need for separate ComputeBudget program instructions.

#### Changes

  New module (message/src/versions/v1/):
  - message.rs: core `Message` struct
  - config.rs: `ComputeBudgetConfig` and `ComputeBudgetConfigMask` for inline compute budget
  - serialization.rs: custom binary serialization/deserialization
  - runtime.rs: `LoadedMessage` wrapper with precomputed writability for runtime
  - builder.rs: builder API for constructing messages
  - constants.rs: format limits and sizes
  - error.rs: deserialization error types

New module (transaction/src/v1.rs):
  - `Transaction`: transaction wrapper combining `Message` with signatures
  - Custom wire format: `[message bytes][signatures]` (no signature length prefix)

### DIFF
--- a/transaction/src/v1.rs
+++ b/transaction/src/v1.rs
@@ -82,10 +82,10 @@ impl From<MessageError> for TransactionError {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Transaction {
     /// The V1 message containing instructions and accounts.
-    message: Message,
+    pub message: Message,
     /// Transaction signatures, one per required signer.
     /// Order matches the first `num_required_signatures` accounts in the message.
-    signatures: Vec<Signature>,
+    pub signatures: Vec<Signature>,
 }
 
 impl Transaction {
@@ -95,25 +95,6 @@ impl Transaction {
 
     pub fn signatures(&self) -> &[Signature] {
         &self.signatures
-    }
-
-    /// Create a V1 transaction from a message and pre-existing signatures.
-    /// Returns an error if the signature count doesn't match `num_required_signatures`.
-    pub fn from_signatures(
-        message: Message,
-        signatures: Vec<Signature>,
-    ) -> Result<Self, TransactionError> {
-        let expected = message.header.num_required_signatures as usize;
-        if signatures.len() != expected {
-            return Err(TransactionError::SignatureCountMismatch {
-                expected,
-                actual: signatures.len(),
-            });
-        }
-        Ok(Self {
-            message,
-            signatures,
-        })
     }
 
     /// Sign a V1 message and create a transaction.
@@ -372,32 +353,6 @@ mod tests {
             .unwrap()
     }
 
-    #[test]
-    fn from_signatures_rejects_too_few_signatures() {
-        let message = create_test_message(); // requires 1 signature
-
-        assert_eq!(
-            Transaction::from_signatures(message, vec![]),
-            Err(TransactionError::SignatureCountMismatch {
-                expected: 1,
-                actual: 0
-            })
-        );
-    }
-
-    #[test]
-    fn from_signatures_rejects_too_many_signatures() {
-        let message = create_test_message(); // requires 1 signature
-
-        assert_eq!(
-            Transaction::from_signatures(message, vec![test_signature(0x01), test_signature(0x02)]),
-            Err(TransactionError::SignatureCountMismatch {
-                expected: 1,
-                actual: 2
-            })
-        );
-    }
-
     #[cfg(feature = "bincode")]
     #[test]
     fn try_sign_rejects_too_many_signers() {
@@ -567,7 +522,10 @@ mod tests {
     fn serialize_produces_correct_wire_format() {
         let message = create_test_message();
         let sig = test_signature(0xAA);
-        let tx = Transaction::from_signatures(message.clone(), vec![sig]).unwrap();
+        let tx = Transaction {
+            message: message.clone(),
+            signatures: vec![sig],
+        };
 
         let bytes = tx.serialize().unwrap();
 
@@ -595,7 +553,10 @@ mod tests {
     fn from_bytes_rejects_truncated_signatures() {
         let message = create_test_message();
         let sig = test_signature(0xCC);
-        let tx = Transaction::from_signatures(message, vec![sig]).unwrap();
+        let tx = Transaction {
+            message,
+            signatures: vec![sig],
+        };
 
         let mut bytes = tx.serialize().unwrap();
         bytes.truncate(bytes.len() - 10); // Remove part of signature
@@ -610,7 +571,10 @@ mod tests {
     fn from_bytes_rejects_trailing_data() {
         let message = create_test_message();
         let sig = test_signature(0xDD);
-        let tx = Transaction::from_signatures(message, vec![sig]).unwrap();
+        let tx = Transaction {
+            message,
+            signatures: vec![sig],
+        };
 
         let mut bytes = tx.serialize().unwrap();
         bytes.push(0xFF); // Extra byte
@@ -625,7 +589,10 @@ mod tests {
     fn from_bytes_roundtrip_single_signature() {
         let message = create_test_message();
         let sig = test_signature(0xBB);
-        let tx = Transaction::from_signatures(message, vec![sig]).unwrap();
+        let tx = Transaction {
+            message,
+            signatures: vec![sig],
+        };
 
         let bytes = tx.serialize().unwrap();
         let parsed = Transaction::from_bytes(&bytes).unwrap();
@@ -659,7 +626,10 @@ mod tests {
             test_signature(0x22),
             test_signature(0x33),
         ];
-        let tx = Transaction::from_signatures(message, signatures.clone()).unwrap();
+        let tx = Transaction {
+            message,
+            signatures: signatures.clone(),
+        };
 
         let bytes = tx.serialize().unwrap();
         let parsed = Transaction::from_bytes(&bytes).unwrap();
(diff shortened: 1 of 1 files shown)


packet/src/lib.rs
```diff
@@ -26,12 +26,15 @@ use {
...
-/// Maximum over-the-wire size of a Transaction
+/// Maximum over-the-wire size of a Transaction (legacy and v0)
 ///   1280 is IPv6 minimum MTU
 ///   40 bytes is the size of the IPv6 header
 ///   8 bytes is the size of the fragment header
 pub const PACKET_DATA_SIZE: usize = 1280 - 40 - 8;
 
+/// Maximum transaction size for V1 transactions (SIMD-0296/SIMD-0385)
+pub const PACKET_DATA_SIZE_V1: usize = 4096;
+
```
tao-stones, line 37: curious on your thoughts on picking which SIZE to use in the environment that supports both legacy/v0 and v1 (when feature gate is activated)? or it will be be one SIZE that is 4096 (kind like you did in `sdk-wasm-js`)?

message/src/versions/v1/config.rs
```diff
@@ -13,9 +13,9 @@ use std::mem::size_of;
 #[cfg_attr(feature = "frozen-abi", derive(AbiExample))]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
-pub struct TransactionConfigMask(pub u32);
+pub struct ComputeBudgetConfigMask(pub u32);
```
tao-stones: SIMD calls for `TransactionConfigMask`, it is to replace compute_budget for now, but has room left for adding future configurations. I'd suggest to keep original name, so no more `ComputeBudget` in V1 land 😄

transaction/src/v1.rs
```diff
@@ -0,0 +1,1104 @@
...
+    /// expected signers (first `num_required_signatures` accounts in the message)
+    /// by public key.
+    #[cfg(feature = "bincode")]
+    pub fn try_sign<T: Signers + ?Sized>(
+        message: Message,
+        keypairs: &T,
+    ) -> Result<Self, SignerError> {
+        let num_required_signatures = message.header.num_required_signatures as usize;
+
+        if message.account_keys.len() < num_required_signatures {
```
tao-stones: this probably belongs to message.sanitize(). try_sign() can either assume `message` is sanitized, or sanitize before signing it.

transaction/src/v1.rs
```diff
@@ -0,0 +1,1104 @@
...
+    /// Transaction signatures, one per required signer.
+    /// Order matches the first `num_required_signatures` accounts in the message.
+    pub signatures: Vec<Signature>,
+}
+
+impl Transaction {
+    /// Create a V1 transaction from a message and existing signatures.
+    ///
+    /// Returns an error if the signature count doesn't match `num_required_signatures`.
+    pub fn from_signatures(
```
tao-stones: try to think if there are use cases for this, is it for tests only?

transaction/src/v1.rs
```diff
@@ -0,0 +1,1104 @@
...
+
+    /// Parse a V1 transaction from wire format bytes.
+    ///
+    /// Expects format: `[message bytes][signatures]`
+    /// Returns an error if there is trailing data after the transaction.
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, TransactionError> {
+        let (tx, consumed) = Self::from_bytes_partial(bytes)?;
+        if consumed != bytes.len() {
+            return Err(TransactionError::TrailingData);
+        }
```
tao-stones: nit: catching `TrailingData` can be done within `from_bytes_partial()`, then it doesn't need to return "consumed bytes".

transaction/src/v1.rs
```diff
@@ -0,0 +1,1104 @@
...
+    /// Serialize the transaction to wire format per SIMD-0385.
+    ///
+    /// Wire format: `[message bytes][signatures]`
+    /// - No length prefix on signatures
+    /// - Signature count determined by `num_required_signatures` in message header
+    pub fn serialize(&self) -> Result<Vec<u8>, TransactionError> {
+        let mut out = self.message.to_bytes()?;
+        let signature_bytes = self
+            .signatures
+            .len()
```
tao-stones: nit: perhaps use `message.header.num_required_signatures` to be consistent?

message/src/versions/v1/config.rs
grod220: Good catch, will rename 👍

packet/src/lib.rs
grod220: It feels context dependent. We should try to not loses the ability to reject oversized legacy/v0 transactions. Basically we need to expose new v1 code paths everywhere (via new fns, enum variations, optional args, etc) or check the version byte first, then validate against the appropriate size limit.

In fact, `sdk-wasm-js` is not doing this correctly I'd say. Going to revert this crate back to master and save it for a future PR (feels independent enough).

transaction/src/v1.rs
grod220: Thinking offline signing/multisig workflows, but realizing the current struct has public fields. So this method is sort of optional right now. However, constructing manually gets around the validation check. Leaning toward making these fields private and forcing method construction. What do you think?

transaction/src/v1.rs
grod220: True, will sanitize here first

transaction/src/v1.rs
grod220: Hmm, I'm not sure if we have the usecases for from_bytes_partial() I originally thought. One simple from_bytes() may just be better.

transaction/src/v1.rs
tao-stones: It makes sense to keep fields private and do "raw" construction via this api. 

My train-of-thoughts is: `V1::Transaction` type is temporarily; it will, or can, be merged into `Transaction` - after all, they have same structure; but managing APIs might be bit messy. That's why I feel don't need to introduce `from_signature()`, but keep fields public as legacy/v0 Transactions currently are.

grod220: @febo subbing you in if that's ok! 🙏  

Totally open for whatever API design changes ya'll find are best.

febo: @tao-stones Made the fields public and removed `from_signatures`.

transaction/src/versioned/mod.rs
apfitzge, line 1: I think the actual `VersionedTransaction` needs to be modified. As is, the signatures would still be at the front of the serialized bytes, right?

transaction/src/v1.rs
```diff
@@ -0,0 +1,1085 @@
...
+}
+
+/// A V1 transaction per SIMD-0385.
+///
+/// Wire format: `[message bytes][signatures]`
+/// - Message bytes include version byte (0x81) through instruction payloads
+/// - Signatures are appended directly with NO length prefix
+/// - Signature count is determined by `num_required_signatures` from the message header
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Transaction {
```
tao-stones, line 83: what do you think about not having `V1::Transaction`? So from user perspective, using `bench-tps` as example, it only knows one transaction type that is `solana-transaction::Transaction`, which can be constructed from `V1::Message`, then goes through same `Transaction -> VersionTransaction -> SanitizedVersionTransaction` lifespan. In another words, we only introduce new Message v1, and Transaction (somewhat) automatically supports that. @febo

transaction/src/versioned/mod.rs
tao-stones: in current impl, `v1::Transaction::serialize()` [puts sigs after message](https://github.com/anza-xyz/solana-sdk/pull/531/changes#diff-a1f13155c4fcb094817d267382eb91d052b86f4a2bdb656d3157ecb9ee9c206fR166). But if we don't introduces `V1::Transaction` as new type, then (de)ser for v1 can happen elsewhere.

transaction/src/v1.rs
febo: Makes sense to me. We would need to make `Transaction` generic over `Message` right? Seems like `VersinedTransaction` and `SanitizedVersionedTransaction` would be fine since they use enum representations for the message.

apfitzge: `Transaction` is the wrong type I think. It only supports Legacy iirc.

`VersionedTransaction`, `SanitizedVersionedTransaction`, `SanitizedTransaction` are what we need to work if we want a drop in replacement in the validator (ignoring users for my perspective only).

transaction/src/versioned/mod.rs
apfitzge: right but for it to actually work in validator we need `VersionedTransaction`; or we've got to change over to some new type wrapping the different message types - which imo doesn't make much sense

transaction/src/v1.rs
febo: Those ones should support `V1` with the changes on this PR. But `Transaction` is being duplicated at the moment. Perhaps we don't need the duplication after all?

tao-stones: yea, remove (or "merge") duplicated "Transaction" is what I was thinking.

apfitzge: > Those ones should support V1 with the changes on this PR.

I think the `Message` types are correctly handled, but `VersionedTransaction` doesn't have a new variant afaict, so don't see how that could be working given the format difference in signature placement (leading vs trailing)

febo: <del>`VersionedTransaction` uses `VersionedMessage`, which has a variant for `V1`. So you can create a new one `VersionedTransaction::try_new` passing a `VersionedMessage::V1`.</del>

I see what you mean now. That has to change indeed.

febo: @tao-stones I am reworking the implementation to avoid so much duplication and address Andrew's comment. I will put a separate PR with it and then we can decide which approach works best.

tao-stones: thanks @febo

febo:
Closed in favour of #538

PR #534, opened 2026-01-20
SIMD-0433: Loader V3: Set Program Data to ELF Length

buffalojoec:
Docs-only change to document the new behavior for `Upgrade` as of [SIMD-0433](https://github.com/solana-foundation/solana-improvement-documents/pull/433).


buffalojoec:
@Lichtso you can review if you want, but marking draft until https://github.com/solana-foundation/solana-improvement-documents/pull/433 is merged.

PR #536, opened 2026-01-26
message: rename confusing param name in compiled_instruction

cryptopapi997:
The program_id() function wants to index the whole account array of the transaction it's part of, not just the program ids, yet the parameter is called `program_ids`. This is made extra misleading by the existence of https://github.com/anza-xyz/solana-sdk/blob/master/message/src/legacy.rs#L477-L482 which makes it seem like the "right" thing to do would be 

```
let prog_id = ix.program_id(tx.message.program_ids())
```

Not a huge deal, but made me have to double check the docs just now which could've been avoided with a better param name.

### DIFF
--- a/message/src/compiled_instruction.rs
+++ b/message/src/compiled_instruction.rs
@@ -55,7 +55,7 @@ impl CompiledInstruction {
         }
     }
 
-    pub fn program_id<'a>(&self, program_ids: &'a [Address]) -> &'a Address {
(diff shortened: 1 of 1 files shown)


joncinque (commented):
This is fine by me, but we need approval from @anza-xyz/tx-metadata

joncinque:
Merging on red, audit failure was fixed with #556

PR #538, opened 2026-01-27
Add support for V1 transactions

febo:
### Add V1 transaction format

Reference: [SIMD-0296](https://github.com/solana-foundation/solana-improvement-documents/pull/296) & [SIMD-0385](https://github.com/solana-foundation/solana-improvement-documents/pull/385)

V1 transactions increase the maximum transaction size from 1232 bytes to 4096 bytes and embed compute budget configuration directly in the message header, eliminating the need for separate ComputeBudget program instructions.

Changes:
* Add a new `versions/v1` module to message implementing V1 message format.
* Add a new `V1` variant to `VersionedMessage` enum.
* Implement wincode support for `VersionedTransaction` supporting V1 transaction format. 

Co-authored-by: @grod220


message/src/versions/v1/error.rs
```diff
@@ -0,0 +1,98 @@
...
+            Self::TooManyInstructions => write!(f, "too many instructions (max 64)"),
+            Self::TooManySignatures => write!(f, "too many signatures (max 12)"),
+            Self::TrailingData => write!(f, "unexpected trailing data"),
+            Self::TransactionTooLarge => write!(f, "transaction exceeds max size (4096 bytes)"),
+            Self::ZeroSigners => write!(f, "must have at least one signer (fee payer)"),
+        }
+    }
+}
+
+impl std::error::Error for MessageError {}
```
joncinque: nit: how about `core::error::Error` instead? This crate already has a minimum rust version of 1.81
```suggestion
impl core::error::Error for MessageError {}
```

message/src/versions/v1/error.rs
```diff
@@ -0,0 +1,98 @@
...
+    TooManySignatures,
+    /// Unexpected trailing data after message.
+    TrailingData,
+    /// Transaction exceeds maximum size (4096 bytes).
+    TransactionTooLarge,
+    /// Must have at least one signer (fee payer).
+    ZeroSigners,
+}
+
+impl std::fmt::Display for MessageError {
```
joncinque: nit: same here, core?
```suggestion
impl core::fmt::Display for MessageError {
```

message/src/sanitized.rs
```diff
@@ -78,7 +79,9 @@ pub enum SanitizedMessage {
     /// Sanitized legacy message
     Legacy(LegacyMessage<'static>),
     /// Sanitized version #0 message with dynamically loaded addresses
-    V0(v0::LoadedMessage<'static>),
+    V0(LoadedMessage<'static>),
+    /// Sanitized version #1 message (4KB transactions, no address lookup tables)
+    V1(LoadedMessage<'static>),
```
joncinque: This will be a question for the tx-metadata folks -- since Legacy wraps a `LegacyMessage` and V0 wraps a `LoadedMessage`, do we want a new type for V1 which doesn't include things like `loaded_addresses`?

message/src/versions/mod.rs
```diff
@@ -137,28 +150,47 @@ impl VersionedMessage {
...
+                let total = message.size();
+                let mut buffer: Vec<u8> = Vec::with_capacity(1 + total);
+                // SAFETY: buffer has sufficient capacity for serialization.
+                unsafe {
+                    let ptr = buffer.as_mut_ptr();
+                    ptr.write(MESSAGE_VERSION_PREFIX | 1);
+                    v1::serialize_into(message, ptr.add(1));
+                    buffer.set_len(1 + total);
+                }
+                buffer
```
joncinque: nit: we might want to also make this logic available in `v1::Message`, outside this function, in case someone wants to serialize a message without bincode. It also avoids the copied code with the wincode trait implementation

message/src/versions/mod.rs
```diff
@@ -173,7 +205,17 @@ impl VersionedMessage {
...
+
+        // The first byte on a versioned message indicates the version.
+        if let Some(&first_byte) = message_bytes.first() {
+            if first_byte & MESSAGE_VERSION_PREFIX != 0 {
+                let version = first_byte & !MESSAGE_VERSION_PREFIX;
+                hasher.update(format!("solana-tx-message-v{version}").as_bytes());
+            }
+            // TODO: Legacy messages (first byte < 0x80) get no domain prefix.
+            // This does not seem to be the current behaviour.
+        }
```
joncinque: This will change message hashes for all transactions, so would need to be properly feature-gated. Is this change required at all? I don't see it mentioned in the SIMDs

message/src/versions/mod.rs
```diff
@@ -204,6 +246,24 @@ impl serde::Serialize for VersionedMessage {
...
+                    seq.end()
+                } else {
+                    // Messages in V1 format cannot be binary serialized via bincode
+                    // because the wire format (per SIMD-0385) is incompatible with bincode's
+                    // data model. Use `VersionedMessage::serialize()` to get the correct
+                    // wire bytes.
+                    Err(serde::ser::Error::custom(
+                        "V1 messages cannot be serialized via bincode. Use VersionedMessage::serialize() for wire bytes.",
+                    ))
+                }
```
joncinque: This should have nothing to do with bincode, we just need to hand-write the implementation of `serde::Serialize` to put the fields in the correct order

message/src/versions/mod.rs
```diff
@@ -316,6 +378,20 @@ impl<'de> serde::Deserialize<'de> for VersionedMessage {
...
+                                        seq.next_element()?
+                                            .ok_or_else(|| de::Error::invalid_length(1, &self))?,
+                                    ))
+                                } else {
+                                    // V1 messages cannot be deserialized via bincode because the wire format (per SIMD-0385)
+                                    // is incompatible. Use `v1::Message::deserialize()` to parse wire-format bytes.
+                                    Err(de::Error::custom(
+                                        "V1 messages cannot be serialized via bincode. Use `v1::Message::deserialize()` to parse wire-format bytes.",
+                                    ))
+                                }
```
joncinque: Same with this, we should just need to hand-write the `Deserialize` implementation if the auto-generated one isn't correct

message/src/versions/mod.rs
```diff
@@ -384,6 +479,17 @@ impl<'de> SchemaRead<'de> for VersionedMessage {
                     dst.write(VersionedMessage::V0(msg));
                     Ok(())
                 }
+                1 => {
+                    // -1 for already-read variant byte
+                    let bytes = reader.fill_buf(v1::MAX_TRANSACTION_SIZE - 1)?;
+                    let (message, consumed) =
+                        deserialize(bytes).map_err(|_| invalid_tag_encoding(1))?;
+
+                    reader.consume(consumed)?;
```
joncinque: nit: can this use the unchecked variant since `deserialize` is assumed to be correct?

message/src/versions/mod.rs
```diff
@@ -507,4 +649,293 @@ mod tests {
         let message_from_string: VersionedMessage = serde_json::from_str(&string).unwrap();
         assert_eq!(message, message_from_string);
     }
+
+    #[test]
+    fn test_v1_message_raw_bytes_roundtrip() {
```
joncinque: nit: we could eventually make these into proptests, or use some randomization to make sure the roundtrip always works

message/src/versions/mod.rs
```diff
@@ -507,4 +649,293 @@ mod tests {
...
+            let deserialized: VersionedMessage =
+                wincode::deserialize(&bytes).expect("Wincode deserialize failed");
+
+            match deserialized {
+                VersionedMessage::V1(parsed) => {
+                    assert_eq!(parsed.header, message.header);
+                    assert_eq!(parsed.lifetime_specifier, message.lifetime_specifier);
+                    assert_eq!(parsed.account_keys, message.account_keys);
+                    assert_eq!(parsed.config, message.config);
+                    assert_eq!(parsed.instructions, message.instructions);
```
joncinque: nit: can we just do `assert_eq!(parsed, message)`?

message/src/versions/mod.rs
```diff
@@ -507,4 +649,293 @@ mod tests {
...
+        assert_ne!(v0_hash, legacy_hash);
+    }
+
+    #[cfg(feature = "blake3")]
+    #[test]
+    fn test_hash_domain_separation_prevents_collision() {
+        // Same payload bytes but different version prefixes must produce different hashes
+        let payload = &[0x01, 0x02, 0x03, 0x04, 0x05];
+
+        // Create messages with same payload but different versions
```
joncinque: Since the message already contains a version byte, the hashes should differ anyway, which makes me think the hash should be maintained as it is, but I'll let the experts weigh on that one

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1600 @@
...
+                .sum::<usize>() // instruction payloads
+    }
+}
+
+impl Sanitize for Message {
+    fn sanitize(&self) -> Result<(), SanitizeError> {
+        // Must have at least one signer (the fee payer)
+        if self.header.num_required_signatures == 0 {
+            return Err(SanitizeError::InvalidValue);
+        }
```
joncinque: nit: this check is handled by `if self.header.num_readonly_signed_accounts >= self.header.num_required_signatures` later, since `num_required_signatures` must be greater than 0 for that check to pass

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1600 @@
...
+
+        // num_required_signatures <= 12
+        if self.header.num_required_signatures > MAX_SIGNATURES {
+            return Err(SanitizeError::IndexOutOfBounds);
+        }
+
+        // Lifetime specifier must not be zero
+        if self.lifetime_specifier == Hash::default() {
+            return Err(SanitizeError::InvalidValue);
+        }
```
joncinque: Not sure if this one is needed either, since it's not on the others and not specified in the SIMD

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1600 @@
...
+    let lifetime_specifier = unsafe {
+        input_ptr = input_ptr.add(4);
+        Hash::new_from_array(*(input_ptr as *const [u8; 32]))
+    };
+
+    // counts
+    //
+    // SAFETY: input length has been checked against `FIXED_HEADER_SIZE`.
+    let num_instructions = unsafe {
+        input_ptr = input_ptr.add(32);
```
joncinque: nit: these adds are a little tough to read since they refer to the previous value that was just read -- can you move them up to after the value is read, as it's done later in the function?

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1600 @@
...
+
+        // Validate instruction constraints
+        for ix in &self.instructions {
+            if ix.accounts.len() > u8::MAX as usize {
+                return Err(MessageError::InstructionAccountsTooLarge);
+            }
+            if ix.data.len() > u16::MAX as usize {
+                return Err(MessageError::InstructionDataTooLarge);
+            }
+        }
```
joncinque: nit: there's a bit of duplication here with the sanitize logic, so it might be nice to combine those into one function, since it would be easy to add or change a check in one place and forget the other

message/src/versions/mod.rs
febo: There is public `serialize` helper: https://github.com/anza-xyz/solana-sdk/blob/a41d3e81fbf7a19a2d47cc87312a582559e82077/message/src/versions/v1/message.rs#L350-L362

It does not write the version, since the `Message` object is not "versioned".

message/src/versions/mod.rs
febo: Yeah, I was not sure about this part. It is not mentioned on the SIMD but it is on #531.

transaction/src/versioned/mod.rs
```diff
@@ -46,31 +45,80 @@ impl TransactionVersion {
...
-#[derive(Debug, PartialEq, Default, Eq, Clone)]
+#[derive(Debug, PartialEq, Eq, Clone)]
 pub struct VersionedTransaction {
-    /// List of signatures
-    #[cfg_attr(feature = "serde", serde(with = "short_vec"))]
-    #[cfg_attr(feature = "wincode", wincode(with = "containers::Vec<_, ShortU16Len>"))]
-    pub signatures: Vec<Signature>,
-    /// Message to sign.
-    pub message: VersionedMessage,
+    pub payload: TransactionPayload,
```
joncinque: Rather than introduce the new payload inner type, the other option would be to customize the `Serialize` and `Deserialize` implementations to change how the signatures are serialized.

~This~ (edit The current) approach is probably simpler at the cost of introducing extra boilerplate everywhere, but I wanted to bring it up to see how others felt.

message/src/versions/mod.rs
febo: I tried, but the format is not really compatible with the way bincode/serde works. There are "fields" that don't appear on the serialization based on values of other fields.

@apfitzge mentioned that bincode is not being used so maybe we should remove the bincode support altogether – I was not sure whether the json encoding is used or not.

transaction/src/versioned/mod.rs
febo: I am pretty sure this part is not working properly at the moment. 😅 It is not clear how you can determine the "version" of the transaction based only on bytes. I started another alternative where the fields don't change, but need to figure it out how to determine the version of the transaction based on the input bytes.

febo: Removed the `Payload` struct and implemented the wincode logic for (de-)serialization in https://github.com/anza-xyz/solana-sdk/pull/538/commits/f6e41a98e38d660e8982653c340aa6a80dd8704f.

message/src/sanitized.rs
apfitzge: Likely we will, there's no `loaded_addresses` when we have no ALTs

transaction/src/versioned/mod.rs
```diff
@@ -160,15 +162,16 @@ impl VersionedTransaction {
...
     }
 
     /// Returns a legacy transaction if the transaction message is legacy.
     pub fn into_legacy_transaction(self) -> Option<Transaction> {
         match self.message {
             VersionedMessage::Legacy(message) => Some(Transaction {
-                signatures: self.signatures,
-                message,
+                signatures: self.signatures.clone(),
+                message: message.clone(),
```
apfitzge: why must we clone now?

message/src/sanitized.rs
apfitzge: For clarity, I think what is here works. But would likely make it a harder transition to deprecate v0 txs than if we just had a separate message type. wdyt?

internally the `LoadedMessage` would just have empty vecs for `LoadedAddresses` which would have minimal perf impact - but is logically a bit confusing.

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1600 @@
...
+
+    /// Configuration for transaction parameters.
+    pub config: TransactionConfig,
+
+    /// The lifetime specifier (blockhash) that determines when this transaction expires.
+    pub lifetime_specifier: Hash,
+
+    /// All account addresses referenced by this message.
+    ///
+    /// The legnth should be specified as an `u8`. Unlike V0, V1 does not support
```
apfitzge: typo:
```suggestion
    /// The length should be specified as an `u8`. Unlike V0, V1 does not support
```

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1600 @@
...
+    ///   - `num_readonly_signed_accounts` addresses for which the transaction contains
+    ///     signatures and are loaded as readonly.
+    ///
+    ///   - `num_addresses-num_required_signatures-num_readonly_unsigned_accounts` addresses
+    ///     for which the transaction does not contain signatures and are loaded as writable.
+    ///
+    ///   - `num_readonly_unsigned_accounts` addresses for which the transaction does not
+    ///     contain signatures and are loaded as readonly.
+    #[cfg_attr(feature = "serde", serde(with = "solana_short_vec"))]
+    #[cfg_attr(feature = "wincode", wincode(with = "containers::Vec<_, ShortU16Len>"))]
```
apfitzge: we do not encode with short-u16 anywhere in the v1 transactions. We either use a u8 or u16 where appropriate.

transaction/src/versioned/mod.rs
febo: We don't. 😅

message/src/versions/v1/message.rs
apfitzge: this looks like it doesn't actually happen for serialization? I'm not sure how the annotations work with the manual implementation lol

febo: This is not used in the `serialize`/`deserialize` helpers, which is how the message is serialized in a transaction. It is here only because there is some logic to support json (de-)serialization using bincode in `VersionedMessage`. Can we drop support for bincode altogether?

message/src/versions/v1/message.rs
febo: Good call! Moved the logic to a `validate` [helper](https://github.com/febo/solana-sdk/blob/21859710f147648b100712fd7b93c58961fd1f3e/message/src/versions/v1/message.rs#L402).

message/src/sanitized.rs
febo: Would that be a type like `LoadedMessage` but without the `loaded_addresses` field?

transaction/src/sanitized.rs
```diff
@@ -201,13 +206,15 @@ impl SanitizedTransaction {
     pub fn to_versioned_transaction(&self) -> VersionedTransaction {
         let signatures = self.signatures.clone();
         match &self.message {
-            SanitizedMessage::V0(sanitized_msg) => VersionedTransaction {
-                signatures,
-                message: VersionedMessage::V0(v0::Message::clone(&sanitized_msg.message)),
-            },
+            SanitizedMessage::V0(sanitized_msg) | SanitizedMessage::V1(sanitized_msg) => {
+                VersionedTransaction {
+                    message: sanitized_msg.message.clone().into_owned(),
```
tao-stones: note to self: 
so `v1` message is folded into `v0`? what's motivation for that, why not keep them separated?

message/src/sanitized.rs
tao-stones: Second that, i lean to introduce a new message type for `v1` instead of sharing `LoadedMessage` with `v0`.

message/src/sanitized.rs
```diff
@@ -94,15 +97,20 @@ impl SanitizedMessage {
...
+                    sanitized_msg.message,
                     loaded_addresses,
                     reserved_account_keys,
                 ))
             }
+            VersionedMessage::V1(_) => SanitizedMessage::V1(LoadedMessage::new(
+                sanitized_msg.message,
+                LoadedAddresses::default(),
+                reserved_account_keys,
+            )),
```
tao-stones: In same subject of `LoadedMessage`,  `v1` doesn't have to be constructed from it, could be something like  `SanitizedMessage::V1(v1::Message::new(...))`. What would we lost if doing so?

message/src/sanitized.rs
```diff
@@ -131,7 +141,7 @@ impl SanitizedMessage {
     pub fn header(&self) -> &MessageHeader {
         match self {
             Self::Legacy(legacy_message) => &legacy_message.message.header,
-            Self::V0(loaded_msg) => &loaded_msg.message.header,
+            Self::V0(loaded_msg) | Self::V1(loaded_msg) => loaded_msg.message.header(),
```
tao-stones: not sure if I like combining `v0` and `v1` logic like this, would think separate them (even without some duplicated code) is cleaner, perhaps also make sunsetting `v0` easier?

message/src/versions/v0/loaded.rs
```diff
@@ -11,53 +9,17 @@ use {
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct LoadedMessage<'a> {
     /// Message which loaded a collection of lookup table addresses
-    pub message: Cow<'a, v0::Message>,
+    pub message: Cow<'a, VersionedMessage>,
```
tao-stones: Ah, maybe the reason `v0` and `v1` shares `LoadedMessage` because they are all `VersionedMessage`?

I think `LoadedAddress` is `v0` only thing, i'll rather not to expose `v1` with any thing `loaded address`.

message/src/versions/mod.rs
apfitzge: We should 100% not change hashing for existing transaction formats. I don't think we need to change this? I also don't see why we push this string into the hasher at all though.

message/src/sanitized.rs
apfitzge: Right, `LoadedMessage` seems like the wrong name, maybe just `v1::Message`? But yeah seems cleaner to me to have the be independent just like the `LegacyMessage` and `LoadedMessage`.

the v1 is actually more similar to the `LegacyMessage` 😄

febo: Added a `CachedMessage` for v1 in https://github.com/anza-xyz/solana-sdk/pull/538/commits/fd6a4645b51cb3d4689c8f42e8f609059c905726 based on `LegacyMessage`. We already have a v1 `Message` type.

message/src/versions/mod.rs
febo: Reverted in https://github.com/anza-xyz/solana-sdk/pull/538/commits/5c10eee805505e4357e3174f9dadfa49e8efa2a6.

message/src/versions/v1/cached.rs
```diff
@@ -0,0 +1,61 @@
...
+            .collect::<Vec<_>>();
+        Self {
+            message: Cow::Owned(message),
+            is_writable_account_cache,
+        }
+    }
+
+    /// Returns true if any account keys are duplicates
+    pub fn has_duplicates(&self) -> bool {
+        let mut uniq = HashSet::new();
```
apfitzge: minor - allocate with size of account_keys()

message/src/versions/v0/loaded.rs
apfitzge: I think now with `CachedMessage` this can/should go back to `v0::Message`

message/src/versions/v1/cached.rs
febo: Done in https://github.com/anza-xyz/solana-sdk/pull/538/commits/52a563c6d2ceeabade3baaa7a42463d60ee098a8 - also updated the similar code in `LoadedMessage`.

message/src/versions/v0/loaded.rs
febo: Good point – done in https://github.com/anza-xyz/solana-sdk/pull/538/commits/d69393518414f8852253d5b24099a4ea96d3afce.

message/src/sanitized.rs
febo: The logic is now separated.

message/src/sanitized.rs
febo: Now using the new `CachedMessage` type.

message/src/versions/v0/loaded.rs
```diff
@@ -109,7 +109,7 @@ impl<'a> LoadedMessage<'a> {
 
     /// Returns true if any account keys are duplicates
     pub fn has_duplicates(&self) -> bool {
-        let mut uniq = HashSet::new();
+        let mut uniq = HashSet::with_capacity(self.account_keys().len());
```
febo, line 112: We could move this change to a separate PR - it is not related to v1.

message/src/versions/v1/config.rs
```diff
@@ -0,0 +1,316 @@
...
+use wincode::{SchemaRead, SchemaWrite};
+
+/// Compute budget configuration for V1 transactions.
+#[cfg_attr(feature = "frozen-abi", derive(AbiExample))]
+#[cfg_attr(
+    feature = "serde",
+    derive(Serialize, Deserialize),
+    serde(rename_all = "camelCase")
+)]
+#[cfg_attr(feature = "wincode", derive(SchemaWrite, SchemaRead))]
```
febo: We don't serialize the `TransactionConfig` using the "default" derive, so technically we could remove these derives.

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1647 @@
...
+mod tests {
+    use super::*;
+
+    /// Builder for constructing V1 messages.
+    ///
+    /// This is used in tests to simplify message construction and validation. For
+    /// client code, users should construct messages using `try_compile` or
+    /// `try_compile_with_config`.
+    #[derive(Debug, Clone, Default)]
+    pub struct MessageBuilder {
```
febo, line 845: This type is not super useful as a public type since it would require passing `CompileInstruction`. Probably the `try_compile*` ones cover the construction of the message.

transaction/Cargo.toml
```diff
@@ -76,7 +76,7 @@ solana-nonce = { workspace = true }
 solana-packet = { workspace = true }
 solana-presigner = { workspace = true }
 solana-pubkey = { workspace = true, features = ["rand"] }
-solana-sha256-hasher = { workspace = true }
+solana-sha256-hasher = { workspace = true, features = ["sha2"] }
```
febo, line 79: Had to enable this for a test that uses the `hash`.

message/src/versions/mod.rs
febo: Done in https://github.com/anza-xyz/solana-sdk/pull/538/commits/38c39f70d15d003bf3db9c19b2169214b97ad571

message/src/versions/v0/loaded.rs
apfitzge: seems straight-forward enough to just pull out

message/src/versions/v1/config.rs
apfitzge: I'd lean toward removing them since it might be confusing to have them when we shouldn't be serializing/deserializing this directly.

message/src/versions/v1/config.rs
```diff
@@ -0,0 +1,316 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct TransactionConfig {
+    /// Priority fee in lamports.
+    pub priority_fee: Option<u64>,
+
+    /// Maximum compute units. None means use runtime default.
+    pub compute_unit_limit: Option<u32>,
+
+    /// Maximum bytes of account data that may be loaded. None means use runtime default.
+    pub loaded_accounts_data_size_limit: Option<u32>,
```
apfitzge: `None means use runtime default`.

In both these cases the default is 0, which is different than the current runtime defaults given for legacy/v0. The SIMD specified 0, so think it's safe to just do so here.

message/src/versions/v1/config.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// Bit 3 (4 bytes as u32 LE).
+    pub const LOADED_ACCOUNTS_DATA_SIZE: u32 = 0b1000;
+
+    /// Mask for requested heap size.
+    ///
+    /// Bit 4 (4 bytes as u32 LE).
+    pub const HEAP_SIZE: u32 = 0b10000;
+
+    /// Mask of all known/supported bits (bits 0-4).
+    pub const KNOWN_BITS: u32 = 0b11111;
```
apfitzge: nit: should we | all the known field masks together here instead of manually recreating the bits?

message/src/versions/v1/config.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    pub const fn new(mask: u32) -> Self {
+        Self(mask)
+    }
+
+    /// Returns true if any unknown bits are set (bits 5-31).
+    ///
+    /// Unknown bits indicate a config field the parser doesn't recognize.
+    /// Since config values are packed sequentially and the parser doesn't
+    /// know the byte size of unknown fields, it cannot skip them correctly.
```
apfitzge: I don't think this is true. The SIMD states that each bit corresponds to 4-bytes (this is why the prio fee takes 2 bits).

It's left unspecified in SIMD right now on how we should handle unspecified bits. I've opened a discussion on discord to figure out how we should handle this: https://discord.com/channels/428295358100013066/1040620771938156584/1468981916161278371

message/src/versions/v1/config.rs
```diff
@@ -0,0 +1,316 @@
...
+
+        if self.has_loaded_accounts_data_size() {
+            size = size.saturating_add(size_of::<u32>());
+        }
+
+        if self.has_heap_size() {
+            size = size.saturating_add(size_of::<u32>());
+        }
+
+        size
```
apfitzge, line 205: Each bit is just 4 bytes, so we can just count the bits and multiply by 4 (`size_of::<u32>()`):
```suggestion
    self.0.count_ones().saturating_mul(size_of::<u32>())
```

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1647 @@
...
+    ///
+    ///   - `num_readonly_signed_accounts` addresses for which the transaction contains
+    ///     signatures and are loaded as readonly.
+    ///
+    ///   - `num_addresses-num_required_signatures-num_readonly_unsigned_accounts` addresses
+    ///     for which the transaction does not contain signatures and are loaded as writable.
+    ///
+    ///   - `num_readonly_unsigned_accounts` addresses for which the transaction does not
+    ///     contain signatures and are loaded as readonly.
+    #[cfg_attr(feature = "serde", serde(with = "solana_short_vec"))]
```
apfitzge: I don't think there's a good reason to use solana_short_vec here; the actual serialization format just uses a u8 for length. If serde wants to use a u64 i think that's more consistent than doing the shortu16 encoding stuff.

Same thing for instructions

message/src/versions/v1/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+mod cached;
+mod config;
+mod error;
+mod message;
+
+use solana_hash::Hash;
+pub use {cached::*, config::*, error::*, message::*};
+
+/// Version byte for V1 messages (decimal 129).
+pub const V1_VERSION_BYTE: u8 = MESSAGE_VERSION_PREFIX | 1;
```
apfitzge: this isn't used anywhere?

transaction/src/versioned/mod.rs
```diff
@@ -225,14 +235,121 @@ impl VersionedTransaction {
...
+            let dst_ptr = dst.as_mut_ptr();
+
+            <containers::Vec<Signature, ShortU16Len> as SchemaRead<'de>>::read(reader, unsafe {
+                &mut *(addr_of_mut!((*dst_ptr).signatures)).cast::<MaybeUninit<Vec<Signature>>>()
+            })?;
+
+            <VersionedMessage as SchemaRead<'de>>::read(reader, unsafe {
+                &mut *(addr_of_mut!((*dst_ptr).message)).cast::<MaybeUninit<_>>()
+            })?;
+        } else {
```
apfitzge: probably need an else if here to make sure it's actually a v1

message/src/versions/mod.rs
```diff
@@ -49,41 +52,48 @@ pub const MESSAGE_VERSION_PREFIX: u8 = 0x80;
...
 impl VersionedMessage {
     pub fn sanitize(&self) -> Result<(), SanitizeError> {
         match self {
             Self::Legacy(message) => message.sanitize(),
             Self::V0(message) => message.sanitize(),
+            Self::V1(message) => message.sanitize(),
         }
     }
 
+    #[inline(always)]
```
apfitzge: is this and the other new inline annotations necessary now?

transaction/src/versioned/mod.rs
```diff
@@ -225,14 +235,121 @@ impl VersionedTransaction {
...
 
+#[cfg(feature = "wincode")]
+impl SchemaWrite for VersionedTransaction {
+    type Src = Self;
+
+    #[allow(clippy::arithmetic_side_effects)]
+    #[inline]
+    fn size_of(src: &Self::Src) -> WriteResult<usize> {
+        Ok(
+            <containers::Vec<Signature, ShortU16Len> as SchemaWrite>::size_of(&src.signatures)?
```
apfitzge: should we match on version here since we do not have a len prefix for the v1 array?

it will in any case only lead to 1 extra byte for the size_of call

transaction/src/versioned/mod.rs
```diff
@@ -225,14 +235,121 @@ impl VersionedTransaction {
...
+                &mut *(addr_of_mut!((*dst_ptr).signatures)).cast::<MaybeUninit<Vec<Signature>>>()
+            })?;
+
+            <VersionedMessage as SchemaRead<'de>>::read(reader, unsafe {
+                &mut *(addr_of_mut!((*dst_ptr).message)).cast::<MaybeUninit<_>>()
+            })?;
+        } else {
+            // V1 transaction
+
+            let message = VersionedMessage::get(reader)?;
```
apfitzge: clarifying for myself - we know the byte > 128 so it's NOT a legacy/v0 in this else. We do not know yet that it's a v1 transaction until the call to `VersionedMessage::get`. Is that correct?

transaction/src/versioned/mod.rs
```diff
@@ -392,6 +515,18 @@ mod tests {
             solana_signature::SIGNATURE_BYTES,
         };
 
+        // Bincode version of VersionedTransaction for cross-checking serialization
+        // with wincode.
```
apfitzge: let's specify that it is only for the legacy/v0 since v1 is not compatible with bincode?

message/src/versions/v1/message.rs
febo: How would this affect the json output? We will probably keep serde for that (#381).

message/src/versions/v1/mod.rs
febo: Maybe we should use it. 😊 There are a few places that we are using `MESSAGE_VERSION_PREFIX | 1`.

message/src/versions/mod.rs
febo: Probably not, I will remove them.

transaction/src/versioned/mod.rs
febo: Good catch!

message/src/versions/v1/message.rs
apfitzge: i think json would correctly support normal vecs instead of short ones, right?

transaction/src/versioned/mod.rs
febo: Yes, but we could find out if it is v1 by checking the `discriminator` value (`discriminator == 129` is v1). Maybe we should validate this and fail otherwise?

message/src/versions/v1/config.rs
febo: Beautiful!

message/src/versions/v1/config.rs
febo: Updated the comment. Also, I am tempted to change the implementation from:
```rust
(self.0 & !Self::KNOWN_BITS) != 0
```
to:
```rust
(self.0 | Self::KNOWN_BITS) != Self::KNOWN_BITS
```
I find it a little bit clearer since we avoid having the double negation. What do you think?

tao-stones: vouch for second one to avoid double negation.

message/src/versions/v1/config.rs
febo: Hmmm, so this leads to a difference in implementation. Because priority fee must use 2 bits, the previous implementation would return a size `0` for it if only one of the bits were set.

We could validate the mask value at creation, so that we know we always have a valid mask. Or keep the existing implementation? I have a preference to do the validation and then we can go with your suggestion. Thoughts?

transaction/src/versioned/mod.rs
febo: It is validating the version byte now.

message/src/versions/v1/message.rs
febo: Yeah, in the end the value on the json will be the same.

message/src/versions/v1/config.rs
tao-stones: i prefer existing implementation, it self-documents what it's doing; If a 3-bits config is introduced, all changes (validation by `has_...()` and taking size) will be in one place.

message/src/versions/mod.rs
```diff
@@ -49,41 +52,46 @@ pub const MESSAGE_VERSION_PREFIX: u8 = 0x80;
 /// format.
 #[cfg_attr(
     feature = "frozen-abi",
-    frozen_abi(digest = "Hndd1SDxQ5qNZvzHo77dpW6uD5c1DJNVjtg8tE6hc432"),
+    frozen_abi(digest = "6CoVPUxkUvDrAvAkfyVXwVDHCSf77aufm7DEZy5mBVeX"),
```
febo, line 55: This was needed since we remove the short-vec serde attribute.

message/src/versions/v1/cached.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl CachedMessage<'_> {
+    pub fn new(message: Message, reserved_account_keys: &HashSet<Address>) -> Self {
+        let is_writable_account_cache = message
+            .account_keys
+            .iter()
+            .enumerate()
+            .map(|(i, _key)| {
+                message.is_writable_index(i)
+                    && !reserved_account_keys.contains(&message.account_keys[i])
```
joncinque: I got worried that this could panic -- how about using `_key` directly rather than going back into the array?

message/src/versions/v1/config.rs
apfitzge: Keeping existing is fine!

message/src/versions/v0/loaded.rs
```diff
@@ -109,7 +109,7 @@ impl<'a> LoadedMessage<'a> {
 
     /// Returns true if any account keys are duplicates
     pub fn has_duplicates(&self) -> bool {
-        let mut uniq = HashSet::with_capacity(self.account_keys().len());
+        let mut uniq = HashSet::new();
```
apfitzge: hehe let's not revert the PR we just separated out!

message/src/versions/v1/config.rs
```diff
@@ -0,0 +1,313 @@
...
+
+impl TransactionConfig {
+    pub const fn new() -> Self {
+        Self {
+            priority_fee: None,
+            compute_unit_limit: None,
+            loaded_accounts_data_size_limit: None,
+            heap_size: None,
+        }
+    }
```
joncinque: nit: clippy would normally complain that this should be a `Default` implementation, but since traits can't have const functions, can we rename this to `empty` or `default`?

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1645 @@
...
+    }
+
+    /// Returns true if the BPF upgradeable loader is present in the account keys.
+    pub fn is_upgradeable_loader_present(&self) -> bool {
+        self.account_keys
+            .iter()
+            .any(|&key| key == bpf_loader_upgradeable::id())
+    }
+
+    /// Returns true if the account at the specified index was requested as writable.
```
apfitzge: This makes it seem the same as `is_writable_index`? We should add that this includes demotion


Honestly the naming of these 2 functions seems very odd to me.
`is_writable_index` is the one that is maybe not true.
`is_maybe_writable` is the one that actually includes demotion, and for non-v0 transactions will always return the truth.

message/src/versions/v1/cached.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+impl CachedMessage<'_> {
+    pub fn new(message: Message, reserved_account_keys: &HashSet<Address>) -> Self {
+        let is_writable_account_cache = message
+            .account_keys
+            .iter()
+            .enumerate()
+            .map(|(i, _key)| {
+                message.is_writable_index(i)
```
apfitzge, line 23: Do we not have a function that returns actual writability properly?

apfitzge: it seems like `is_maybe_writable` should (confusing as fuck, but consistent with previous namings)

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1645 @@
...
+                program_id_index: 1,
+                accounts: vec![0, 2],
+                data: vec![1, 2, 3],
+            })
+            .collect();
+        assert!(message.sanitize().is_ok());
+    }
+
+    #[test]
+    fn builder_requires_lifetime_specifier() {
```
apfitzge: we have a ton of tests for this test-only thing

message/src/versions/v0/loaded.rs
febo: Oh, what!? 😅

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1645 @@
...
+    ///
+    ///     Ok(tx)
+    /// }
+    /// #
+    /// # let client = RpcClient::new(String::new());
+    /// # let payer = Keypair::new();
+    /// # let instruction = Instruction::new_with_bincode(Address::new_unique(), &(), vec![
+    /// #   AccountMeta::new(Address::new_unique(), false),
+    /// # ]);
+    /// # Ok::<(), anyhow::Error>(())
```
joncinque, line 206: Should this example be calling `create_v1_tx` at some point?

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1645 @@
...
+    ///     )?;
+    ///
+    ///     Ok(tx)
+    /// }
+    /// #
+    /// # let client = RpcClient::new(String::new());
+    /// # let payer = Keypair::new();
+    /// # let instruction = Instruction::new_with_bincode(Address::new_unique(), &(), vec![
+    /// #   AccountMeta::new(Address::new_unique(), false),
+    /// # ]);
```
joncinque, line 292: Same with this one, should it actually create the transaction?

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1645 @@
...
+        // for the config.
+        let value = unsafe { u32::from_le_bytes(*(input_ptr as *const [u8; 4])) };
+        config = config.with_heap_size(value);
+        input_ptr = unsafe { input_ptr.add(size_of::<u32>()) };
+    }
+
+    // instruction headers
+
+    // Define instruction header type (program_id_index, num_accounts, data_len).
+    type InstructionHeader = (u8, u8, [u8; 2]);
```
apfitzge: we manually create this elsewhere for the size-calculation. maybe just define it somewhere above to re-use via size_of?

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1645 @@
...
+    /// # let instruction = Instruction::new_with_bincode(Address::new_unique(), &(), vec![
+    /// #   AccountMeta::new(Address::new_unique(), false),
+    /// # ]);
+    /// # Ok::<(), anyhow::Error>(())
+    /// ```
+    pub fn try_compile_with_config(
+        payer: &Address,
+        instructions: &[Instruction],
+        recent_blockhash: Hash,
+        config: &TransactionConfig,
```
joncinque: nit: how about taking this by value since it'll just be copied anyway?

apfitzge (commented):
only minor comments left

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1645 @@
...
+    pub fn is_upgradeable_loader_present(&self) -> bool {
+        self.account_keys
+            .iter()
+            .any(|&key| key == bpf_loader_upgradeable::id())
+    }
+
+    /// Returns true if the account at the specified index was requested as writable.
+    ///
+    /// The `reserved_account_keys` parameter allows demoting reserved accounts to readonly.
+    pub fn is_maybe_writable(
```
joncinque, line 373: nit: the `maybe` part in the function name came from v0 messages, because loaded addresses weren't included -- I think this can be more forceful and just be called `is_writable`

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1645 @@
...
+            }
+        }
+
+        // Demote program IDs, unless the upgradeable loader is present
+        // (upgradeable programs need to be writable for upgrades)
+        if self.is_key_called_as_program(key_index) && !self.is_upgradeable_loader_present() {
+            return false;
+        }
+
+        true
```
joncinque: nit: up to you, but this could just become `!self.demote_program_id(key_index)`

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1645 @@
...
+    }
+
+    pub fn demote_program_id(&self, i: usize) -> bool {
```
joncinque, line 387: nit: this function name isn't very clear. Since it's only checking, it could be called something like `is_program_id_write_demoted`

message/src/versions/v1/message.rs
febo: Well, it was not a test-only thing but I think it should be. I don't think clients will use a builder that expects `CompileInstruction`. So we could either move it back to be a public type or we can remove the tests?

message/src/versions/v1/message.rs
febo: Looked at the `v0` implementation. `is_writable_index` is meant to be private, so I changed to `pub(crate)` since there are some tests for it. Also updated the comments for `is_maybe_writable`. I did not change the name of these functions since they are the same as `v0`, but happy to do that if needed.

apfitzge: that's fine if it's not public! It would be super confusing if public.

Should we change `is_writable_index` to something more like `is_requested_writable_index` on all 3 variants if it's a private fn? (separate PR if we do)

message/src/versions/v1/message.rs
apfitzge: yeah think we could remove the tests, seems weird to be testing a test structure

message/src/versions/v1/message.rs
febo: Sounds good!

message/src/versions/v1/message.rs
febo: Done in https://github.com/anza-xyz/solana-sdk/pull/538/commits/acc1b50243c68bf72012796e4aef393752965a6a

message/src/versions/mod.rs
```diff
@@ -507,4 +572,192 @@ mod tests {
...
+                    priority_fee,
+                    compute_unit_limit,
+                    program_id_index,
+                    instr_accounts,
+                    data,
+                }
+            }
+        }
+
+        proptest!(|(test_data in generate_message_data())| {
```
joncinque: nit: This is typically done a bit differently -- the `propcompose!` is on its own, and the test is defined as:

```
proptest! {
    #[test]
    fn test_v1_message_raw_bytes_roundtrip(
        test_data in generate_message_data()
    ) {
    }
}
```

This is just following the example from the proptest book: https://proptest-rs.github.io/proptest/proptest/tutorial/macro-prop-compose.html

message/src/versions/mod.rs
```diff
@@ -507,4 +572,192 @@ mod tests {
...
+                )
+                (
+                    // The `program_id_index` cannot be 0 (payer).
+                    program_id_index in 1u8..accounts.len() as u8,
+                    // we need to have at least `required_signatures` accounts.
+                    instr_accounts in prop::collection::vec(
+                        0u8..accounts.len() as u8,
+                        (required_signatures as usize)..=accounts.len(),
+                    ),
+                    data in prop::collection::vec(any::<u8>(), 0..=2048),
```
joncinque: Is this limit to make sure it stays under 4kib total?

message/src/versions/mod.rs
```diff
@@ -507,4 +572,192 @@ mod tests {
...
+
+    #[test]
+    fn test_v1_message_raw_bytes_roundtrip() {
+        #[derive(Clone, Debug)]
+        struct TestMessageData {
+            required_signatures: u8,
+            lifetime: [u8; 32],
+            accounts: Vec<[u8; 32]>,
+            priority_fee: u64,
+            compute_unit_limit: u32,
```
joncinque: how about making these `Option`s? That way the mask is tested too

message/src/versions/mod.rs
```diff
@@ -507,4 +572,192 @@ mod tests {
...
+            program_id_index: u8,
+            instr_accounts: Vec<u8>,
+            data: Vec<u8>,
+        }
```
joncinque: How about adding the requested heap size and loaded data size as `Option<u32>`s?

message/src/versions/mod.rs
```diff
@@ -507,4 +572,192 @@ mod tests {
...
+
+        prop_compose! {
+            fn generate_message_data()
+                (
+                    // Generate between 12 and 64 accounts since we need at least the
+                    // amount of `required_signatures`.
+                    accounts in prop::collection::vec(any::<[u8; 32]>(), 12..=64),
```
joncinque: We could generate `accounts` based on a random number between `required_signatures` and 64 instead.

message/src/versions/mod.rs
```diff
@@ -507,4 +572,192 @@ mod tests {
...
+                    lifetime in any::<[u8; 32]>(),
+                    priority_fee in 0u64..=1_000_000u64,
```
joncinque: It probably shouldn't make a difference, but how about making this any u64?

transaction/src/versioned/mod.rs
```diff
@@ -304,17 +435,24 @@ mod tests {
...
     #[test]
     fn tx_uses_nonce_bad_prog_id_idx_fail() {
         let (_, _, mut tx) = nonced_transfer_tx();
-        match &mut tx.message {
-            VersionedMessage::Legacy(message) => {
+        match &mut tx {
+            VersionedTransaction {
+                message: VersionedMessage::Legacy(message),
+                ..
+            } => {
```
joncinque: nit: Is this change needed? If not, can we keep the old code?

transaction/src/versioned/mod.rs
```diff
@@ -485,26 +635,100 @@ mod tests {
...
+
+    #[test]
+    fn v1_transaction_at_max_size() {
+        // Calculate exact max data size for a transaction at the limit:
+        // - 1 signature
+        // - Fixed header (version + MessageHeader + config mask + lifetime + num_ix + num_addr)
+        // - 2 addresses
+        // - No config values (mask = 0)
+        // - 1 instruction header
+        // - 1 account index in instruction
```
joncinque: This test is great! Can you also add a case of adding one more byte and failing?

joncinque (commented):
Looking great overall! Mostly tiny things on my side.

(I hope this review goes through, I don't see a lot of my comments after there were more pushes to the branch 🙏 edit: it all went through!)

message/src/versions/v1/cached.rs
febo: Oh, that looks strange. Found out that it comes from here: https://github.com/anza-xyz/solana-sdk/blob/master/message/src/sanitized.rs#L40-L42

We can fix there as well in a separate PR.

message/src/versions/v1/config.rs
febo: Renamed to `empty` since the type already derives `Default`.

message/src/versions/v1/message.rs
febo: Yep, good catch!

message/src/versions/v1/message.rs
febo: Yep!

message/src/versions/v1/message.rs
febo: Makes sense.

message/src/versions/v1/message.rs
febo: "Legacy" `Message` and `VersionedMessage` also use the name `is_maybe_writable`. Do we want to change those names as well? Just asking to see whether do the single change here or we can change all of them in a separate PR.

message/src/versions/v1/message.rs
febo: This one also appear in all of the `Message` types. Should we change all of them in a separate PR or just this one?

message/src/versions/mod.rs
febo: Yes – added a comment to make it more explicit.

message/src/versions/mod.rs
febo: The issue that I found is that both appear on the same argument list, so they can't reference each other. And we can't move any of them to the second argument list.

transaction/src/versioned/mod.rs
febo: Nope, reverted the change.

message/src/versions/v1/message.rs
joncinque: I'd vote to change it here at the very least, since there's no more concept of `maybe`, but maybe @apfitzge has an opinion?

message/src/versions/v1/message.rs
joncinque: We can keep it for a separate PR

edit: but we should do it!

transaction/src/versioned/mod.rs
```diff
@@ -671,8 +669,10 @@ mod tests {
         });
     }
 
-    #[test]
-    fn v1_transaction_at_max_size() {
+    #[test_case(0 ; "at max size")]
+    #[test_case(1 ; "over by one")]
```
joncinque, line 673: I'm probably missing how this is used, but shouldn't this test-case fail? Or is the idea to have agave fail on transactions that are too big rather than failing at the sdk level?

message/src/versions/mod.rs
joncinque: It should be possible to do it using the `Just` terminology. Here's an example of generating a number and a number in between in the stake pool program: https://github.com/solana-program/stake-pool/blob/8b4ddc69f782b7d3274bc0955aba6cd9c906b40d/program/src/state.rs#L1248

But anyway, I don't want to block this PR on that, we can go forward

joncinque (commented):
Looks good to me! Just one last question, but nothing needs to block this going forward.

Future work from this PR:
* update `is_maybe_writable` -> `is_writable`
* update `demote_program_id` -> some better name

Is that it?

joncinque (dismissed):
Oops sorry, I'll also approve, in case my questions aren't valid 😄

transaction/src/versioned/mod.rs
febo: I had the same question to @apfitzge and @tao-stones just the other day since I did not see a size validation for either v0 or v1. It seems that this is done externally, like agave will validate the size.

febo: I had the same question to @apfitzge and @tao-stones just the other day since I did not see a size validation for either v0 or v1. It seems that this is done externally, like agave will validate the size.

apfitzge: Yeah I think it's fine to do this in agave, as we do for legacy/v0.

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1489 @@
...
+        } else {
+            false
+        }
+    }
+
+    /// Returns `true` if the account at the specified index was requested to be
+    /// writable.
+    ///
+    /// This method should not be used directly.
+    pub(crate) fn is_writable_index(&self, key_index: usize) -> bool {
```
apfitzge: The implementatin of this function should be exactly the same as for the `Legacy` transaction, but looks like you've re-implemented it. Seems like we could share this definition in a free function we just call (it is only different for v0 txs due to ALTs).

message/src/versions/v1/message.rs
```diff
@@ -0,0 +1,1489 @@
...
+    ///
+    /// Before loading addresses, we can't demote write locks properly so this should
+    /// not be used by the runtime. The `reserved_account_keys` parameter is optional
+    /// to allow clients to approximate writability without requiring fetching the latest
+    /// set of reserved account keys.
+    ///
+    /// Program accounts are demoted from writable to readonly, unless the upgradeable
+    /// loader is present in which case they are left as writable since upgradeable
+    /// programs need to be writable for upgrades.
+    pub fn is_maybe_writable(
```
apfitzge, line 373: similarly here - should be the same as the legacy message, let's try to share there with some common free function.

apfitzge (commented):
small comment on having shared impl for a couple functions - the indexing is tricky and it seems better if we can share the tried and true impl instead of writing it in a different way.

message/src/versions/mod.rs
febo: Hmmm, but they are in separate argument lists, right?
```rust
fn total_stake_and_rewards()(total_lamports in 1..u64::MAX)(
        total_lamports in Just(total_lamports),
        rewards in 0..=total_lamports,
) -> (u64, u64) {
        (total_lamports - rewards, rewards)
}
```
`total_lamports` is defined in the first one and then reused in the second.

The tricky thing that we have here is that both `accounts` and `required_signatures` need to be defined in the first one, since they are reused in the second.

message/src/versions/v1/message.rs
febo: Had to move a few more functions – done in https://github.com/anza-xyz/solana-sdk/pull/538/commits/ce6ae519cce00219c45c6263e5971656bc93148f.

message/src/versions/v1/message.rs
febo: Same, done in https://github.com/anza-xyz/solana-sdk/pull/538/commits/ce6ae519cce00219c45c6263e5971656bc93148f.

apfitzge: perfect!

apfitzge (approved):

joncinque (approved):

PR #539, opened 2026-01-27
commitment-level: add default implementation which satisfies clippy

fanatid:


program/src/lib.rs
```diff
@@ -675,8 +675,6 @@ pub use solana_example_mocks as example_mocks;
 
 #[cfg(test)]
 mod tests {
-    use super::unchecked_div_by_const;
-
```
joncinque, line 679: The change is fine, but did you mean to include it here?

fanatid: yes, i want to include it too, but did not put to pr description :/

PR #532, opened 2026-01-15
Add wincode support for more types

4r33x:
Implemented types:
    Account
    AccountSharedData
    AccountMeta
    Instruction
    InstructionError
    TransactionError
    
Added "wincode" feature to Pubkey to forward Address wincode impl

PR #542, opened 2026-01-29
transaction!: Replace bincode for wincode

febo:
#### Problem

With `bincode` not being maintained anymore and `wincode` having better performance, it is a good time to replace the use of `bincode` with `wincode`. Additionally, the upcoming transaction v1 format is not compatible with `bincode` serialization.

#### Solution

This PR removes the `bincode` feature of both `message` and `transaction` packages, replacing `bincode` by `wincode`.

### DIFF
--- a/scripts/build-sbf.sh
+++ b/scripts/build-sbf.sh
@@ -12,7 +12,6 @@ build_sbf_excludes=(
   --exclude solana-file-download
   --exclude solana-genesis-config
   --exclude solana-keypair
-  --exclude solana-message
   --exclude solana-offchain-message
   --exclude solana-presigner
   --exclude solana-sdk-wasm-js
--- a/sdk-wasm-js/Cargo.toml
+++ b/sdk-wasm-js/Cargo.toml
@@ -27,7 +27,7 @@ solana-message = { workspace = true }
 solana-packet = { workspace = true }
 solana-signature = { workspace = true }
 solana-signer = { workspace = true }
-solana-transaction = { workspace = true, features = ["wincode", "verify"] }
+solana-transaction = { workspace = true, features = ["verify", "wincode"] }
 wincode = { workspace = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]


fanatid:
is it possible to migrate to wincode all crates (PR #532)? or plan to migrate one crate after each other?

febo:
> is it possible to migrate to wincode all crates (PR #532)? or plan to migrate one crate after each other?

The plan is to eventually migrate all of them. I will have a look at #532.

transaction/src/lib.rs
```diff
@@ -180,6 +181,7 @@ const NONCED_TX_MARKER_IX_INDEX: u8 = 0;
     solana_frozen_abi_macro::frozen_abi(digest = "ADDDuk3dAZJ5hDxue8v4btH7nhEyngxUpXaC7A4k8gyQ")
 )]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
```
febo, line 183: We still need to derive serde for frozen-abi macro to work.

scripts/build-sbf.sh
```diff
@@ -12,6 +12,7 @@ build_sbf_excludes=(
   --exclude solana-file-download
   --exclude solana-genesis-config
   --exclude solana-keypair
+  --exclude solana-message
```
apfitzge: Can you explain this change?

sdk-wasm-js/Cargo.toml
```diff
@@ -28,7 +27,8 @@ solana-message = { workspace = true }
 solana-packet = { workspace = true }
 solana-signature = { workspace = true }
 solana-signer = { workspace = true }
-solana-transaction = { workspace = true, features = ["bincode", "verify"] }
+solana-transaction = { workspace = true, features = ["wincode", "verify"] }
```
apfitzge: should features be in alphabetic order? I saw you re-order elsewhere with these, so not sure which we prefer. but best to be consistent in the ordering!

transaction/src/versioned/mod.rs
apfitzge, line 53: out of scope for this PR. Does deriving the serde traits here pose a concern when we add support for txv1?

scripts/build-sbf.sh
febo: I reverted it – it was needed because of the MSRV of the message crate, but Jon bumped the platform-tools version so now it should be ok.

sdk-wasm-js/Cargo.toml
febo: Yeah, fixed it.

transaction/src/versioned/mod.rs
febo: As far as I can tell, `Transaction` only supports the legacy format so it won't interfere with txv1. But also the serde traits will be needed for the frozen-abi macro, which should be ok even with txv1. The only thing that does not work is to serialize txv1 using serde/bincode.

apfitzge: this is the `VersionedTransaction`; I saw you had made a comment on `Transaction` which does only support the legacy/original format.

Thanks for the explanation!

febo: Oh, right, but the comment still applies. 😅 We need it for frozen-abi.

PR #544, opened 2026-01-29
rust: Bump nightly compiler to 2026-01-22, Solana CLI to 3.1.8

joncinque:
#### Problem

The nightly compiler hasn't been bumped in a long time.

#### Summary of changes

To go with #543, bump the nightly compiler version. There are a few new lints to fix:

* the arithmetic side effects lint picks up some tests, so just use checked math
* the `struct_extensions` feature on wincode uses some `const` functions that aren't stable until 1.85.0, so bump MSRV on solana-message. Otherwise we get:

```
error: current MSRV (Minimum Supported Rust Version) is `1.81.0` but this item is stable in a `const` context since `1.83.0`
   --> message/src/lib.rs:119:25
    |
119 |     derive(SchemaWrite, SchemaRead),
    |                         ^^^^^^^^^^ in this derive macro expansion
```

### DIFF
--- a/frozen-abi/src/abi_digester.rs
+++ b/frozen-abi/src/abi_digester.rs
@@ -697,7 +697,7 @@ mod tests {
 
         #[serde_as]
         #[derive(serde_derive::Serialize, AbiExample)]
-        #[frozen_abi(digest = "CVqaXh4pWCiUyAuZ6dZPCmbCEtJyNH3e6uwUpJzymT6b")]
+        #[frozen_abi(digest = "8q6BU8QUrApZkpcAanNv7DppqddY9tjDBhJWpGDmYmZu")]
         struct U8ArrayWithGenericAs {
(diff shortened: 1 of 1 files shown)


joncinque:
Sorry, gonna move this to draft, need a bit more work

joncinque:
Gonna wait on #545 before landing this

joncinque:
The frozen-abi digest updates are expected due to some new lifetimes that become explicit. A type that used to be:

```
updating with:             variant(1) Some(serde_with::ser::SerializeAsWrap<[u8; 48], [serde_with::Same; 48]>) (newtype)
```

Becomes:

```
updating with:             variant(1) Some(serde_with::ser::SerializeAsWrap<'_, [u8; 48], [serde_with::Same; 48]>) (newtype)
```

So once the mac job decides to kick off, this should be good!

PR #545, opened 2026-01-29
example-mocks!: Remove dependencies

joncinque:
#### Problem

The example-mocks crate has loads of circular dependencies, which is partly by design, but it has some pretty high-level dependencies, such as solana-message, which is causing issues for bumping rust-version in that crate.

#### Summary of changes

A lot of overdue cleanup on the example mocks, which includes:

* removing dependencies and modules that no longer exist in solana-sdk
* remove `VersionedTransaction` and inline it in solana-message
* use broken-up crates in solana-system-interface's doctests
* remove solana-message dependency by making the mock transaction use more generic parameters on the message

Unblocks the nightly rust upgrade and wincode bumps

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4004,9 +4004,13 @@ dependencies = [
  "k256",
  "rand 0.8.5",
  "solana-define-syscall",
+ "solana-example-mocks",
  "solana-frozen-abi",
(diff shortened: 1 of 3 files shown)


joncinque:
Ok, sorry for all the additional pushes, this should be good now

PR #547, opened 2026-01-30
feature-gate / vote: Deprecate functions that take in `Rent`

joncinque:
#### Problem

The `Rent` sysvar type changed, which means that any downstream users of the type must also get a major version bump. However, not all of these usages of the `Rent` type are really needed.

#### Summary of changes

Deprecate functions in `feature-gate-interface` and `vote-interface` that take in `Rent` directly. None of these functions are used in agave currently.

Since solana-rent v4 was already published, we'll need to backport these changes to v3, publish them, then make the breaking change in these crates.

### DIFF
--- a/feature-gate-interface/src/instruction.rs
+++ b/feature-gate-interface/src/instruction.rs
@@ -11,7 +11,7 @@ use {
 /// Activate a feature
 #[cfg(feature = "bincode")]
 #[deprecated(
-    since = "3.0.1",
+    since = "3.1.0",
     note = "Use `activate_with_lamports` with `rent.minimum_balance(Feature::size_of())` instead"
 )]
 pub fn activate(feature_id: &Pubkey, funding_address: &Pubkey, rent: &Rent) -> Vec<Instruction> {
--- a/vote-interface/src/state/vote_state_1_14_11.rs
+++ b/vote-interface/src/state/vote_state_1_14_11.rs
@@ -47,7 +47,7 @@ pub struct VoteState1_14_11 {
 
 impl VoteState1_14_11 {
     #[deprecated(
-        since = "5.0.1",
+        since = "5.1.0",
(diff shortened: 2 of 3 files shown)


feature-gate-interface/src/instruction.rs
```diff
@@ -8,8 +8,12 @@ use {
     solana_system_interface::instruction as system_instruction,
 };
 
-#[cfg(feature = "bincode")]
 /// Activate a feature
+#[cfg(feature = "bincode")]
+#[deprecated(
+    since = "3.0.1",
```
febo: I think this needs to be a minor bump according to semver.

_Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backward compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated._ 

I don't feel stronger about it, so happy either way.

vote-interface/src/state/vote_state_1_14_11.rs
```diff
@@ -46,6 +46,10 @@ pub struct VoteState1_14_11 {
 }
 
 impl VoteState1_14_11 {
+    #[deprecated(
+        since = "5.0.1",
```
febo: Same here about the minor/patch bump.

vote-interface/src/state/vote_state_v3.rs
```diff
@@ -88,6 +88,10 @@ impl VoteStateV3 {
         }
     }
 
+    #[deprecated(
+        since = "5.0.1",
```
febo: And here, minor/patch bump.

feature-gate-interface/src/instruction.rs
joncinque: Oh gotcha, I didn't realize that, thanks!

febo (dismissed):
Looks good – just left an optional comment about the version bump.

joncinque:
Updated the version numbers, thanks for the information!

PR #540, opened 2026-01-27
account: impl ReadableAccount for Deref<Target = impl ReadableAccount>

fanatid:
Need `impl ReadableAccount for &Account` for https://github.com/jup-ag/jupiter-amm-interface/ 🥲

PR #548, opened 2026-01-30
v3: account: impl ReadableAccount for Deref<Target = impl ReadableAccount> (#540)

joncinque:
This is a manual backport of #540 because the cherry-pick failed.

solana-account currently contains unpublished breaking changes, and v4 would require changes to a few downstream crates. To get the change out more quickly, we can just backport the PR and publish a new v3.

Note that this removes an optimized implementation of `to_account_shared_data` for `Ref<AccountSharedData>`, but that won't be a problem because the removal of `to_account_shared_data` is already part of v3.1 with https://github.com/anza-xyz/agave/pull/8668


joncinque:
Actually, I'm just realizing that the version of solana-account in the v3 branch is a bit out of date, so we'll do this in a different order, by reverting the breaking changes on master, publishing, then reapplying

PR #550, opened 2026-01-30
deps: Bump short-vec to v3.2.0 in repo

joncinque:
#### Problem

solana-short-vec v3.2.0 added the `serde` feature, which many other SDK crates depend on. However, the repo depends on the minimum version of solana-short-vec, which doesn't have the feature.

#### Summary of changes

Bump the version in the repo to v3.2.0


febo (approved):
Looks good!

PR #553, opened 2026-02-03
feat(nonce): support wincode serialization for Versions

kskalski:
`verify_nonce_account` comes up in profiles.

Changes:
* Add `SchemaRead` and `SchemaWrite` implementations through automatic derives for `solana_nonce::versions::Versions` and nested types.
* if `wincode` feature on `solana-nonce-account` is enabled, deserialize `Versions` in `verify_nonce_account` with wincode

### DIFF
--- a/nonce/src/state.rs
+++ b/nonce/src/state.rs
@@ -9,6 +9,7 @@ use {
 
 const DURABLE_NONCE_HASH_PREFIX: &[u8] = "DURABLE_NONCE".as_bytes();
 
+#[repr(transparent)]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[cfg_attr(feature = "wincode", derive(SchemaWrite, SchemaRead))]
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
@@ -17,6 +18,7 @@ pub struct DurableNonce(Hash);
 /// Initialized data of a durable transaction nonce account.
 ///
 /// This is stored within [`State`] for initialized nonce accounts.
+#[repr(C)]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 #[cfg_attr(feature = "wincode", derive(SchemaWrite, SchemaRead))]
 #[derive(Debug, Default, PartialEq, Eq, Clone)]


nonce/src/state.rs
```diff
@@ -8,13 +10,15 @@ use {
 const DURABLE_NONCE_HASH_PREFIX: &[u8] = "DURABLE_NONCE".as_bytes();
 
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+#[cfg_attr(feature = "wincode", derive(SchemaWrite, SchemaRead))]
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
 pub struct DurableNonce(Hash);
```
kskalski, line 16: Is it ok if I put `#[repr(C)]` on this struct and `Data` + `FeeCalculator`?
This will allow wincode to derive zero-copy marker for `Data` when compiled on little endian, at least with the wincode 3.x.

nonce/src/versions.rs
```diff
@@ -11,6 +13,7 @@ use {
     feature = "serde",
     derive(serde_derive::Deserialize, serde_derive::Serialize)
 )]
+#[cfg_attr(feature = "wincode", derive(SchemaWrite, SchemaRead))]
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum Versions {
     Legacy(Box<State>),
```
kskalski, line 19: I wonder what is the minimum extra boilerplate that would allow us to deserialize `Versions` with borrowing, no-allocations and zero-copy. I'm thinking of:

```rust
pub enum VersionsRef<'a> {
  Legacy(StateRef<'a>),
  Current(StateRef<'a>),
}
pub enum StateRef<'a> {
    #[default]
    Uninitialized,
    Initialized(&'a Data),
}
```
+ with `repr(C)` and little-endian target `Data` should be zero-copy, then deserializing `VersionsRef` would basically mean reading two enum discriminants and storing a ref to the input account's `Vec<u8>` buffer.

Is there a better way?

nonce/src/state.rs
joncinque: We typically use `repr(transparent)` for these kinds of types.

As for `Data`, it looks like it's only being serialized with `bincode` in Agave, which means it can break ABI with `repr(C)`.

`FeeCalculator` already has `repr(C)`, so we should be good there

nonce/src/state.rs
```diff
@@ -106,4 +111,16 @@ mod test {
         let size = bincode::serialized_size(&data).unwrap();
         assert_eq!(State::size() as u64, size);
     }
+
+    #[test]
+    #[cfg(feature = "wincode")]
```
joncinque, line 118: Since this test gated on the `wincode` feature, can you add `solana-nonce` as a dev-dependency with the `wincode` feature enabled? That way we're sure the test is actually running. We do this for some other crates, ie https://github.com/anza-xyz/solana-sdk/blob/185c1c2ff4805d5fcf4f6728ad5690de57c7ff3c/account/Cargo.toml#L48

nonce/src/versions.rs
joncinque: I'm not totally sure why this has a `Box` at all, so you can definitely add a new type!

however, wincode doesn't seem to allow for enums to derive `ZeroCopy` by default: https://github.com/anza-xyz/wincode/blob/b8e4d148047b338b858f22aafaeb8f10262631e5/wincode-derive/src/assert_zero_copy.rs#L14, so we've had to implement those manually, like at https://github.com/anza-xyz/solana-sdk/blob/185c1c2ff4805d5fcf4f6728ad5690de57c7ff3c/message/src/versions/mod.rs#L341

I'm not sure of the canonical way to implement this, because `State` is either 1 byte for `Uninitialized` or 73 for `Initialized`. Constructing an enum with ref variants, as you've laid out, makes the most sense to me. But since the variants are the same, we could probably simplify VersionsRef into a struct like:

```
struct VersionsRef<'a> {
    version: u8, // Legacy or Current
    state: StateRef<'a>,
}
```

Either way, I'll let @cpubot weigh in for that part and learn from him 😁

joncinque (commented):
Looks good to me overall, feel free to go forward with the `repr` changes, and please get some feedback from @cpubot for how best to do the enums and avoid allocations

nonce/src/state.rs
cpubot: Agree, `#[repr(transparent)]` would be sufficient here for enabling zero-copy

nonce/src/versions.rs
cpubot: A zero-copy type in wincode means you can deserialize a reference to that type. So, enums _themselves_ are never zero-copy, but they can contain references to zero-copy types. Deserializing enum discriminant values will proceed as normal, but the rest of the zero-copy machinery can take over for producing a reference to variant data.

@kskalski's proposed design with `VersionsRef` and `StateRef` look canonical to me as far as leveraging zero-copy machinery with enums. If we want to store the `version` as a byte, we could accomplish that with `#[wincode(tag_encoding = "u8")]` on the enum, which can avoid any manual branching on that byte later on

kskalski: Ok, this is a potential future improvement, since it depends on:
* switching solana-sdk to 3.x
* actually there is a bug in wincode that prevents this pattern https://github.com/anza-xyz/wincode/issues/145

nonce/src/state.rs
kskalski: Would breaking ABI for `Data` be in an issue for agave? 
It would need to send transmuted bytes over the wire / persistence for this to be a problem, right?
`bincode` serialization format won't be affected, since it's based on source code field order.

nonce/src/state.rs
kskalski: done

nonce/src/state.rs
joncinque: From what I see, it shouldn't be an issue, since it's always serialized through bincode, and not stored in other formats

kskalski: Ok, I added the `repr(transpart)` / `repr(C)` annotations as necessary, so I think this PR is good to go (also verified that all tests pass merged with upstream wincode changes)

joncinque (approved):
Looks great to me!

PR #555, opened 2026-02-03
repo: Update solana-short-vec to properly publish

joncinque:
#### Problem

It isn't possible to publish crates because there are two versions of solana-short-vec in the lockfile.

#### Summary of changes

Use the same version from the repo and from crates.io. We could also add a [patch.crates-io] in the top-level Cargo.toml, but then cargo-semver-checks will fail.


febo:
I will merge this so we can publish the message and transaction crates.

PR #557, opened 2026-02-03
CI: Don't document dependencies

joncinque:
#### Problem

The docs check in CI is quite slow, probably because we're also documenting dependencies.

#### Summary of changes

Don't document dependencies using `--no-deps`


joncinque:
Beautiful, the job goes down to 7 minutes from 54 minutes

PR #559, opened 2026-02-04
Add wincode support for Transaction

apfitzge:
### Problem
- `Transaction` unlike `VersionedTransaction` does not support wincode (de)serialization.
- This means I cannot share code for deserialization of the two types (solana-rpc currently shares deser code using bincode)

### Changes
- Add wincode support for `Transaction`
  - use container attributes consistent with `VersionedTransaction`


apfitzge:
clsoing in favor of #542

PR #562, opened 2026-02-07
instruction-view!: Use`AsRef<AccountView>` in invoke helpers

febo:
### Problem

Currently, `invoke*` expect an array or slice of `&AccountView`. This is because in general program entrypoints consume references from the `accounts` slice.
```rust
pub fn process_instruction(
   program_id: &Address,
   accounts: &[AccountView],
   instruction_data: &[u8],
) -> ProgramResult {
   // - both `from` and `to` are `&AccountView` references.
   let [from, to, ..] = accounts else {
      return Err(ProgramError::NotEnoughAccountKeys);
   };
   ...
}
```
But there are scenarios where a program might end up with a `&[AccountView]`, in addition to the original `accounts` slice.
```rust
pub fn process_instruction(
   program_id: &Address,
   accounts: &[AccountView],
   instruction_data: &[u8],
) -> ProgramResult {
   // - both `from` and `to` are `&AccountView` references.
   // - `remaining` is a `&[AccountView]` slice.
   let [from, to, remaining @ ..] = accounts else {
      return Err(ProgramError::NotEnoughAccountKeys);
   };
   ...
}
```
In this case, the `remaining` slice cannot be passed directly to `invoke*` to perform a cross-program invocation since these helpers expect a `&[&AccountView]`.

### Solution

This PR adds an `impl AsRef<AccountView> for AccountView` and updates the expected argument for `invoke*` to `&[A: AsRef<AccountView>]`. With these changes, `invoke*` can be called with either `&[&AccountView]` or `&[AccountView]`.

### DIFF
--- a/instruction-view/src/cpi.rs
+++ b/instruction-view/src/cpi.rs
@@ -329,7 +329,7 @@ pub fn invoke_signed<const ACCOUNTS: usize, A: AsRef<AccountView>>(
         );
     }
 
-    let mut accounts = [MaybeUninit::<CpiAccount>::uninit(); ACCOUNTS];
+    let mut accounts = [const { MaybeUninit::<CpiAccount>::uninit() }; ACCOUNTS];
 
     // SAFETY: The array of `AccountView`s will be checked to ensure that it has
     // the same number of accounts as the instruction – this indirectly validates
@@ -403,7 +403,7 @@ pub fn invoke_signed_with_bounds<const MAX_ACCOUNTS: usize, A: AsRef<AccountView
         return Err(ProgramError::InvalidArgument);
     }
 
-    let mut accounts = [MaybeUninit::<CpiAccount>::uninit(); MAX_ACCOUNTS];
+    let mut accounts = [const { MaybeUninit::<CpiAccount>::uninit() }; MAX_ACCOUNTS];
 
     // SAFETY: The stack allocated account storage `MAX_ACCOUNTS` was validated
     // to be sufficient for the number of accounts expected by the instruction.


LStan:
This is an interesting idea, but I wonder if it can be done without enabling the `copy` feature for `solana-account-view`. At least, it can be enabled by a downstream project if necessary.

febo:
> This is an interesting idea, but I wonder if it can be done without enabling the `copy` feature for `solana-account-view`. At least, it can be enabled by a downstream project if necessary.

The requirement is for this part:
```rust
let mut accounts = [MaybeUninit::<CpiAccount<A>>::uninit(); ACCOUNTS];
```
At this point, `A` must be copy. References are copy by default, so all good; but since `A` can also be `AccountInfo`, then the type needs to be `Copy`. We could allow downstream users to enabled, but not sure whether that would be unexpected or not.

instruction-view/Cargo.toml
```diff
@@ -16,7 +16,11 @@ all-features = true
 rustdoc-args = ["--cfg=docsrs"]
 
 [features]
-cpi = ["dep:solana-define-syscall", "dep:solana-program-error"]
+cpi = [
+    "dep:solana-define-syscall",
+    "dep:solana-program-error",
+    "solana-account-view/copy",
```
febo: It was necessary to enable copy on `AccountView` for the `MaybeUninit` use:
```rust
let mut accounts = [MaybeUninit::<CpiAccount<A>>::uninit(); MAX_ACCOUNTS];
```
Another alternative is to document the requirement and downstream users need to enable the feature when using `&[AccountInfo]` slices.

update: Having `Copy` on `AccountView` is not too bad. The type has the same size when passed as a reference or as copy; also, most of the time a program will have references to `AccountView`.

LStan: I can't come up with another use case except remaining_accounts, which is a rare case. So, enabling `copy` feature for everyone who does cpi (and also adding a possibility of a sudden unnecessary copy in their code) feels a bit overkill.
Enabling `copy` also adds `Copy` to `RuntimeAccount`, but probably it is not usually used.

instruction-view/src/cpi.rs
```diff
@@ -65,12 +65,13 @@ pub struct CpiAccount<'a> {
     /// `&'a AccountView` lives. Instead of holding a reference to the actual `AccountView`,
     /// which would increase the size of the type, we claim to hold a reference without
     /// actually holding one using a `PhantomData<&'a AccountView>`.
-    _account_view: PhantomData<&'a AccountView>,
+    _account_view: PhantomData<A>,
```
LStan: The comment above should probably be fixed.

febo: Reverted the change on the type, so the comment don't need update now.

instruction-view/Cargo.toml
febo: Actually, we don't need to change `CpiAccount` to be generic, which means we don't need to enable `"copy"` on the `AccountView` type.

instruction-view/src/cpi.rs
```diff
@@ -329,15 +329,19 @@ pub fn invoke_signed<const ACCOUNTS: usize>(
         );
     }
 
-    const UNINIT: MaybeUninit<CpiAccount> = MaybeUninit::<CpiAccount>::uninit();
-    let mut accounts = [UNINIT; ACCOUNTS];
+    let mut accounts = [MaybeUninit::<CpiAccount>::uninit(); ACCOUNTS];
```
joncinque: Does this change incur any cost? I always thought we define the `const` separately so that the array is also created at compile-time, but now that I write this out, it makes no sense since it's an array of `MaybeUninit`, which are explicitly _not_ initialized 😅

febo: Oh, good point, this change is not required anymore since `CpiAccount` is being made generic. If I recall correctly, there was something related the Rust version, but I can't find exactly what it was.

We could revert the change or use `const { MaybeUninit::<CpiAccount>::uninit() }` or leave it as it is. 😊 Any preferences?

joncinque: I'd say leave it as it is here, unless there's a performance penalty

febo: I am tempted to use `const { ... }` notation to be explicit that the expression is evaluated at compile time. It should not change the generated code.

PR #563, opened 2026-02-07
Upgrade to wincode 0.4.1

cpubot:
[wincode `0.3.0`](https://github.com/anza-xyz/wincode/releases/tag/wincode%40v0.3.0):
- adds a new `Config` type parameter to `SchemaRead` and `SchemaWrite`
- marks `SchemaRead` and `SchemaWrite` as `unsafe` traits
- provides built-in support for customizable length encodings

[wincode `0.4.0`](https://github.com/anza-xyz/wincode/releases/tag/wincode%40v0.4.0):
- enables deriving `SchemaRead` and `SchemaWrite` on types with with generic type parameters 
- adjusts `SchemaRead` and `SchemaWrite` trait definitions to use `impl Reader` and `impl Writer` rather than `&mut impl Reader` and `&mut impl Writer`, respectively

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -203,7 +203,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "62945a2f7e6de02a31fe400aa489f0e0f5b2502e69f95f853adb82a96c7a6b60"
 dependencies = [
  "quote",
- "syn 2.0.111",
+ "syn 2.0.114",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


apfitzge:
lgtm from tx-metadata. @febo I think this will conflict with #538.

Waiting for @joncinque to give approval before I press appove, only question I have is on the other cargo updates - do we usually update in a group like this or separately (do we care?)

febo:
> lgtm from tx-metadata. @febo I think this will conflict with #538.

Yes, indeed.

joncinque (dismissed):
Looks good to me! Just so I'm sure I understand though, this will be a breaking change across all types that implement the wincode traits, correct?

febo (approved):
Looks great!

PR #564, opened 2026-02-09
v2: clock: Restore consts that are cfg-ed out

joncinque:
#### Problem

#173 inadvertently removed a bunch of consts when it just wanted to deprecate them, which created breaking changes in v2.2.2 of clock.

#### Summary of changes

Restore the consts by removing the `#[cfg(test)]` annotation.

cc @cavemanloverboy thanks for noticing!


joncinque:
Merging on red, failures are unrelated, due to this being a pretty old branch

PR #565, opened 2026-02-10
transaction: Advance reader after signatures

febo:
### Problem

When reading a transaction v1, the reader is not being advanced after the signatures.

### Solution

Call `consume` on the reader with the number of bytes read for the signatures.


febo:
> Ah darn, nice catch

Credits to @cpubot – he found it out during his rebase.

apfitzge (approved):

PR #566, opened 2026-02-10
transaction: Add message validation

febo:
### Problem

There are now two different transaction formats with the addition of v1: one that support legacy and v0 messages and another v1 messages. But it is still possible to mix the message format with a different transaction format.

### Solution

Add a validation after reading the message to check it matches the expected format based on the transaction format.

### DIFF
--- a/transaction/src/versioned/mod.rs
+++ b/transaction/src/versioned/mod.rs
@@ -762,4 +762,41 @@ mod tests {
             "Deserialized payload should match original"
         );
     }
+
+    #[test]
+    fn test_v1_message_in_legacy_transaction() {
+        #[rustfmt::skip]
+        let malformed_input: &[u8] = &[
+            0x00,                   // 0 signatures via ShortU16 -> takes Legacy/V0 path
+            0x81,                   // V1 message prefix
+            // V1 LegacyHeader (3 bytes)
+            0x01,                   // num_required_signatures = 1
+            0x00,                   // num_readonly_signed_accounts = 0
+            0x00,                   // num_readonly_unsigned_accounts = 0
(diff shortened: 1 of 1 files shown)


joncinque (commented):
Ah very clever! This code bothered me a bit since it was always deserializing `VersionedMessage`, even though we've already checked the version.

We can add these defensive checks since it's an easy fix, but it would be more robust to actually deserialize the expected message variant in each branch rather than `VersionedMessage`. Meaning, if `discriminator == V1_PREFIX`, then we do `MessageV1::get`, which checks the discriminator.

What do you think?

febo:
> Ah very clever! This code bothered me a bit since it was always deserializing `VersionedMessage`, even though we've already checked the version.
> 
> We can add these defensive checks since it's an easy fix, but it would be more robust to actually deserialize the expected message variant in each branch rather than `VersionedMessage`. Meaning, if `discriminator == V1_PREFIX`, then we do `MessageV1::get`, which checks the discriminator.
> 
> What do you think?

We could, the only caveat is that legacy don't have a "version" prefix so we will need to replicate the logic [here](https://github.com/anza-xyz/solana-sdk/blob/master/message/src/versions/mod.rs#L425).

joncinque:
> We could, the only caveat is that legacy don't have a "version" prefix so we will need to replicate the logic here

Mm right, I'm a bit torn, but I guess it's better to add a defensive check than to replicate the version byte logic. Let's see what the tx-metadata team thinks too

joncinque:
Oh also, can you add a test for this?

apfitzge (approved):

joncinque (approved):
Alright, let's go with it!

PR #567, opened 2026-02-11
bump wincode to 0.4.3; drop safety for VersionedTransaction

cpubot:
wincode-derive 0.4.1 deprecates the `struct_extensions` derive attribute and exposes it as a standalone derive macro, `#[derive(UninitBuilder)]`.

`UninitBuilder` can significantly simplify manual `SchemaRead` implementations, as it provides methods for dealing with uninitialized fields and provides implicit drop-safety through an internal initialization bitset. It also plugs directly into derive attributions like `#[wincode(with)]` such that custom serialization (e.g., `Vec<u8, ShortU16>`) is respected in builder methods.

This PR does two things:
1. Converts existing `#[wincode(struct_extensions)]` to use `#[derive(UninitBuilder)]`
2. Updates the `SchemaRead` impl of `VersionedTransaction` to use the uninit builder.

This change also fixes a potential memory-leak in the current implementation.

https://github.com/anza-xyz/solana-sdk/blob/a1ca3879cbcd79cea2fa2a52af19049c2986e627/transaction/src/versioned/mod.rs#L309-L319

This (correctly) initializes the `Vec<Signature>` directly into the destination. But, if the second `VersionedMessage::read` were to fail, that `Vec<Signature>` would not be dropped, since it is initialized directly on the destination, which is a `MaybeUninit<VersionedTransaction>`.

The `UninitBuilder` solves this with its initialization tracking, such that on drop, all initialized fields are dropped.

### DIFF
--- a/transaction/src/versioned/mod.rs
+++ b/transaction/src/versioned/mod.rs
@@ -10,10 +10,7 @@ use {
 };
 #[cfg(feature = "wincode")]
 use {
-    core::{
-        mem::MaybeUninit,
-        ptr::{addr_of_mut, copy_nonoverlapping},
-    },
+    core::{mem::MaybeUninit, ptr::copy_nonoverlapping},
     solana_message::v1::SIGNATURE_SIZE,
     solana_message::MESSAGE_VERSION_PREFIX,
     solana_signer::{signers::Signers, SignerError},
@@ -22,7 +19,7 @@ use {
         containers,
         io::{Reader, Writer},
         len::ShortU16,
-        ReadError, ReadResult, SchemaRead, SchemaWrite, WriteResult,
+        ReadError, ReadResult, SchemaRead, SchemaWrite, UninitBuilder, WriteResult,
     },
 };
 #[cfg(feature = "serde")]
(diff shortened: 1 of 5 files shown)


transaction/src/versioned/mod.rs
```diff
@@ -300,29 +302,22 @@ unsafe impl<'de, C: Config> SchemaRead<'de, C> for VersionedTransaction {
...
-                    &mut *(addr_of_mut!((*dst_ptr).signatures))
-                        .cast::<MaybeUninit<Vec<Signature>>>()
-                },
-            )?;
-
-            <VersionedMessage as SchemaRead<'de, C>>::read(reader, unsafe {
-                &mut *(addr_of_mut!((*dst_ptr).message)).cast::<MaybeUninit<_>>()
-            })?;
+            builder.read_signatures(&mut reader)?;
+            builder.read_message(reader)?;
```
apfitzge, line 311: this looks a lot cleaner!

apfitzge (approved):

transaction/src/versioned/mod.rs
```diff
@@ -300,23 +302,14 @@ unsafe impl<'de, C: Config> SchemaRead<'de, C> for VersionedTransaction {
...
```
apfitzge, line 311: @febo we need to make sure this is legacy or v0 right?

cpubot: Check remains from the rebase below

transaction/src/versioned/mod.rs
```diff
@@ -300,23 +302,14 @@ unsafe impl<'de, C: Config> SchemaRead<'de, C> for VersionedTransaction {
...
+            builder.finish();
```
0xalpharush: I think this needs to move after the error path as it seems to prevent dropping the initialized fields and leak
```
                return Err(ReadError::Custom("invalid message version"));
```

cpubot: Yep, good call

transaction/src/versioned/mod.rs
febo: The check is still there, so all good.

PR #572, opened 2026-02-12
feat: Allows using wincode in bls-signatures

akhi3030:
We want to be able to use `wincode` in various `votor` related structs and from a manual inspection, it seems like we need to add `wincode` support for this struct at least.


akhi3030:
sorry for the noise.  Looks like this needs a bit more work.

akhi3030:
actually, we can [de]serialize this as raw bytes so we do not need to do anything here.  Sorry for the noise.

Issue #571, opened 2026-02-11
Next round of breaking changes


febo:
### Problem

We have a few PRs with breaking changes to release.
* #567 
* #563 
* #538 
* #532 

### Proposed Solution

Here is the order of publishing and their version bump:

* [x] hash  (MINOR): wincode
* [x] address (MINOR): wincode, new helpers
* [x] pubkey (MINOR): wincode
* [x] signature  (MINOR): wincode
* [x] fee-calculator (MINOR): wincode
* [x] nonce (MINOR): wincode
* [x] message (MAJOR): txv1 + wincode
* [x] nonce-account (MAJOR): wincode + breaking change on `Account`
* [x] transaction (MAJOR): txv1 + wincode
* [x] client-traits (MAJOR): depends on transaction)
* [x] sdk-wasm-js (MAJOR): depends on transaction + wincode
* [x] sdk (MAJOR): depends on transaction

joncinque:
Here's the list of the rest of the crates, and the release that they need:

* [x] atomic-u64 -- patch fix, just get things to compile no-std
* [x] sdk-macro -- patch fix, copy change
* [x] borsh -- patch fix, deps change
* [x] commitment-config -- patch fix, satisfy clippy
* [x] file-download -- patch fix, use buffered writer
* [x] frozen-abi -- minor fix, should be major, but do minor since a private module was removed
* [x] msg -- minor fix, alloc instead of std
* [x] program-pack -- minor fix, add no-std
* [x] instruction-error -- minor upgrade, wincode
* [x] hard-forks -- patch fix, enable alloc on serde
* [x] reward-info -- major upgrade, removed struct
* [x] transaction-error -- minor upgrade, enable deps + wincode
* [x] serde-varint -- patch upgrade, serde on short-vec
* [x] cluster-type -- minor fix, technically major, but no user would break, so minor?
* [x] shred-version -- patch fix, tiny changes
* [x] instruction -- minor upgrade, wincode
* [x] packet -- minor upgrade, new constructor
* [x] serialize-utils -- patch fix, only test changes / new rand
* [x] epoch-stake -- patch fix, make no-std
* [x] package-metadata -- patch fix, enable std on pubkey explicitly
* [x] clock -- patch fix, doc comment
* [x] epoch-rewards -- patch fix, dep changes
* [x] slot-hashes -- patch fix, dep changes
* [x] stable-layout -- patch fix, dep changes
* [x] rent -- minor fix, add fields as deprecated
[...]

PR #573, opened 2026-02-12
Adding best practices to the release flow

nbelenkov:
**Problem**

We should use trusted publishing and environments across all our repos, so adding them here. Very similar to Pinocchio one, but with the addition of SLSA and a GitHub app for releases, similar to WinCode.

**Solution**

I am adding the following:
1. `prod` environment that needs 2 approvals before running
2. OIDC trusted publishing that authenticates the action directly with crates.io
3. Adding SLSA provenance, so that we can verify that the crate has been built on our action in our repo.
4. tags and releases will be limited to only the GitHub app, so the action makes an app token to create a release

**What needs to be done before merging:**

1. Move all crates that are published here to trusted publishing
2. Set up `prod` env
3. Create a gh release app


.github/workflows/publish-rust.yml
```diff
@@ -207,13 +207,29 @@ jobs:
     name: Publish crate
     runs-on: ubuntu-latest
     needs: [format, clippy, detached-minimal-versions, semver, docsrs]
+    environment: prod
     permissions:
       contents: write
+      id-token: write
+      attestations: write
```
joncinque, line 229: I'm assuming we don't want artifact metadata storage records, but just making sure: does this also need `artifact-metadata: write`?

https://github.com/actions/attest-build-provenance?tab=readme-ov-file#usage

joncinque (commented):
Makes sense to me! We have a bunch of publishes to get out with https://github.com/anza-xyz/solana-sdk/issues/571, so can we wait until those publishes are done before adding this?

I have a small worry about how annoying this process will be if we ever need to make a breaking change at `solana-address` or something, which essentially requires publishing ~100 crates. Doing that by myself took a few days of clicking a button every 5 minutes.

Maybe at that point we'll need to create a new workflow which just does all the publishes in order. Is that the plan for Agave?

.github/workflows/publish-rust.yml
nbelenkov: good spot, yep we need this here

PR #574, opened 2026-02-13
Bump wincode version

febo:
### Problem

Since wincode-derive `0.4.2` was released, the minimal versions check in CI is broken because it picks up the new wincode-derive version with an incompatible wincode one. A new `by_ref` method was added to wincode `0.4.4`, which is used by wincode-derive `0.4.2`, but currently the workspace is using wincode `0.4.3`.

### Solution

Bump wincode to `0.4.4`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5006,9 +5006,9 @@ dependencies = [
 
 [[package]]
 name = "wincode"
-version = "0.4.3"
+version = "0.4.4"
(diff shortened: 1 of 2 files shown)


febo:
@joncinque @cpubot I think another alternative would be to add wincode-derive to the workspace dependencies with an specific version, but I guess we want to have the fixes in wincode `0.4.4` instead.

PR #577, opened 2026-02-17
system-interface: Add wincode support

febo:
### Problem

The new version of `solana-message` dropped support for bincode in favour of wincode, but `SystemInstruction` does not have support for wincode. This makes it awkward to create `CompiledInstruction` for system instructions.

### Solution

Add wincode support in `system-interface`. This allows to serialize `SystemInstruction` using wincode and to use any of the instruction builder helpers.

### DIFF
--- a/system-interface/Cargo.toml
+++ b/system-interface/Cargo.toml
@@ -20,7 +20,7 @@ crate-type = ["rlib"]
 
 [features]
 alloc = []
-bincode = ["alloc", "serde", "solana-instruction/bincode", "solana-instruction/std"]
+bincode = ["alloc", "serde", "solana-instruction/bincode"]
 frozen-abi = [
     "dep:solana-frozen-abi",
     "dep:solana-frozen-abi-macro",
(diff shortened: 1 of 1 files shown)


scripts/check-no-std.sh
```diff
@@ -36,7 +36,6 @@ no_std_crates=(
   -p solana-sha256-hasher
   -p solana-signature
   -p solana-sysvar-id
-  -p solana-system-interface
```
febo, line 39: I did not want to exclude the "wincode" feature, so removed from here.

apfitzge (approved):
lgtm - would wait on jon's approval for the no-std change,

scripts/check-no-std.sh
joncinque: That makes sense to me -- the crate was never really intended for usage in no-std environments, and instead people should use some pinocchio helpers

apfitzge:
> Looks good to me, but just for my own knowledge, will we have to do something similar for all of the other programs like stake? Or is this particular to the system program because it's so tied into other things?

I'd argue we probably want to **eventually** replace bincode with wincode everywhere.
This one just pops up a bunch in agave.

joncinque:
> I'd argue we probably want to eventually replace bincode with wincode everywhere.

Totally agreed that we should do it eventually, my question was whether we need to do it immediately -- is that the case?

PR #583, opened 2026-02-24
frozen-abi: add imbl::GenericHashMap

2501babe:
https://github.com/anza-xyz/agave/pull/10762 intends to replace `im` with `imbl` as the backing store of `StakesCache`. the current crate is no longer maintained, while the new crate is a drop-in replacement under active development, with significant improvements

testing with my own benchmarks (https://github.com/anza-xyz/agave/pull/10760) and confirmed by alessandro on mainnet, this change speeds up epoch cutover by 30-50%. however, we must implement `AbiExample` for `imbl::HashMap` to satify the frozen-abi contract on `Bank`. the way this is done for `im` is a simple impl in sdk

one of the challenges with copying this for `imbl` is the perf improvements we want are in `imbl@v7.0.0`, but this version depends on rust 1.85, while sdk enforces a minimum supported version of 1.81. the alternative is to put `imbl::HashMap` under a newtype wrapper and implement `AbiExample` in agave, with no sdk changes. this ends up being quite unpleasant because it requires us to write a different `Serialize` for the `frozen-abi` feature gate that returns some other type, because a wrapper over `imbl::HashMap` does not save us from needing `AbiExample` on every underlying type. aside from being ugly, i feel this violates the spirit of `frozen-abi` because we are explicitly lying to it about the serde representation of our type

"if only there were a better way!" there is. but i leave it to you to decide if i am a clever sort or if im guilty of Crate Crimes. this pr does the simple `AbiExample` for `imbl::HashMap` in sdk, identical to the treatment of `im::HashMap`. but i allow both `imbl@v6` and `imbl@7` in `Cargo.toml`, and give sdk a `Cargo.lock` enforcing 6. this means sdk is happy on rust 1.81. but the open version allows agave to use 7 without issue, getting us the perf we desire

none of this should affect the actual thing these tests care about, which is the snapshot format. `im::HashMap` and `imbl::HashMap` both go to the serde serialized representation of a platonic hashmap

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -180,7 +180,7 @@ group = "0.13.0"
 hex = "0.4.3"
 hmac = "0.12.1"
 im = "15.1.0"
-imbl = ">= 6.0.0, < 8.0.0"
+imbl = ">= 6.0.0, < 8.0.0" # agave needs v7 for perf, but sdk needs v6 for min rust version 1.81; change to "7.0.0" when sdk >= rust 1.85
 indicatif = "0.18.4"
 itertools = "0.12.1"
 js-sys = "0.3.77"


frozen-abi/src/abi_example.rs
```diff
@@ -431,6 +431,21 @@ impl<
     }
 }
 
+#[cfg(not(target_os = "solana"))]
+impl<K, V, S, P> AbiExample for imbl::GenericHashMap<K, V, S, P>
+where
```
2501babe, line 436: `imbl::HashMap` is just a type alias for this struct with two of the generic params held fixed

Cargo.toml
```diff
@@ -180,6 +180,7 @@ group = "0.13.0"
 hex = "0.4.3"
 hmac = "0.12.1"
 im = "15.1.0"
+imbl = ">= 6.0.0, < 8.0.0"
```
joncinque: Can you just add a comment here so we don't mess this up in the future? Something like

```
# Use v6 in the SDK to maintain a lower rust-version, but allow v7 for Agave which uses a higher rust-version and needs v7 performance
```

2501babe: [73b053c](https://github.com/anza-xyz/solana-sdk/pull/583/commits/73b053c3af0c406e9313d9be82e8877347c9413f) i put it to the right rather than above to copy the style of the serde comments

joncinque (approved):
Looks great, thanks!

PR #588, opened 2026-02-26
vote interface: moar test coverage

buffalojoec:
The serialization API in the vote-interface is pretty under-tested.

This is a conglomerate PR where we move tests into proper modules and then
add more test coverage to each one.

I'll break this PR up into smaller ones and link back here.


buffalojoec:
All wrapped!

PR #592, opened 2026-02-26
vote-interface: add test coverage for authorized_voters

buffalojoec:
#### Problem
The `AuthorizedVoters` API is pretty old code, but it's still used by the Vote
program, and will continue to be used after the transition to Vote State V4.

It's fairly undocumented and has zero test coverage, despite having a somewhat
sensitive relationship between its methods.

#### Summary of Changes

Document the API completely.
Add copious test coverage, including roundtrip tests for `get & cache -> purge`,
which is what the Vote program does with this API.


joncinque (approved):
Looks great!

PR #600, opened 2026-03-02
Revert "vote: use `InvalidAccountData` for v0_23_5 vsv deserializing (#447)"

buffalojoec:
Unfortunately this change would result in a consensus issue if we bumped the validator to use solana-vote-interface >5.1.0 and the vote state v4 feature was active.

https://github.com/anza-xyz/agave/blob/b26916630e415c9b526fb317263ae90627b7a859/programs/vote/src/vote_state/mod.rs#L75-L78

Note we should yank 5.1.0 and 5.1.1 and release a new 5.2.0.

### DIFF
--- a/vote-interface/src/state/vote_state_versions.rs
+++ b/vote-interface/src/state/vote_state_versions.rs
@@ -351,7 +351,7 @@ mod tests {
         // Tag 0 (V0_23_5 — explicitly rejected).
         assert_eq!(
             VoteStateVersions::deserialize(&buf),
-            Err(InstructionError::InvalidAccountData)
+            Err(InstructionError::UninitializedAccount)
         );
 
         // Tag 4 (unknown).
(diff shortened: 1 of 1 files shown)


joncinque (commented):
Are error codes part of consensus for native programs? I know they aren't for BPF programs, and I always thought it was the same for native programs.

buffalojoec:
> Are error codes part of consensus for native programs? I know they aren't for BPF programs, and I always thought it was the same for native programs.

OK fair enough! I wanted to dig up the real reason as to why this is the case, since I've known they weren't consensus-critical, but I wasn't sure exactly on the scope.

Here's a writeup as to why and how error codes returned by programs are not part of consensus (bank hash): https://gist.github.com/buffalojoec/6b463e94637deeb2e82af434214ecc64

With that, I'll close this. Thanks!

joncinque:
Very cool writeup, thanks!

PR #603, opened 2026-03-03
account-view!: Require `mut` references

febo:
### Problem

Currently, Rust borrow checker cannot help much when manipulating accounts since the API expects only `&self` references, while many of the methods mutate the account data. For example, the following examples compile:
* Using data references after the account data has been modified:
```rust
let [from, to, _system_program] = accounts else {
    return Err(ProgramError::NotEnoughAccountKeys);
};

let data = unsafe { to.borrow_unchecked_mut() };

CreateAccount {
    from,
    to,
    lamports: 1_000_000_000,
    space: 10,
    owner: program_id,
}
.invoke()?;

data[0] = 10;
```
* Double account borrow via unchecked method:
```rust
let reference = unsafe { to.borrow_unchecked_mut() };
let another_reference = unsafe { to.borrow_unchecked_mut() };
```

An alternative is to require `&mut` references in methods that mutate the underlying account, so that Rust borrow checker can prevent UB situations like the ones above.

### Solution

Modify the following methods to require `&mut` references.

Note that the borrow checker cannot prevent double borrowing a duplicated account via different `AccountView` instances. So while requiring `&mut` references offers a bit more protection, a program is still necessary to check that accounts are not duplicated.

### DIFF
--- a/account-view/src/lib.rs
+++ b/account-view/src/lib.rs
@@ -177,7 +177,7 @@ impl AccountView {
 
     /// Set the lamports in the account.
     #[inline(always)]
-    pub fn set_lamports(&self, lamports: u64) {
+    pub fn set_lamports(&mut self, lamports: u64) {
         // SAFETY: The `raw` pointer is guaranteed to be valid.
         unsafe {
             (*self.raw).lamports = lamports;


account-view/src/lib.rs
```diff
@@ -124,15 +124,16 @@ impl AccountView {
 
     /// Return a reference to the address of the program that owns this account.
     ///
-    /// For ownership checks, use the safe `owned_by` method instead.
+    /// For ownership checks, use the safe [`Self::owned_by`] method instead.
```
LStan: Shouldn't the word "safe" be removed? The method `owner` is also safe now, which could cause confusion.

account-view/src/lib.rs
joncinque: Should this be `&mut self`?

febo: Yep, good point! Updated in https://github.com/anza-xyz/solana-sdk/pull/603/commits/47568b2aef1327b5610b8c1c46fcbef39c2b7df3.

PR #605, opened 2026-03-04
`solana-zero-copy` crate for zero-copy types and traits

grod220:
References: https://discord.com/channels/428295358100013066/774014770402689065/1478447978745303183

Adds `solana-zero-copy` and `solana-nullable` crates. All copied from [spl-pod](https://github.com/solana-program/libraries/tree/main/pod/src). 

Also implements Nullable in solana-address and solana-program-option. 

In follow-up PRs, `spl-pod` will re-export from `solana-zero-copy`.


febo:
> Now that I see it like this, I'm less confident about the name `solana-pod`. We used that name because of bytemuck's trait, but more fundamentally, this crate is providing two things:
> 
> * a zero-copy version of `Option`
> * unaligned primitives (for zero-copy types)
> 
> So the thing that ties these together is less their `Pod`-ness, but more their `zero-copy`-ness. But I'm not too sure about the name `solana-zero-copy`, since that might imply actual zero copy serde.
> 
> On the other hand, we could eventually include some `Transmute` trait to reinterpret bytes like @febo had prototyped, but I think wincode ended up being faster and safer.
> 
> Any thoughts?

This is a good point. Perhaps we could include a `Transmute` trait as well, so the library can be used for zero-copy. It will be a "lightweight" alternative in case you don't need/want to bring any dependencies. In terms of performance, wincode is the same as doing a manual transmute.

For example, in a program you might decide to use the `Transmute` trait since you have greater control of how data is manipulated. On a client, you can then enable the wincode feature to have a safer API.

febo:
Couple of other points to consider:
1. We should bring the commit history.
2. Are we keeping the `Pod` prefix for types?

joncinque:
> Are we keeping the Pod prefix for types?

If we lean into the idea of including a `Transmute` trait, then we could rename the types `Transmutable*` instead, ie `TransmutableOption` / `TransmutableU64`

Otherwise, `Unaligned*` would be very accurate, ie `UnalignedU64`, or `UnalignedLeU64` to specify its endianness

grod220:
Migrated all commits from original `spl-pod`. Just need to remember to rebase the commits instead of merging this PR. Also renamed crate to `solana-zero-copy`.

@febo any thoughts on proposed naming for the primitives?

scripts/patch-crates-functions.sh
```diff
@@ -59,6 +59,7 @@ all_crate_dirs=(
   package-metadata
   package-metadata-macro
   packet
+  pod
```
febo: nit:
```suggestion
  zero-copy
```

nullable/src/maybe_null.rs
```diff
@@ -0,0 +1,412 @@
...
+/// requiring extra space to indicate if the value is `Some` or `None`.
+///
+/// This can be used when a specific value of `T` indicates that its
+/// value is `None`.
+#[repr(transparent)]
+#[cfg_attr(
+    feature = "borsh",
+    derive(BorshDeserialize, BorshSerialize, BorshSchema)
+)]
+#[cfg_attr(feature = "wincode", derive(SchemaRead, SchemaWrite))]
```
febo, line 32: Not sure if the assert will work here, but we need to make sure that it implements `wincode::ZeroCopy`.
```suggestion
#[cfg_attr(feature = "wincode", derive(SchemaRead, SchemaWrite))]
#[cfg_attr(feature = "wincode", wincode(assert_zero_copy))]
```

zero-copy/src/option.rs
```diff
@@ -0,0 +1,412 @@
...
+            assert!(borsh::from_slice::<PodOption<u64>>(&[]).is_err());
+        }
+    }
+
+    #[cfg(feature = "wincode")]
+    mod wincode_tests {
+        use super::*;
+
+        #[test]
+        fn test_wincode_pod_option_roundtrip_and_size() {
```
febo: We probably need some tests with `ZeroCopy::from_bytes` here.

zero-copy/src/unaligned.rs
```diff
@@ -0,0 +1,381 @@
...
+
+        #[test_case(PodBool::from_bool(true))]
+        #[test_case(PodBool::from_bool(false))]
+        #[test_case(PodU16::from_primitive(u16::MAX))]
+        #[test_case(PodI16::from_primitive(i16::MIN))]
+        #[test_case(PodU32::from_primitive(u32::MAX))]
+        #[test_case(PodU64::from_primitive(u64::MAX))]
+        #[test_case(PodI64::from_primitive(i64::MIN))]
+        #[test_case(PodU128::from_primitive(u128::MAX))]
+        fn wincode_roundtrip<
```
febo, line 374: We should have tests with `ZeroCopy::from_bytes`.

joncinque:
`MaybeNull` is pretty good to me!

grod220:
Ok, did a sweep with renaming. We've got these now:

```
MaybeNull
UnalignedBool
UnalignedU16
UnalignedI16
UnalignedU32
UnalignedU64
UnalignedI64
UnalignedU128
```

A couple things I noticed.

- `MaybeNull` doesn't really convey that one value is sort of the "reserved" null value. Feel like `SentinelOption` or `ReservedValueOption` (a mouthful) would be more accurate.
- `UnalignedBool` isn't really correct. The native `bool` already is alignment 1. It's more a representation thing. I wonder if `ByteBool` is better for that.
- Is it more useful to convey the layout over the alignment property (i.e. `LeU64` versus `UnalignedU64`)?

As a family though, the names would be a bit scattered though:
```
SentinelOption
ByteBool
LeU16
LeI16
LeU32
LeU64
LeI64
LeU128
```
Also thinking about it, isn't the goal to communicate: "these are safe to use in zero copy contexts"?. And somehow to do that in a concise way. What if we went with small prefixes like:
```
(Zc = zero copy)

ZcOption
ZcBool
ZcU16
ZcI16
ZcU32
ZcU64
ZcI64
ZcU128
```
or
```
(Tm = transmutable)

TmOption
TmBool
TmU16
TmI16
TmU32
TmU64
TmI64
TmU128
```

nullable/src/maybe_null.rs
grod220: Does not compile, but took inspiration [from this](https://github.com/anza-xyz/wincode/blob/6feb75ff4c5b26275b0ee8bc4a12ea906d996a02/wincode/src/schema/impls.rs#L427) for the generics code below

febo:
> * `MaybeNull` doesn't really convey that one value is sort of the "reserved" null value. Feel like `SentinelOption` or `ReservedValueOption` (a mouthful) would be more accurate.

I don't think it should convey the idea of having a "reserved" value, that is the `Nullable` trait role. It should say "this value might be null (none)". So the name matches the requirement that a value is maybe null when it is `Nullable`. So perhaps the `Nullable` trait needs a new name?

> * `UnalignedBool` isn't really correct. The native `bool` already is alignment 1. It's more a representation thing. I wonder if `ByteBool` is better for that.

I would argue that the `Unaligned` "qualifier" means a value that can be stored without assuming natural/native alignment. Since `UnalignedBool` is effectively stored as a `u8` and not `bool`, it kind of fits this definition.

> * Is it more useful to convey the layout over the alignment property (i.e. `LeU64` versus `UnalignedU64`)?

I think the unaligned property for zero-copy is more important, since it is the functionality that the crate is providing. Programs tend to (always) use little-endian.

Another thought: what if we drop the prefixes altogether?

- `MaybeNull`
- `Bool`
- `U16`
- `I16`
- `U32`
- `U64`
- `I64`
- `U128`

The names won't clash with existing types, would they? When there is a clash, we can use the qualified name: `solana_zero_copy::U64`.

joncinque:
> Another thought: what if we drop the prefixes altogether?

I like it! Non-standard acronyms will be more annoying than anything, so we may as well be clear. We could even put them in a module called `unaligned` so people import `solana_zero_copy::unaligned::U64` -- that looks pretty neat to me.

As for the `Nullable`, what we're really providing is a union on `T`, so we could call it `OptionalUnion` or `NullableUnion`, but I would still go with `MaybeNull` over the others, I think it's the clearest name.

grod220:
I like the shortened naming too! The lower amount of characters is quite nice. Also scoped to the `unaligned::` module 👍

zero-copy/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+#![cfg_attr(docsrs, feature(doc_cfg))]
+
+#[cfg(feature = "borsh")]
+extern crate alloc;
+
+mod option;
+mod primitives;
+
+pub mod unaligned {
+    pub use crate::{option::*, primitives::*};
```
febo: Does the `option` module needs to be under `unaligned`. They could be independent of alignment, no?

zero-copy/src/option.rs
```diff
@@ -0,0 +1,425 @@
...
+    #[inline]
+    pub fn get(self) -> Option<T> {
+        if self.0.is_none() {
+            None
+        } else {
+            Some(self.0)
+        }
+    }
+
+    /// Returns the contained value as an `Option`.
```
febo:
```suggestion
    /// Returns a reference to the contained value as an `Option`.
```

zero-copy/src/option.rs
```diff
@@ -0,0 +1,425 @@
...
+    #[inline]
+    pub fn as_ref(&self) -> Option<&T> {
+        if self.0.is_none() {
+            None
+        } else {
+            Some(&self.0)
+        }
+    }
+
+    /// Returns the contained value as a mutable `Option`.
```
febo:
```suggestion
    /// Returns a mutable reference to the contained value as an `Option`.
```

zero-copy/src/option.rs
```diff
@@ -0,0 +1,425 @@
...
+pub enum MaybeNullError {
+    /// Attempted to wrap a none-equivalent value in `Some`.
+    NoneValueInSome,
+}
+
+impl core::fmt::Display for MaybeNullError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            Self::NoneValueInSome => {
+                write!(f, "cannot wrap none-equivalent value in Some")
```
febo:
```suggestion
                write!(f, "cannot wrap None-equivalent value in Some")
```

zero-copy/src/option.rs
```diff
@@ -0,0 +1,425 @@
...
+    T: Nullable + Deserialize<'de>,
+{
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let option = Option::<T>::deserialize(deserializer)?;
+        match option {
+            Some(value) if value.is_none() => Err(serde::de::Error::custom(
+                "Invalid MaybeNull encoding: Some(value) cannot equal the none marker.",
```
febo:
```suggestion
                "Invalid MaybeNull encoding: Some(value) cannot equal the None marker.",
```

nullable/src/maybe_null.rs
```diff
@@ -0,0 +1,425 @@
...
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct MaybeNull<T: Nullable>(T);
+
+#[cfg(feature = "wincode")]
```
febo, line 39:
```suggestion
/// # Safety
///
/// `MaybeNull<T>` where `T: ZeroCopy` is trivially zero-copy.
#[cfg(feature = "wincode")]
```

zero-copy/src/primitives.rs
```diff
@@ -0,0 +1,386 @@
...
+use serde_derive::{Deserialize, Serialize};
+#[cfg(feature = "wincode")]
+use wincode::{SchemaRead, SchemaWrite};
+#[cfg(feature = "borsh")]
+use {
+    alloc::string::ToString,
+    borsh::{BorshDeserialize, BorshSchema, BorshSerialize},
+};
+
+/// The standard `bool` is not a `bytemuck::Pod`, define an unaligned replacement.
```
febo:
```suggestion
/// The standard `bool` is not naturally zero-copy, define an unaligned replacement.
///
/// Any nonzero value is interpreted as `true`; only `0` is interpreted as `false`.
```

address/src/lib.rs
```diff
@@ -379,6 +379,14 @@ pub fn address_eq(a1: &Address, a2: &Address) -> bool {
     }
 }
 
+/// Implementation of `Nullable` for `Address`.
+///
+/// The zero address (`[0u8; 32]`) is the `None` value.
+#[cfg(feature = "zero-copy")]
+impl solana_zero_copy::option::Nullable for Address {
+    const NONE: Self = Address::new_from_array([0u8; ADDRESS_BYTES]);
+}
```
joncinque, line 388: Seeing this makes me think we should move the option type to a new crate. It's a bit confusing to enable the "zero-copy" feature on Address, when the type is already zero-copy -- people will likely blindly enable the feature without knowing what they're getting.

Let's move `Nullable` to a new crate, maybe called `solana-nullable` or `solana-nullable-option` or `solana-optional-union` or `solana-union-option` or `solana-zero-copy-option` -- what do you think @febo ?

zero-copy/src/unaligned.rs
joncinque, line 1: How about we just call this `unaligned.rs`?

address/src/lib.rs
febo: Would that be just the trait `Nullable`? One alternative is to rename the feature from `"zero-copy"` to `"nullable"` to avoid the confusion.

joncinque: Yeah I think so, and implementing the trait for any types that we don't have control over. I could get on board with a different name to the feature too

febo: > Yeah I think so, and implementing the trait for any types that we don't have control over.

This is a good point, let's go with a separate `solana-nullable` crate.

zero-copy/src/unaligned.rs
grod220: Whoops, yes

address/src/lib.rs
grod220: Just updated. However, `solana-program-option` is a bit different given it imports `MaybeNull` which remained in the zero copy crate.

nullable/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+    ($($ty:ty),* $(,)?) => {
+        $(
+            impl Nullable for $ty {
+                const NONE: Self = 0;
+            }
+        )*
+    };
+}
+
+impl_zero_as_none!(i8, i16, i32, i64, i128, isize, u8, u16, u32, u64, u128, usize);
```
febo: I am not sure whether upstream BPF supports `u128` already or not.
```suggestion
impl_zero_as_none!(i8, i16, i32, i64, isize, u8, u16, u32, u64, usize);
#[cfg(not(target_arch = "bpf"))]
impl_zero_as_none!(i128, u128);
```

program-option/src/lib.rs
```diff
@@ -10,6 +10,11 @@ use core::{
     convert, mem,
     ops::{Deref, DerefMut},
 };
+#[cfg(feature = "maybe-null")]
+use {
+    solana_nullable::Nullable,
+    solana_zero_copy::option::{MaybeNull, MaybeNullError},
+};
```
febo: Since the use of `MaybeNull` is tied to the use of `Nullable`, it would be easier for downstream crates if we re-export `solana-nullable` from `solana-zero-copy`. The this would look like:
```suggestion
use {
    solana_zero_copy::option::{MaybeNull, MaybeNullError, Nullable},
};
```

zero-copy/src/option.rs
```diff
@@ -0,0 +1,406 @@
...
+//! For example, a 64-bit unsigned integer can designate `0` as a `None` value.
+//! This would be equivalent to
+//! [`Option<NonZeroU64>`](https://doc.rust-lang.org/std/num/type.NonZeroU64.html)
+//! and provide the same memory layout optimization.
+
+#[cfg(feature = "bytemuck")]
+use bytemuck::{Pod, Zeroable};
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use solana_nullable::Nullable;
```
febo: As I mentioned in another comment, we could re-export `solana-nullable` in this module.
```suggestion
pub use solana_nullable::Nullable;
```

zero-copy/src/unaligned.rs
```diff
@@ -0,0 +1,388 @@
...
+#[cfg_attr(
+    feature = "borsh",
+    derive(BorshDeserialize, BorshSerialize, BorshSchema)
+)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(from = "u128", into = "u128"))]
+#[cfg_attr(feature = "bytemuck", derive(Pod, Zeroable))]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+#[repr(transparent)]
+pub struct U128(pub [u8; 16]);
```
febo, line 161: Same comment about upstream BPF `u128` support. Maybe we need a `#[cfg(not(target_arch = "bpf"))]` here as well.

nullable/src/lib.rs
joncinque: It definitely does not unfortunately :sweat_smile:

joncinque: Is there a reason to implement these at all? It would make more sense to do it if we expose a set of unaligned non-zero integer types similar to `NonZeroU64`, but for these I worry it'll just create an unnecessary footgun.

nullable/src/maybe_null.rs
joncinque, line 1: Sorry for all the continued comments on this, but why don't we put this type in the `solana-nullable` crate? Having just the trait in a separate crate will probably be more confusing, since the trait is useless without this type

program-option/Cargo.toml
```diff
@@ -11,3 +11,13 @@ edition = { workspace = true }
 
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
+
+[features]
+maybe-null = ["dep:solana-nullable", "dep:solana-zero-copy"]
```
joncinque: nit: how about calling this feature `nullable` for consistency?

nullable/src/lib.rs
grod220: I ran into the orphan rule when trying to do some simple tests for it when it was in a split crate and figured to do the impl here. But your reasoning is fair, will remove.

program-option/src/lib.rs
grod220: Now co-located so no longer an issue

program-option/Cargo.toml
grod220: when MaybeNull was separate, the different naming made more sense in my head. But now plain nullable is definitely best.

nullable/src/maybe_null.rs
grod220: Think I was confused by the response to Febo's question: "Would that be just the trait Nullable?". Yes would simplify things to keep these together. Fixing.

zero-copy/src/unaligned.rs
grod220: Yep, makes sense!

grod220:
Oh dear 😅 

> Merging a pull request using the "Rebase and merge" option is limited to 100 commits. - [docs](https://docs.github.com/en/repositories/creating-and-managing-repositories/repository-limits#rebase-limits
)

In an effort to preserve the history, going to squash dependabot commits if adjacent to another into one commit. That should get us under the 100 mark.

joncinque:
> In an effort to preserve the history, going to squash dependabot commits if adjacent to another into one commit. That should get us under the 100 mark.

Ohhhh that explains why sometimes my merges wouldn't work when importing history into new repos. I would just push directly to main in those cases :smiling_imp: but better to not do that here

grod220:
Package updates:

- solana-zero-copy 1.0.0 first release
- solana-nullable 1.0.0 first release
- solana-address 2.3.0 -> 2.4.0
- solana-program-option 3.0.1 -> 3.1.0

To come:

- spl-pod 0.7.2 -> 0.7.3 (after zero copy re-export PR)

PR #607, opened 2026-03-05
bls cache

zz-sol:
This branch introduces a substantial BLS-signature verification refactor.

- Added `PreparedHashedMessage` support so callers can reuse pre-hashed/prepared message data.
- Added new APIs for aggregate/distinct verification across:
  - raw messages
  - pre-hashed messages
  - pre-hashed + prepared messages
- Updated both serial and `parallel` verification paths to support these forms.
- Added message-hash grouping in distinct aggregated verification (identical message hashes are merged by summing pubkeys before pairing checks), improving efficiency.
- Refactored pubkey verification internals to expose “prepared” verification path directly.

~Bumped `solana-bls-signatures` crate version **from 3.1.0 to 3.2.0**.~

### DIFF
--- a/bls-signatures/src/signature/points.rs
+++ b/bls-signatures/src/signature/points.rs
@@ -9,7 +9,7 @@ use {
         signature::bytes::{Signature, SignatureCompressed},
     },
     blstrs::{Bls12, G1Affine, G1Projective, G2Affine, G2Prepared, G2Projective, Gt, Scalar},
-    group::{prime::PrimeCurveAffine, Group, GroupEncoding},
+    group::{prime::PrimeCurveAffine, Group},
     pairing::{MillerLoopResult, MultiMillerLoop},
 };
 #[cfg(all(feature = "parallel", not(target_os = "solana")))]
@@ -443,7 +443,7 @@ impl SignatureProjective {
         let neg_g1_generator = &neg_g1_generator_val;
 
         let (grouped_pubkeys_affine, grouped_prepared_hashes) = Self::group_prepared_terms(
-            pubkeys_affine.into_iter().zip(prepared_refs.into_iter()),
+            pubkeys_affine.into_iter().zip(prepared_refs),
             public_keys_len,
         );
 


bls-signatures/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "solana-bls-signatures"
 description = "Solana BLS Signatures"
 documentation = "https://docs.rs/solana-bls-signatures"
-version = "3.1.0"
+version = "3.2.0"
```
samkim-crypto: Ah we generally do publishing via the CI and the CI script actually bumps in `Cargo.toml` for us. Let's keep the current version and let the CI bump this when it does the publishing. You can browse our CI logic [here](https://github.com/anza-xyz/solana-sdk/blob/master/.github/workflows/publish-rust.yml) if you are curious.

bls-signatures/src/signature/points.rs
```diff
@@ -45,6 +38,44 @@ pub struct SignatureProjective(pub(crate) G2Projective);
...
+        let mut grouped_pubkeys = alloc::vec::Vec::with_capacity(entries.len());
+        let mut grouped_prepared_hashes = alloc::vec::Vec::with_capacity(entries.len());
+
+        for (key, pubkey, prepared_hash) in entries {
+            if grouped_keys.last().is_some_and(|last| *last == key) {
+                let idx = grouped_pubkeys.len().saturating_sub(1);
+                #[allow(clippy::arithmetic_side_effects)]
+                {
+                    grouped_pubkeys[idx] += pubkey;
+                }
```
samkim-crypto, line 69: Nit: Should we just do something like
```
if let Some(last_pubkey) = grouped_pubkeys.last_mut() {
    *last_pubkey += pubkey;
}
```

To avoid clippy warning and slightly more idiomatic?

bls-signatures/src/signature/points.rs
```diff
@@ -45,6 +38,44 @@ pub struct SignatureProjective(pub(crate) G2Projective);
...
+        }
+
+        entries.sort_unstable_by(|a, b| a.0.cmp(&b.0));
+
+        let mut grouped_keys = alloc::vec::Vec::with_capacity(entries.len());
+        let mut grouped_pubkeys = alloc::vec::Vec::with_capacity(entries.len());
+        let mut grouped_prepared_hashes = alloc::vec::Vec::with_capacity(entries.len());
+
+        for (key, pubkey, prepared_hash) in entries {
+            if grouped_keys.last().is_some_and(|last| *last == key) {
```
samkim-crypto: Seems like we can write `if grouped_keys.last() == Some(&key)` to compare directly.

bls-signatures/src/signature/points.rs
```diff
@@ -45,6 +38,44 @@ pub struct SignatureProjective(pub(crate) G2Projective);
...
+    fn group_prepared_terms<'a>(
+        pairs: impl Iterator<Item = (G1Affine, &'a PreparedHashedMessage)>,
+        capacity: usize,
+    ) -> (alloc::vec::Vec<G1Affine>, alloc::vec::Vec<G2Prepared>) {
+        let mut entries = alloc::vec::Vec::with_capacity(capacity);
+        for (pubkey_affine, prepared_hashed_message) in pairs {
+            entries.push((
+                prepared_hashed_message.hashed_message.0.to_bytes(),
+                G1Projective::from(pubkey_affine),
+                prepared_hashed_message.prepared.clone(),
```
samkim-crypto: Do we need the vector to own `G2Prepared`. My understanding was the `G2Prepared` is 19kb, so cloning a bunch of these could be quite expensive. It seems like the `multi_miller_loop` itself takes in vector of references, so maybe we should store a vector of references instead.

bls-signatures/src/hash.rs
```diff
@@ -32,6 +32,29 @@ impl HashedMessage {
     }
 }
 
+/// A pre-hashed and prepared message (G2 prepared point) for optimized verification.
+/// This type is useful when the same message is verified repeatedly against many signatures.
```
samkim-crypto: Nit: Let's make the docs for the `PreparedHashedMessage` and `HashedMessage` consistent. For each of these, let's also mention the memory implications: `PreparedHashedMessage` is larger ~19kb(?) for each element, but that we get to skip the pairing preparation step.

zz-sol:
> Looks fine to me overall! Some comments below. I was considering caching the `G2Prepared` as well, but my understanding was that `G2Prepared` is quite large ~19kb, so I was originally thinking of just caching `G2Affine` instead.
> 
> But maybe in the grand scheme of things, 19kb is not so bad. I guess it could be an issue if someone constantly spams the network with distinct messages, but we can probably take care of it by limiting the size of the cache.

good catch. i refactored the code so now it caches affine form instead of prepared ones

bls-signatures/src/signature/points.rs
```diff
@@ -147,7 +208,64 @@ impl SignatureProjective {
...
+            return Err(BlsError::InputLengthMismatch);
+        }
+        if public_keys.len() == 0 {
+            return Err(BlsError::EmptyAggregation);
+        }
+        let aggregate_signature = SignatureProjective::aggregate(signatures)?;
+        Self::verify_distinct_aggregated_pre_hashed(
+            public_keys,
+            &aggregate_signature,
+            prepared_hashed_messages.map(|prepared| &prepared.hashed_message),
```
samkim-crypto: Here, it seems like the API is a little weird. The `verify_distinct_prepared` takes in `PreparedHashedMessage`, but inside the function, we drop the `G2Prepared` point and let the `verify_distinct_aggregated_pre_hashed` recompute it again.

It seems like we should either remove the `G2Prepared` API or make functions like `verify_distinct_aggreagated` generic so that it can either take in pre-hashed or pre-hashed-prepared points.

bls-signatures/src/signature/points.rs
```diff
@@ -45,6 +38,45 @@ pub struct SignatureProjective(pub(crate) G2Projective);
...
 impl SignatureProjective {
+    /// Group pairing terms by message hash in `O(n log n)` time by sorting keys.
+    fn group_hashed_terms(
+        pairs: impl Iterator<Item = (G1Affine, HashedMessage)>,
+        capacity: usize,
+    ) -> (alloc::vec::Vec<G1Affine>, alloc::vec::Vec<HashedMessage>) {
+        let mut entries = alloc::vec::Vec::with_capacity(capacity);
+        for (pubkey_affine, hashed_message) in pairs {
+            entries.push((
+                hashed_message.0.to_bytes(),
```
samkim-crypto: Sorry I missed commenting on this before, but I believe the `to_bytes()` will compress the point, which will perform division. It might not make that much of a difference, but it might be worth checking if `to_uncompressed()` performs better here. The `to_uncompressed` will skip division, but can make sorting slightly slower, so not clear which is better.

bls-signatures/src/signature/points.rs
```diff
@@ -45,6 +38,45 @@ pub struct SignatureProjective(pub(crate) G2Projective);
...
+                G1Projective::from(pubkey_affine),
```
samkim-crypto: Can you also double check if `G1Projective` conversion is really need here? `blst` supports mixed addition, so maybe we can get away with just using affine, which should be faster?

bls-signatures/src/signature/points.rs
```diff
@@ -307,52 +585,75 @@ impl SignatureProjective {
...
+                }
+                Ok::<G1Affine, BlsError>(affine.0)
+            })
+            .collect::<Result<_, _>>()?;
 
-        // Check for errors from the parallel operations and unwrap the results.
-        let pubkeys_affine = pubkeys_affine_res?;
-        let prepared_hashes = prepared_hashes_res?;
+        let public_keys_len = public_keys.len();
+        let (grouped_pubkeys_affine, grouped_hashed_messages) = Self::group_hashed_terms(
```
samkim-crypto, line 703: The `group_hashed_terms` is largely a sequential function. I wonder if it makes a difference to use something like rayon's `par_sort_unstable_by`. Might be worth testing in a follow-up.

bls-signatures/src/signature/points.rs
```diff
@@ -45,6 +38,45 @@ pub struct SignatureProjective(pub(crate) G2Projective);
...
+            entries.push((
+                hashed_message.0.to_bytes(),
+                G1Projective::from(pubkey_affine),
+                hashed_message,
+            ));
+        }
+
+        entries.sort_unstable_by(|a, b| a.0.cmp(&b.0));
+
+        let mut grouped_keys = alloc::vec::Vec::with_capacity(entries.len());
```
samkim-crypto: I thought a bit again about simplifying the loop below a bit. What do you think about the following?

```
        let mut grouped_pubkeys = alloc::vec::Vec::new();
        let mut grouped_hashed_messages = alloc::vec::Vec::new();

        for chunk in entries.chunk_by(|a, b| a.0 == b.0) {
            let mut aggregate_pubkey = chunk[0].1;
            
            // Sum any remaining pubkeys in this identical chunk
            for item in &chunk[1..] {
                #[allow(clippy::arithmetic_side_effects)]
                {
                    aggregate_pubkey += item.1;
                }
            }
            
            grouped_pubkeys.push(aggregate_pubkey);
            grouped_hashed_messages.push(chunk[0].2);
        }
```

It seems like we can remove the `grouped_keys` this way and save on memory allocation.

PR #608, opened 2026-03-06
file-download: allow selection of tls backend

t-nelson:
#### Problem
the `reqwest` bump in 3.1.3 unintentionally swapped out the default tls backend (formerly ring, now aws-lc-rs). feature set merging causes a conflict in `rustls` configuration, which breaks a monorepo doctest :upside_down_face: 

#### Changes
expose `reqwest`'s tls backend crate features to allow configuration

### DIFF
--- a/file-download/Cargo.toml
+++ b/file-download/Cargo.toml
@@ -12,8 +12,13 @@ edition = { workspace = true }
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
 
+[features]
+default = ["native-tls"]
+rustls = ["reqwest/rustls"]
+native-tls = ["reqwest/native-tls"]
+
 [dependencies]
 console = { workspace = true }
 indicatif = { workspace = true }
 log = { workspace = true }
-reqwest = { workspace = true, features = ["blocking", "brotli", "deflate", "gzip", "rustls", "json"] }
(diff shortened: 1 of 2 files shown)


file-download/Cargo.toml
```diff
@@ -12,8 +12,13 @@ edition = { workspace = true }
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
 
+[features]
+default = ["rustls"]
```
t-nelson: i chose to retain the `reqwest` default backend as our default backend. this is probably a minor(?) bump. if we want to retain the `solana-file-download` default backend i can swap this and then we retain patch compatibility. either way we probably want to yank 3.1.3

joncinque: If we want to use reqwest's default, how about just doing that explicitly by enabling `reqwest/default-tls` here? That way if the default changes, we get that immediately, *and* we can still configure it as needed downstream.

Either way, I agree with yanking and doing a minor release

t-nelson: the more i think about this the more i think we should maintain `solana-file-download`'s default and just deprecate the crate. it's useless phone guy abstraction. reqwest is easy to use

i'll make the change now

t-nelson:
> Ok this works!

who has authority to yank 3.1.3 once 3.1.4 is out?

PR #613, opened 2026-03-06
address: Add derive program address helper

febo:
### Problem

Currently the only way to find a program address is to use `Address::try_find_program_address`. After PR #333 introduced a helper to derive an address when the bump value is known and #526 added the ability to check whether an address is on curve or not, both can be combined to find a program address consuming less CUs.

### Solution

This PR adds a new helper `derive_program_address` that instead of using the `try_find_program_address`, it uses the `sol_sha256` and `sol_curve_validate_point` syscalls to find a program address by testing all possible bump values.

Below is a CU comparison for a derivation that takes one iteration (`255` bump value):
| function       | CUs |
|------------| --- |
| `try_find_program_address` | 🟥 1528 (+1193) |
| (*new*) `derive_program_address`  |  🟩 335 |

Note: Each evaluation of a bump value consumes about `335` CUs, so an address with a bump value of `254` consumes about `770` CUs and so on; similarly, each evaluation in `try_find_program_address` consumes about `1528`, so two iterations consumes about `3056` and so on.

### DIFF
--- a/address/Cargo.toml
+++ b/address/Cargo.toml
@@ -33,11 +33,7 @@ frozen-abi = [
 rand = ["dep:rand", "atomic", "std"]
 sanitize = ["dep:solana-sanitize"]
 serde = ["dep:serde", "dep:serde_derive"]
-sha2 = [
-    "dep:sha2-const-stable",
-    "dep:solana-sha256-hasher",
-    "syscalls",
-]
+sha2 = ["dep:sha2-const-stable", "dep:solana-sha256-hasher", "syscalls"]
 std = ["decode", "borsh?/std", "serde?/std", "wincode?/std", "alloc"]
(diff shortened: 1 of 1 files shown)


address/Cargo.toml
```diff
@@ -35,9 +35,8 @@ sanitize = ["dep:solana-sanitize"]
 serde = ["dep:serde", "dep:serde_derive"]
 sha2 = [
     "dep:sha2-const-stable",
-    "dep:solana-define-syscall",
     "dep:solana-sha256-hasher",
-    "error",
+    "syscalls",
```
febo: The `"syscalls"` feature has the same dependencies and it is needed for the `Address::is_on_curve` to be visible.

PR #616, opened 2026-03-10
cleanup(message): peek denominator to decode LegacyMessage

kskalski:
Simplify implementation of wincode decoding `VersionedMessage` by first peeking the denominator byte and consuming it immediately if message format is not `Legacy`.

In legacy case, since no bytes were consumed so far, the implementation can proceed with using regular generated implementation for `LegacyMessage`.


cpubot (commented):
LGTM assuming we want to keep buffering `Reader` methods in v0.5

kskalski:
Since it only uses peek for 1 byte, yes, this shouldn't be a problem.
I would be more diligent about looking for cases where we need to peek more than 1 byte, which will force more costly buffer manipulations.

PR #618, opened 2026-03-11
[bls-signatures] Add `PopVerified` trait and re-organize Pop and signature verification traits

samkim-crypto:
#### Problem

Currently, the crate allows raw, unverified public keys to be used directly in aggregate signature verification. This exposes downstream users to potential Rogue Key Attacks if they forget to manually verify a Proof of Possession (PoP). Because the compiler doesn't complain when unverified keys are aggregated, it is easy to accidentally introduce a critical vulnerability. We need the API to strictly enforce cryptographic safety so these unsafe aggregations are caught at compile time. By leveraging the Rust type system, we can make this invalid state completely unrepresentable. This effectively shifts the burden of remembering PoP checks from the developer to the compiler.

#### Summary of Changes

* Introduced `PopVerified` and `AggregatePubkey` wrapper types to track the trusted cryptographic state of public keys at compile time. 
* Split the previous `VerifiablePubkey` trait into `VerifyPop` (for proving raw keys) and `VerifySignature` (restricted exclusively to the safe wrapper types). 
* Updated the aggregation and verification methods to strictly require these safe wrappers, while providing an `unsafe { PopVerified::new_unchecked() }` escape hatch for loading trusted keys from local storage.

### DIFF
--- a/bls-signatures/src/pubkey/points.rs
+++ b/bls-signatures/src/pubkey/points.rs
@@ -1,3 +1,4 @@
+use core::ops::Deref;
 #[cfg(all(feature = "parallel", not(target_os = "solana")))]
 use rayon::prelude::*;
 #[cfg(all(not(target_os = "solana"), feature = "std"))]
@@ -13,7 +14,6 @@ use {
         signature::{AsSignatureAffine, SignatureAffine},
     },
     blstrs::{Bls12, G1Affine, G1Projective, G2Prepared, Gt, Scalar},
-    core::ops::Deref,
     group::{prime::PrimeCurveAffine, Group},
     pairing::{MillerLoopResult, MultiMillerLoop},
 };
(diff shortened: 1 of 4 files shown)


bls-signatures/src/pubkey/points.rs
```diff
@@ -414,3 +422,66 @@ impl_add_to_accumulator!(
...
+#[repr(transparent)]
+pub struct PopVerified<T: ?Sized>(pub(crate) T);
+
+impl<T: ?Sized> PopVerified<T> {
+    /// Bypasses the cryptographic PoP check.
+    ///
+    /// # Safety
+    /// Only use this if the public key was loaded from a trusted source
+    /// where its Proof of Possession was previously verified (e.g., a local validator ledger).
+    pub unsafe fn new_unchecked(inner: T) -> Self
```
zz-sol, line 437: nice!

zz-sol:
LGTM. 

Only note that i want to add is that maybe we can do a readme.md for glossary? there are quite a few different struct/trait with similar names and it may be a bit confusing

samkim-crypto:
> LGTM.
> 
> Only note that i want to add is that maybe we can do a readme.md for glossary? there are quite a few different struct/trait with similar names and it may be a bit confusing

Yeah this was something that was suggested by the auditors as well. I am currently working on a README with small examples on how to use the crate. I will add this in a follow-up PR.

PR #620, opened 2026-03-12
account-info: add no_std support

grod220:
Convert `solana-account-info` to a `#![no_std]` crate. No public API changes.

### DIFF
--- a/account-info/src/lib.rs
+++ b/account-info/src/lib.rs
@@ -1,15 +1,21 @@
 //! Account information.
+#![no_std]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+
+extern crate alloc;
+
+#[cfg(feature = "bincode")]
+use alloc::boxed::Box;
 use {
+    alloc::rc::Rc,
+    core::{
+        cell::{Ref, RefCell, RefMut},
+        fmt,
+        slice::from_raw_parts_mut,
+    },
     solana_address::Address,
     solana_program_error::ProgramError,
     solana_program_memory::sol_memset,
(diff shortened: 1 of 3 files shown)


scripts/check-no-std.sh
```diff
@@ -38,6 +38,13 @@ no_std_crates=(
   -p solana-signature
   -p solana-sysvar-id
 )
+
+# Crates that are no_std but unconditionally require alloc.
+# These are only checked in the alloc+core pass, not the core-only pass.
+no_std_alloc_crates=(
+  -p solana-account-info
+)
+
```
grod220, line 47: Up for discussion on this. `account-info` doesn't really serve a purpose without alloc. But if so, we need an exception in this script.

joncinque: I like this approach! We'll have a bunch more of these, like `solana-instruction` and `solana-program-entrypoint`, so best start somewhere. @febo what do you think?

account-info/src/lib.rs
```diff
@@ -1,15 +1,21 @@
 //! Account information.
+#![no_std]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+
+extern crate alloc;
+
+#[cfg(feature = "bincode")]
+use alloc::boxed::Box;
```
joncinque, line 8: We really shot ourselves in the foot with this one 😅

scripts/check-no-std.sh
febo: I like this approach as well.

PR #621, opened 2026-03-12
wincode API compatible v1::Message SchemaRead impl

cpubot:
wincode's `Reader` trait currently allows arbitrary buffering from its implementation back-end (via `fill_buf`, `fill_exact`, etc). This is being removed in the next version so that wincode can support a wider variety of `Reader` back-ends, and is now deprecated as of `0.4.6`.

This requires rewriting the current `v1::Message` `SchemaRead` implementation because it currently uses `fill_buf` and operates directly on that slice.
https://github.com/anza-xyz/solana-sdk/blob/94e0202ceab167596a176c25ca09b96755885b24/message/src/versions/v1/message.rs#L540-L543

The rewrite follows the same deserialization scheme as the existing manual `deserialize` function, but uses wincode's supported `Reader` APIs. The manual reference implementation can be seen here:
https://github.com/anza-xyz/solana-sdk/blob/94e0202ceab167596a176c25ca09b96755885b24/message/src/versions/v1/message.rs#L652-L834

The resulting assembly of both implementations is nearly identical, with the `SchemaRead` implementation having a slightly smaller stack frame. Local micro-benchmarks corroborate this, as performance is basically identical, with the `SchemaRead` implementation being slightly faster on larger payloads and the manual impl being slightly faster on smaller payloads, but the difference was on the order of nanoseconds, so likely benchmark noise.
```
v1_deserialize/manual/small
                        time:   [36.012 ns 36.278 ns 36.567 ns]
                        thrpt:  [2.9035 GiB/s 2.9266 GiB/s 2.9482 GiB/s]
v1_deserialize/schema/small
                        time:   [39.239 ns 39.523 ns 39.792 ns]
                        thrpt:  [2.6682 GiB/s 2.6863 GiB/s 2.7057 GiB/s]
v1_deserialize/manual/large
                        time:   [164.67 ns 166.15 ns 167.68 ns]
                        thrpt:  [9.8920 GiB/s 9.9832 GiB/s 10.073 GiB/s]
v1_deserialize/schema/large
                        time:   [148.30 ns 149.25 ns 150.39 ns]
                        thrpt:  [11.029 GiB/s 11.114 GiB/s 11.185 GiB/s]
```

There are some other very minor deprecations with the 0.4.6 release -- those have been included here as well since they are just a simple rename `peek` -> `peek_byte`, or switching from `fill_exact` to `take_scoped`. I can do these changes in a separate PR, but we will have to `#[allow(deprecated)]` on not-yet-migrated parts.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -203,7 +203,7 @@ proc-macro2 = "1.0.106"
 proptest = "1.9"
 qstring = "0.7.2"
 qualifier_attr = { version = "0.2.2", default-features = false }
-quote = "1.0.44"
+quote = "1.0.45"
 rand = "0.9.2"
 rand_chacha = "0.9.0"
 rayon = "1.10.0"
@@ -329,7 +329,7 @@ static_assertions = "1.1.0"
 strum = "0.24"
 strum_macros = "0.24"
 subtle = "2.6.1"
-syn = "2.0.114"
+syn = "2.0.117"
 tempfile = "3.20.0"
 test-case = "3.3.1"
 thiserror = { version = "2.0.18", default-features = false }
@@ -337,7 +337,7 @@ tiny-bip39 = "2.0.0"
 toml = "0.8.23"
 uriparse = "0.6.4"
 wasm-bindgen = "0.2.100"
-wincode = { version = "0.4.4", features = ["derive"], default-features = false }
+wincode = { version = "0.4.7", features = ["derive"], default-features = false }
 
 [profile.release]
 split-debuginfo = "unpacked"
--- a/message/src/versions/v1/message.rs
+++ b/message/src/versions/v1/message.rs
@@ -41,11 +41,10 @@ use serde_derive::{Deserialize, Serialize};
 use solana_frozen_abi_macro::AbiExample;
 #[cfg(feature = "wincode")]
 use {
-    crate::v1::MAX_TRANSACTION_SIZE,
-    core::slice::from_raw_parts,
+    crate::v1::{InstructionHeader, FIXED_HEADER_SIZE},
+    core::{mem::MaybeUninit, ptr::copy_nonoverlapping, slice::from_raw_parts},
     wincode::{
-        config::ConfigCore,
-        error::invalid_tag_encoding,
+        config::{Config, ConfigCore},
         io::{Reader, Writer},
         ReadResult, SchemaRead, SchemaWrite, WriteResult,
     },
@@ -55,17 +54,17 @@ use {
         compiled_instruction::CompiledInstruction,
         compiled_keys::CompiledKeys,
         v1::{
-            InstructionHeader, MessageError, TransactionConfig, TransactionConfigMask,
-            FIXED_HEADER_SIZE, MAX_ADDRESSES, MAX_INSTRUCTIONS, MAX_SIGNATURES,
+            MessageError, TransactionConfig, TransactionConfigMask, MAX_ADDRESSES,
+            MAX_INSTRUCTIONS, MAX_SIGNATURES,
         },
         AccountKeys, CompileError, MessageHeader,
     },
-    core::{mem::size_of, ptr::copy_nonoverlapping},
+    core::mem::size_of,
     solana_address::Address,
     solana_hash::Hash,
     solana_instruction::Instruction,
     solana_sanitize::{Sanitize, SanitizeError},
-    std::{collections::HashSet, mem::MaybeUninit},
+    std::collections::HashSet,
 };
 
 /// A V1 transaction message (SIMD-0385) supporting 4KB transactions with inline compute budget.
@@ -516,7 +515,6 @@ impl Sanitize for Message {
 unsafe impl<C: ConfigCore> SchemaWrite<C> for Message {
     type Src = Self;
 
-    #[allow(clippy::arithmetic_side_effects)]
     #[inline(always)]
     fn size_of(src: &Self::Src) -> WriteResult<usize> {
         Ok(src.size())
@@ -575,23 +573,6 @@ unsafe impl<C: ConfigCore> SchemaWrite<C> for Message {
     }
 }
 
-#[cfg(feature = "wincode")]
-unsafe impl<'de, C: ConfigCore> SchemaRead<'de, C> for Message {
-    type Dst = Self;
-
-    fn read(mut reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
-        let bytes = reader.fill_buf(MAX_TRANSACTION_SIZE)?;
-        let (message, consumed) = deserialize(bytes).map_err(|_| invalid_tag_encoding(1))?;
-
-        // SAFETY: `deserialize` validates that we read `consumed` bytes.
-        unsafe { reader.consume_unchecked(consumed) };
-
-        dst.write(message);
-
-        Ok(())
-    }
-}
-
 /// Serialize the message.
 #[cfg(feature = "wincode")]
 #[inline]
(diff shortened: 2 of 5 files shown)


transaction/src/versioned/mod.rs
```diff
@@ -353,9 +353,6 @@ unsafe impl<'de, C: Config> SchemaRead<'de, C> for VersionedTransaction {
                     expected_signatures_len,
                 );
                 signatures.set_len(expected_signatures_len);
-                // Advance the reader by the number of bytes we just consumed
-                // for the signatures.
-                reader.consume_unchecked(bytes_to_read);
```
cpubot, line 358: [`take_scoped`](https://docs.rs/wincode/0.4.7/wincode/io/trait.Reader.html#method.take_scoped) automatically advances the reader, so `consume*` is no longer needed here.

Cargo.toml
apfitzge, line 1: bumping the non-wincode versions here on purpose?

message/src/versions/v1/message.rs
```diff
@@ -646,6 +628,110 @@ pub(crate) unsafe fn serialize_into(message: &Message, mut dst: *mut u8) {
     }
 }
 
+#[cfg(feature = "wincode")]
+unsafe impl<'de, C: Config> SchemaRead<'de, C> for Message {
+    type Dst = Message;
+
+    #[expect(clippy::arithmetic_side_effects)]
+    fn read(mut reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
```
apfitzge, line 588: we've now got essentially the same logic duplicated here and in deserialize. Can these somehow be shared so we don't have to maintain in 2 places?

Cargo.toml
apfitzge, line 1: are the non-wincode updates intentional here? are they necessary?

cpubot: Necessary for the `minimal-version` CI check

message/src/versions/v1/message.rs
cpubot: In theory `deserialize` could call `wincode`, but that forces a `wincode` dependency in the `message` crate.

These two paths have different contracts:

- `SchemaRead::read` operates on a generic `Reader`
- `deserialize` operates on a raw `&[u8]`

Sharing logic would require some kind of cursoring abstraction that could handle both cases, which likely would add more code, logic / complexity  than it would save

apfitzge: 😕 does deserialize not require wincode already? I thought it did? Doesn't serialize require wincode?

I don't see a reason to not require wincode for deserialization.

cpubot: > 😕 does deserialize not require wincode already? I thought it did? Doesn't serialize require wincode?

It does not -- it is implemented manually over a slice https://github.com/anza-xyz/solana-sdk/blob/94e0202ceab167596a176c25ca09b96755885b24/message/src/versions/v1/message.rs#L652-L834

Same story with serialize
https://github.com/anza-xyz/solana-sdk/blob/94e0202ceab167596a176c25ca09b96755885b24/message/src/versions/v1/message.rs#L554-L647


> I don't see a reason to not require wincode for deserialization.

I can't speak to this as I don't maintain the sdk crates, but perhaps @joncinque can weigh in

apfitzge: /cc @febo

cpubot: It looks like we actually gate a `serialize` method on on `VersionedMessage` by the `wincode` feature. So perhaps it's reasonable to remove the manual implementation and just use `wincode`. But need more guidance from a maintainer here.

https://github.com/anza-xyz/solana-sdk/blob/94e0202ceab167596a176c25ca09b96755885b24/message/src/versions/mod.rs#L170-L173

febo: Agree – we can remove the manual ones (both `serialize` and `deserialize`) and just use `wincode` impls. It was separate before since we were trying to use bincode as well.

cpubot: Done

message/src/versions/v1/message.rs
```diff
@@ -646,191 +628,116 @@ pub(crate) unsafe fn serialize_into(message: &Message, mut dst: *mut u8) {
...
-        offset,
-    ))
+        });
+
+        Ok(())
+    }
+}
+
+/// Deserialize the message from the provided input buffer, returning the message and
+/// the number of bytes read.
```
febo, line 688:
```suggestion
/// Deserialize the message from the provided input buffer.
```

message/src/versions/v1/message.rs
```diff
@@ -646,191 +628,116 @@ pub(crate) unsafe fn serialize_into(message: &Message, mut dst: *mut u8) {
...
+                lifetime_specifier,
+                config_mask,
+                num_instructions,
+                num_addresses,
+            )
+        };
 
-        if input.len() < offset {
-            return Err(MessageError::BufferTooSmall);
+        let account_keys_bytes = reader.take_scoped(num_addresses * size_of::<Address>())?;
```
kskalski, line 611: we could avoid using take_scoped here with `containers::decode_into_slice_t`, right?

cpubot: The advantage of `take_scoped` here is that it does a bounds check before allocating, which is consistent with the reference implementation

kskalski: I guess this does the trick of checking the input for available data before allocating:
```rust
        // `as_trusted_for(num_addresses * size_of::<Address>())` returns
        //   exactly the requested number of bytes, or errors.
        let account_keys_reader =
            unsafe { reader.as_trusted_for(num_addresses * size_of::<Address>()) }?;
        let mut account_keys = Vec::with_capacity(num_addresses);
        wincode::containers::decode_into_slice_t::<Address, C>(
            account_keys_reader,
            &mut account_keys.spare_capacity_mut()[..num_addresses],
        )?;
        // SAFETY:
        // - `decode_into_slice_t` initilszes exactly `num_addresses` elements into slice.
        // - `account_keys` was allocated with capacity for `num_addresses` elements.
        unsafe {
            account_keys.set_len(num_addresses);
        }
```

but clearly given this code and uses of `take_scoped` below we should expose API that will create and deserialize vector of given size without reading the `len`. This is one of the implementations of reading with context that you proposed, but it still feels more natural to just add as a public function in `wincode::schema::containers::Vec` first before fitting it into a new trait.

febo: Does using `decode_into_slice_t` provides an advantage over `take_scoped`? E.g., less copies?

cpubot: The disadvantage to `take_scoped` is that it will make this `SchemaRead` implementation compatible with fewer `Reader` back-ends. Not all back-ends will be able to support scoped borrows. But, this implementation already necessitates stable borrows (which is a stricter requirement than call scoped borrows) given how it [interacts with the instruction headers](https://github.com/cpubot/solana-sdk/blob/831225303e771d81c21113115adbdfe9df840887/message/src/versions/v1/message.rs#L699-L706).

My sense is that it's safe to assume that this message payload will always be deserialized from in-memory sources, and if that is correct, it's worth using the borrowing machinery to avoid heap allocating for instruction headers.

As the above relates to `take_scoped` -- if we're using `take_borrowed`, it's safe to assume `take_scoped` is available. `decode_into_slice_t` doesn't have any performance advantage over the current implementation -- we'd just have to preallocate the `Vec` before calling it. `take_scoped` can bounds check before we allocate, which is a nice feature because we can validate the input has enough bytes before allocating.

tao-stones (dismissed):
looks good from v1 message format perspective. Not familiar enough to commit on wincode APIs.

message/src/versions/v1/message.rs
kskalski: I agree in this case we already use `take_borrowed` down below for instruction headers, so this pretty much guarantees that `take_scoped` will work fine. I wonder though, if we can figure out ways to remove `take_scoped` from API and replace it with `take_borrowed` or `as_trusted_for` - I am yet to see a real example where we can't.

IMHO the other advantage of using `decode_into_slice_t` is getting rid of the manual  `copy_nonoverlapping` and comments explaining why zero copy is fine here (`decode_into_slice_t` will automatically detect if it's fine to do raw copy or decode elements one by one).

A comment for separate discussion - this would break compatibility due to changed struct types, but seems like `Address` is zero copy and this whole collection could be zero-copy converted to `&'a [Address]`

cpubot: For the purposes of this PR, I don't want to get bogged down here on wincode semantic adjustments. I agree directionally with the points here, for the most part. I'm personally okay with the `copy_nonoverlapping` and the safety comment. It's small and isolated enough that I'm not worried about maintenance burden. Unless you feel strongly otherwise, will leave as is and merge.

PR #625, opened 2026-03-13
deprecate solana-packet

alexpyattaev:
- Solana-packet will no longer be used in agave once https://github.com/anza-xyz/agave/pull/6309 lands
- We can deprecate it from sdk already today?


packet/src/lib.rs
```diff
@@ -1,6 +1,8 @@
 //! The definition of a Solana network packet.
 #![cfg_attr(feature = "frozen-abi", feature(min_specialization))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![deprecated(since = "4.2.0", note = "Scheduled for removal")]
```
t-nelson, line 4: we need a separate deprecation warning on the `PACKET_DATA_SIZE` constant pointing to the analogue in transaction. even though both are a bug

alexpyattaev: 100% they are. will update.

PR #627, opened 2026-03-17
instruction: Add `no_std` support

febo:
### Problem

Currently the instruction crate gates all types under `"std"` feature, but it could use the `alloc` crate and be compatible with `no_std`.

### Solution

Use `alloc` to import the required types instead of `std`. It was necessary to [update](https://github.com/anza-xyz/wincode/pull/235) wincode to avoid a dependency to `alloc::sync`, which is not available in upstream BPF target.

### DIFF
--- a/instruction/Cargo.toml
+++ b/instruction/Cargo.toml
@@ -20,8 +20,9 @@ borsh = ["dep:borsh"]
 default = ["std"]
 frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro", "serde", "std"]
 serde = ["serde/alloc", "dep:serde_derive", "solana-pubkey/serde"]
-std = ["serde?/std", "borsh?/std"]
-syscalls = ["std"]
+# The `std` feature is currently unused and retained for backwards compatibility.
+std = []
+syscalls = []
 wincode = ["dep:wincode", "solana-pubkey/wincode"]
 
 [dependencies]


instruction/Cargo.toml
```diff
@@ -19,7 +19,7 @@ bincode = ["dep:bincode", "dep:serde"]
 borsh = ["dep:borsh"]
 default = ["std"]
 frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro", "serde", "std"]
-serde = ["dep:serde", "dep:serde_derive", "solana-pubkey/serde"]
+serde = ["serde/alloc", "dep:serde_derive", "solana-pubkey/serde"]
 std = ["serde?/std", "borsh?/std"]
 syscalls = ["std"]
```
febo: This does not need to enable `"std"`. Left as is otherwise it can be a breaking change.

joncinque: My attitude has typically been "if someone is depending on incorrect feature activations, it's not our fault if we break them". And I've never really gotten complaints about that.

Looking at this more, I'm now wondering if we need the `std` feature at all? Removing it would be a breaking change, but we probably don't need to enable all the std stuff, right?

febo: Yep, we don't need the `std` at all.

joncinque: So to be clear, can we remove this enabling of std and also the features enabled under std? ie do the following
```
std = []
syscalls = []
```

febo: Yes. The only thing that might change is if a downstream crate was taking advantage of serde/borsh having `"std"` enabled by default. I guess the risk is minimal.

joncinque: I'm ok with breaking those users -- they'll figure out what they need and fix their own build

febo: We could remove the `"std"` feature since it is not being used. Given that it was been enabled by default, maybe it is not going to have a big impact. Or should we be more conservative? 😅

joncinque: Let's be conservative on that part at least haha, I think a few places were adding it by hand

febo: Done in https://github.com/anza-xyz/solana-sdk/pull/627/commits/4bab2943e3f605a7caff8c04b2a4abb3349a6ebc – added a comment about it as well.

PR #629, opened 2026-03-18
epoch-rewards: remove assertion inside distribute

mjain-jump:
When fuzzing FD + Agave at the block level, ensuring that this assertion doesn't trigger is incredibly difficult. Within the fuzzer itself, it involves parsing through all vote / stake accounts / delegations, recomputing partitions for epoch rewards, calculating the total distributed amount, and parsing the epoch rewards sysvar to make sure the `total_rewards` field is properly bounded. It is a lot easier to simply remove this assertion, or handling it in a more graceful manner.

### DIFF
--- a/epoch-rewards/src/lib.rs
+++ b/epoch-rewards/src/lib.rs
@@ -60,7 +60,6 @@ pub struct EpochRewards {
 impl EpochRewards {
     pub fn distribute(&mut self, amount: u64) {
         let new_distributed_rewards = self.distributed_rewards.saturating_add(amount);
-        assert!(new_distributed_rewards <= self.total_rewards);
         self.distributed_rewards = new_distributed_rewards;
     }
 }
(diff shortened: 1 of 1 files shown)


epoch-rewards/src/lib.rs
```diff
@@ -60,7 +60,6 @@ pub struct EpochRewards {
 impl EpochRewards {
     pub fn distribute(&mut self, amount: u64) {
         let new_distributed_rewards = self.distributed_rewards.saturating_add(amount);
-        assert!(new_distributed_rewards <= self.total_rewards);
```
buffalojoec, line 63: @joncinque is this one of those "if there's ever an issue with the lamport accounting, halt the chain as a last resort rather than create an LoF scenario"?

joncinque: We were discussing this briefly -- there's only one callsite for this function, so we could hoist the checks into agave to make sure or have this return an error, but I think that would just move the panic up higher into the runtime eventually.

Would that work for the fuzzer @mjain-jump ?

PR #630, opened 2026-03-19
refactor(message): use wincode::SchemaReadContext to read vectors with known len

kskalski:
Changes:
* remove unsafe code used to deserialize `Vec<Account>`
* use new `wincode::SchemaReadContext` API for constructing vector with already read / known `len`
* add manual prealloc checks to control given len meets allocation size limits (this part is optional - current code only checks that input has appropriate number of bytes before allocating, but given low max counts involved here 256 / u16::MAX we could as well skip the checks similarly to https://github.com/anza-xyz/solana-sdk/pull/640)
* removes use of `take_scoped` wincode reader's API, which in general is supported only for some readers (not an issue strictly in this case, given that this code also uses `take_borrowed` the `take_scoped` is most likely also going to be available)

### DIFF
--- a/message/src/versions/v1/message.rs
+++ b/message/src/versions/v1/message.rs
@@ -42,11 +42,13 @@ use solana_frozen_abi_macro::AbiExample;
 #[cfg(feature = "wincode")]
 use {
     crate::v1::{InstructionHeader, FIXED_HEADER_SIZE},
-    core::{mem::MaybeUninit, ptr::copy_nonoverlapping, slice::from_raw_parts},
+    core::{mem::MaybeUninit, slice::from_raw_parts},
     wincode::{
         config::{Config, ConfigCore},
+        context,
         io::{Reader, Writer},
-        ReadResult, SchemaRead, SchemaWrite, WriteResult,
+        len::SeqLen,
+        ReadResult, SchemaRead, SchemaReadContext, SchemaWrite, WriteResult,
     },
 };
 use {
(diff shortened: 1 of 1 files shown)


message/src/versions/v1/message.rs
```diff
@@ -608,24 +610,11 @@ unsafe impl<'de, C: Config> SchemaRead<'de, C> for Message {
...
-        // - `account_keys` was allocated with capacity for `num_addresses` elements.
-        unsafe {
-            copy_nonoverlapping(
-                account_keys_bytes.as_ptr().cast::<Address>(),
-                account_keys.as_mut_ptr(),
-                num_addresses,
-            );
-            account_keys.set_len(num_addresses);
-        }
+        <C::LengthEncoding as SeqLen<C>>::prealloc_check::<Address>(num_addresses)?;
```
apfitzge, line 613: how do i know what the preallocation limit is here?

kskalski: It is a const parameter of the configuration, by default it's 4MiB (https://github.com/anza-xyz/wincode/blob/c20ec4bc05892fbe69e2dca33af03ddbddaa062e/wincode/src/config/mod.rs#L18) so in fact when not setting the limit specifically to a lower value it is way above the possible values that will come up in those counters (2^8 or 2^16) and the check will not do anything (good question if it actually gets optimized out, probably yes).

cpubot: Agree we should check if the preallocation check gets DCEd when the allocation size can never exceed the preallocation limit.

apfitzge:
> Probably a good question for @apfitzge -- do we anticipate versioned messages being deserialized from anything other than a byte slice or `Cursor`?

In agave I'd expect only from a Vec<u8> or byte slice. This is in SDK though...idk what other users will think of haha.

kskalski:
I verified that all prealloc checks are eliminated from compiled output for default config - including the addresses collection (since `Address` is 32 bytes, the byte-length permitted by 1-byte len is up to 8160, this is still skipped for default prealloc limit being 4MiB).

PR #631, opened 2026-03-19
curve25519: Add `no_std` support

LStan:
`curve25519` can be `no_std` with `alloc`.


Cargo.toml
```diff
@@ -328,7 +328,7 @@ solana-vote-interface = { path = "vote-interface", version = "5.0.0" }
 static_assertions = "1.1.0"
 strum = "0.24"
 strum_macros = "0.24"
-subtle = "2.6.1"
+subtle = {version = "2.6.1", default-features = false }
```
LStan: `subtle` is used in `curve25519` and `bls-signatures`. They both don't need default features (`std` and `i128`).

PR #632, opened 2026-03-23
borsh: no-std option

grod220:
Right now there isn't a way to opt-out of std lib features on this crate. It's currently blocking no-std efforts for crates like https://github.com/solana-program/token-metadata/pull/57.

With this, consumers can now pass `--no-default-features`.

### DIFF
--- a/borsh/Cargo.toml
+++ b/borsh/Cargo.toml
@@ -12,10 +12,6 @@ edition = { workspace = true }
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
 
-[features]
-default = ["std"]
-std = ["borsh/std"]
-
 [dependencies]
 borsh = { workspace = true }
 


borsh/Cargo.toml
```diff
@@ -12,8 +12,12 @@ edition = { workspace = true }
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
 
+[features]
+default = ["std"]
+std = ["borsh/std"]
```
joncinque: Are these features actually needed for anything? The crate isn't actually doing anything different based on this feature being enabled. If people need std on borsh enabled in their build, they can probably do it themselves, no?

grod220: True, I'd be in favor of removing it. But would that be a breaking change though? Currently, a consumer could not enable borsh/std themselves and get it through solana-borsh doing it transitively.

joncinque: I don't think it's really possible to use `solana-borsh` without also having `borsh` in some way, so if someone really needs the `std` feature of borsh, they should enable the feature themselves.

When we clean up dependencies and features, I think it's fine for downstream users to break on cases like this. If we break them, their build only worked because they explicitly depended on something incorrect / implicit on our side.

grod220: That's fair! Will update.

PR #634, opened 2026-03-23
reward-info: Add variant for deactivated stake

joncinque:
#### Problem

As mentioned during this [SIMD-0392 amendment](https://github.com/solana-foundation/solana-improvement-documents/pull/488#discussion_r2962812011), block metadata should carry the information of when a stake account is deactivating.

#### Summary of changes

Add a new `DeactivatingStake` variant of `RewardType`, to be used in https://github.com/anza-xyz/agave/pull/11332.

### DIFF
--- a/reward-info/src/lib.rs
+++ b/reward-info/src/lib.rs
@@ -14,7 +14,7 @@ pub enum RewardType {
     Rent,
     Staking,
     Voting,
-    DeactivatingStake,
+    DeactivatedStake,
 }
 
 impl fmt::Display for RewardType {
@@ -27,7 +27,7 @@ impl fmt::Display for RewardType {
                 RewardType::Rent => "rent",
                 RewardType::Staking => "staking",
                 RewardType::Voting => "voting",
-                RewardType::DeactivatingStake => "deactivating-stake",
+                RewardType::DeactivatedStake => "deactivated-stake",
             }
         )
     }


reward-info/src/lib.rs
```diff
@@ -26,6 +27,7 @@ impl fmt::Display for RewardType {
                 RewardType::Rent => "rent",
                 RewardType::Staking => "staking",
                 RewardType::Voting => "voting",
+                RewardType::DeactivatingStake => "deactivating-stake",
```
t-nelson: how do we plan to use this wrt multi-epoch deactivations?

joncinque: My thinking is that it has multiple "deactivating" entries, since it's technically still deactivating

t-nelson: i guess my original thought was that they'd get the special designation only in their last deactivation epoch. that is when active delegation goes to zero

...then we can get cute and do the same for first epoch with rewards and never have to scan the account's tx history again :wink:

joncinque: Fine by me! Either way :heavy_check_mark: please so I can implement it in Agave

t-nelson:
```suggestion
                RewardType::DeactivatedStake => "deactivated-stake",
```

change in usage, change in tense 😉

joncinque: Done!

PR #636, opened 2026-03-24
[bls-signatures] Remove `Debug` for secret key and implement `Zeroize` and `ZeroizeOnDrop`

samkim-crypto:
#### Summary of Changes

- Right now, the BLS secret key implements `Debug`, which can reveal secret key information on screen or log files. I updated the code with a custom `Debug` implementation.
- I also implemented the traits `Zeroize` and `ZeroizeOnDrop` for BLS secret and BLS keypair to zeroize memory on drop as an extra safety measure.

### DIFF
--- a/bls-signatures/src/secret_key.rs
+++ b/bls-signatures/src/secret_key.rs
@@ -11,6 +11,7 @@ use {
     core::ptr,
     ff::Field,
     rand::rngs::OsRng,
+    zeroize::{Zeroize, ZeroizeOnDrop},
 };
 #[cfg(feature = "solana-signer-derive")]
 use {solana_signature::Signature, solana_signer::Signer, subtle::ConstantTimeEq};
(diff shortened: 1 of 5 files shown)


bls-signatures/src/secret_key.rs
```diff
@@ -19,9 +20,31 @@ use {solana_signature::Signature, solana_signer::Signer, subtle::ConstantTimeEq}
...
+#[derive(Clone, Eq, PartialEq)]
 pub struct SecretKey(pub(crate) Scalar);
 
+impl core::fmt::Debug for SecretKey {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "SecretKey(<hidden>)")
+    }
+}
+
+impl Zeroize for SecretKey {
```
zz-sol, line 32: can we do 

```rust
#[derive(Zeroize)]
#[zeroize(drop)]
struct MySecret {
    key: [u8; 32],
}
```

samkim-crypto: I tried exactly this, but the `Scalar` type from the `blstrs` library does not implement `Zeroize`, so we cannot just derive it.

We can't implement `Zeroize` for `Scalar` in this crate either, because both are foreign types/traits. We can consider updating `SecretKey` to wrap `[u8; 32]`, but this is probably outside the scope of this PR.

PR #638, opened 2026-03-24
Serialize `Address` as base58 string

grod220:
Reference: https://github.com/solana-program/token-metadata/pull/57#discussion_r2974368421

Currently, serde serializes Address as a 32 integer array. This is not idiomatic as addresses have always historically been serialized as base58 strings. This PR adds custom serde impl's that replace the default derives.

Change considered a bug fix and will go out as a patch release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -26,7 +26,7 @@ dependencies = [
  "cfg-if",
  "once_cell",
  "version_check",
- "zerocopy",
+ "zerocopy 0.7.35",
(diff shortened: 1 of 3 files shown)


grod220:
Misunderstood the history on this, thought the original serialization was via base58 on Pubkey, but in fact it's always been the plain derive: https://github.com/solana-labs/solana-program-library/pull/3291. Not worth breaking.

PR #639, opened 2026-03-24
nullable: serde_with support

grod220:
Inspired from: https://github.com/solana-program/token-metadata/pull/57#discussion_r2974368421

This lets downstream consumers customize how the inner `T` serializes without hand-rolling a per-field serde module. For example, `MaybeNull<Address>` can serialize the `Some` case as a base58 string and `None` as `null` via:

  ```rust
  #[serde(with = "serde_with::As::<Option<DisplayFromStr>>")]
  pub new_authority: MaybeNull<Address>
```

### DIFF
--- a/nullable/src/serde_with.rs
+++ b/nullable/src/serde_with.rs
@@ -4,10 +4,32 @@
 //! that any `serde_with` adapter can be composed with `MaybeNull<T>` using the
 //! `Option<Strategy>` pattern.
 //!
-//! # Usage
+//! ```rust
+//! # pub mod solana_address {
+//! #     #[derive(PartialEq, serde_derive::Serialize, serde_derive::Deserialize)]
+//! #     pub struct Address([u8; 32]);
+//! #     impl solana_nullable::Nullable for Address {
+//! #         const NONE: Self = Address([0u8; 32]);
+//! #     }
+//! #     impl core::str::FromStr for Address {
+//! #         type Err = String;
+//! #         fn from_str(s: &str) -> Result<Self, Self::Err> {
+//! #             Ok(Address([0u8; 32]))
+//! #         }
+//! #     }
+//! #     impl core::fmt::Display for Address {
+//! #         fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+//! #             Ok(())
+//! #         }
+//! #     }
+//! # }
 //!
-//! ```ignore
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: nullable/src/serde_with.rs [hunk 1]
```diff
@@ -0,0 +1,309 @@
+//! [`serde_with`] integration for [`MaybeNull`].
+//!
+//! Provides blanket [`SerializeAs`] and [`DeserializeAs`] implementations so
+//! that any `serde_with` adapter can be composed with `MaybeNull<T>` using the
+//! `Option<Strategy>` pattern.
+//!
+//! # Usage
+//!
+//! ```ignore
```
joncinque: This example is really good! I took the liberty of making it work:

```
//! ```rust
//! # pub mod solana_address {
//! #     #[derive(PartialEq, serde_derive::Serialize, serde_derive::Deserialize)]
//! #     pub struct Address([u8; 32]);
//! #     impl solana_nullable::Nullable for Address {
//! #         const NONE: Self = Address([0u8; 32]);
//! #     }
//! #     impl core::str::FromStr for Address {
//! #         type Err = String;
//! #         fn from_str(s: &str) -> Result<Self, Self::Err> {
//! #             Ok(Address([0u8; 32]))
//! #         }
//! #     }
//! #     impl core::fmt::Display for Address {
//! #         fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
//! #             Ok(())
//! #         }
//! #     }
//! # }
//! use serde_derive::{Deserialize, Serialize};
//! use serde_with::{serde_as, DisplayFromStr};
//! use solana_address::Address;
//! use solana_nullable::MaybeNull;
//!
//! #[serde_as]
//! #[derive(Serialize, Deserialize)]
//! struct MyStruct {
//!     #[serde_as(as = "Option<DisplayFromStr>")]
//!     pub field: MaybeNull<Address>,
//! }
//!
//! // Or without the proc macro:
//! #[derive(Serialize, Deserialize)]
//! struct MyStruct2 {
//!     #[serde(with = "serde_with::As::<Option<DisplayFromStr>>")]
//!     pub field: MaybeNull<Address>,
//! }
//! ```
```

joncinque (commented):
The code looks great! That wrapper type is really cool. Let's just make the doc comment actually work, since I think it'll be useful to people

PR #640, opened 2026-03-24
chore(wincode): remove `peek_byte` usage from `VersionedMessage`

cpubot:
`peek` and `consume` `Reader` methods are being removed in wincode `v0.5`.

To accommodate, `v0.4.9` introduces a `SchemaReadContext` trait, which can be used to create `SchemaRead` implementations that can use pre-fetched (i.e., already read) input.

This PR removes usage of `peek_byte` from the `SchemaRead` implementation of `VersionedMessage` and `VersionedTransaction`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3434,6 +3434,7 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.18",
  "wincode",
+ "zeroize",
 ]
 
 [[package]]
@@ -4261,7 +4262,7 @@ dependencies = [
 
 [[package]]
 name = "solana-reward-info"
-version = "5.0.0"
+version = "6.0.0"
 dependencies = [
  "serde",
  "serde_derive",
@@ -5808,9 +5809,9 @@ dependencies = [
 
 [[package]]
 name = "zeroize"
-version = "1.8.1"
+version = "1.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ced3678a2879b30306d323f4542626697a464a97c0a07c9aebf7ebca65cd4dde"
+checksum = "b97154e67e32c85465826e8bcc1c59429aaaf107c1e4a9e53c8d8ccd5eff88d0"
 dependencies = [
  "zeroize_derive",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -295,7 +295,7 @@ solana-program-option = { path = "program-option", version = "3.0.0" }
 solana-program-pack = { path = "program-pack", version = "3.0.0" }
 solana-pubkey = { path = "pubkey", version = "4.1.0", default-features = false }
 solana-rent = { path = "rent", version = "4.0.0", default-features = false }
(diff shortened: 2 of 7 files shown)


transaction/src/versioned/mod.rs
```diff
@@ -322,41 +321,35 @@ unsafe impl<'de, C: Config> SchemaRead<'de, C> for VersionedTransaction {
...
-            unsafe {
-                let signatures_ptr = signatures.as_mut_ptr();
-                copy_nonoverlapping(
-                    bytes.as_ptr() as *const Signature,
-                    signatures_ptr,
-                    expected_signatures_len,
-                );
-                signatures.set_len(expected_signatures_len);
-            }
+            let num_signatures = message.header().num_required_signatures as usize;
```
kskalski, line 343: should we do the pre-alloc check here?

cpubot: I think we're fine without. `Signature` is 64 bytes, `num_required_signatures` is one byte. Theoretical max allocation is `16320` bytes, which is well below our default preallocation limit

transaction/src/versioned/mod.rs
```diff
@@ -322,41 +321,35 @@ unsafe impl<'de, C: Config> SchemaRead<'de, C> for VersionedTransaction {
...
+            let signatures = <Vec<Signature> as SchemaReadContext<C, _>>::get_with_context(
+                context::Len(num_signatures),
+                reader.by_ref(),
```
kskalski: drop `.by_ref()` here

transaction/src/versioned/mod.rs
```diff
@@ -301,18 +301,17 @@ unsafe impl<'de, C: Config> SchemaRead<'de, C> for VersionedTransaction {
...
-
-            // SAFETY: The transaction is fully initialized at this point since
-            // both `signatures` and `message` have been read.
-            let message = unsafe { builder.uninit_message_mut().assume_init_ref() };
+            let signatures = <Vec<Signature> as SchemaReadContext<C, _>>::get_with_context(
+                // Here `discriminator < 0x80`, so it is a canonical one-byte `ShortU16`.
+                context::Len(discriminator as usize),
+                reader.by_ref(),
+            )?;
+            let message = <VersionedMessage as SchemaRead<C>>::get(reader.by_ref())?;
```
kskalski: drop `.by_ref()` here

transaction/src/versioned/mod.rs
cpubot: Fixed

transaction/src/versioned/mod.rs
cpubot: Fixed

transaction/src/versioned/mod.rs
cpubot: I could however revert back to `take_scoped` if we deem this sensitive. Better generally to avoid the allocation if we know the input bytes can't support it. Trade-off is `Reader` compatibility. Unclear if we need to support `Reader`s other than byte slices / `Cursor`s here. But, 16320 bytes is small enough that it's likely a non-issue.

PR #643, opened 2026-03-26
feat(wincode-varint): add leb128 varint helper for wincode schemas

kskalski:
In order to support wincode (de)serialization for structs like
```rust
    struct LockoutOffset {
        #[serde(with = "serde_varint")]
        offset: Slot,
        confirmation_count: u8,
    }
```
the LEB128 (encoding used by `serde_varint`) schema for encoding integers using wincode is necessary.

This PR adds a separate crate `solana-wincode-varint` with implementation of `Leb128Int` that provides this functionality, e.g.
```rust
    struct LockoutOffset {
        #[wincode(with = "Leb128Int<Slot>")]
        offset: Slot,
        confirmation_count: u8,
    }
```

### DIFF
--- a/wincode-varint/Cargo.toml
+++ b/wincode-varint/Cargo.toml
@@ -8,7 +8,7 @@ repository = { workspace = true }
 homepage = { workspace = true }
 license = { workspace = true }
 edition = { workspace = true }
-rust-version = "1.85.0"
+rust-version = { workspace = true }
 
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]


wincode-varint/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+        unsafe impl<C: ConfigCore> SchemaWrite<C> for Leb128Int<$type> {
+            type Src = $type;
+
+            fn size_of(src: &$type) -> WriteResult<usize> {
+                let bits = <$type>::BITS - src.leading_zeros();
+                Ok(bits.div_ceil(7).max(1) as usize)
+            }
+
+            fn write(mut writer: impl Writer, src: &$type) -> WriteResult<()> {
+                let mut value = *src;
```
cpubot, line 79: Since we can reasonably efficiently compute the final length, it may be worth considering `as_trusted_for` here.

Probably doesn't help much on a smaller integer type like `u16`, since we only save at most 2 bounds checks. But on something like a `u64`, we may see some benefit, as we can save at most 9 bounds checks. Probably worth an experiment.

serde-varint/Cargo.toml
```diff
@@ -12,11 +12,16 @@ edition = { workspace = true }
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
 
+[features]
+wincode = ["dep:wincode"]
+
 [dependencies]
 serde = { workspace = true }
+wincode = { workspace = true, features = ["alloc"], optional = true }
```
cpubot: Do we need `alloc` outside of `dev-dependencies`?

wincode-varint/src/lib.rs
kskalski: I added bench, which provides some insights:
* current impl seems to be the fastest, but concrete numbers can differ a lot on how exactly we bench the functions
* `as_trusted_for` is slower in all cases, except on u64 random when it gets same / comparable perf
* I also tried variations of looping `0..len` either using `as_trusted_for` writer or stack created uninit array (+single write of initialized len bytes at the end), but they are slower than baseline, often by much
* bench loop assumptions influence perf further -> e.g. removing `black_box` from `&mut buf` speeds up `small` cases by 50%

serde-varint/Cargo.toml
kskalski: removed

kskalski:
> Is it perhaps a bit awkward to put this into the `serde-varint` crate? I wonder if it makes sense to rename this crate to something more general like `solana-varint` or `solana-leb128`. It seems 

I think I will duck the rename for now. The addition here supports very close use case to current serde attributes and are tied closely to their migration. Moreover we implement the same byte format as serde+bincode. The "serde" name over-use seems acceptable to me - the name leaked even into `wincode` itself, where I suppose it relates to "serialization - deserialization", not the `serde` crate / APIs, so let's pretend we relate to the same here and maybe do some wider cleanup in the future.

> reasonable to mark the `serde` feature optional -- if I'm _only_ using `wincode` functionality, it would be nice to be able to avoid pulling in `serde`, but `serde` being optional with the crate name being `serde-varint` feels awkward.

Ok, moved serde to sub-mod controlled by (default-enabled) feature

wincode-varint/src/lib.rs
cpubot: Cool, thanks for checking 🙂

cpubot (dismissed):
LGTM

wincode-varint/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+name = "solana-wincode-varint"
+description = "Wincode schemas for LEB128 variable-length integer encoding"
+documentation = "https://docs.rs/solana-wincode-varint"
+version = "1.0.0"
+authors = { workspace = true }
+repository = { workspace = true }
+homepage = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+rust-version = "1.85.0"
```
joncinque: Wanna just make this 1.89 from the start? That way we won't have a merge issue later 😄

kskalski: I think I will aim for `{ workspace = true }` here once https://github.com/anza-xyz/solana-sdk/pull/648 is in, which I would like to merge first.

kskalski:
> Looks great! I've reserved the name and added anza-team as an owner

So right now this crate on crates.io is not published (version 0.0.0) - the publish action failed with
```
[3](https://github.com/anza-xyz/solana-sdk/actions/runs/24131449475/job/70409211616#step:8:54)

Caused by:
  the remote server responded with an error (status 403 Forbidden): this crate exists but you don't seem to be an owner. If you believe this is a mistake, perhaps you need to accept an invitation to be an owner before publishing.
```

Also, which options should be used in action such that we publish at version 1.0.0 without any bump? Or this needs to be done manually for the first itme?

kskalski:
Ah, I guess anza-team didn't accept the ownership transfer if you sent it already, since you are still an owner.

joncinque:
I've sent another invite for anza-team to take over the crate

> Also, which options should be used in action such that we publish at version 1.0.0 without any bump? Or this needs to be done manually for the first itme?

The easiest thing is to commit the crate with some v0 (like v0.1.0), and then do a major release

kskalski:
Hm, it won't let me, is it trying to find the previous (non-existent 0.1.0) tag or something?
```
Run ${GITHUB_ACTION_PATH}/run.sh --config=scripts/cliff.toml wincode-varint@v0.1.0..HEAD --include-path "wincode-varint/**" --github-repo anza-xyz/solana-sdk
Error: SetCommitRangeError("wincode-varint@v0.1.0..HEAD", Error { code: -3, klass: 4, message: "revspec 'wincode-varint@v0.1.0' not found" })
Error: SetCommitRangeError("wincode-varint@v0.1.0..HEAD", Error { code: -3, klass: 4, message: "revspec 'wincode-varint@v0.1.0' not found" })
stat: cannot statx 'TEMP_CHANGELOG.md': No such file or directory
/home/runner/work/_actions/orhun/git-cliff-action/v4/run.sh: line 71: [: : integer expression expected
Error: Process completed with exit code 1.
```
(https://github.com/anza-xyz/solana-sdk/actions/runs/24148080566/job/70467394811)

kskalski:
Well, the crate got published and Cargo.toml updated, so I suppose it is some final step of generating release, which I don't mind missing...

joncinque:
> Well, the crate got published and Cargo.toml updated, so I suppose it is some final step of generating release, which I don't mind missing...

Yeah that part isn't such a big deal, so we can consider the release a success! I created a small release by hand, feel free to edit as you wish: https://github.com/anza-xyz/solana-sdk/releases/tag/wincode-varint%40v1.0.0

PR #644, opened 2026-03-26
Adds AddressHasherBuilder::with_offset()

brooksprumo:
In agave's accounts index, we'd like to use AddressHasherBuilder as the mechanism to handle mapping from pubkeys to index bins[^1]. One missing feature in AddressHasherBuilder is visibility and opt-in determinism.

Specifically, when agave generates its accounts index, we'd like to know the value of `AddressHasherBuilder::offset`. We would then use this value as a seed to reproduce state, in case we need to debug/test/investigate anything later.

This PR adds two two pieces.

1. A method to return the builder's `offset` value.
2. A method to construct a builder with a specific offset.

[^1]: https://github.com/anza-xyz/agave/pull/11528

### DIFF
--- a/address/src/hasher.rs
+++ b/address/src/hasher.rs
@@ -54,6 +54,20 @@ pub struct AddressHasherBuilder {
     offset: usize,
 }
 
+impl AddressHasherBuilder {
+    /// Constructs a builder with a specific offset.
+    ///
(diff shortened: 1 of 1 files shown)


brooksprumo:
@joncinque @febo Added y'all as reviewers based on authors of recent commits to this code. Please do add/remove reviewers as necessary. Thanks!

PR #646, opened 2026-03-27
instruction: Remove unused frozen-abi dependencies

LStan:
#### Problem
`frozen-abi` logic was removed from `solana-instruction` in #12, but the dependencies and feature were not removed from `Cargo.toml`.

#### Summary of changes
Remove them.

### DIFF
--- a/instruction/src/lib.rs
+++ b/instruction/src/lib.rs
@@ -170,7 +170,7 @@ impl Instruction {
     /// ```
     /// # use solana_pubkey::Pubkey;
     /// # use solana_instruction::{AccountMeta, Instruction};
-    /// # use serde::{Serialize, Deserialize};
+    /// # use serde_derive::{Serialize, Deserialize};
     /// #
     /// #[derive(Serialize, Deserialize)]
     /// pub struct MyInstruction {


instruction/Cargo.toml
```diff
@@ -18,7 +18,6 @@ rustdoc-args = ["--cfg=docsrs"]
 bincode = ["dep:bincode", "dep:serde"]
 borsh = ["dep:borsh"]
 default = ["std"]
-frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro", "serde", "std"]
```
joncinque, line 21: Can we keep the feature? Otherwise this is a breaking change, can add a comment like `# Remove this feature with the next breaking release`

kskalski:
I tried to publish minor for `instruction`, but it actually marked the crate as requiring major bump due to:
```
  Summary semver requires new major version: 1 major and 0 minor checks failed
    Finished [  23.951s] solana-instruction
--- failure feature_no_longer_enables_feature: package feature no longer enables another feature ---
```

Do you think we should adjsut the default features until we are ready to do major release, do such release now or use some other work-around?

joncinque:
Eh, I would say to just skip the semver checks for that one -- it only affects the frozen-abi feature, which is not meant for downstream usage outside of agave

PR #647, opened 2026-03-28
chore: Bump wincode to 0.5.0

cpubot:


kskalski (approved):

kskalski:
Looks like the minimum rust version checked by CI is still 1.81, due to 
`minimum_version=$(readCargoVariable rust-version "$base/../program/Cargo.toml")` in `scripts/rust-version.sh`

joncinque:
Just coming in with a fly-by check on this.

Considering usage of wincode in sdk types is probably still limited to agave, this shouldn't be too destructive, but will require 18 re-publishes. I'm seeing:

```
account
address
bls-signatures
fee-calculator
hash
instruction-error
instruction
message
nonce-account
nonce
nullable
reward-info
sdk-wasm-js
signature
system-interface
transaction-error
transaction
zero-copy
```

I took a glance at the commits between 0.4.9 and 0.5.0, and from my limited view, there aren't any breaking changes in the traits implemented by these types, but rather in the reader API.

Has there been any consideration in splitting these two parts into separate crates? That way we can avoid unnecessary breaking changes in the future when this usage is more widespread.

Feel free to completely disregard if this doesn't work in the design of wincode. I don't know the code nearly as well as you do!

joncinque:
Sorry, that was a silly suggestion -- the SchemaRead / SchemaWrite traits rely on the Reader and Writer traits (of course), so there's no real way to split them. Never mind!

kskalski:
> this shouldn't be too destructive, but will require 18 re-publishes. I'm seeing:

I played a bit with setting agave to use git version and it quickly escalates to needing >30 SDK crates to be updated plus problem with some other external crates. Is there an established way to go over the dependency graph to publish new version of crates and updating them in chunks?

I suppose due to MSVR update and wincode breaking dependency change most (all?) SDK crates will need to go to a new major release, right?

joncinque:
When we added wincode support in the sdk, to avoid having to do a breaking release of the sdk crates whenever wincode bumped to a breaking release, we decided to treat it as "experimental", meaning "this can break at any time, beware". Most likely, only Agave will be affected since it's the primary user of wincode.

Once wincode is more widely used though, we'll need to avoid breaking so often, and likely implement multiple versions of the wincode traits at the same time.

Doing a full breaking release of the whole sdk is extremely time-consuming, so I don't want to do it more than once per year, roughly.

So let's just do minor releases for all of the crates that explicitly use wincode. @cpubot does that sound correct based on previous conversations?

kskalski:
So if I understand correctly we will just do minor versions for sdk crates, since 
* all typical users are on toolchain 1.89, we don't treat MSVR as breaking
* wincode usually (always?) is an optional feature and when it's disabled we don't observe incompatibilities

For agave we will need to wait for all crates where we enabled wincode to get a new release and upgrade them in one go, possibly with a few other crates if any hard of logical dependency arise.

joncinque:
Yep that's all correct!

PR #649, opened 2026-03-30
feat(address-lookup-table-interface): support wincode

kskalski:
Add `wincode` feature.

Note: when both `bincode` and `wincode` features are enabled, the `wincode` based code is used

### DIFF
--- a/address-lookup-table-interface/src/instruction.rs
+++ b/address-lookup-table-interface/src/instruction.rs
@@ -1,13 +1,20 @@
 #[cfg(feature = "serde")]
 use serde_derive::{Deserialize, Serialize};
 use {solana_clock::Slot, solana_pubkey::Pubkey, solana_sdk_ids::address_lookup_table::id};
-#[cfg(feature = "bincode")]
+#[cfg(all(not(feature = "wincode"), feature = "bincode"))]
 use {
     solana_instruction::{AccountMeta, Instruction},
     solana_sdk_ids::system_program,
 };
+#[cfg(feature = "wincode")]
+use {
+    solana_instruction::{AccountMeta, Instruction},
+    solana_sdk_ids::system_program,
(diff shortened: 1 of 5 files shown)


address-lookup-table-interface/src/instruction.rs
```diff
@@ -125,7 +132,7 @@ pub fn create_lookup_table(
 /// Constructs an instruction that freezes an address lookup
 /// table so that it can never be closed or extended again. Empty
 /// lookup tables cannot be frozen.
-#[cfg(feature = "bincode")]
+#[cfg(all(not(feature = "wincode"), feature = "bincode"))]
```
apfitzge: is this the prefered style in sdk? Seems weird to me to have the fn definitions in separate places instead of just having interior impl blocks.

address-lookup-table-interface/src/instruction.rs
```diff
@@ -113,7 +120,7 @@ fn create_lookup_table_common(
 /// This instruction doesn't require the authority to be a signer but
 /// until v1.12 the address lookup table program still requires the
 /// authority to sign the transaction.
-#[cfg(feature = "bincode")]
+#[cfg(all(not(feature = "wincode"), feature = "bincode"))]
 pub fn create_lookup_table(
```
apfitzge: afaict this fn is exactly the same

address-lookup-table-interface/src/instruction.rs
kskalski: true, I guess it might be better to do `or` at the fn definition and just gate impl specific calls in the body, will update tomorrow

address-lookup-table-interface/src/instruction.rs
kskalski: updated the code to avoid duplicating functions

PR #650, opened 2026-03-30
instruction-view: Add initialize helper

febo:
### Problem

Currently, the only way to create a `CpiAccount` is using the `CpiAccount::from` conversion. In most situations, this does not lead to an optimal bytecode generation, which results in more CU consumption.

### Solution

Add an `init_from_account_view` constructor that performs an in-place initialization of a `CpiAccount`. It also improves how the account flags (signer, writable and executable) are copied over. Instead of copying each value individual, all 3 values are copied in a single operation.

cc: @nagisa @LucasSte

### DIFF
--- a/instruction-view/src/cpi.rs
+++ b/instruction-view/src/cpi.rs
@@ -79,6 +79,7 @@ pub struct CpiAccount<'account> {
 
 // Make sure the layout of `CpiAccount` and `RuntimeAccount` are compatible for the fields
 // that are copied over as a single value in `CpiAccount::init_from_account_view`.
+#[allow(clippy::arithmetic_side_effects)]
 const _: () = {
     const RUNTIME_SIGNER_OFFSET: usize = offset_of!(RuntimeAccount, is_signer);
     const CPI_SIGNER_OFFSET: usize = offset_of!(CpiAccount<'static>, is_signer);


instruction-view/src/cpi.rs
```diff
@@ -53,39 +59,86 @@ pub struct CpiAccount<'a> {
...
+    assert!(
+        offset_of!(RuntimeAccount, executable) - RUNTIME_SIGNER_OFFSET
+            == offset_of!(CpiAccount<'static>, executable) - CPI_SIGNER_OFFSET
+    );
+
+    assert!(
+        offset_of!(RuntimeAccount, padding) - RUNTIME_SIGNER_OFFSET
+            == offset_of!(CpiAccount<'static>, _padding) - CPI_SIGNER_OFFSET
+    );
+};
```
LucasSte, line 101: Because of the padding at the end, you could also include a const assert for the alignment.

`assert_eq!(align_of::<RuntimeAccount>(), align_of::<CpiAccount>)`

Or, you could assert the size of the entire structs matches. According to [Rust's size_of](https://doc.rust-lang.org/std/mem/fn.size_of.html) function,
> More specifically, this is the offset in bytes between successive elements in an array with that item type including alignment padding.

`assert_eq!(size_of::<RuntimeAccount>(), size_of::<CpiAccount>)`
`assert_eq!(align_of::<RuntimeAccount>(), 56)`.

febo: > `assert_eq!(size_of::<RuntimeAccount>(), size_of::<CpiAccount>)` `assert_eq!(align_of::<RuntimeAccount>(), 56)`.

I think you mean:
```rust
assert_eq!(align_of::<RuntimeAccount>(), align_of::<CpiAccount>);
assert_eq!(size_of::<RuntimeAccount>(), 56);
```
But the size is actually `52` bytes – in memory it is `56`. We could add another `4` bytes of padding to match it, but is that needed?

LucasSte: If both structures have the same members and same alignment, they should have the same size too no?

The `size_of` function returns 56, not 52:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=930d675fa9d7962dbf3b6d7f4efc7adc

febo: > If both structures have the same members and same alignment, they should have the same size too no?

Nope, they are different. One follows the layout of the program input and the other the layout for `sol_invoke_signed_c`.

> The `size_of` function returns 56, not 52:
> 
> https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=930d675fa9d7962dbf3b6d7f4efc7adc

Yeah, that is the size in memory, but the sum of the fields is `52`. I wonder whether adding a size assert will give the wrong impression or not – e.g., when you add another `u32` field at the end, the assert will still be valid.

Since we are not strictly dependent on alignment or size, I will keep the asserts as they are. The important part is to ensure that those 4 fields are contiguous in memory in both structs.

PR #654, opened 2026-04-01
account: remove loader v4 from PROGRAM_OWNERS

buffalojoec:
#### Problem
As per [RFC-0423](https://github.com/solana-foundation/solana-improvement-documents/discussions/423), Loader V4 will be redesigned and reintroduced as an on-chain program. The builtin implementation should be removed.

#### Summary of Changes
Remove Loader V4's ID from `PROGRAM_OWNERS`, which is used in Agave to key on program account ownership.

### DIFF
--- a/account/src/lib.rs
+++ b/account/src/lib.rs
@@ -15,7 +15,7 @@ use {
     solana_clock::{Epoch, INITIAL_RENT_EPOCH},
     solana_instruction_error::LamportsError,
     solana_pubkey::Pubkey,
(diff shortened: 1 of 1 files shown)


buffalojoec:
@joncinque if you could please publish `solana-account` with this change that would be great. I'll use the new version in another PR to remove Loader V4 codepaths from `solana-svm`

joncinque (approved):
Let's make this change with a minor release -- it's not breaking per se, but we'll need to be careful about the upgrade in agave

PR #657, opened 2026-04-01
clock: Deprecate `NUM_CONSECUTIVE_LEADER_SLOTS` constant

vadorovsky:
It was moved to solana-leader-schedule in https://github.com/anza-xyz/agave/pull/11397. There is no use of the constant in the SDK repository.


clock/src/lib.rs
```diff
@@ -70,6 +70,7 @@ static_assertions::const_assert_eq!(DEFAULT_SLOTS_PER_EPOCH, 432_000);
 pub const DEFAULT_SLOTS_PER_EPOCH: u64 = 2 * TICKS_PER_DAY / DEFAULT_TICKS_PER_SLOT;
 
 // leader schedule is governed by this
+#[deprecated(since = "3.1.0", note = "Moved to solana-leader-schedule crate")]
```
vadorovsky, line 73: I'm assuming the next minor version of this crate will be 3.1.0, but lmk if I should change it to something else.

PR #675, opened 2026-04-10
remove loader-v4-interface crate

buffalojoec:
#### Problem
As per [RFC-0423](https://github.com/solana-foundation/solana-improvement-documents/discussions/423), Loader V4 will be redesigned and reintroduced as an on-chain program. The builtin implementation should be removed.

#### Summary of Changes
Remove the Loader V4 interface from the SDK. We have no idea if this will be the same interface for the on-chain version, and there's no sense maintaining it until we've had a chance to redesign it.


joncinque (approved):
Makes sense to me! We do still have ownership of the crate name and can reintroduce it in the future if we want, or go with a whole other name

PR #678, opened 2026-04-12
fix: Add sha2 feature where needed

abelmarnk:
This PR adds the sha2 feature to some crates for which it was not previously enabled.

### DIFF
--- a/nonce/Cargo.toml
+++ b/nonce/Cargo.toml
@@ -42,5 +42,6 @@ wincode = { workspace = true, optional = true, features = ["alloc"] }
 
 [dev-dependencies]
 bincode = { workspace = true }
+solana-sha256-hasher = { workspace = true, features = ["sha2"] }
 solana-nonce = { path = ".", features = ["dev-context-only-utils", "wincode"] }
 solana-pubkey = { workspace = true, features = ["std"], default-features = false }
--- a/nonce/Cargo.toml
+++ b/nonce/Cargo.toml
@@ -42,6 +42,6 @@ wincode = { workspace = true, optional = true, features = ["alloc"] }
 
 [dev-dependencies]
 bincode = { workspace = true }
-solana-sha256-hasher = { workspace = true, features = ["sha2"] }
 solana-nonce = { path = ".", features = ["dev-context-only-utils", "wincode"] }
 solana-pubkey = { workspace = true, features = ["std"], default-features = false }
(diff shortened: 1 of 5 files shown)


nonce/Cargo.toml
```diff
@@ -37,7 +37,7 @@ serde_derive = { workspace = true, optional = true }
 solana-fee-calculator = { workspace = true }
 solana-hash = { workspace = true, features = ["copy", "decode"] }
 solana-pubkey = { workspace = true, default-features = false }
-solana-sha256-hasher = { workspace = true }
+solana-sha256-hasher = { workspace = true, features = ["sha2"] }
```
joncinque: I'm not sure about this change -- `solana-nonce` can be used in on-chain programs, so it shouldn't always enable the sha2 feature. Can you move the sha2 feature enabling to a dev-dependency?

message/src/legacy.rs
joncinque, line 1: Can you make the solana-message changes in a different PR?

joncinque (commented):
Thanks for your contribution! Most of these make sense, and please move the wincode-related to changes to a separate PR

message/src/legacy.rs
abelmarnk: Sure

nonce/Cargo.toml
abelmarnk: Thanks for the quick feedback, I've reverted it.
I originally left it there because the sha2 feature appears to be target-gated, so I assumed it was safe for on-chain use. My concern was that if solana-nonce is also used in off-chain builds, the same test failure would still surface unless another dependency happens to pull in the sha2 feature.

nonce/Cargo.toml
```diff
@@ -42,5 +42,6 @@ wincode = { workspace = true, optional = true, features = ["alloc"] }
 
 [dev-dependencies]
 bincode = { workspace = true }
+solana-sha256-hasher = { workspace = true, features = ["sha2"] }
```
joncinque: Ah sorry, there's a sorting issue here, this needs to go last in the list

joncinque (commented):
Just one last thing

nonce/Cargo.toml
abelmarnk: thanks, resolved

joncinque (approved):
Thanks again!

PR #679, opened 2026-04-13
sha512 syscall

deanmlittle:
This PR lays the groundwork for the implementation of the upcoming [SHA512 syscall](https://github.com/solana-foundation/solana-improvement-documents/pull/512). It introduces two new crates to the solana-sdk repo:

1. `solana-hash-512` - a 64-byte variant of `solana-hash` that duplicates all existing functionality, but references 64-byte specific functions in serialization methods such as `five8`, and with a max base58 string length of 88 instead of 44. It shares the existing `HashError` enum type with `solana-hash` so as to not create a new error enum just for this use case, as the enum variants and corresponding error messages are identical.
2. `solana-sha512-hasher` - a duplicate of the existing `solana-sha256-hasher` API, but for SHA512. Returns a `Hash512` instead of a `Hash`.

Along with these changes, we also introduce the `sol_sha512` syscall in `define-syscalls`, as well as handling the hasher reexport and all necessary APIs for use with `register_feature_gated_function` in Agave.

### DIFF
--- a/hash-512/Cargo.toml
+++ b/hash-512/Cargo.toml
@@ -1,0 +1,49 @@
+[package]
+name = "solana-hash-512"
+description = "Solana wrapper for the 64-byte output of the SHA-512 hashing algorithm."
+documentation = "https://docs.rs/solana-hash-512"
+version = "0.0.0"
+rust-version = { workspace = true }
+authors = { workspace = true }
+repository = { workspace = true }
+homepage = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu", "wasm32-unknown-unknown"]
+all-features = true
+rustdoc-args = ["--cfg=docsrs"]
+
+[features]
+atomic = ["dep:solana-atomic-u64"]
+borsh = ["dep:borsh"]
+bytemuck = ["copy", "dep:bytemuck", "dep:bytemuck_derive"]
+copy = []
+decode = ["dep:five8", "solana-hash/decode"]
+default = []
+frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro", "std"]
+serde = ["dep:serde", "dep:serde_derive", "dep:serde-big-array"]
+std = ["borsh?/std", "serde?/std", "wincode?/std"]
+wincode = ["dep:wincode"]
+
+[dependencies]
+borsh = { workspace = true, optional = true }
+bytemuck = { workspace = true, optional = true }
+bytemuck_derive = { workspace = true, optional = true }
+five8 = { workspace = true, optional = true }
+serde = { workspace = true, optional = true }
+serde-big-array = { workspace = true, optional = true }
+serde_derive = { workspace = true, optional = true }
+solana-atomic-u64 = { workspace = true, optional = true }
+solana-frozen-abi = { workspace = true, optional = true, features = ["frozen-abi"] }
+solana-frozen-abi-macro = { workspace = true, optional = true, features = ["frozen-abi"] }
+solana-hash = { workspace = true, optional = true }
+wincode = { workspace = true, optional = true }
+
+[dev-dependencies]
+bs58 = { workspace = true, default-features = false, features = ["alloc"] }
+solana-hash-512 = { path = ".", features = ["atomic", "copy", "decode"] }
+
+[lints]
+workspace = true
--- a/hash-512/src/lib.rs
+++ b/hash-512/src/lib.rs
@@ -1,0 +1,190 @@
+#![no_std]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(feature = "frozen-abi", feature(min_specialization))]
+#[cfg(feature = "borsh")]
+use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
+#[cfg(feature = "std")]
+extern crate std;
+#[cfg(feature = "bytemuck")]
+use bytemuck_derive::{Pod, Zeroable};
+#[cfg(feature = "decode")]
+use core::{
+    fmt,
+    str::{from_utf8_unchecked, FromStr},
+};
+#[cfg(feature = "serde")]
+use {
+    serde_big_array::BigArray,
+    serde_derive::{Deserialize, Serialize},
+};
+#[cfg(feature = "borsh")]
+extern crate alloc;
+#[cfg(feature = "borsh")]
+use alloc::string::ToString;
+#[cfg(feature = "decode")]
+pub use solana_hash::ParseHashError;
+#[cfg(feature = "wincode")]
+use wincode::{SchemaRead, SchemaWrite};
+
+/// Size of a hash in bytes.
+pub const HASH_BYTES: usize = 64;
+/// Maximum string length of a base58 encoded 64-byte hash.
+pub const MAX_BASE58_LEN: usize = 88;
+
+/// A hash; the 64-byte output of the SHA-512 hashing algorithm.
+#[cfg_attr(feature = "frozen-abi", derive(solana_frozen_abi_macro::AbiExample))]
+#[cfg_attr(
+    feature = "borsh",
+    derive(BorshSerialize, BorshDeserialize),
+    borsh(crate = "borsh")
+)]
+#[cfg_attr(feature = "borsh", derive(BorshSchema))]
+#[cfg_attr(feature = "bytemuck", derive(Pod, Zeroable))]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "wincode", derive(SchemaWrite, SchemaRead))]
+#[cfg_attr(feature = "copy", derive(Copy))]
+#[cfg_attr(not(feature = "decode"), derive(Debug))]
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+#[repr(transparent)]
+pub struct Hash512(
+    #[cfg_attr(feature = "serde", serde(with = "BigArray"))] pub(crate) [u8; HASH_BYTES],
+);
+
+impl Default for Hash512 {
+    fn default() -> Self {
+        Self([0u8; HASH_BYTES])
+    }
+}
+
+impl From<[u8; HASH_BYTES]> for Hash512 {
+    fn from(from: [u8; HASH_BYTES]) -> Self {
+        Self(from)
+    }
+}
+
+impl AsRef<[u8]> for Hash512 {
+    fn as_ref(&self) -> &[u8] {
+        &self.0[..]
+    }
+}
+
+#[cfg(feature = "decode")]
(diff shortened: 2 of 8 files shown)


deanmlittle:
Crates have already been reserved and assigned to `anza-team` with version 0.0.0:

https://crates.io/crates/solana-hash-512
https://crates.io/crates/solana-sha512-hasher

Provided we don't need to make any drastic changes, we can just bump the version, pull 0.0.0 and release at whatever version is deemed reasonable to match the other hashers.

hash-512/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+bytemuck = { workspace = true, optional = true }
+bytemuck_derive = { workspace = true, optional = true }
+five8 = { workspace = true, optional = true }
+serde = { workspace = true, optional = true }
+serde-big-array = { workspace = true, optional = true }
+serde_derive = { workspace = true, optional = true }
+solana-atomic-u64 = { workspace = true, optional = true }
+solana-frozen-abi = { workspace = true, optional = true, features = ["frozen-abi"] }
+solana-frozen-abi-macro = { workspace = true, optional = true, features = ["frozen-abi"] }
+solana-hash = { workspace = true, features = ["decode"] }
```
samkim-crypto: Is hardcoding the `decode` feature here intended? It seems like the `[features]` block properly manages the conditional activation of the feature, but I might be missing something.

hash-512/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_new_unique() {
+        assert!(Hash512::new_unique() != Hash512::new_unique());
```
samkim-crypto: Nit: use `assert_ne!` to make it more canonical?

samkim-crypto (commented):
Looks good to me! Just a question and a nit. 

I wonder if we can just duplicate the error type in `solana-hash-512` since it does seem like importing an entirely separate crate just for the error enum seems like an anti-pattern. Either way is fine with me though.

deanmlittle:
> I wonder if we can just duplicate the error type in `solana-hash-512` since it does seem like importing an entirely separate crate just for the error enum seems like an anti-pattern. Either way is fine with me though.

I also have no strong opinions here. The most "solana-sdk" way is to invent a third crate just for the error type which I think we can all agree would be seriously overkill 😅 .

I don't think it's worth duplicating. Imo, it's totally fine to import the other crate, as they each have an identical dependency tree. If nobody expresses a strong opinion to the contrary, I'm just going to leave this as is.

hash-512/Cargo.toml
deanmlittle: Nope. Good catch!

deanmlittle: I think I left this in before I defined this:
`decode = ["dep:five8", "solana-hash/decode"]`

Will remove and squash.

hash-512/src/lib.rs
deanmlittle: Fixed ce67ac5463f65636b5eb47c90364a6ef0b3c9e2a

hash-512/Cargo.toml
deanmlittle: Fixed ce67ac5463f65636b5eb47c90364a6ef0b3c9e2a

hash-512/src/lib.rs
```diff
@@ -0,0 +1,195 @@
+#![no_std]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(feature = "frozen-abi", feature(min_specialization))]
+#[cfg(feature = "borsh")]
+use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
+#[cfg(feature = "std")]
+extern crate std;
```
joncinque, line 7: do we need `std` or this feature at all?

hash-512/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+#[cfg(feature = "bytemuck")]
+use bytemuck_derive::{Pod, Zeroable};
+#[cfg(feature = "decode")]
+use core::{
+    fmt,
+    str::{from_utf8_unchecked, FromStr},
+};
+#[cfg(feature = "sanitize")]
+use solana_sanitize::Sanitize;
```
joncinque: Do we need sanitize for this? I'm guessing not, so let's just remove it

hash-512/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+bytemuck = { workspace = true, optional = true }
+bytemuck_derive = { workspace = true, optional = true }
+five8 = { workspace = true, optional = true }
+serde = { workspace = true, optional = true }
+serde-big-array = { workspace = true, optional = true }
+serde_derive = { workspace = true, optional = true }
+solana-atomic-u64 = { workspace = true, optional = true }
+solana-frozen-abi = { workspace = true, optional = true, features = ["frozen-abi"] }
+solana-frozen-abi-macro = { workspace = true, optional = true, features = ["frozen-abi"] }
+solana-hash = { workspace = true }
```
joncinque: It looks like we can make this totally optional, and only added with the `decode` feature

sha512-hasher/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+version = "0.0.0"
+authors = { workspace = true }
+repository = { workspace = true }
+homepage = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
```
joncinque, line 16: To make sure the docs build works properly
```suggestion
[package.metadata.docs.rs]
targets = ["x86_64-unknown-linux-gnu"]
all-features = true
rustdoc-args = ["--cfg=docsrs"]
```

joncinque (commented):
Looks great overall! Just some small stuff on my side.

Also, can you add the two new crates to `scripts/patch-crates-functions.sh`?

Note to self: we should probably have `solana-signature` re-export this type in a breaking change, since there are a ton of similarities between the two.

hash-512/src/lib.rs
deanmlittle: Yay! Removing code!

hash-512/src/lib.rs
deanmlittle: I was simply copying the API for Sha256 in every case you see here. If you don't want it, I'm happy to delete *as much* of this crate as possible!

joncinque: Haha I figured, yes please minimize things

deanmlittle: Answer: Yes, you do (probably). Removing it kills `frozen-abi` and stops you from propagating std to consumers like borsh, etc.

sha512-hasher/Cargo.toml
deanmlittle: addressed 6d286b1

hash-512/Cargo.toml
deanmlittle: done 6d286b1

hash-512/src/lib.rs
deanmlittle: done 6d286b1

hash-512/src/lib.rs
joncinque: oh right, I forgot about the frozen-abi part, sorry -- can you just get gate this on `feature = "frozen-abi"`?

deanmlittle:
> scripts/patch-crates-functions.sh

Done! [13b28aa](https://github.com/anza-xyz/solana-sdk/pull/679/commits/13b28aa86472deefa393e76c46648156e7bbb042)

hash-512/src/lib.rs
deanmlittle: Well, it already does this by virtue of:

It's just a little more granular. I don't think that's so bad? I feel like for the purpose of this PR, we should probably keep it this way to match the `solana-hash` API. If we want to change this up, we should change both in a separate PR for consistency. Wdyt?

joncinque: Eh, I would just do the new code right, but I can also put in a follow-up, no need to block

joncinque (commented):
Looks good to me! I'll let @samkim-crypto give the final approval and merge though

joncinque:
https://github.com/anza-xyz/solana-sdk/pull/682 for the future work on solana-hash, I'll do the same on hash-512 after this merges

hash-512/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+bytemuck_derive = { workspace = true, optional = true }
+five8 = { workspace = true, optional = true }
+serde = { workspace = true, optional = true }
+serde-big-array = { workspace = true, optional = true }
+serde_derive = { workspace = true, optional = true }
+solana-atomic-u64 = { workspace = true, optional = true }
+solana-frozen-abi = { workspace = true, optional = true, features = ["frozen-abi"] }
+solana-frozen-abi-macro = { workspace = true, optional = true, features = ["frozen-abi"] }
+solana-hash = { workspace = true, optional = true }
+solana-sanitize = { workspace = true, optional = true }
```
zz-sol: nit: seems like a dead dependency?

zz-sol (dismissed):
LGTM

hash-512/Cargo.toml
samkim-crypto: Yeah it seems like `solana-sanitize` was not removed in [6d286b1](https://github.com/anza-xyz/solana-sdk/pull/679/commits/6d286b10768495b9401e362b05faef42c7d8a8dc). I'll approve merge once this is removed.

deanmlittle: good catch. fixed and squashed here: https://github.com/blueshift-gg/solana-sdk/commit/5700cf1372eea5bcb4bf10dab8578c27d8b52275

samkim-crypto (approved):

deanmlittle:
@yihau would it be possible for `anza-team` to yank `0.0.0` from crates.io after this gets published?

yihau:
yes, I can do that. feel free to ping me when the first version has been published.

joncinque:
@yihau we're good to go! I accidentally published solana-hash-512 before #684 landed, so please yank v0.0.0 and v1.0.0 of solana-hash-512 and v0.0.0 of solana-sha512-hasher

@deanmlittle , feel free to use v1.1.0 of solana-hash-512 and v1.0.0 of solana-sha512-hasher

deanmlittle:
Perfect. Thanks heaps! @joncinque 🙏

deanmlittle:
Update. We still need to:
1. bump `solana_define_syscall`, as it adds `sol_sha512` to definitions.
2. update `solana_sha512_hasher` to consume the new version
3. bump `solana_sha512_hasher`

joncinque:
Done, this should be good to go with solana-define-syscall v5.1.0 and solana-sha512-hasher v1.0.1

@yihau can you also yank v1.0.0 of solana-sha512-hasher?

yihau:
done!

PR #681, opened 2026-04-14
small update: update message test

abelmarnk:
This PR updates some parts of the `solana-messsage` tests to use use `wincode` instead of `bincode`.

### DIFF
--- a/message/src/legacy.rs
+++ b/message/src/legacy.rs
@@ -764,10 +764,11 @@ mod tests {
         assert!(!message.is_instruction_account(2));
     }
 
+    #[cfg(feature = "wincode")]
     #[test]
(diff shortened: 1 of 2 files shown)


abelmarnk:
@joncinque would it make sense to remove [this](https://github.com/anza-xyz/solana-sdk/blob/master/message/src/inline_nonce.rs#L60) as well? so the `bincode` dev-dependency can be removed from the crate.

joncinque:
Oops sorry, I just noticed this, you don't want to include the change anymore?

abelmarnk:
😅 I was thinking it wasn't correct or something.

joncinque:
@febo what do you think about this change?

PR #682, opened 2026-04-14
hash: Gate std on frozen-abi

joncinque:
#### Problem

The hash crate has an std feature, but it's totally unnecessary.

#### Summary of changes

Keep the std feature for now, but make it no-op. This is technically breaking, but only impacts users that expect `solana-hash/std` to enable std on other crates. We've typically made that kind of breaking change in a minor version.


febo (approved):
Looks great!

PR #684, opened 2026-04-15
hash-512: Remove std feature

joncinque:
#### Problem

The std feature in hash-512 was never needed.

#### Summary of changes

Similar to #682, but for the hash-512 crate.


joncinque:
Rebased for audit failure

febo (approved):
Looks great!

PR #685, opened 2026-04-15
frozen-abi: allow for api_digest, abi_digest in frozen_abi macro separately

puhtaytow:
### Description

Currently, the `frozen_abi`  proc macro requires providing an `api_digest`,  while this behavior is correct in many cases, not all scenarios require strict API compatibility checks—sometimes stable serialization alone is sufficient.

### Summary of changes
- Turned `api_digest` attribute in `frozen_abi` proc macro, as optional, enforcing providing at least one digest, either `api_digest` or `abi_digest`
- Updated docs
- Added test with `StableAbi` only

### DIFF
--- a/frozen-abi-macro/src/lib.rs
+++ b/frozen-abi-macro/src/lib.rs
@@ -311,37 +311,41 @@ pub fn derive_abi_enum_visitor(item: TokenStream) -> TokenStream {
 fn quote_for_test(
     test_mod_ident: &Ident,
     type_name: &Ident,
-    expected_api_digest: &str,
+    expected_api_digest: Option<&str>,
     expected_abi_digest: Option<&str>,
     abi_serializer: AbiSerializer,
 ) -> TokenStream2 {
-    let test_api = quote! {
-            #[test]
-            fn test_api_digest() {
-                use ::solana_frozen_abi::abi_example::{AbiExample, AbiEnumVisitor};
+    let test_api = if expected_api_digest.is_some() {
+        quote! {
+                #[test]
+                fn test_api_digest() {
+                    use ::solana_frozen_abi::abi_example::{AbiExample, AbiEnumVisitor};
 
-                let mut digester = ::solana_frozen_abi::abi_digester::AbiDigester::create();
-                let example = <#type_name>::example();
-                let result = <_>::visit_for_abi(&&example, &mut digester);
-                let mut hash = digester.finalize();
-                if result.is_err() {
-                    ::std::eprintln!("Error: digest error: {:#?}", result);
-                }
-                result.unwrap();
-                let actual_digest = ::std::format!("{}", hash);
-                if ::std::env::var("SOLANA_ABI_BULK_UPDATE").is_ok() {
-                    if #expected_api_digest != actual_digest {
-                        ::std::eprintln!("sed -i -e 's/{}/{}/g' $(git grep --files-with-matches frozen_abi)", #expected_api_digest, hash);
+                    let mut digester = ::solana_frozen_abi::abi_digester::AbiDigester::create();
+                    let example = <#type_name>::example();
+                    let result = <_>::visit_for_abi(&&example, &mut digester);
+                    let mut hash = digester.finalize();
+                    if result.is_err() {
+                        ::std::eprintln!("Error: digest error: {:#?}", result);
                     }
-                    ::std::eprintln!("Warning: Not testing the abi digest under SOLANA_ABI_BULK_UPDATE!");
-                } else {
-                    if let Ok(dir) = ::std::env::var("SOLANA_ABI_DUMP_DIR") {
-                        assert_eq!(#expected_api_digest, actual_digest, "Possibly API changed? Examine the diff in SOLANA_ABI_DUMP_DIR!: \n$ diff -u {}/*{}* {}/*{}*", dir, #expected_api_digest, dir, actual_digest);
+                    result.unwrap();
+                    let actual_digest = ::std::format!("{}", hash);
+                    if ::std::env::var("SOLANA_ABI_BULK_UPDATE").is_ok() {
+                        if #expected_api_digest != actual_digest {
+                            ::std::eprintln!("sed -i -e 's/{}/{}/g' $(git grep --files-with-matches frozen_abi)", #expected_api_digest, hash);
+                        }
+                        ::std::eprintln!("Warning: Not testing the abi digest under SOLANA_ABI_BULK_UPDATE!");
                     } else {
-                        assert_eq!(#expected_api_digest, actual_digest, "Possibly API changed? Confirm the diff by rerunning before and after this test failed with SOLANA_ABI_DUMP_DIR!");
+                        if let Ok(dir) = ::std::env::var("SOLANA_ABI_DUMP_DIR") {
+                            assert_eq!(#expected_api_digest, actual_digest, "Possibly API changed? Examine the diff in SOLANA_ABI_DUMP_DIR!: \n$ diff -u {}/*{}* {}/*{}*", dir, #expected_api_digest, dir, actual_digest);
+                        } else {
+                            assert_eq!(#expected_api_digest, actual_digest, "Possibly API changed? Confirm the diff by rerunning before and after this test failed with SOLANA_ABI_DUMP_DIR!");
+                        }
                     }
                 }
-            }
+        }
+    } else {
+        TokenStream2::new()
     };
 
     let abi_serialize_expr = match abi_serializer {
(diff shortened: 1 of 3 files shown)


puhtaytow:
@kskalski This one for you :pray:  It would allow removing  `serde::Serialize` from types where we don't need to check api compatibility, but only care about serialization. In such cases the new proc macro `stable_serialize` should be used instead of `frozen_abi`.

frozen-abi/src/lib.rs
```diff
@@ -1,29 +1,46 @@
 //! # StableAbi
 //!
-//! The `StableAbi` is an optional extension to `frozen-abi` that provides functionality for
-//! detecting unintended encoding changes. It is designed to be used in conjunction with the
-//! mandatory `AbiExample`.
+//! The `StableAbi` provides functionality for detecting unintended encoding changes.
+//!
+//! There are two supported workflows:
+//! - `#[frozen_abi(...)]`: API digest + optional ABI digest checks
+//! - `#[stable_serialize(...)]`: ABI digest-only checks for types that only need stable
```
kskalski: shouldn't we just make `api_digest` optional? this way we would actually arrive in the semantically correct and consistent usage of `#[frozen_abi(abi_digest = "..")]`, which checks ABI not API :)

puhtaytow: we thinking similar, but Jon is right in this case https://github.com/anza-xyz/solana-sdk/pull/443#discussion_r2550274112 :)

```
What's the motivation for removing these checks? With this PR as it is, someone can use frozen_abi without specifying either an abi_digest or a digest, which would be dangerous if they're expecting something to be tested.
```

puhtaytow: imo frozen-abi proc macro should be renamed to api_compatibility or something that makes more sense. The StableAbi stuff should be removed from it, kept in stable_serialize, AbiExample has to be ported to use wincode, and serializer selection attr also to trash as it wont be needed after migration from bincode; but these are breaking changes and demand full switch to wincode in sdk and agave crates.. so probably 5.0

kskalski: you mean this comment
> What's the motivation for removing these checks? With this PR as it is, someone can use `frozen_abi` without specifying either an `abi_digest` or a `digest`, which would be dangerous if they're expecting something to be tested.
> 
> If that makes sense to you, let's keep the existing check that `api_expected_digest` is set, and we can eventually do a breaking change that also requires `abi_expected_digest` to be set.
?

I think it just says to not allow *not setting any digest*, I think if we have a check that requires *at least one* digest to be set, we should be fine.

kskalski: @joncinque let us know if it makes sense to allow:
* `#[frozen_abi(digest = "..")]` (deprecated)
* `#[frozen_abi(api_digest = "..")]`
* `#[frozen_abi(abi_digest = "..")]`
* `#[frozen_abi(abi_digest = "..", api_digest = "..")]`

But not allow `#[frozen_abi]`

joncinque: Yep that makes sense to me!

As a naming nit, I think the new macro name should start with `frozen_abi_` for consistency -- maybe `frozen_abi_stable_serialize`?

kskalski: I'm in favor of reusing old macros and names, but adding options with _abi / _api suffix :)

joncinque: `frozen_abi_abi` and `frozen_abi_api`? or `frozen_abi` and `frozen_api`? I'd lean toward the second one :smile:

puhtaytow: Gentlemen. I feel like we have a bit of confusion here or.. im being total batator (sorry if thats the case).

What we currently have is [`frozen_abi`](https://github.com/anza-xyz/solana-sdk/blob/master/frozen-abi-macro/src/lib.rs#L507) proc macro, which [requires](https://github.com/anza-xyz/solana-sdk/blob/master/frozen-abi-macro/src/lib.rs#L536-L543) providing [`api_digest`](https://github.com/anza-xyz/solana-sdk/blob/master/frozen-abi-macro/src/lib.rs#L513) in every usage, and has as optional `abi_digest`.

What Kamil wants, is to remove the requirement to provide the `api_digest`, so `frozen_abi` proc macro could be used for testing only serialization without the canonical representation tests in cases which dont demand api compatibility checks. 

The `abi_digest` is used by `StableAbi` in this [test](https://github.com/anza-xyz/solana-sdk/blob/master/frozen-abi-macro/src/lib.rs#L358-L372), where `api_digest` is used by `AbiExample` in this [one](https://github.com/anza-xyz/solana-sdk/blob/master/frozen-abi-macro/src/lib.rs#L319-L344) 

I prepared the PR with separate proc macro, so we could keep `frozen_abi` proc macro requiring the `api_digest` as it is currently, but give option to use `StableAbi` only. 

So, we have to make decision, do we add new proc macro `frozen_abi_stable_serialize` or whatever we decide that makes sense in terms of naming, or we lift the requirement from the existing one `frozen_abi`, to provide the `api_digest`.

--edit--

The latter is the smallest change and isn't breaking, but if somebody remove accidentally `api_digest`, from existing usage/code then, `test_api_digest` for this type won't run and we wont know about it.

kskalski: Having `frozen_abi(digest | abi_digest)` + `frozen_api(digest | api_digest)` would make sense in the ideal world. However, since currently there is only `frozen_abi`, which checks `api`, and there is no `frozen_api` at all and there is no way to actually check `abi-only`, we would need a couple of deprecations and migration steps to land in that ideal world.

What I was suggesting above is to only use existing `frozen_abi` with *at least one* option `digest | api_digest | abi_digest`. That seems like a good first step.

Later we could also add a clone macro `frozen_api` and at some point we could deprecate using `abi_digest` in `frozen_api` and `api_digest` in `frozen_abi`. :)

The existence of `AbiExample` vs `StableAbi` is a separate issue that is confusing (primarily because of naming mismatch), but that is a separate story to solve.

puhtaytow: Agreed :100: let me fix the PR then.

kskalski:
LGTM

frozen-abi-macro/src/lib.rs
```diff
@@ -311,37 +311,41 @@ pub fn derive_abi_enum_visitor(item: TokenStream) -> TokenStream {
...
-                let mut hash = digester.finalize();
-                if result.is_err() {
-                    ::std::eprintln!("Error: digest error: {:#?}", result);
-                }
-                result.unwrap();
-                let actual_digest = ::std::format!("{}", hash);
-                if ::std::env::var("SOLANA_ABI_BULK_UPDATE").is_ok() {
-                    if #expected_api_digest != actual_digest {
-                        ::std::eprintln!("sed -i -e 's/{}/{}/g' $(git grep --files-with-matches frozen_abi)", #expected_api_digest, hash);
+    let test_api = if expected_api_digest.is_some() {
```
joncinque: Sorry for the silly question, but since this is NOT doing `if let Some(expected_api_digest) = expected_api_digest {`, is the `Option<&str>` expanded to the wrapped string?

joncinque (commented):
Looks good to me, just a question on the `Option<...>` unwrapping

frozen-abi-macro/src/lib.rs
puhtaytow: Thats indeed a good question as it might look strange, because in tests there is no explicit unwrapping, but in fact it does happen under the hood.

https://docs.rs/quote/latest/src/quote/to_tokens.rs.html#114-120

So, if `X_digest.is_some()` evaluates to `true`, then in `quote!` responsible for the test generation, we would get expanded the inner value as you wrote.

puhtaytow: It might as well look a bit wasteful, since in `frozen_abi` we check the `Option` variant, so why not to expand it there and pass the inner.. and the reason for that is, it would make harder to decide whether to generate a specific test at the `quote_for_test` level.

If you think it could be done better then please let me know :pray:

puhtaytow: Otherwise lets :shipit:

kskalski: I was actually scratching my head about that too, but figured that `#expected_api_digest` must be splicing `expr` and `Some(expr)` in the same way.
So maybe for readability and avoiding surprises it will be better to use 
`if let Some(expected_api_digest) = .. { } else { }`

kskalski (commented):
LGTM

frozen-abi-macro/src/lib.rs
puhtaytow: Makes sense. Less confusion, less explaining later.

joncinque (approved):
Looks great, thanks!

PR #686, opened 2026-04-16
account: remove PROGRAM_OWNERS

buffalojoec:
#### Problem
This constant is sketchy, since it has consensus implications and one of the programs previously contained in it will be removed. It's also only used in two places in the same crate in Agave, so it can just be inlined.

#### Summary of Changes
Remove `PROGRAM_OWNERS`.

### DIFF
--- a/account/src/lib.rs
+++ b/account/src/lib.rs
@@ -15,7 +15,6 @@ use {
     solana_clock::{Epoch, INITIAL_RENT_EPOCH},
     solana_instruction_error::LamportsError,
     solana_pubkey::Pubkey,
-    solana_sdk_ids::{bpf_loader, bpf_loader_deprecated, bpf_loader_upgradeable, loader_v4},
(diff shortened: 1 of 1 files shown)


joncinque:
Let's mark it deprecated and do a minor release first, and that way we know to fixup all usages

buffalojoec:
@joncinque I landed both https://github.com/anza-xyz/solana-sdk/pull/690 and https://github.com/anza-xyz/agave/pull/11996, so this should be good to go!

Which version should we publish at?

joncinque:
We can publish a breaking release if we want to land this, but to avoid unnecessary downstream annoyance, let's use the semver trick on v4, and re-export everything from v5, and inline `PROGRAM_OWNERS` in a new v4 release.

buffalojoec:
I think I'd rather just leave it deprecated for now and let a major get cut when we actually need some more significant breakage. The deprecation warning should be enough for Agave, and it's already been removed in the validator.

PR #688, opened 2026-04-16
Revert "account: remove loader v4 from PROGRAM_OWNERS"

kskalski:
Reverts anza-xyz/solana-sdk#654

Context: https://github.com/anza-xyz/agave/pull/11798#pullrequestreview-4117842549


buffalojoec:
We might also want to yank 4.2.0 after publishing this change as 4.2.1, but I do not have perms to do that.

buffalojoec:
@kskalski we still want to land this. Reopening.

buffalojoec (approved):

PR #690, opened 2026-04-16
account: deprecate PROGRAM_OWNERS

buffalojoec:
#### Problem
This constant is sketchy, since it has consensus implications and one of the programs previously contained in it will be removed. It's also only used in two places in the same crate in Agave, so it can just be inlined.

#### Summary of Changes
Deprecate `PROGRAM_OWNERS`.

### DIFF
--- a/account/src/lib.rs
+++ b/account/src/lib.rs
@@ -693,7 +693,7 @@ pub fn create_is_signer_account_infos<'a>(
 }
 
 /// Replacement for the executable flag: An account being owned by one of these contains a program.
-#[deprecated(since = "5.0.0", note = "no longer available as a constant")]
+#[deprecated(since = "4.3.0", note = "no longer available as a constant")]
 pub const PROGRAM_OWNERS: &[Pubkey] = &[
     bpf_loader_upgradeable::id(),
     bpf_loader::id(),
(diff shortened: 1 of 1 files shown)


account/src/lib.rs
```diff
@@ -693,6 +693,7 @@ pub fn create_is_signer_account_infos<'a>(
 }
 
 /// Replacement for the executable flag: An account being owned by one of these contains a program.
+#[deprecated(since = "5.0.0", note = "no longer available as a constant")]
```
joncinque: Next version with the deprecation warning will probably be 4.3.0 -- do you want to add back loader v4 while we're at it?

buffalojoec: Ah crap I always get these attributes wrong.

buffalojoec: > Next version with the deprecation warning will probably be 4.3.0 -- do you want to add back loader v4 while we're at it?

Same PR or separate?

joncinque: May as well keep it separate so we can audit better

buffalojoec: Then how about we merge the revert first? https://github.com/anza-xyz/solana-sdk/pull/688
I can rebase onto that.

joncinque: order shouldn't matter -- let's just land them both!

joncinque (approved):
Looks great!

PR #692, opened 2026-04-16
program-log-macro: Use block expression

febo:
### Problem

Currently the `log!` macro expansion behaves differently depending on whether there are placeholders in the message or not. This is problematic since it might lead to an invalid expansion when the macro is used in a `match` arm expression, for example.

### Solution

Update the macro expansion to always return a block.


joncinque (approved):
Makes sense!

PR #693, opened 2026-04-17
program-log: Allow custom crate name in macro expansion

febo:
### Problem

Currently the log macros hardcode the `solana-program-log` crate name, which prevents their use when a different dependency name is used.

### Solution

This PR makes 2 main changes:
* the `log!` macro is wrapped in a `macro_rules` and the `$crate` value is forwarded to the proc_macro.
* the `log_cu_usage` proc_macro_attribute  now has an optional `crate` path attribute, which allows the caller to specify the log crate name.

### DIFF
--- a/program-log-macro/Cargo.toml
+++ b/program-log-macro/Cargo.toml
@@ -19,6 +19,7 @@ rustdoc-args = ["--cfg=docsrs"]
 proc-macro = true
 
 [dependencies]
+proc-macro-crate = "3"
 quote = { workspace = true }
 regex = { workspace = true }
 syn = { workspace = true, features = ["extra-traits", "full"] }
--- a/program-log-macro/Cargo.toml
+++ b/program-log-macro/Cargo.toml
@@ -19,7 +19,6 @@ rustdoc-args = ["--cfg=docsrs"]
 proc-macro = true
 
 [dependencies]
-proc-macro-crate = "3"
 quote = { workspace = true }
 regex = { workspace = true }
 syn = { workspace = true, features = ["extra-traits", "full"] }
--- a/program-log-macro/src/lib.rs
+++ b/program-log-macro/src/lib.rs
@@ -7,7 +7,6 @@ extern crate alloc;
 use {
     alloc::{format, string::ToString, vec::Vec},
     proc_macro::TokenStream,
-    proc_macro_crate::{crate_name, FoundCrate},
     quote::quote,
     regex::Regex,
     syn::{
@@ -23,7 +22,7 @@ const DEFAULT_BUFFER_SIZE: &str = "200";
 
 /// The default name of the `solana-program-log` package to search for when
 /// discovering the crate path.
-const PROGRAM_LOG_PACKAGE_NAME: &str = "solana-program-log";
+const PROGRAM_LOG_PACKAGE_NAME: &str = "::solana_program_log";
 
 /// Represents the input arguments to the `log!` macro.
 struct LogArgs {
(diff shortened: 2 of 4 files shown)


program-log-macro/src/lib.rs
```diff
@@ -81,13 +160,15 @@ impl Parse for LogArgs {
 ///
 /// # Arguments
 ///
+/// - `crate_path`: The path to the crate where the `Logger` struct is defined. This is a required argument.
```
joncinque: Maybe I'm misunderstanding how this is supposed to be used, but isn't this a little ugly?

Does someone need to write: `solana_program_log_macro::log!("solana-program-log", "{}", ...)` every time?

Can this be done with a different macro -- kind of like the difference between `println!` and `write!`? So we could have the simple `log!` and also `log_with_crate!` for those that need it.

program-log-macro/Cargo.toml
```diff
@@ -19,6 +19,7 @@ rustdoc-args = ["--cfg=docsrs"]
 proc-macro = true
 
 [dependencies]
+proc-macro-crate = "3"
```
joncinque: This dep depends on `toml_edit`, which is not no-std -- is that ok for the final build to be no-std?

joncinque (commented):
Just a couple of questions, the CU logging one is definitely good with me

program-log-macro/Cargo.toml
febo: You are right! Even `proc-macro-crate` is no no-std. I removed this dependency and instead of trying to "guess" the crate name, just used `solana_program_log`. This should work in most cases, so we probably don't need anything too fancy.

program-log-macro/src/lib.rs
febo: > Does someone need to write: solana_program_log_macro::log!("solana-program-log", "{}", ...) every time?

The idea is that this macro gets wrapped in a `macro_rules` so then `$crate` is forwarded. `solana-program-log` is doing this now.

We could make the crate name optional, but I guess in most cases the macro is used through `solana-program-log`.

febo:
@joncinque Following up on this, do you think we need to make further changes? For the `proc_macro_attribute`, there is not a very good solution. We either use a default value for the crate if one is not specified or we can use `toml_parser` to try to "discover" the crate.

program-log-macro/src/lib.rs
```diff
@@ -65,14 +76,54 @@ impl Parse for LogArgs {
...
+
+            return Ok(Self { crate_path });
+        }
+
+        // Support for standalone path.
+        let crate_path = input.parse::<Path>()?;
+
+        if !input.is_empty() {
+            return Err(input.error("unexpected tokens after crate path"));
+        }
```
joncinque, line 118: nit: can we just combine these branches?
```suggestion
        // Support for `crate = <PATH>`.
        if input.peek(Token![crate]) {
            input.parse::<Token![crate]>()?;
            input.parse::<Token![=]>()?;
        }

        // Support for standalone path.
        let crate_path = input.parse::<Path>()?;

        if !input.is_empty() {
            return Err(input.error("unexpected tokens after crate path"));
        }
```

joncinque (commented):
Sorry for the slow review, this one fell off the radar. Looks great to me! Just a nit

joncinque (dismissed):
Thanks!

febo:
Sorry @joncinque, made one more change. Decided to make the crate path optional on the `log` proc macro as well, so when we publish the new version it is not a breaking change.

joncinque (approved):
Looks great!

PR #694, opened 2026-04-20
[BLS] add Ethereum/BLS test vectors

zz-sol:
This PR use a python script to fetch and regenerate the BLS tests vectors, put them in `target/` repo, and run rust tests against the test vectors.

### DIFF
--- a/bls-signatures/tests/ethereum_vectors.rs
+++ b/bls-signatures/tests/ethereum_vectors.rs
@@ -1,10 +1,6 @@
 use {
-    blstrs::{Bls12, G1Affine, G1Projective, G2Affine, G2Prepared, G2Projective, Gt, Scalar},
-    group::{prime::PrimeCurveAffine, Group},
-    pairing::{MillerLoopResult, MultiMillerLoop},
     serde_json::Value,
     solana_bls_signatures::{
-        hash::HASH_TO_POINT_DST,
         pubkey::{PopVerified, PubkeyAffineUnchecked, PubkeyCompressed, VerifySignature},
         signature::SignatureAffine,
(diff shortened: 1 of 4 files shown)


bls-signatures/tests/ethereum_vectors.rs
```diff
@@ -0,0 +1,396 @@
...
+            verified,
+            case["output"].as_bool().unwrap(),
+            "fixture {}",
+            path.display()
+        );
+    }
+}
+
+#[test]
+fn ethereum_batch_verify_vectors() {
```
samkim-crypto, line 238: In this test, it seems like we are not testing our current API in the `solana-bls-signatures` crate, but rather manually verifying with the miller loop. Is this intended?

samkim-crypto (commented):
Awesome! Looks good to me, but just one question below.

bls-signatures/tests/ethereum_vectors.rs
zz-sol: yes. this is intensional. Ethereum's batch verification uses RLCs and our API don't support RLC. So to make it almost identical, I had to implement scalar sampling (the scalars are not random though). 

Or do you think we should just use our own API and not worry about RLCs?

samkim-crypto: Okay, my vote would be to use our own API. Maybe we can just test the positive cases? Our batch verify should work for these. For the negative cases, maybe we can skip with comments like:

```
// Our API intentionally omits RLCs and relies on upstream Proofs of Possession.
if case_id == "test_case_name" {
    continue; 
}
```

What do you think?

samkim-crypto (approved):
LGTM!

PR #696, opened 2026-04-21
loader-v3-interface: return default on non-bool trailing bytes

buffalojoec:
#### Problem
#695 updated the `OptionalTrailingBool` schema to reject non-bool arguments, but it's actually easier to work in a consensus environment if we instead treat them as `DEFAULT`.

#### Summary of Changes
Adjust the schema so that non-bool values and the omission of trailing bytes both result in `DEFAULT`.


buffalojoec:
Nevermind I don't like this.

PR #697, opened 2026-04-22
[BLS] fix bls pop with pubkey padding

zz-sol:
```
When payload is present, proof generation in proof_of_possession and verification in
verify_proof_of_possession hash only payload but do not necessarily bind the claimed public key
into the hash input. This weakens the proof from a proof of possession for a specific key into a signature
over a shared message and reintroduces rogue key attacks.
```

In this PR, we address this attack by always pad the payload with the public key.


samkim-crypto (approved):
Looks good to me!

PR #698, opened 2026-04-22
[ed25519] batch verify api

zz-sol:
### Add batch API in SDK. 

Ed25519-dalek has a batch verification API which gives performance improvement. This is not utilized in sdk nor agave. 
This PR exposes that API and allows agave to use batch verification.

### Performance comparison

Benched on MBP M4max

```
cargo bench -p solana-signature --bench verify --features parallel
```

| number of sig | ind | batch | par_ind | par_batch |
|---:|---:|---:|---:|---:|
| 1 | 21.690 µs | 21.815 µs | 21.821 µs | 21.892 µs |
| 2 | 43.897 µs | 43.952 µs | 43.289 µs | 43.835 µs |
| 4 | 91.366 µs | 90.910 µs | 73.452 µs | 72.734 µs |
| 8 | 183.65 µs | 185.20 µs | 77.689 µs | 78.645 µs |
| 16 | 369.58 µs | 184.03 µs | 141.46 µs | 142.42 µs |
| 32 | 738.36 µs | 353.61 µs | 154.93 µs | 232.56 µs |
| 64 | 1480.0 µs | 695.53 µs | 231.35 µs | 369.07 µs |
| 128 | 2951.2 µs | 1355.3 µs | 343.10 µs | 534.64 µs |
| 256 | 5905.4 µs | 2513.1 µs | 599.47 µs | 528.08 µs |
| 512 | 11858 µs | 4760.7 µs | 1106.8 µs | 838.47 µs |


signature/benches/verify.rs
```diff
@@ -0,0 +1,107 @@
...
+struct BenchData {
+    messages: Vec<Vec<u8>>,
+    signatures: Vec<Signature>,
+    pubkeys: Vec<[u8; 32]>,
+}
+
+fn create_bench_data(size: usize) -> BenchData {
+    let signing_keys: Vec<_> = (0..size)
+        .map(|i| {
+            let byte = u8::try_from(i.checked_add(1).unwrap()).unwrap();
```
samkim-crypto: Are we sure that this doesn't error on batch size >= 256? Seems like 257 or 513 will not fit into u8?

signature/Cargo.toml
```diff
@@ -20,18 +20,20 @@ default = ["std", "alloc"]
 alloc = ["wincode?/alloc"]
 bytemuck = ["dep:bytemuck", "dep:bytemuck_derive"]
 frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro", "std"]
+parallel = ["std", "verify", "dep:rayon"]
 rand = ["dep:rand"]
 serde = ["dep:serde", "dep:serde_derive", "dep:serde-big-array"]
 std = ["alloc", "serde?/std", "wincode?/std"]
-verify = ["dep:ed25519-dalek"]
+verify = ["alloc", "dep:ed25519-dalek"]
```
samkim-crypto: What do you think about creating a separate `batch-verify` feature? Adding `alloc` to `verify` is a breaking change, which is not an issue necessarily if we do a major release, but not all downstream might want to perform batch verification (and will be forced to require a heap allocator).

signature/src/lib.rs
```diff
@@ -81,6 +83,147 @@ impl Signature {
...
+        let mut message_bytes = small_messages[..len]
+            .iter()
+            .map(|message| message.unwrap())
+            .collect::<Vec<_>>();
+
+        signatures.push(next_signature.0);
+        pubkey_bytes.push(next_signature.1);
+        message_bytes.push(next_signature.2);
+
+        for ((signature, pubkey), message) in signature_iter.zip(pubkey_iter).zip(message_iter) {
```
samkim-crypto: Is it possible for `zip` to silently end prematurely if the length of the signature, public key, and message iterators have differing lengths? Should we have some custom logic to ensure these iterators have the same lengths?

signature/src/lib.rs
```diff
@@ -81,6 +83,147 @@ impl Signature {
...
     }
+
+    pub fn batch_verify<'a>(
+        signatures: impl Iterator<Item = &'a Signature>,
+        pubkey_bytes: impl Iterator<Item = &'a [u8]>,
+        message_bytes: impl Iterator<Item = &'a [u8]>,
+    ) -> bool {
+        // on my laptop, we start to gain improvement over individual verify
+        // for single-threaded batch: at 2 signatures
+        // for 16 thread parallel batch: at 16 * 16 = 256 signatures
```
samkim-crypto: This comment seems a bit informal. I know that these parameter will depend on the specific hardware, but maybe we can write it slightly more formally as follows for now:

```suggestion
        // Empirical benchmarks indicate that batch verification outperforms 
        // individual verification starting at 2 signatures for single-threaded 
        // execution, and at approximately 256 signatures for highly parallel 
        // environments (e.g., 16-thread execution). These thresholds 
        // can be tuned further based on future profiling and experimentation.
```

wdyt?

signature/src/lib.rs
```diff
@@ -304,4 +448,65 @@ mod tests {
...
+        assert!(!Signature::batch_verify(
+            signatures.iter(),
+            pubkeys.iter().map(|pubkey| pubkey.as_slice()),
+            bad_messages.iter().copied(),
+        ));
+    }
+
+    #[cfg(feature = "parallel")]
+    #[test]
+    fn test_par_batch_verify() {
```
samkim-crypto, line 487: The `batch_verify` logic depends on the `INDIVIDUAL_VERIFY_THRESHOLD`, which is 8 for the parallel feature. Maybe we should test a batch of size at least 8 for completeness.

signature/src/lib.rs
```diff
@@ -304,4 +448,65 @@ mod tests {
         // Sanity check: ensure the pointer is the same.
         assert_eq!(signature.as_array().as_ptr(), signature.0.as_ptr());
     }
+
+    #[test]
+    fn test_batch_verify() {
```
samkim-crypto, line 467: Could be nice to test that signatures, public keys, and message iterators of differing lengths to make sure that it fails in this case.

samkim-crypto (commented):
Looking good! Some minor comments and questions below. Also, given that we changed the API to a single iterator of tuples in https://github.com/anza-xyz/agave/pull/12107, should we have an analogous API here?

signature/src/lib.rs
```diff
@@ -83,6 +113,121 @@ impl Signature {
...
+    /// messages.
+    ///
+    /// Each iterator item is a tuple of `(signature, pubkey_bytes,
+    /// message_bytes)`, where `signature` is the Ed25519 signature to verify,
+    /// `pubkey_bytes` is the signer's public key, and `message_bytes` is the
+    /// exact message that was signed. Returns `false` if any signature fails to
+    /// verify, any public key or signature cannot be parsed, or batch
+    /// verification fails.
+    pub fn batch_verify<'a>(
+        signature_data: impl Iterator<Item = (&'a Signature, &'a [u8], &'a [u8])>,
```
samkim-crypto: I am realizing that if we used an `ExactSizeIterator` instead of an `Iterator`, then a lot of the logic in the function can be quite simplified. For replay sigverify, it does seem like it is sufficient to use an `ExactSizeIterator` (for leader sigverify, we may use `.filter()`, so maybe not applicable there).

What do you think? Even if this is the case, feel free to address this in a follow-up PR.

samkim-crypto (dismissed):
Looks good to me! Just one suggestion regarding the use of `ExactSizeIterator`.

samkim-crypto (approved):
LGTM!

PR #699, opened 2026-04-22
message: Add heap size bounds check for message v1

febo:
### Problem

Message v1 accepts a configurable heap size parameter, but currently the value is not being validated.

### Solution

Add bounds validation for the heap size parameter.

### DIFF
--- a/message/src/versions/v1/message.rs
+++ b/message/src/versions/v1/message.rs
@@ -56,8 +56,8 @@ use {
         compiled_instruction::CompiledInstruction,
         compiled_keys::CompiledKeys,
         v1::{
-            MessageError, TransactionConfig, TransactionConfigMask, DEFAULT_HEAP_SIZE,
-            MAX_ADDRESSES, MAX_HEAP_SIZE, MAX_INSTRUCTIONS, MAX_SIGNATURES,
+            MessageError, TransactionConfig, TransactionConfigMask, MAX_ADDRESSES, MAX_HEAP_SIZE,
+            MAX_INSTRUCTIONS, MAX_SIGNATURES, MIN_HEAP_SIZE,
         },
         AccountKeys, CompileError, MessageHeader,
     },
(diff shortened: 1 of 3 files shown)


message/src/versions/v1/mod.rs
```diff
@@ -32,7 +32,12 @@ pub const MAX_INSTRUCTIONS: u8 = 64;
 pub const MAX_SIGNATURES: u8 = 12;
 
 /// Default heap size in bytes when not specified (32KB).
-pub const DEFAULT_HEAP_SIZE: u32 = 32_768;
+///
+/// This also represents the minimum heap size.
+pub const DEFAULT_HEAP_SIZE: u32 = 32 * 1024;
```
febo: Cosmetic change to make the value be readable like the `MAX_HEAP_SIZE`.

message/src/versions/v1/message.rs
```diff
@@ -462,11 +462,15 @@ impl Message {
...
         }
 
-        // heap size must be a multiple of 1024
+        // if specified, heap size must be a multiple of 1024 and within valid bounds
         if let Some(heap_size) = self.config.heap_size {
             if heap_size % 1024 != 0 {
                 return Err(MessageError::InvalidHeapSize);
             }
+
+            if !(DEFAULT_HEAP_SIZE..=MAX_HEAP_SIZE).contains(&heap_size) {
```
apfitzge: nit: can we just add a third MIN_HEAP_SIZE, even if it's equal to default?

As is this line looks like the validation is incorrect (it's not)

message/src/versions/v1/message.rs
```diff
@@ -462,11 +462,15 @@ impl Message {
...
+                return Err(MessageError::InvalidHeapSize);
+            }
```
tao-stones, line 473: Maybe add a test to cover this new path?

tao-stones (commented):
thanks for adding this check

joncinque (commented):
Agreed with the other comments -- new test and new const would be very useful

tao-stones (approved):
lgtm

message/src/versions/v1/message.rs
febo: Good point!

apfitzge (approved):

joncinque (approved):

PR #702, opened 2026-04-23
[BLS] add compiler_fence to zeroize

zz-sol:
This PR enhances zeroization as follows

- add `compiler_fence` for zeroization
- add zeroization via type protection for sk related json

### DIFF
--- a/bls-signatures/src/secret_key.rs
+++ b/bls-signatures/src/secret_key.rs
@@ -131,23 +131,30 @@ impl SecretKey {
 
 impl TryFrom<&[u8]> for SecretKey {
     type Error = BlsError;
-    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
-        if bytes.len() != BLS_SECRET_KEY_SIZE {
+    fn try_from(src: &[u8]) -> Result<Self, Self::Error> {
+        if src.len() != BLS_SECRET_KEY_SIZE {
             return Err(BlsError::ParseFromBytes);
         }
-        let bytes = Zeroizing::new(<[u8; BLS_SECRET_KEY_SIZE]>::try_from(bytes).unwrap());
+        let mut bytes = Zeroizing::new([0u8; BLS_SECRET_KEY_SIZE]);
+        bytes.copy_from_slice(src);
         Self::parse_scalar(&bytes).map(Self)
     }
 }
 
-impl From<&SecretKey> for [u8; BLS_SECRET_KEY_SIZE] {
-    fn from(secret_key: &SecretKey) -> Self {
-        // WARNING: The returned buffer contains raw secret-key bytes. Callers should zeroize it
-        // as soon as they are done using it.
-        secret_key.0.to_bytes_le()
-    }
-}
-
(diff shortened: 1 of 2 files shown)


bls-signatures/src/secret_key.rs
```diff
@@ -141,3 +147,9 @@ impl From<&SecretKey> for [u8; BLS_SECRET_KEY_SIZE] {
         secret_key.0.to_bytes_le()
     }
 }
+
+impl From<&SecretKey> for Zeroizing<[u8; BLS_SECRET_KEY_SIZE]> {
```
samkim-crypto, line 158: Given that we now implement `From<&SecretKey>` for `Zeroizing<[u8; BLS_SECRET_KEY_SIZE]>`, should we remove the old `From<&SecretKey>` for `[u8; BLS_SECRET_KEY_SIZE]`?

bls-signatures/src/secret_key.rs
```diff
@@ -129,8 +135,8 @@ impl TryFrom<&[u8]> for SecretKey {
         if bytes.len() != BLS_SECRET_KEY_SIZE {
             return Err(BlsError::ParseFromBytes);
         }
-        // unwrap safe due to the length check above
-        Self::parse_scalar(bytes.try_into().unwrap()).map(Self)
+        let bytes = Zeroizing::new(<[u8; BLS_SECRET_KEY_SIZE]>::try_from(bytes).unwrap());
```
samkim-crypto: Can we make sure that `<[u8; BLS_SECRET_KEY_SIZE]>::try_from(bytes).unwrap()` is not passed in by value to `Zeroizing::new`?

This one is probably equivalent to
```
let temp = <[u8; BLS_SECRET_KEY_SIZE]>::try_from(bytes).unwrap();
let bytes = Zeroizing::new(temp);
```

And the `temp` value might not be zeroed out.

It might be safer to use a copy function instead?
```
let mut safe_bytes = Zeroizing::new([0u8; BLS_SECRET_KEY_SIZE]);
safe_bytes.copy_from_slice(bytes); // does not make an owned copy
Self::parse_scalar(&safe_bytes).map(Self)
```

I am not 100% if this is actually an issue, so we can also go with the current implementation and ask about this on the fix review.

samkim-crypto (commented):
Looks good to me! Maybe we should remove the existing `From` implementation and a comment that I am not 100% sure about, so it would be nice to get your take.

bls-signatures/src/secret_key.rs
zz-sol: Good catch. 

I think we should remove it. We now have `From<&SecretKey>` for `Zeroizing<[u8; BLS_SECRET_KEY_SIZE]>` and we should discourage `[u8; 32]` conversion. 

if someone really wants the secret bytes he should do

```
let zeroizing_bytes: Zeroizing<[u8; 32]> = (&secret_key).into();
let mut raw_bytes = [0u8; 32];
raw_bytes.copy_from_slice(zeroizing_bytes.as_slice());
```

in which case they will not access the secret bytes by accident

samkim-crypto (approved):
Looks great!

PR #703, opened 2026-04-23
feat(short-vec): replicode impl from wincode crate

kskalski:
This will allow deprecating solana-short-vec feature in wincode crate (or just not enabling it by users), which causes cyclic dependency (see https://github.com/anza-xyz/wincode/issues/303)

Note, this code is very similar to https://github.com/anza-xyz/wincode/blob/master/wincode/src/len.rs#L383-L894 except:
* nothing is marked as pub, this impl only implements traits 
* `pub const fn decode_short_u16` is skipped, instead the tests and examples use `wincode::deserialize::<ShortU16>(&bytes)` (well, we could possibly add that to lib.rs or improve / replace decode_shortu16_len using its impl - the question if / where this API is needed)
* because of the above, one benchmark (`wincode:decode_short_u16`) was no longer contributing anything new, it's skipped

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4504,6 +4504,7 @@ version = "3.2.0"
 dependencies = [
  "assert_matches",
  "bincode",
+ "criterion",
  "proptest",
  "serde",
  "serde_core",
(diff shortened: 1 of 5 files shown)


cpubot (approved):

joncinque (approved):
Looks great!

PR #704, opened 2026-04-23
[BLS] fix misleading batch verification documentation

zz-sol:
This is part of the auditing findings D that I missed during the first screen -- the documentation is a bit abigious. Now addressed.


samkim-crypto (approved):
Looks good! I attempted to address this in https://github.com/anza-xyz/solana-sdk/pull/674 ([78b5a30](https://github.com/anza-xyz/solana-sdk/pull/674/commits/78b5a30c4b33b336ae6a228808a5f62c71114526)), but your fix is much more complete and comprehensive. Thanks!

PR #706, opened 2026-04-24
refactor: switch to solana-short-vec/wincode feature

kskalski:
https://github.com/anza-xyz/solana-sdk/pull/703 added `wincode` feature to `solana-short-vec` such that we can deprecate `solana-short-vec` feature from `wincode`.

Switch sdk crates to use that instead of relying on wincode's optional dependency (towards fixing https://github.com/anza-xyz/wincode/issues/303)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3987,7 +3987,7 @@ dependencies = [
  "solana-nonce",
  "solana-sanitize",
  "solana-sdk-ids",
- "solana-short-vec 3.2.1",
+ "solana-short-vec",
  "solana-system-interface",
  "solana-transaction-error",
  "static_assertions",
@@ -4179,7 +4179,7 @@ dependencies = [
  "solana-serde-varint",
  "solana-serialize-utils",
  "solana-sha256-hasher",
- "solana-short-vec 3.2.1",
+ "solana-short-vec",
  "solana-slot-hashes",
  "solana-slot-history",
  "solana-stable-layout",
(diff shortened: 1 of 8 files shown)


joncinque (approved):
Looks great! Just to be sure, we'll start using this in agave as soon as possible, correct?

I think we need to avoid a situation where we're enabling the `solana-short-vec` feature on wincode and the `wincode` feature on `solana-short-vec`

kskalski:
Yes, I want to get rid of solana-short-vec feature enabled on wincode ASAP, which I think now blocks using new solana-stake-inteface.
Currently agave has it enabled, but even if it's removed the feature kicks in, since those SDK crates force it.

The only concern about those changes is the patch vs minor release - this PR makes solana-message and solana-transaction depend on solana-short-vec 3.2.1 (which I just published) if they enable wincode... possibly we should use solana-short-vec 3.3.0 and bump that dependency here too.

joncinque:
> The only concern about those changes is the patch vs minor release - this PR makes solana-message and solana-transaction depend on solana-short-vec 3.2.1 (which I just published) if they enable wincode... possibly we should use solana-short-vec 3.3.0 and bump that dependency here too.

I'm not too fussed either way on 3.2.1 vs 3.3.0, since you bumped the whole repo to use 3.2.1, which was the correct move. We'll just need to make sure to publish both of these crates and use them

kskalski:
Yeah, I guess the prerequisite is - ensure `solana-short-vec` crate is >=3.2.1 (this is already the case in agave)
and then the flip need to happen in one go:
* disable `solana-short-vec` feature in `wincode` dependency across the workspace
* bump `solana-message` and `solana-transaction` to new versions (with this PR)

apfitzge (approved):

PR #707, opened 2026-04-28
program-log: Remove explicit use of memory syscalls

febo:
### Problem

Currently, the `Logger` implementation is calling memory syscalls explicitly. This is less efficient than allowing the compiler to decide whether they are needed or not.

### Solution

Remove the explicit use of memory syscall. The compiler can decide to use them or not.


joncinque (approved):
Makes sense!

PR #708, opened 2026-04-30
frozen-abi: add stable abi sample derive macro

puhtaytow:
### Description

There was no way to derive rand distribution implementation, which is required by `StableAbi` tests.

### Summary of changes
- extended docs and tests
- added stable abi sample derive macro, which allow for providing `with=` parameter with override as needed
- changed literal to expression in digests

### DIFF
--- a/frozen-abi-macro/src/lib.rs
+++ b/frozen-abi-macro/src/lib.rs
@@ -29,6 +29,12 @@ pub fn derive_stable_abi(_item: TokenStream) -> TokenStream {
     "".parse().unwrap()
 }
 
+#[cfg(not(feature = "frozen-abi"))]
+#[proc_macro_derive(StableAbiSample, attributes(stable_abi_sample))]
+pub fn derive_stable_abi_sample(_item: TokenStream) -> TokenStream {
+    "".parse().unwrap()
+}
+
 #[cfg(feature = "frozen-abi")]
 #[proc_macro_derive(StableAbi)]
 pub fn derive_stable_abi(item: TokenStream) -> TokenStream {
@@ -56,12 +62,28 @@ pub fn derive_stable_abi(item: TokenStream) -> TokenStream {
 }
 
 #[cfg(feature = "frozen-abi")]
+#[proc_macro_derive(StableAbiSample, attributes(stable_abi_sample))]
+pub fn derive_stable_abi_sample(item: TokenStream) -> TokenStream {
+    let item = parse_macro_input!(item as Item);
+    match item {
+        Item::Struct(input) => derive_stable_abi_sample_struct_type(input),
+        Item::Enum(input) => derive_stable_abi_sample_enum_type(input),
+        _ => Error::new_spanned(
+            item,
+            "StableAbiSample can only be derived for struct or enum",
+        )
+        .to_compile_error()
+        .into(),
+    }
+}
+
+#[cfg(feature = "frozen-abi")]
 use proc_macro2::{Span, TokenStream as TokenStream2, TokenTree};
 #[cfg(feature = "frozen-abi")]
 use quote::{quote, ToTokens};
 #[cfg(feature = "frozen-abi")]
 use syn::{
-    parse_macro_input, Attribute, Error, Fields, Ident, Item, ItemEnum, ItemStruct, ItemType,
+    parse_macro_input, Attribute, Error, Expr, Fields, Ident, Item, ItemEnum, ItemStruct, ItemType,
     LitStr, Variant,
 };
 
@@ -112,6 +134,154 @@ fn filter_allow_attrs(attrs: &mut Vec<Attribute>) {
         let ss = &attr.path().segments.first().unwrap().ident.to_string();
         ss.starts_with("allow")
     });
+}
+
+#[cfg(feature = "frozen-abi")]
+fn parse_stable_abi_sample_override(field: &syn::Field) -> Result<Option<TokenStream2>, Error> {
+    let mut override_expr: Option<TokenStream2> = None;
+    for attr in &field.attrs {
+        if !attr.path().is_ident("stable_abi_sample") {
+            continue;
+        }
+        attr.parse_nested_meta(|meta| {
+            if meta.path.is_ident("with") {
+                // reject duplicate `with` on the same field
+                if override_expr.is_some() {
+                    return Err(meta.error("duplicate `with` in `#[stable_abi_sample(...)]`"));
+                }
+                let value = meta.value()?.parse::<LitStr>()?;
+                let expr = value.parse::<Expr>().map_err(|err| {
+                    Error::new(value.span(), format!("invalid `with` expression: {err}"))
+                })?;
+                override_expr = Some(quote! { #expr });
+                Ok(())
+            } else {
+                Err(meta.error("unsupported `stable_abi_sample` option; expected `with = \"...\"`"))
+            }
+        })?;
+    }
+    Ok(override_expr)
+}
+
+#[cfg(feature = "frozen-abi")]
+fn stable_abi_sample_field_expr(field: &syn::Field) -> Result<TokenStream2, Error> {
+    Ok(match parse_stable_abi_sample_override(field)? {
+        Some(expr) => expr,
+        None => quote! { rng.random() },
+    })
+}
+
+#[cfg(feature = "frozen-abi")]
+fn derive_stable_abi_sample_struct_type(input: ItemStruct) -> TokenStream {
+    let type_name = &input.ident;
+    let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
+    let turbofish = ty_generics.as_turbofish();
+    let sample_expr = match &input.fields {
+        Fields::Named(named_fields) => {
+            let mut fields = quote! {};
+            for field in &named_fields.named {
+                let field_name = &field.ident;
+                let field_expr = match stable_abi_sample_field_expr(field) {
+                    Ok(field_expr) => field_expr,
+                    Err(err) => return err.to_compile_error().into(),
+                };
+                fields.extend(quote! {#field_name: #field_expr,});
+            }
+            quote! {#type_name #turbofish { #fields }}
+        }
+        Fields::Unnamed(unnamed_fields) => {
+            let mut fields = quote! {};
+            for field in &unnamed_fields.unnamed {
+                let field_expr = match stable_abi_sample_field_expr(field) {
+                    Ok(field_expr) => field_expr,
+                    Err(err) => return err.to_compile_error().into(),
+                };
+                fields.extend(quote! {#field_expr,});
+            }
+            quote! {#type_name #turbofish ( #fields )}
+        }
+        Fields::Unit => quote! {#type_name #turbofish},
+    };
+    quote! {
+        #[automatically_derived]
+        impl #impl_generics ::solana_frozen_abi::rand::distr::Distribution<#type_name #ty_generics>
+            for ::solana_frozen_abi::rand::distr::StandardUniform
+            #where_clause
+        {
+            fn sample<R: ::solana_frozen_abi::rand::Rng + ?Sized>(
+                &self,
+                rng: &mut R,
+            ) -> #type_name #ty_generics {
+                #sample_expr
+            }
+        }
+    }
+    .into()
+}
+
+#[cfg(feature = "frozen-abi")]
+fn stable_abi_sample_enum_variant_expr(
+    type_name: &Ident,
+    ty_generics: &syn::TypeGenerics,
+    variant: &Variant,
+) -> Result<TokenStream2, Error> {
+    let variant_name = &variant.ident;
+    let turbofish = ty_generics.as_turbofish();
+    match &variant.fields {
+        Fields::Unit => Ok(quote! {#type_name #turbofish::#variant_name}),
+        Fields::Named(variant_fields) => {
+            let mut fields = quote! {};
+            for field in &variant_fields.named {
+                let field_name = &field.ident;
+                let field_expr = stable_abi_sample_field_expr(field)?;
+                fields.extend(quote! {#field_name: #field_expr,});
+            }
+            Ok(quote! {#type_name #turbofish::#variant_name { #fields }})
+        }
+        Fields::Unnamed(variant_fields) => {
+            let mut fields = quote! {};
+            for field in &variant_fields.unnamed {
+                let field_expr = stable_abi_sample_field_expr(field)?;
+                fields.extend(quote! {#field_expr,});
+            }
+            Ok(quote! {#type_name #turbofish::#variant_name( #fields )})
+        }
+    }
+}
+
+#[cfg(feature = "frozen-abi")]
+fn derive_stable_abi_sample_enum_type(input: ItemEnum) -> TokenStream {
+    let type_name = &input.ident;
+    let variants = &input.variants;
+    let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
+    let variant_count = variants.len();
+    let mut match_arms = quote! {};
+    for (index, variant) in variants.iter().enumerate() {
+        let sample_expr =
+            match stable_abi_sample_enum_variant_expr(type_name, &ty_generics, variant) {
+                Ok(sample_expr) => sample_expr,
+                Err(err) => return err.to_compile_error().into(),
+            };
+        match_arms.extend(quote! {#index => #sample_expr,});
+    }
+    quote! {
+        #[automatically_derived]
+        impl #impl_generics ::solana_frozen_abi::rand::distr::Distribution<#type_name #ty_generics>
+            for ::solana_frozen_abi::rand::distr::StandardUniform
+            #where_clause
+        {
+            fn sample<R: ::solana_frozen_abi::rand::Rng + ?Sized>(
+                &self,
+                rng: &mut R,
+            ) -> #type_name #ty_generics {
+                match rng.random_range(0..#variant_count) {
+                    #match_arms
+                    _ => unreachable!(),
+                }
+            }
+        }
+    }
+    .into()
 }
 
 #[cfg(feature = "frozen-abi")]
--- a/frozen-abi-macro/src/lib.rs
+++ b/frozen-abi-macro/src/lib.rs
@@ -55,13 +55,34 @@ pub fn derive_stable_abi(item: TokenStream) -> TokenStream {
     expanded.into()
 }
 
+#[cfg(not(feature = "frozen-abi"))]
+#[proc_macro_derive(StableAbiSample, attributes(stable_abi_sample))]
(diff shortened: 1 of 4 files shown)


frozen-abi/src/stable_abi.rs
```diff
@@ -105,43 +105,49 @@ mod tests {
...
-    struct TestStructAbiDigestOnly {
+    struct TestStableAbiSample {
         a: u64,
         b: bool,
         c: [u8; 32],
         d: (u8, u8),
-    }
-
-    impl crate::rand::distr::Distribution<TestStructAbiDigestOnly>
-        for crate::rand::distr::StandardUniform
```
kskalski, line 116: I suppose this way of generating samples will still be supported? we should leave this test as is and add the one for TestStableAbiSample

frozen-abi/src/stable_abi.rs
```diff
@@ -105,43 +105,49 @@ mod tests {
...
+        #[stable_abi_sample(with = "rng.random::<[bool; 4]>().to_vec()")]
+        e: Vec<bool>,
+        f: bool,
+        #[stable_abi_sample(
+            with = "{ let mut m = std::collections::HashMap::new(); m.insert(rng.random(), rng.random()); m }"
+        )]
+        g: std::collections::HashMap<u64, bool>,
+        #[stable_abi_sample(
+            with = "(rng.random(), vec![TestStableAbiSampleInnerStruct { a: rng.random::<[u16; 2]>().to_vec(), b: rng.random() }])"
+        )]
```
kskalski, line 207: previously the `sample` implementation used simply `rng.random()`, since I guess each of those types do have distribution implementation, is still still supported and the `with = ` options above optional? If so, let's have separate test for those

puhtaytow: yea, the override is optional and saver in case of type not having distribution impl; splitted tests and added a bit more examples; thank you :pray:

puhtaytow:
@kskalski whenever you find a moment :pray:

frozen-abi/src/stable_abi.rs
```diff
@@ -124,4 +124,128 @@ mod tests {
...
+        solana_frozen_abi_macro::frozen_abi(
+            abi_digest = "EVb5Tgy4rZzifSr6cE8HX3K8Zy6w9PD9hamvo3G9Ue7D",
+            abi_serializer = "wincode",
+        )
+    )]
+    struct TestStableAbiSampleOverride {
+        #[stable_abi_sample(with = "rng.random::<[bool; 4]>().to_vec()")]
+        a: Vec<bool>,
+        #[stable_abi_sample(
+            with = "{ let mut m = std::collections::HashMap::new(); m.insert(rng.random(), rng.random()); m }"
```
kskalski: let's have examples that can exercise varying collection sizes, e.g.
```
std::collections::HashMap::from_iter((0..rng.random::<u8>() % 4).map(|_| (rng.random(), rng.random())))
```
or even better if we could skip the output type using inference on the output:
```
(0..rng.random::<u8>() % 4).map(|_| (rng.random(), rng.random())).collect()
```

frozen-abi/Cargo.toml
```diff
@@ -45,7 +45,7 @@ memmap2 = { workspace = true }
 # to avoid version skew and keep digest computation stable regardless of consumer dependencies.
 rand = { workspace = true }
 rand_chacha = { workspace = true }
-wincode = { workspace = true, features = ["alloc", "derive"], optional = true }
+wincode = { workspace = true, features = ["std", "alloc", "derive"], optional = true }
```
kskalski: is it only used by tests? if so, it should rather be an override in dev-dependencies

frozen-abi/src/stable_abi.rs
```diff
@@ -124,4 +124,128 @@ mod tests {
...
+            abi_digest = "AgNkEpErnFBuy7iTAEUUAC1fbvokEkhbsfFnx4DtXAvY",
+            abi_serializer = "wincode",
+        )
+    )]
+    struct TestStableAbiSampleSimple {
+        a: u64,
+        b: bool,
+        c: [u8; 32],
+        d: (u8, u8),
+    }
```
kskalski, line 148: we could add a test that checks that the same abi_digest is generated for different types that serialize in the same way - bincode / wincode is a simple serial format where e.g. `struct A { x: u8, b: u64, c: [u8; 4], d: HashMap<u8, u64> }` serializes the same as `(u8, u64, u8, [u8; 3], Vec<(u8, u64)>)` (or various variants of fields / tuples)

frozen-abi/Cargo.toml
puhtaytow: Yeah the `std` feature is used only by frozen-abi crate tests so moved it to dev-dependencies. Thanks.

frozen-abi/src/stable_abi.rs
puhtaytow: HashMap isnt good candidate for this tbh because it doesnt guarantee deterministic ordering; thats why it should have only 1 element.

puhtaytow: I took inspiration from the first suggestion tho as it looks cleaner than block of code i had previously. Thanks.

puhtaytow: I added to `Vec<bool>` the trick u suggested.

frozen-abi/src/stable_abi.rs
puhtaytow: Should be good now :pray:

frozen-abi/src/stable_abi.rs
```diff
@@ -124,4 +124,198 @@ mod tests {
...
+                    &self,
+                    rng: &mut R,
+                ) -> TestEquivalentWincodeStruct {
+                    TestEquivalentWincodeStruct {
+                        a: rng.random(),
+                        b: rng.random(),
+                        c: (rng.random(), rng.random()),
+                    }
+                }
+            }
```
kskalski: do we need those manual impls for the equivalence to hold? I was hoping to assert that derived impl behaves in this way

frozen-abi/src/stable_abi.rs
```diff
@@ -124,4 +124,198 @@ mod tests {
...
+        C {
+            #[stable_abi_sample(with = "rng.random::<[bool; 4]>().to_vec()")]
+            a: Vec<bool>,
+            b: u16,
+        },
+    }
+
+    // Keep these fixtures wire-compatible
+    #[rustfmt::skip]
+    macro_rules! linked_equivalent_pair {
```
kskalski: I would call it something like `check_equal_wire_format`

frozen-abi/src/lib.rs
```diff
@@ -28,15 +26,51 @@
...
+//! `rng.random()` (for example `Vec<_>` or `HashMap<_, _>`), or when you want a specific shape.
+//!
+//! ```rust,ignore
+//! #[derive(StableAbi, StableAbiSample)]
+//! #[frozen_abi(abi_digest = "...", abi_serializer = "wincode")]
+//! struct MyTypeWithOverride {
+//!     #[stable_abi_sample(with = "rng.random::<[bool; 4]>().to_vec()")]
+//!     a: Vec<bool>,
+//!     #[stable_abi_sample(
+//!         with = "{ let mut m = std::collections::HashMap::new(); m.insert(rng.random(), rng.random()); m }"
```
kskalski: docs could also use the expression style value

frozen-abi/src/stable_abi.rs
kskalski: Let's use `BTreeMap` vs `Vec` then, we will still need some trick to generate sorted keys, but we could do something like

```
(0..rng.random::<u16>() % 4).map(|i| (i << 8 + rng.random::<u8>(), rng.random())).collect()
```

Or just use two sequence collections like `VecDeque` and `Vec`, but I think maps vs vec is more interesting, since for anyone not knowing bincode it wouldn't be obvious maps have the same wire format as sequence of pairs

frozen-abi/src/stable_abi.rs
puhtaytow: derive suffice, i just thought it would be easier to understand whats going on with manuals; fixed

frozen-abi/src/stable_abi.rs
puhtaytow: Should be good now :+1:

frozen-abi-macro/src/lib.rs
```diff
@@ -29,6 +29,12 @@ pub fn derive_stable_abi(_item: TokenStream) -> TokenStream {
     "".parse().unwrap()
 }
 
+#[cfg(not(feature = "frozen-abi"))]
+#[proc_macro_derive(StableAbiSample, attributes(stable_abi_sample))]
+pub fn derive_stable_abi_sample(_item: TokenStream) -> TokenStream {
```
kskalski: would be nice to move this declaration next to the #[cfg(feature = "frozen-abi")] version

frozen-abi-macro/src/lib.rs
```diff
@@ -114,6 +136,154 @@ fn filter_allow_attrs(attrs: &mut Vec<Attribute>) {
...
+    let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
+    let variant_count = variants.len();
+    let mut match_arms = quote! {};
+    for (index, variant) in variants.iter().enumerate() {
+        let sample_expr =
+            match stable_abi_sample_enum_variant_expr(type_name, &ty_generics, variant) {
+                Ok(sample_expr) => sample_expr,
+                Err(err) => return err.to_compile_error().into(),
+            };
+        match_arms.extend(quote! {#index => #sample_expr,});
```
kskalski: I suspect you could collect a vector of expressions here and splice them later, something along the lines of:
```
let match_arms: Vec<proc_macro2::TokenStream> = variants
    .iter()
    .enumerate()
    .map(|(index, variant)| {
        let sample_expr =
            match stable_abi_sample_enum_variant_expr(type_name, &ty_generics, variant) {
                Ok(sample_expr) => sample_expr,
                Err(err) => return err.to_compile_error(),
            };
        quote! { #index => #sample_expr, }
    })
    .collect();
```
Then splice it with the repetition syntax `#( #match_arms )*`:
```
quote! {
    match rng.random_range(0..#variant_count) {
        #( #match_arms )*
        _ => unreachable!(),
    }
}
```

should work too in other places you use extend on the single expression

frozen-abi/src/stable_abi.rs
```diff
@@ -124,4 +124,240 @@ mod tests {
...
+            #[stable_abi_sample(with = "rng.random::<[bool; 4]>().to_vec()")]
+            a: Vec<bool>,
+            b: u16,
+        },
+    }
+
+    // Keep these fixtures wire-compatible
+    macro_rules! check_equal_wire_format {
+        (
+            abi_digest_pair_a = $abi_a:literal,
```
kskalski: it would be cleaner if we could just invoke this macro 3 times with separate test cases... but I'm not sure how easy it would be to have the whole structs defined in the macro parameter.
 
Maybe we don't really need a macro and could use a const &str for abi_digest that would be used on the corresponding struct pairs?

frozen-abi/src/stable_abi.rs
```diff
@@ -252,112 +252,101 @@ mod tests {
...
-        (
-            abi_digest_pair_a = $abi_a:literal,
-            abi_digest_pair_b = $abi_b:literal,
-            abi_digest_pair_c = $abi_c:literal
-        ) => {
-            #[derive(Debug, wincode::SchemaWrite)]
-            #[cfg_attr(
+    const ABI_DIGEST_PAIR_A: &str = "G7kuFGzwY6HwSytv6UsjWVEAbhrfv2n2gmchE27mSRiM";
+    const ABI_DIGEST_PAIR_B: &str = "14qLvWX4UebbLBaKi6v31A8xDfXU8ifX8DqCGbpAwjtD";
+    const ABI_DIGEST_PAIR_C: &str = "9pGP5GGD2HxDRCQeDv3rPGTfVH9SzkZCXMSGHpZnzz4G";
```
kskalski: let's place the constants next to relevant structs declarations and give them meaningful names:
* ABI_DIGEST_EQUIVALENT_FIELD_STRUCTURES
* ABI_DIGEST_EQUIVALENT_BYTE_SEQUENCES
* ABI_DIGEST_EQUIVALENT_KEY_VALUE_SEQUENCES

puhtaytow:
Looks much better after suggestions. Thanks :pray:

frozen-abi-macro/src/lib.rs
```diff
@@ -114,6 +136,166 @@ fn filter_allow_attrs(attrs: &mut Vec<Attribute>) {
...
+                        .map(|field_expr| quote! {#field_name: #field_expr,})
+                })
+                .collect::<Result<Vec<_>, _>>()?;
+            Ok(quote! {#type_name #turbofish::#variant_name { #(#fields)* }})
+        }
+        Fields::Unnamed(variant_fields) => {
+            let fields = variant_fields
+                .unnamed
+                .iter()
+                .map(|field| stable_abi_sample_field_expr(field).map(|field_expr| quote! {#field_expr,}))
```
kskalski: could we skip the map into `#field_expr,` and add comma on the use site like in https://github.com/anza-xyz/wincode/blob/068bcf2983b621fcd602d2176afa5a26816f1529/wincode-derive/src/common.rs#L233 ?

frozen-abi/src/stable_abi.rs
```diff
@@ -124,4 +124,231 @@ mod tests {
...
+                solana_frozen_abi_macro::frozen_abi(
+                    abi_digest = ABI_DIGEST_EQUIVALENT_BYTE_SEQUENCES,
+                    abi_serializer = "bincode",
+                )
+            )]
+    struct TestEquivalentCollectionsBincode(
+        #[stable_abi_sample(
+            with = "(0..rng.random::<u8>() % 4).map(|_| rng.random::<u8>()).collect()"
+        )]
+        std::collections::VecDeque<u8>,
```
kskalski: let's import the type

frozen-abi/src/stable_abi.rs
```diff
@@ -124,4 +124,231 @@ mod tests {
...
+                solana_frozen_abi_macro::frozen_abi(
+                    abi_digest = ABI_DIGEST_EQUIVALENT_KEY_VALUE_SEQUENCES,
+                    abi_serializer = "wincode",
+                )
+            )]
+    struct TestEquivalentBTreeMapVsVecWincode {
+        #[stable_abi_sample(
+            with = "(0..rng.random::<u16>() % 4).map(|i| (((i << 8) + rng.random::<u8>() as u16), rng.random())).collect()"
+        )]
+        a: std::collections::BTreeMap<u16, u8>,
```
kskalski: import the type

frozen-abi-macro/src/lib.rs
```diff
@@ -114,6 +136,167 @@ fn filter_allow_attrs(attrs: &mut Vec<Attribute>) {
...
+            quote! {#type_name #turbofish { #(#fields)* }}
+        }
+        Fields::Unnamed(unnamed_fields) => {
+            let fields = match unnamed_fields
+                .unnamed
+                .iter()
+                .map(stable_abi_sample_field_expr)
+                .collect::<Result<Vec<_>, _>>()
+            {
+                Ok(fields) => fields,
```
kskalski: instead of the match, I guess 
`let fields = ..collect..().map_err(|err| err.to_compile_error())?` 
would be cleaner

kskalski (commented):
It's getting pretty good, some last nits

kskalski:
LGTM

frozen-abi/src/stable_abi.rs
```diff
@@ -124,4 +126,231 @@ mod tests {
...
+        "G7kuFGzwY6HwSytv6UsjWVEAbhrfv2n2gmchE27mSRiM";
+    #[derive(Debug, wincode::SchemaWrite)]
+    #[cfg_attr(
+                feature = "frozen-abi",
+                derive(solana_frozen_abi_macro::StableAbi, solana_frozen_abi_macro::StableAbiSample),
+                solana_frozen_abi_macro::frozen_abi(
+                    abi_digest = ABI_DIGEST_EQUIVALENT_FIELD_STRUCTURES,
+                    abi_serializer = "wincode",
+                )
+            )]
```
joncinque, line 267: Not blocking, but it is surprising that rustfmt doesn't try to change this

frozen-abi/src/lib.rs
```diff
@@ -28,15 +26,52 @@
...
-//! For types which don't, you must provide as well a custom implementation:
+//! Field override is optional and only needed for fields that cannot be sampled with plain
+//! `rng.random()` (for example `Vec<_>` or `HashMap<_, _>`), or when you want a specific shape.
+//!
+//! ```rust,ignore
+//! #[derive(StableAbi, StableAbiSample)]
+//! #[frozen_abi(abi_digest = "...", abi_serializer = "wincode")]
+//! struct MyTypeWithOverride {
+//!     #[stable_abi_sample(
+//!         with = "(0..rng.random::<u8>() % 4).map(|_| rng.random::<bool>()).collect()")]
```
joncinque, line 64: These `with` statements essentially make us re-export `rand`, since if someone uses a different (breaking) version of rand than what's expected, they'll get errors.

That essentially means that we would need breaking releases of frozen-abi, and thus *a lot* of breaking changes in the SDK, anytime we update to a new breaking version of `rand`.

This is a bit annoying, but we could also just not respect semver for this feature since frozen-abi is meant for internal use only, and if someone is using `with = `, they should be aware of potential breakage. Does that sound ok @kskalski ?

joncinque (approved):
Looks great to me! Sorry for the slow review

frozen-abi/src/lib.rs
kskalski: It's probably not worse than what `StableAbi` implied already, since it required random distribution trait to be implemented for types...
Given how frequently and severely `rand` crate breaks compatibility and thus how hard it is to upgrade its versions, this could be a concern... There is already lots of code in the ecosystem that expose rand in public APIs though, so maybe we can just live with that.

joncinque: On our side we should aim to remove any exposure to `rand` in our public APIs, but it's true that it isn't too much worse than what we have already.

Either way, we can consider `frozen-abi` as private, since no downstream user should be enabling it other than agave.

puhtaytow: Thank you for the reviews :pray: :pray:  ill think how to reduce the re-exports. 

> we can consider frozen-abi as private

It would be cool tho if we could make it useful for wider audience.

frozen-abi/src/stable_abi.rs
puhtaytow: This is because i intentionally put there format skip coz it was messing with format strange way in the tests block; ill fix it by the way in a future pr.

PR #709, opened 2026-05-01
[BLS] update screening doc; improve security boundary

zz-sol:
This PR addresses audit documentation findings around BLS verification semantics.

- It clarifies that `verify_distinct` and related APIs provide aggregate screening guarantees, not CHP12-style weighted batch verification or per-signer validity guarantees. 
- It also clarifies the `VerifySignature` safety boundary while keeping `VerifySignature` public: crate-provided implementations are limited to `PopVerified<T>` and aggregate public-key wrappers, but downstream implementations remain an explicit extension point.


samkim-crypto (approved):
LGTM!

PR #710, opened 2026-05-06
impl SIMD-0529: big_int_mod_exp syscall api refactoring

zz-sol:
This PR implements the `big_int_mod_exp ` syscall as per [SIMD-0529](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0529-big-integer-modexp-syscall.md).

---- old pr description ----
- this PR refactors the API to allow for both big and little endian. 
- i also thought about the backend changes. we have the following options

| library | 2048-bit | 4096-bit |
| --- | ---: | ---: |
| `crypto-bigint` | 3.34 ms | 27.6 ms |
| `num-bigint` | 2.15 ms | 18.4 ms |
| `rug` | 1.25 ms | 8.49 ms |

- crypto-bigint is constant time but prob not required
- num-bigint is currently being used
- rug is the fastest -- but it depends on `gmp` and `cc` (wrapped by rust) so there maybe some additional system requirement. 

------

follow up: `rug` requires `LGPL-3.0+` license. we prob don't want to use it. so we stick with `num-bigint`.


samkim-crypto (commented):
Yeah in this case, let's stick with `num-bigint`. The changes looks good to me overall. Let's wait on merging this until https://github.com/solana-foundation/solana-improvement-documents/pull/529 is merged. I do have some comments regarding the syscall function signature, but I commented on the SIMD side.

samkim-crypto (commented):
This looks good overall.

I briefly looked at the num_bigint source code and it doesn't seem like they special case the following cases (please double check me), so I think we should short-circuit:
- If exponent is 0, then just return 1 (I think we originally had this logic, so you may have removed it for a reason I am overlooking)
- If exponent is 1, then we should just do base % modulus (again, please double check that num_bigint doesn't take care of this already)
- If base is 0, then we should just return 0

Also, I think num_bigint does pre-reduction only if the byte length of the base is greater than the byte length of modulus. If the byte lengths are equal, but base is greater, then it doesn't do pre-reduction, so I think we should do pre-reduction in our crate.

It could also be a micro optimization, but maybe we should terminate early on even modulus by just looking at the least significant bit. Right now, even if we get an even modulus, the `validate_modulus` can potentially scan through the entire modulus.

big-mod-exp/src/lib.rs
```diff
@@ -1,88 +1,419 @@
...
+/// Big integer modular exponentiation.
+///
+/// Inputs and output are little-endian unsigned integers. The returned value is
+/// padded to exactly `modulus.len()` bytes with trailing zeroes.
+///
+/// # Panics
+///
+/// Panics if any operand is longer than [`BIG_MOD_EXP_MAX_BYTES`] or if
+/// `modulus` is empty, zero, one, or even.
 pub fn big_mod_exp(base: &[u8], exponent: &[u8], modulus: &[u8]) -> Vec<u8> {
```
samkim-crypto: What do you think about returning `Option<Vec<u8>>`? Functions like `validate_modulus` and `validate_inputs` have packing behavior, which we probably don't want.

big-mod-exp/src/lib.rs
```diff
@@ -1,88 +1,419 @@
...
+
+    validate_modulus(modulus);
+}
+
+fn validate_modulus(modulus: &[u8]) {
+    let Some((&least_significant_byte, more_significant_bytes)) = modulus.split_first() else {
+        panic!("modulus length must be nonzero");
+    };
+
+    let has_nonzero_more_significant_byte = more_significant_bytes.iter().any(|byte| *byte != 0);
```
samkim-crypto: This logic takes `O(N)`, so let's perform the odd/even check (which is `O(1)`) before this logic.

big-mod-exp/src/lib.rs
```diff
@@ -1,88 +1,419 @@
...
-        if modulus.is_zero() || modulus.is_one() {
-            return vec![0_u8; modulus_len];
+        if is_zero_le(exponent) {
+            return padded_one(modulus_len);
+        }
+        if is_zero_le(base) {
+            return vec![0; modulus_len];
+        }
+
+        let should_reduce_base = base_needs_reduction(base, modulus);
```
samkim-crypto: I think the `BigUint` implementation of `PartialOrd` in the `num_bigint` library already has a very optimized word-sized limb comparisons, so what do you think about just doing something like:

```
        let modulus = BigUint::from_bytes_le(modulus);
        let mut base = BigUint::from_bytes_le(base);

        if base >= modulus {
            base %= &modulus;
        }

        if base.is_zero() {
            return vec![0; modulus_len];
        }

        if base.is_one() || is_one_le(exponent) {
            return padded_to_modulus_len(base.to_bytes_le(), modulus_len);
        }
```

samkim-crypto (commented):
Looking good. Just a few more comments!

big-mod-exp/src/lib.rs
```diff
@@ -1,88 +1,359 @@
...
+
+        if base >= modulus {
+            base = core::ops::Rem::rem(base, &modulus);
         }
 
+        if base == BigUint::ZERO {
+            return Some(vec![0; modulus_len]);
+        }
+
+        if base == BigUint::from(1_u8) || is_one_le(exponent) {
```
samkim-crypto: Sorry just one last comment. I believe `BigUint::from(1_u8)` will allocate `Vec<u32>` on the heap, so we can probably avoid this. I think we can make the check `base.bits() == 1 || is_one_le(exponent)`. What do you think?

samkim-crypto (commented):
One last comment and I think this should be ready to go!

samkim-crypto (approved):
Looks good to me!

big-mod-exp/Cargo.toml
```diff
@@ -12,13 +12,12 @@ edition = { workspace = true }
 [package.metadata.docs.rs]
 targets = ["x86_64-unknown-linux-gnu"]
 
-[target.'cfg(not(target_os = "solana"))'.dependencies]
-num-bigint = { workspace = true }
-num-traits = { workspace = true }
-
-[target.'cfg(target_os = "solana")'.dependencies]
+[dependencies]
```
joncinque, line 15: This dependency isn't needed for non-solana builds, right?
```suggestion
[target.'cfg(any(target_os = "solana", target_arch = "bpf"))'.dependencies]
```

big-mod-exp/src/lib.rs
```diff
@@ -1,88 +1,359 @@
...
 
+/// Parameters for the `sol_big_mod_exp` syscall.
+///
+/// The pointed-to input slices are encoded as little-endian unsigned integers.
 #[repr(C)]
+#[derive(Clone, Copy, Debug)]
 pub struct BigModExpParams {
-    pub base: u64,
+    /// VM pointer to the base bytes.
+    pub base: *const u8,
```
joncinque, line 10: We've had a discussion about this in other PRs, and it's not a huge deal, but a pointer input is not an ABI-stable input since its width can vary with different targets, so a `u64` is typically a better option.

joncinque (commented):
Just a couple of fly-by comments, I hope that's ok! Looks great overall

PR #711, opened 2026-05-07
[ed25519] fix batch_verify to matches verify_strict logic

zz-sol:
dalek's `batch_verify` function does not perform malleability check as in `verify_strict`. this PR fix this issue.


samkim-crypto (approved):
Looks fine to me!

PR #713, opened 2026-05-13
big_mod_exp: make BigModExpParams layout independent of target

nagisa:
Otherwise the crate compiled for non-64-bit pointer targets will have representation incompatible with the representation on solana VM guests.

### DIFF
--- a/big-mod-exp/src/lib.rs
+++ b/big-mod-exp/src/lib.rs
@@ -1,11 +1,12 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
+
 #[repr(C)]
 pub struct BigModExpParams {
-    pub base: *const u8,
+    pub base: u64,
     pub base_len: u64,
-    pub exponent: *const u8,
+    pub exponent: u64,
     pub exponent_len: u64,
-    pub modulus: *const u8,
+    pub modulus: u64,
     pub modulus_len: u64,
 }
 
@@ -39,11 +40,11 @@ pub fn big_mod_exp(base: &[u8], exponent: &[u8], modulus: &[u8]) -> Vec<u8> {
         let mut return_value = vec![0_u8; modulus.len()];
 
         let param = BigModExpParams {
-            base: base as *const _ as *const u8,
+            base: base.as_ptr().expose_provenance() as u64,
             base_len: base.len() as u64,
-            exponent: exponent as *const _ as *const u8,
+            exponent: exponent.as_ptr().expose_provenance() as u64,
             exponent_len: exponent.len() as u64,
-            modulus: modulus as *const _ as *const u8,
+            modulus: modulus.as_ptr().expose_provenance() as u64,
             modulus_len: modulus.len() as u64,
         };
         unsafe {


Lichtso:
Just remembered, didn't we want to remove this syscall entirely? (see https://github.com/anza-xyz/agave/pull/12192)

joncinque:
I believe the syscall is being revived by @zz-sol with https://github.com/anza-xyz/agave/pull/12321, probably with a better implementation. It's one of the pieces needed to perform RSA operations on-chain.

With that said, to avoid SDK details leaking into Agave, I would lean towards making this struct private in the `#[target_os = "solana"]` branch and avoid the problem entirely. Agave can define exactly the expected input type for the syscall.

Does that sound reasonable?

nagisa:
> With that said, to avoid SDK details leaking into Agave, I would lean towards making this struct private in the #[target_os = "solana"] branch and avoid the problem entirely. Agave can define exactly the expected input type for the syscall.

That was my initial suggestion as well, but @Lichtso thought fixing it in SDK was better. A problem of sorts is that pretty much every syscall that deals with typed inputs in memory uses sdk types rather than redefining their own, so this would be a divergence in approach and style (even though I personally think agave should be defining its own types, rather than pulling sdk ones.)

Lichtso:
> think agave should be defining its own types, rather than pulling SDK ones

if we duplicate all types the problem of keeping them in sync gets even worse.

Still if this syscall gets reworked then this PR might not make sense right now.

big-mod-exp/src/lib.rs
LucasSte: We could change the type here to `u64` so that this structure is also compatible with the Agave codebase.

nagisa: That's sorta what I did in the previous iteration of this PR (https://github.com/anza-xyz/solana-sdk/commit/5b5d8bd19fc50d99f96936e7e625629f026adc0d) before Jon's comment.

joncinque:
> if we duplicate all types the problem of keeping them in sync gets even worse.

Yeah definitely, I'm not suggesting to do it everywhere. In this case, since the type is an implementation detail, and not as part of the user interface, it seemed like a good reason to gate the type.

Although looking at it more, since the type is an implementation detail, maybe just moving to a `u64` would be the simplest thing? (Sorry for the back and forth)

nagisa:
Alright, adjusted back to using u64, this time unconditionally. Assuming the CI tests the sbpf target compiles :)

Lichtso (approved):

joncinque (approved):
Thanks!

big-mod-exp/src/lib.rs
```diff
@@ -1,11 +1,12 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
+
 #[repr(C)]
```
LucasSte, line 3: Nit: There is an extra line here.

nagisa: Intentional change on my part. This adds visual separation for attributes that apply to distinct scopes (the `#!` one for the whole module, the `#` just for the struct.)

PR #715, opened 2026-05-14
transaction: prepend v1 message byte in SanitizedTransaction::message_data

mcintyre94:
While testing v1 transactions I found that the preflight sigverify would fail. Transactions would land and confirm, but only with `skipPreflight: true` in the RPC request. This is because the bytes being signed at the RPC sigverify were just `v1::serialize`, which excludes the version byte.

This PR adds `serialize` for the v1 versioned message, which prepends the v1 message byte to `v1::serialize`. We then update `SanitizedTransaction::message_data` to use `cached_msg.message.serialize()`, matching the v0 branch, instead of calling `v1::serialize` directly.

Testing locally, this enables sending v1 transactions without `skipPreflight`.

### DIFF
--- a/message/src/versions/v1/message.rs
+++ b/message/src/versions/v1/message.rs
@@ -391,6 +391,12 @@ impl Message {
         crate::is_program_id_write_demoted(i, &self.account_keys, &self.instructions)
     }
 
+    /// Serialize this message with the V1 version prefix byte.
+    #[cfg(feature = "wincode")]
+    pub fn serialize(&self) -> Vec<u8> {
+        wincode::serialize(&(crate::v1::V1_PREFIX, self)).unwrap()
+    }
+
     /// Calculate the serialized size of the message in bytes.
     #[allow(clippy::arithmetic_side_effects)]
     #[inline(always)]
--- a/message/src/versions/v1/message.rs
+++ b/message/src/versions/v1/message.rs
@@ -588,6 +588,7 @@ unsafe impl<C: ConfigCore> SchemaWrite<C> for Message {
 /// Serialize the message.
 #[cfg(feature = "wincode")]
 #[inline]
+#[deprecated(since = "4.1.2", note = "use `Message::serialize` instead")]
 pub fn serialize(message: &Message) -> Vec<u8> {
     wincode::serialize(message).unwrap()
 }
@@ -1233,7 +1234,7 @@ mod tests {
         ];
 
         for message in &test_cases {
-            assert_eq!(message.size(), serialize(message).len());
+            assert_eq!(message.size(), wincode::serialize(message).unwrap().len());
         }
     }
 
@@ -1255,7 +1256,7 @@ mod tests {
             .build()
             .unwrap();
 
-        let bytes = serialize(&message);
+        let bytes = wincode::serialize(&message).unwrap();
 
         // Build expected bytes manually per SIMD-0385
         //
@@ -1298,7 +1299,7 @@ mod tests {
             .build()
             .unwrap();
 
-        let bytes = serialize(&message);
+        let bytes = wincode::serialize(&message).unwrap();
 
         let mut expected = vec![1, 0, 0];
         // ConfigMask: priority fee (bits 0,1) + CU limit (bit 2) = 0b111 = 7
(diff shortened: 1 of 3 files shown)


tao-stones (dismissed):
LG for tx-metadata

apfitzge:
I'm confused - how is it that the transactions can land correctly if the signature is incorrect? or is this to do with the rpc implementation of signature verification?

mcintyre94:
@apfitzge Yep this is just the RPC signature verification - the preflight check. If you do `skipPreflight: true` then it bypasses this and lands. But if you don't skip preflight then it uses this code to verify the signature and fails.

transaction/src/sanitized.rs
```diff
@@ -453,4 +454,30 @@ mod tests {
...
+        let sanitized = SanitizedTransaction::try_create(
+            versioned_tx,
+            MessageHash::Compute,
+            None,
+            SimpleAddressLoader::Disabled,
+            &HashSet::default(),
+        )
+        .unwrap();
+
+        sanitized.verify().unwrap();
```
febo, line 485: Shouldn't the test explicitly check that the first byte of the serialized message is the version? Not sure if the `verify` is doing that.

apfitzge:
> @apfitzge Yep this is just the RPC signature verification - the preflight check. If you do `skipPreflight: true` then it bypasses this and lands. But if you don't skip preflight then it uses this code to verify the signature and fails.

what is rpc doing differently that makes this fail?

edit: nvm I see it.

febo (commented):
Looks good! Left a comment about the test. Also, there seems to be some overlap with `VersionedMessage`, since that one explicitly write the version byte. But the implementation of serialize added in this PR is similar to the one that exists on `v0`, so probably ok.

transaction/src/sanitized.rs
mcintyre94: Thanks, I've added an explicit assert for the prefix. The `verify` ensures that the signing code and verifying code match, so the test would have failed before but was a bit indirect/could have been made to pass by different means.

apfitzge (dismissed):

febo (dismissed):
Looks good to me. Complementing my previous comment, we "lose" the `VersionedMessage` information when creating the `SanitizedTransaction`, so the approach followed in this PR seems to be inline with the existing `v0` behaviour.

Let's wait for @joncinque to have a look and give the final approval.

message/src/versions/v1/message.rs
```diff
@@ -391,6 +391,12 @@ impl Message {
         crate::is_program_id_write_demoted(i, &self.account_keys, &self.instructions)
     }
 
+    /// Serialize this message with the V1 version prefix byte.
+    #[cfg(feature = "wincode")]
+    pub fn serialize(&self) -> Vec<u8> {
+        wincode::serialize(&(crate::v1::V1_PREFIX, self)).unwrap()
+    }
```
joncinque, line 398: It's going to be a bit confusing to have `v1::Message::serialize` and `v1::serialize`, and someone will get this wrong in the future almost 100% -- what should we do with `v1::serialize`?

Our options are:
* leave it alone -- I don't love this, someone will get tripped up by this in the future
* remove it, force people to use `wincode::serialize` directly -- this is a breaking change, but since this shouldn't be widely used, maybe it's ok?
* deprecate it -- like the previous, but a little better since there's no breaking change
* fix it -- might be ok since it only has a few usages. But in that case, we shouldn't add this function

Of all those, I lean towards deprecating. Thoughts @apfitzge ?

joncinque (commented):
Great find! Just a nit about the existing function. We can go forward with this either way if we're pressed

message/src/versions/v1/message.rs
febo: Agree on the deprecate idea. We should also update the implementation to use `self.serialize()` instead of `wincode::serialize`.

apfitzge: Deprecation seems fine to me

mcintyre94: I've deprecated it, I think that makes sense. @febo I don't think we can use `self.serialize()` for it because that would include the v1 prefix byte which would change its behaviour.

joncinque (commented):
Looks good to me! We just need to use `wincode::serialize` instead of `v1::serialize` in a couple of tests it seems

joncinque (approved):
Looks great!

apfitzge (approved):

apfitzge:
is there a hold up on merging this?

mcintyre94:
Nope sorry, I was OOO end of last week. Merging now :)

PR #717, opened 2026-05-15
frozen-abi: switch to RngCore in StableAbi trait (and more)

puhtaytow:
### Issue

Direct reliance on rand StandardUniform in sampling mechanism for StableAbi made the trait hard to extend for types that need custom sampling, e.g. usize, which revealed the design flaw fixed in this pr.

### Summary of changes

Changes to the signature of the `StableAbi` trait and macro were made in a backward-compatible way. They do not require any action from users who used the old `StableAbi` trait, either with a manual `StandardUniform` implementation per type or through the derive macro, even when using the `with` field attribute.

- replaced the StableAbi default StandardUniform implementation with rand RngCore
- added impl_stable_abi_via_standard_uniform for types that can still delegate directly to rng.random()
- added impl_stable_abi_via_size_of_from_bytes for usize and isize
- added impl_stable_abi_for_tuples
- added StableAbi implementations for arrays and Option<T>
- updated stable_abi_sample_field_expr to use <FieldTy as ::solana_frozen_abi::stable_abi::StableAbi>::random(rng) instead of calling rng.random() directly
- updated the StableAbi derive output to generate a random() implementation that delegates to rand
- updated the frozen-abi docs to reflect the new StableAbi derive behavior
- added test coverage for platform-sized integers, tuples, and NonZero sampling

### DIFF
--- a/frozen-abi-macro/src/lib.rs
+++ b/frozen-abi-macro/src/lib.rs
@@ -50,7 +50,11 @@ pub fn derive_stable_abi(item: TokenStream) -> TokenStream {
 
     let expanded = quote! {
         #[automatically_derived]
-        impl #impl_generics ::solana_frozen_abi::stable_abi::StableAbi for #ident #ty_generics #where_clause {}
+        impl #impl_generics ::solana_frozen_abi::stable_abi::StableAbi for #ident #ty_generics #where_clause {
+            fn random(rng: &mut (impl ::solana_frozen_abi::rand::RngCore + ?Sized)) -> Self {
+                ::solana_frozen_abi::rand::Rng::random::<Self>(rng)
+            }
+        }
     };
     expanded.into()
 }
@@ -166,7 +170,12 @@ fn parse_stable_abi_sample_override(field: &syn::Field) -> Result<Option<TokenSt
 fn stable_abi_sample_field_expr(field: &syn::Field) -> Result<TokenStream2, Error> {
     Ok(match parse_stable_abi_sample_override(field)? {
         Some(expr) => expr,
-        None => quote! { rng.random() },
+        None => {
+            let ty = &field.ty;
+            quote! {
+                <#ty as ::solana_frozen_abi::stable_abi::StableAbi>::random(rng)
+            }
+        }
     })
 }
 
--- a/frozen-abi/src/lib.rs
+++ b/frozen-abi/src/lib.rs
@@ -29,7 +29,11 @@
 //! Deriving `StableAbi` adds:
 //!
 //! ```rust,ignore
-//! impl ::solana_frozen_abi::stable_abi::StableAbi for MyType {}
+//! impl ::solana_frozen_abi::stable_abi::StableAbi for MyType {
+//!     fn random(rng: &mut (impl ::solana_frozen_abi::rand::RngCore + ?Sized)) -> Self {
+//!         ::solana_frozen_abi::rand::Rng::random::<Self>(rng)
+//!     }
+//! }
 //! ```
 //!
 //! The `StableAbi::random()` default implementation calls `rng.random::<MyType>()`, so your type
--- a/frozen-abi/src/stable_abi.rs
+++ b/frozen-abi/src/stable_abi.rs
@@ -1,17 +1,118 @@
-use rand::{distr::StandardUniform, Rng, RngCore};
+use {
+    core::{array, num::NonZero},
+    rand::{Rng, RngCore},
+};
 
 pub trait StableAbi: Sized {
-    fn random(rng: &mut impl RngCore) -> Self
-    where
-        StandardUniform: rand::distr::Distribution<Self>,
-    {
-        rng.random::<Self>()
+    fn random(rng: &mut (impl RngCore + ?Sized)) -> Self;
+}
+
+macro_rules! impl_stable_abi_via_standard_uniform {
+    ($($t:ty),* $(,)?) => {
+        $(
+            impl StableAbi for $t {
+                fn random(rng: &mut (impl RngCore + ?Sized)) -> Self {
+                    rng.random::<Self>()
+                }
+            }
+        )*
+    };
+}
+
+macro_rules! impl_stable_abi_via_size_of_from_bytes {
+    ($from_bytes:ident, $($t:ty),* $(,)?) => {
+        $(
+            impl StableAbi for $t {
+                fn random(rng: &mut (impl RngCore + ?Sized)) -> Self {
+                    Self::$from_bytes(rng.random::<[u8; core::mem::size_of::<Self>()]>())
+                }
+            }
+        )*
+    };
+}
+
+macro_rules! impl_stable_abi_for_tuples {
+    ($(($($t:ident),+ $(,)?)),* $(,)?) => {
+        $(
+            impl<$($t),+> StableAbi for ($($t,)+)
+            where
+                $($t: StableAbi),+
+            {
+                fn random(rng: &mut (impl RngCore + ?Sized)) -> Self {
+                    ($($t::random(rng),)+)
+                }
+            }
+        )*
+    };
+}
+
+impl_stable_abi_via_standard_uniform!(
+    u8,
+    u16,
+    u32,
+    u64,
+    u128,
+    i8,
+    i16,
+    i32,
+    i64,
+    i128,
+    f32,
+    f64,
+    bool,
+    char,
+    NonZero<u8>,
+    NonZero<u16>,
+    NonZero<u32>,
+    NonZero<u64>,
+    NonZero<u128>,
+    NonZero<i8>,
+    NonZero<i16>,
+    NonZero<i32>,
+    NonZero<i64>,
+    NonZero<i128>,
+);
+impl_stable_abi_via_size_of_from_bytes!(from_le_bytes, usize);
+impl_stable_abi_via_size_of_from_bytes!(from_le_bytes, isize);
+impl_stable_abi_for_tuples!(
+    (A),
+    (A, B),
+    (A, B, C),
+    (A, B, C, D),
+    (A, B, C, D, E),
+    (A, B, C, D, E, F),
+    (A, B, C, D, E, F, G),
+    (A, B, C, D, E, F, G, H),
+    (A, B, C, D, E, F, G, H, I),
+    (A, B, C, D, E, F, G, H, I, J),
+    (A, B, C, D, E, F, G, H, I, J, K),
+    (A, B, C, D, E, F, G, H, I, J, K, L),
+);
+
+impl<T, const N: usize> StableAbi for [T; N]
+where
+    T: StableAbi,
+{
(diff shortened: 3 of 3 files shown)


puhtaytow:
@kskalski :pray:

kskalski:
This won't work for type aliases. We need to find a way to implement required trait for such types.

frozen-abi/src/stable_abi.rs
```diff
@@ -9,6 +9,44 @@ pub trait StableAbi: Sized {
...
+    R: Rng + ?Sized,
+{
+    T::sample_for_stable_abi(rng)
+}
+
+// exclude usize from the blanket as it has dedicated impl
+impl !SampleWithStandardUniform for usize {}
+
+#[cfg(not(target_pointer_width = "64"))]
+compile_error!("StableAbiSample only supports usize fields on 64-bit targets");
```
kskalski: Why do we forbid that?

There is a catch that abi digests will differ between 64-bit and 32-bit platforms (i.e. wincode always encodes with 64-bit, but the random generator will generate different values), but I guess this is acceptable. Alternatively, we could try to limit random usize values to smaller range, but this would lose our ability to exercise corner cases.

frozen-abi/src/stable_abi.rs
```diff
@@ -9,6 +9,44 @@ pub trait StableAbi: Sized {
...
+
+#[cfg(target_pointer_width = "64")]
+impl StableAbiSampleField for usize {
+    fn sample_for_stable_abi<R: Rng + ?Sized>(rng: &mut R) -> usize {
+        rng.random::<u64>() as usize
```
kskalski: you could do
```
let bytes: [u8; mem::size_of::<usize>()] = rand::random();
usize::from_ne_bytes(bytes)
```

frozen-abi/src/stable_abi.rs
```diff
@@ -339,18 +377,37 @@ mod tests {
 
     #[derive(Debug, serde_derive::Serialize)]
     #[cfg_attr(
-                feature = "frozen-abi",
```
kskalski, line 342: the reformat should be a separate PR

frozen-abi/src/stable_abi.rs
```diff
@@ -9,6 +9,44 @@ pub trait StableAbi: Sized {
...
+// covers every type that are supported by rand directly
+pub auto trait SampleWithStandardUniform {}
+
+pub trait StableAbiSampleField: Sized {
+    fn sample_for_stable_abi<R: Rng + ?Sized>(rng: &mut R) -> Self;
+}
+
+impl<T> StableAbiSampleField for T
+where
+    T: SampleWithStandardUniform,
```
kskalski: Would `StableAbiSampleField` blanket impl work without this constraint? I'm wondering why do we really need new `StableAbiSampleField` trait.

Could we impl `StableAbiSampleField` directly for `usize`?

frozen-abi/src/stable_abi.rs
puhtaytow: > There is a catch that abi digests will differ between 64-bit and 32-bit platforms

That was exactly the reason, lifted the constraints.

frozen-abi/src/stable_abi.rs
puhtaytow: Yeah, im still working on it :sweat_smile: 

The constraint is a trick to filter out `usize` from blanket implementation, so i could avoid conflicting impls. Agreed it does look bad, ill try to figure out a better way.

frozen-abi/src/lib.rs
```diff
@@ -113,7 +121,10 @@
 
 #![allow(incomplete_features)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
-#![cfg_attr(feature = "frozen-abi", feature(specialization))]
+#![cfg_attr(
+    feature = "frozen-abi",
+    feature(auto_traits, negative_impls, specialization)
```
kskalski: this can be reverted now, right?

frozen-abi-macro/src/lib.rs
```diff
@@ -48,10 +48,14 @@ pub fn derive_stable_abi(item: TokenStream) -> TokenStream {
     };
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
 
-    let expanded = quote! {
-        #[automatically_derived]
-        impl #impl_generics ::solana_frozen_abi::stable_abi::StableAbi for #ident #ty_generics #where_clause {}
-    };
+        let expanded = quote! {
```
kskalski: fix identation?

frozen-abi/src/lib.rs
```diff
@@ -29,11 +29,19 @@
 //! Deriving `StableAbi` adds:
 //!
 //! ```rust,ignore
-//! impl ::solana_frozen_abi::stable_abi::StableAbi for MyType {}
+//! impl ::solana_frozen_abi::stable_abi::StableAbi for MyType {
+//!     fn random(rng: &mut impl ::solana_frozen_abi::rand::RngCore) -> Self
+//!     where
+//!         ::solana_frozen_abi::rand::distr::StandardUniform:
+//!             ::solana_frozen_abi::rand::distr::Distribution<Self>,
```
kskalski: we no longer have this constraint, right?

frozen-abi/src/stable_abi.rs
```diff
@@ -1,11 +1,49 @@
...
+
+impl_stable_abi_via_standard_uniform!(
+    u8, u16, u32, u64, u128,
+    i8, i16, i32, i64, i128,
+    f32, f64, bool, char,
+);
+
+impl StableAbi for usize {
+    fn random(rng: &mut (impl RngCore + ?Sized)) -> Self {
+        Self::from_ne_bytes(rng.random::<[u8; std::mem::size_of::<Self>()]>())
```
kskalski: we could have a macro like
```
impl_stable_abi_via_size_of_from_bytes!(from_ne_bytes, usize);
impl_stable_abi_via_size_of_from_bytes!(from_ne_bytes, isize);
```

frozen-abi/src/stable_abi.rs
```diff
@@ -1,11 +1,83 @@
...
+        )*
+    };
+}
+
+macro_rules! impl_stable_abi_via_size_of_from_bytes {
+    ($from_bytes:ident, $($t:ty),* $(,)?) => {
+        $(
+            impl StableAbi for $t {
+                fn random(rng: &mut (impl RngCore + ?Sized)) -> Self {
+                    Self::$from_bytes(rng.random::<[u8; std::mem::size_of::<Self>()]>())
```
kskalski: not sure if there are other issues for using StableAbi in no-std env, but here we could use `core::mem::size_of` (could import `core::mem` in the file)

frozen-abi/src/stable_abi.rs
```diff
@@ -1,11 +1,83 @@
...
+    (A, B, C, D, E, F, G, H, I, J, K),
+    (A, B, C, D, E, F, G, H, I, J, K, L),
+);
+
+impl<T, const N: usize> StableAbi for [T; N]
+where
+    T: StableAbi,
+{
+    fn random(rng: &mut (impl RngCore + ?Sized)) -> Self {
+        std::array::from_fn(|_| T::random(rng))
```
kskalski: import array with `use core::array;`

frozen-abi-macro/src/lib.rs
```diff
@@ -47,10 +47,13 @@ pub fn derive_stable_abi(item: TokenStream) -> TokenStream {
         }
     };
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
-
```
kskalski, line 50: could also get rid of the whiespace diff

kskalski (commented):
should we also implement StableAbi for NonZero types?

could you add a summary on how this affects compatibility and whether existing code using StableAbi is affected?

kskalski:
LGTM

frozen-abi/src/stable_abi.rs
```diff
@@ -353,4 +444,74 @@ mod tests {
...
+    )]
+    struct TestPlatformDependent {
+        a: usize,
+        b: AliasUsize,
+        c: Option<usize>,
+        d: Option<AliasUsize>,
+        e: isize,
+        f: AliasIsize,
+        g: Option<AliasIsize>,
+    }
```
joncinque, line 483: Should this test be gated on `target_pointer_width` of 64?

frozen-abi/src/stable_abi.rs
```diff
@@ -1,17 +1,108 @@
...
+    ($($t:ty),* $(,)?) => {
+        $(
+            impl StableAbi for NonZero<$t> {
+                fn random(rng: &mut (impl RngCore + ?Sized)) -> Self {
+                    Self::new(<$t as StableAbi>::random(rng)).unwrap_or(NonZero::new(1).unwrap())
+                }
+            }
+        )*
+    };
+}
```
joncinque: Sorry if this is a silly question, but why is the fallback behavior needed? It looks like the NonZero types all implement `StandardUniform` already: https://docs.rs/rand/latest/rand/distr/struct.StandardUniform.html#impl-Distribution%3CNonZero%3Ci8%3E%3E-for-StandardUniform

joncinque (commented):
Looks good to me overall, just a couple of questions

frozen-abi/src/stable_abi.rs
puhtaytow: You are right, we could remove the `impl_stable_abi_via_standard_uniform_nonzero` entirely :) thanks!

frozen-abi/src/stable_abi.rs
puhtaytow: Makes sense as it would produce falsepositives on different bit platforms.

joncinque: Sorry, it looks like the change got added to the non-zero tests, which should be fine regardless of width.

And actually, should we also limit this test to `target_endian` = `"little"`?

puhtaytow: omg sorry i didnt notice! added the little endian, but its really cosmetics imo as it seem like only powerpc has big endian among modern-ish platforms and its rather rare might be wrong ofc :) anyway thank you

kskalski: actually, maybe we should use from_le_bytes instead of native, this way endianness won't affect abi_digest

puhtaytow: sounds good we could offload it to compiler

joncinque (approved):
Looks good to me!

brooksprumo:
I'm going to bow out of reviewing, as I really don't know macros very well. I defer to Jon and Kamil, and looks like they both said LGTM, so :shipit:

kskalski (approved):
@brooksprumo sure, just a heads-up - the possible extension here is that we will support collections also, so I wanted you to be familiar with how sampling for basic types works.

PR #718, opened 2026-05-15
zero-copy: Add unaligned arithmetic impl

febo:
### Problem

While the zero-copy crate provides unaligned wrappers for primitive numeric types, it is necessary to manually convert them to perform arithmetic operations.

### Solution

Implement the relevant traits so arithmetic operations with unaligned types are less verbose.

### DIFF
--- a/zero-copy/src/unaligned.rs
+++ b/zero-copy/src/unaligned.rs
@@ -414,6 +414,14 @@ macro_rules! impl_int_conversion {
                 self.partial_cmp(&o)
             }
         }
+        impl core::cmp::PartialOrd<$P> for $P {
+            #[inline(always)]
+            fn partial_cmp(&self, other: &$P) -> Option<core::cmp::Ordering> {
+                let s: $I = (*self).into();
+                let o: $I = (*other).into();
+                s.partial_cmp(&o)
+            }
+        }
     };
 }
 
@@ -423,7 +431,7 @@ macro_rules! impl_int_conversion {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(from = "u16", into = "u16"))]
 #[cfg_attr(feature = "bytemuck", derive(Pod, Zeroable))]
-#[derive(Clone, Copy, Debug, Default, PartialEq, PartialOrd)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
 #[repr(transparent)]
 pub struct U16(pub [u8; 2]);
 impl_int_conversion!(U16, u16);
@@ -434,7 +442,7 @@ impl_int_conversion!(U16, u16);
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(from = "i16", into = "i16"))]
 #[cfg_attr(feature = "bytemuck", derive(Pod, Zeroable))]
-#[derive(Clone, Copy, Debug, Default, PartialEq, PartialOrd)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
 #[repr(transparent)]
 pub struct I16(pub [u8; 2]);
 impl_int_conversion!(I16, i16);
@@ -449,7 +457,7 @@ impl_int_conversion!(I16, i16);
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(from = "u32", into = "u32"))]
 #[cfg_attr(feature = "bytemuck", derive(Pod, Zeroable))]
-#[derive(Clone, Copy, Debug, Default, PartialEq, PartialOrd)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
 #[repr(transparent)]
 pub struct U32(pub [u8; 4]);
 impl_int_conversion!(U32, u32);
@@ -464,7 +472,7 @@ impl_int_conversion!(U32, u32);
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(from = "u64", into = "u64"))]
 #[cfg_attr(feature = "bytemuck", derive(Pod, Zeroable))]
-#[derive(Clone, Copy, Debug, Default, PartialEq, PartialOrd)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
 #[repr(transparent)]
 pub struct U64(pub [u8; 8]);
 impl_int_conversion!(U64, u64);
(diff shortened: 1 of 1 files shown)


zero-copy/src/unaligned.rs
```diff
@@ -58,16 +63,87 @@ impl From<Bool> for bool {
 /// Simple macro for implementing conversion functions between unaligned
 /// integers and standard integers.
 ///
-/// The standard integer types can cause alignment issues when placed in a
-/// `bytemuck::Pod`, so these replacements are usable in all bytemuck `Pod`
-/// types.
+/// When using standard integer types in a struct, it mgith be required
```
grod220:
```diff
- mgith
+ might
```

zero-copy/src/unaligned.rs
```diff
@@ -58,16 +63,87 @@ impl From<Bool> for bool {
...
+/// to add padding to match their alignment requirements. Unaligned types
+/// avoid this since their alignment requirement is `1`.
 #[macro_export]
```
grod220, line 65: Strange case, but with #[macro_export], someone re-using this macro would get an import err from:

```rust
impl<T: Into<$I>> Add<T> for $P {
    type Output = Self;
```

as the imports are global. To prevent that you'd need to use the fully qualified name: `impl<T: Into<$I>> core::ops::Add<T>`

zero-copy/src/unaligned.rs
```diff
@@ -79,6 +155,142 @@ macro_rules! impl_int_conversion {
...
+                *self = *self - rhs;
+            }
+        }
+        impl<T: Into<$I> + Copy> PartialEq<T> for $P {
+            #[inline(always)]
+            fn eq(&self, other: &T) -> bool {
+                let s: $I = (*self).into();
+                let other: $I = (*other).into();
+                s == other
+            }
```
grod220: This is flexible, but we do lose the ability now to do this (after removing `#[derive(PartialEq)]`):

```rust
const ANSWER: U16 = U16::from_primitive(42);

pub fn is_answer(value: U16) -> bool {
    matches!(value, ANSWER)
}
```

Maybe we can keep `#[derive(PartialEq)]` and do something like this instead:

```rust
impl core::cmp::PartialEq<$I> for $P {
    fn eq(&self, other: &$I) -> bool {
        <$I>::from(*self) == *other
    }
}

impl core::cmp::PartialEq<$P> for $I {
    fn eq(&self, other: &$P) -> bool {
        *self == <$I>::from(*other)
    }
}
```

so we'd get things like 

```
U16::from_primitive(42) == 42u16
and
42u16 == U16::from_primitive(42)
```

zero-copy/src/unaligned.rs
```diff
@@ -79,6 +155,142 @@ macro_rules! impl_int_conversion {
...
+            fn add(self, rhs: T) -> Self {
+                let s: $I = self.into();
+                let other: $I = rhs.into();
+                #[allow(
+                    clippy::arithmetic_side_effects,
+                    reason = "add follows primitive integer behavior and wraps on overflow"
+                )]
+                Self::from(s + other)
+            }
+        }
```
grod220: Something like this would fail right?

```rust
let value = U16::from_primitive(41);
let result = value + 1;
```

As 1 is treated as an i32 by default and then `i32: Into<u16>` would fail.

Can we re-work this to get it to work?

zero-copy/src/unaligned.rs
```diff
@@ -392,4 +604,215 @@ mod tests {
             assert_eq!(&value, zero_copy_ref);
         }
     }
+
```
grod220, line 738: Could we get some divide by zero and overflow tests?

zero-copy/src/unaligned.rs
febo: Hmmm, we don't need to use macro_export right? In any case, I will add the qualified name but we should consider removing the export in a future release.

zero-copy/src/unaligned.rs
febo: Good point! I had to remove the use of generic `impl Into` to make it work. 

The only ones that don't work are the type methods `checked_*` and `saturated_*`, e.g.:
```rust
let value = U16::from_primitive(41);
let result = value.saturating_add(1);
```
We could remove the generic there as well, but we will need another method then:
```rust
pub fn saturating_add(self, rhs: u16) -> Self;
pub fn saturating_add_unaligned(self, rhs: U16) -> Self;
```
or the alternative is to qualify the literal:
```rust
let value = U16::from_primitive(41);
let result = value.saturating_add(1u64);
```
I am a little bit inclined to keep it as it is so the literal needs to be "typed". What do you think?

zero-copy/src/unaligned.rs
febo: I added a divide by zero to the `ArithmeticMethod::CheckedDiv` tests. There are a few overflow tests for `CheckedAdd`, `CheckedMul` and `CheckedSub`.  Did you have a particular overflow test in mind?

zero-copy/src/unaligned.rs
febo: Another alternative is to define a trait for `Saturating*` operations, then we can have both implementations. Maybe this would be the best way.

update: just did a small test and it works. The caveat is that client code needs to import the trait.

joncinque (commented):
Looks great to me! Nothing to add on my side, I'll let @grod220 give the approval

zero-copy/src/unaligned.rs
```diff
@@ -88,7 +423,7 @@ macro_rules! impl_int_conversion {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[cfg_attr(feature = "serde", serde(from = "u16", into = "u16"))]
 #[cfg_attr(feature = "bytemuck", derive(Pod, Zeroable))]
-#[derive(Clone, Copy, Debug, Default, PartialEq)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, PartialOrd)]
 #[repr(transparent)]
 pub struct U16(pub [u8; 2]);
```
grod220: I don't think we can derive `PartialOrd` here. This orders the bytes and not the number it represents. Just tried this out via the test:

```rust
#[test]
fn ordering() {
  let u16_255 = U16::from_primitive(255);
  let u16_256 = U16::from_primitive(256);
  // [0, 1] < [255, 0]
  assert!(u16_256 < u16_255);
}
```

and it passes. We need an impl that converts both sides to the number before comparing.

zero-copy/src/unaligned.rs
grod220: Ah missed it! 👍

zero-copy/src/unaligned.rs
grod220: I see. No strong opinion from my end, happy to go with your preference.

zero-copy/src/unaligned.rs
febo: Good catch! I replaced the derive with a manual implementation that converts the values to numbers first.

grod220 (approved):
Nice job!

zero-copy/src/unaligned.rs
febo: Let's leave it as it is, adding a type suffix is probably easier than having to import a trait.

PR #705, opened 2026-04-23
define-syscall: accept trailing comma in macro argument list

at264939-ctrl:
## Summary
Fixes #701 by allowing `define_syscall!` to accept an optional trailing comma in argument lists.

## Changes
- Updated macro patterns in `define-syscall/src/lib.rs` from:
  - `($($arg:ident: $typ:ty),*)`
  to:
  - `($($arg:ident: $typ:ty),* $(,)?)`
- Applied this to all relevant macro arms (with return type and without, in both cfg branches).
- Added regression test: `define-syscall/tests/trailing_comma.rs`.

## Why
Rust function-style argument lists support trailing commas, and `rustfmt` commonly inserts them for multi-line formatting. The macro should match that syntax.

## Testing
- `cargo test -p solana-define-syscall`

PR #719, opened 2026-05-18
define-syscall: allow trailing commas in macros

### LINKED ISSUE #701: `define-syscall!` macro reject trailing comma in argument list
# Description

the macro's `$($arg:ident: $typ:ty),*` pattern accepts comma between args but not after the last one.

# Context

In the definition of the macro https://github.com/anza-xyz/solana-sdk/blob/b71750f2befd31c11ab4a91876768751588836c5/define-syscall/src/lib.rs#L12

if we insert a comma in the last argument while using the macro, we hit an error.

### Example
```rust
solana_define_syscall::define_syscall!(fn sol_curve_multiscalar_mul(
                  curve_id: u64,
                  scalars_addr: *const u8,
                  points_addr: *const u8,
                  points_len: u64,
                  result_point_addr: *mut u8,  <----------- this comma
              ) -> u64);
```

causes
```
   |               ) -> u64);
   |               ^ no rules expected this token in macro call
   |
note: while trying to match meta-variable `$arg:ident`
   |
   |     ($(#[$attr:meta])* fn $name:ident($($arg:ident: $typ:ty),*) -> $ret:ty) => {
   |                                         ^^^^^^^^^^

```

# Solution
Allow an optional trailing comma in the matcher:
```rust
$($arg:ident: $typ:ty),* $(,)? 
```
( In both arms )
# Considerations

Rust's function syntax accepts a trailing comma, and rustfmt even adds one by default on multi-line argument lists, so the macro rejecting it is a little bit weird.

BretasArthur1:
This pull request updates the `define_syscall!` macro to allow for an optional trailing comma in the argument list. 

This change improves flexibility and consistency with Rust's macro conventions.

Closes: #705 Fixes #701

cc: @joncinque

### DIFF
--- a/define-syscall/src/lib.rs
+++ b/define-syscall/src/lib.rs
@@ -9,7 +9,7 @@ pub mod definitions;
 ))]
 #[macro_export]
 macro_rules! define_syscall {
-    ($(#[$attr:meta])* fn $name:ident($($arg:ident: $typ:ty),*) -> $ret:ty) => {
+    ($(#[$attr:meta])* fn $name:ident($($arg:ident: $typ:ty),* $(,)?) -> $ret:ty) => {
         $(#[$attr])*
         #[inline]
         pub unsafe fn $name($($arg: $typ),*) -> $ret {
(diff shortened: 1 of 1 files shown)


BretasArthur1:
Hey @joncinque I didn't add a test in this PR(since it's a simple change), but I tested locally to make sure it's working. If you want I can add one here as well!

joncinque (approved):
Looks great, thanks!

PR #720, opened 2026-05-18
feat: support wincode for sysvar dependencies

kskalski:
Support `wincode` serialization through optional feature.

### DIFF
--- a/slot-hashes/src/lib.rs
+++ b/slot-hashes/src/lib.rs
@@ -40,6 +40,7 @@ pub type SlotHash = (u64, Hash);
     feature = "serde",
     derive(serde_derive::Deserialize, serde_derive::Serialize)
 )]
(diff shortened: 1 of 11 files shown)


slot-hashes/src/lib.rs
```diff
@@ -40,6 +40,7 @@ pub type SlotHash = (u64, Hash);
     feature = "serde",
     derive(serde_derive::Deserialize, serde_derive::Serialize)
 )]
+#[cfg_attr(feature = "wincode", derive(wincode::SchemaWrite, wincode::SchemaRead))]
 #[derive(PartialEq, Eq, Debug, Default)]
 pub struct SlotHashes(Vec<SlotHash>);
```
kskalski, line 45: This will benefit from making `SlotHash` a `repr(C)` tuple struct, since wincode is able to deserialize Vec of such items with single memcopy. However there are public function using `SlotHash`, so this will require a bit of extra work - to do in follow up

grod220 (approved):
LGTM

PR #723, opened 2026-05-19
no-std feature: solana-serialize-utils

grod220:
Putting standard library usage behind a feature flag so consumers can do a no-std build. 

This is a dep of solana-instructions-sysvar which is a dep of token-2022-interface.

### DIFF
--- a/serialize-utils/src/lib.rs
+++ b/serialize-utils/src/lib.rs
@@ -1,11 +1,11 @@
 //! Helpers for reading and writing bytes.
 
-#![cfg_attr(not(feature = "std"), no_std)]
+#![no_std]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![allow(clippy::arithmetic_side_effects)]
 
 extern crate alloc;
-#[cfg(test)]
+#[cfg(feature = "std")]
 extern crate std;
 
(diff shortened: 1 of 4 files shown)


serialize-utils/src/lib.rs
```diff
@@ -1,9 +1,16 @@
 //! Helpers for reading and writing bytes.
 
+#![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![allow(clippy::arithmetic_side_effects)]
-use {solana_pubkey::Pubkey, solana_sanitize::SanitizeError};
 
+extern crate alloc;
+#[cfg(test)]
+extern crate std;
```
joncinque: I believe this needs to be gated on the std feature
```suggestion
#[cfg(feature = "std")]
extern crate std;
```

serialize-utils/src/lib.rs
```diff
@@ -1,9 +1,16 @@
 //! Helpers for reading and writing bytes.
 
+#![cfg_attr(not(feature = "std"), no_std)]
```
joncinque: All of our crates just enable this always, and bring in the std crate if the std feature is enabled
```suggestion
#![no_std]
```

joncinque (commented):
Looks good! Just a small point

joncinque (approved):
Looks good to me!

PR #724, opened 2026-05-19
[ed25519] swtich to cryptograph/solana-ed25519 dep

zz-sol:
This is the first step towards heea integration. 

# performance

```
cargo bench -p solana-signature --bench verify --features=batch-verify --features=parallel
```

**Single Verify**

| Size | main individual | branch individual | Δ | main par_individual | branch par_individual | Δ |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 22.71 µs | 20.17 µs | -11.2% | 22.64 µs | 20.14 µs | -11.0% |
| 2 | 45.05 µs | 40.18 µs | -10.8% | 43.68 µs | 39.73 µs | -9.0% |
| 4 | 91.57 µs | 81.15 µs | -11.4% | 89.21 µs | 86.93 µs | -2.6% |
| 8 | 183.8 µs | 162.4 µs | -11.6% | 103.2 µs | 99.67 µs | -3.4% |
| 16 | 368.5 µs | 325.6 µs | -11.6% | 157.6 µs | 144.9 µs | -8.1% |
| 32 | 746.6 µs | 652.3 µs | -12.6% | 161.5 µs | 146.6 µs | -9.2% |
| 64 | 1.476 ms | 1.307 ms | -11.4% | 245.1 µs | 227.1 µs | -7.4% |
| 128 | 2.947 ms | 2.615 ms | -11.3% | 355.4 µs | 322.6 µs | -9.2% |
| 256 | 5.896 ms | 5.234 ms | -11.2% | 612.3 µs | 572.2 µs | -6.6% |
| 512 | 11.892 ms | 10.503 ms | -11.7% | 1.131 ms | 995.7 µs | -11.9% |

**Batch Verify**

| Size | main batch | branch batch | Δ | main par_batch | branch par_batch | Δ |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 22.65 µs | 20.13 µs | -11.1% | 22.60 µs | 20.10 µs | -11.1% |
| 2 | 44.92 µs | 38.15 µs | -15.1% | 43.91 µs | 38.40 µs | -12.5% |
| 4 | 91.57 µs | 59.76 µs | -34.7% | 89.20 µs | 61.05 µs | -31.6% |
| 8 | 183.6 µs | 103.9 µs | -43.4% | 103.2 µs | 105.8 µs | +2.5% |
| 16 | 225.4 µs | 190.6 µs | -15.4% | 158.8 µs | 194.6 µs | +22.5% |
| 32 | 441.3 µs | 364.8 µs | -17.3% | 235.4 µs | 220.3 µs | -6.4% |
| 64 | 865.4 µs | 727.3 µs | -16.0% | 347.2 µs | 267.4 µs | -23.0% |
| 128 | 1.690 ms | 1.297 ms | -23.3% | 516.6 µs | 359.9 µs | -30.3% |
| 256 | 3.196 ms | 2.432 ms | -23.9% | 597.9 µs | 516.8 µs | -13.6% |
| 512 | 6.129 ms | 4.552 ms | -25.7% | 1.006 ms | 835.2 µs | -16.9% |


zz-sol:
Hmmm CI is failing because solana-ed25519 is not published. will be fixed later

samkim-crypto (commented):
Looks good to me, but since this is consensus critical code (though we are not cutting right away), do you think we can wait on merging this until we have our audit complete?

zz-sol:
sounds good

keypair/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "solana-keypair"
```
zz-sol, line 2: I also updated dependencies for `solana-keypair`. if we still want to use `dalek` for this I can revert my changes

keypair/src/seed_derivable.rs
```diff
@@ -35,15 +43,74 @@ pub fn keypair_from_seed_and_derivation_path(
...
+        mac.update(&[0u8]);
+        mac.update(&secret_key);
+        mac.update(&index.to_bits().to_be_bytes());
+        let bytes = mac.finalize().into_bytes();
+
+        secret_key.copy_from_slice(&bytes[..Keypair::SECRET_KEY_LENGTH]);
+        chain_code.copy_from_slice(&bytes[Keypair::SECRET_KEY_LENGTH..]);
+    }
+
+    Ok(Keypair::new_from_array(secret_key))
```
samkim-crypto: It seems like it would be nice to refactor this logic out into the `ed25519-solana` crate for the next version.

keypair/src/seed_derivable.rs
```diff
@@ -35,15 +43,74 @@ pub fn keypair_from_seed_and_derivation_path(
...
-) -> Result<Keypair, Bip32Error> {
-    let extended = ed25519_dalek_bip32::ExtendedSigningKey::from_seed(seed)
-        .and_then(|extended| extended.derive(&derivation_path))?;
-    Ok(Keypair(extended.signing_key))
+) -> Result<Keypair, Box<dyn error::Error>> {
+    let mut mac = new_hmac_sha512(ED25519_BIP32_NAME)?;
+    mac.update(seed);
+    let bytes = mac.finalize().into_bytes();
+
+    let mut secret_key = [0u8; Keypair::SECRET_KEY_LENGTH];
```
samkim-crypto: I think we should be zeroizing out these temporary buffers after use.

keypair/src/lib.rs
```diff
@@ -92,29 +95,30 @@ impl TryFrom<&[u8]> for Keypair {
...
-            })
-            .map(Self)
+        let keypair_bytes: &[u8; KEYPAIR_LENGTH] = bytes.try_into().map_err(|_| {
+            SignatureError::from_source(String::from(
+                "candidate keypair byte array is the wrong length",
+            ))
+        })?;
+        let mut secret_key = [0u8; Keypair::SECRET_KEY_LENGTH];
+        secret_key.copy_from_slice(&keypair_bytes[..Keypair::SECRET_KEY_LENGTH]);
+        let signing_key = SigningKey::from_bytes(&secret_key);
```
samkim-crypto: I think `SigningKey::from_bytes` already accepts `&[u8; 32]`, so we can avoid a copy by doing the following:

```
let secret_key: &[u8; Keypair::SECRET_KEY_LENGTH] = keypair_bytes[..Keypair::SECRET_KEY_LENGTH]
    .try_into()
    .unwrap();
let signing_key = SigningKey::from_bytes(secret_key);
```

Cargo.toml
```diff
@@ -173,12 +173,11 @@ console = "0.15.10"
 console_error_panic_hook = "0.1.7"
 console_log = "0.2.2"
 criterion = "0.5.1"
+curve25519 = { package = "solana-ed25519", version = "0.2.0", default-features = false, features = ["digest", "precomputed-tables", "rand_core", "zeroize"] }
```
samkim-crypto: Maybe we should just make this `solana-ed25519` instead of specifying `package = "solana-ed25519"` to be consistent with the rest of the toml file.

samkim-crypto (commented):
Just some nits on the code logic itself.

If it is possible, I think it would be better if we break this up into two separate PR, one for the keypair and one for signature.

Can you also post benchmark results before and after this PR for comparison?

keypair/src/seed_derivable.rs
zz-sol: see https://github.com/anza-xyz/cryptography/pull/72

we prob will need to release a solana-ed25519 v0.2.1 for this

zz-sol:
so I have reverted the KeyPair changes -- this will be done in a follow up PR.

As a result, this PR no longer depends on https://github.com/anza-xyz/cryptography/pull/72

samkim-crypto (dismissed):
Looks fine to me, but let's wait in case @joncinque wants to take a look as well.

Also, it would be helpful if you can run the benches on devserver and post the differences in performance in the PR description.

samkim-crypto:
Oh but since we are cutting https://github.com/anza-xyz/cryptography/pull/74, it would be better to use ed25519-solana v2.1 here instead.

zz-sol:
> Can you also post benchmark results before and after this PR for comparison?

added to the PR description

signature/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "solana-signature"
 description = "Solana 64-byte signature type"
 documentation = "https://docs.rs/solana-signature"
-version = "3.4.1"
+version = "3.5.0"
```
joncinque: We don't need to do the version bump here, the publish job handles that

signature/Cargo.toml
```diff
@@ -18,27 +18,34 @@ rustdoc-args = ["--cfg=docsrs"]
...
 
 [dependencies]
 bytemuck = { workspace = true, optional = true }
 bytemuck_derive = { workspace = true, optional = true }
-curve25519-dalek = { workspace = true, optional = true }
-ed25519-dalek = { workspace = true, optional = true }
 five8 = { workspace = true }
 rand = { workspace = true, optional = true }
+rand_chacha_0_3 = { package = "rand_chacha", version = "0.3", default-features = false, optional = true }
+rand_core_0_6 = { package = "rand_core", version = "0.6.4", features = ["getrandom"], optional = true }
```
joncinque: Are there plans to update these to newer versions? These crates are quite old

signature/Cargo.toml
```diff
@@ -48,10 +55,11 @@ wincode = { workspace = true, optional = true }
 bincode = { workspace = true }
 bs58 = { workspace = true, features = ["std"] }
 criterion = { workspace = true }
-curve25519-dalek = { workspace = true }
-ed25519-dalek = { workspace = true, features = ["batch"] }
+rand_chacha_0_3 = { package = "rand_chacha", version = "0.3", default-features = false }
+rand_core_0_6 = { package = "rand_core", version = "0.6.4", features = ["getrandom"] }
```
joncinque: Rather than redefining these deps, it might be easier to just enable the `batch-verify` feature in this crate under `dev-dependencies`

signature/src/lib.rs
```diff
@@ -168,42 +144,45 @@ impl Signature {
...
+            ));
         }
 
-        ed25519_dalek::verify_batch(&message_bytes, &parsed_signatures, &parsed_pubkeys).is_ok()
+        let Ok(rng) = <rand_chacha_0_3::ChaCha20Rng as rand_core_0_6::SeedableRng>::from_rng(
+            rand_core_0_6::OsRng,
+        ) else {
+            return false;
+        };
+        batch.verify(rng).is_ok()
```
joncinque: Including `rand` types in the public API requires us to publish a new breaking release anytime we bump to a new breaking release of `rand`.

This can be fine in some situations -- for example, the rand usage is an implementation detail in this crate, so doesn't require a breaking release for here.

When possible though, we should avoid using external types in our public APIs.

signature/src/lib.rs
```diff
@@ -139,24 +84,55 @@ impl Signature {
...
     }
 
     /// Verifies this signature against `message_bytes` using `pubkey_bytes`.
     ///
     /// `pubkey_bytes` is the signer's Ed25519 public key, and `message_bytes`
     /// is the exact message that was signed. Returns `false` if verification
     /// fails or the public key cannot be parsed.
+    ///
+    /// TODO: switch to zebra since SIMD-0376 is merged.
+    /// This will be in a follow-up PR.
```
joncinque: nit: we shouldn't be adding TODOs. Let's create Github issues to track instead, or mark functions as deprecated.

signature/src/lib.rs
```diff
@@ -139,24 +84,55 @@ impl Signature {
...
     pub(self) fn verify_verbose(
         &self,
         pubkey_bytes: &[u8],
         message_bytes: &[u8],
-    ) -> Result<(), ed25519_dalek::SignatureError> {
-        let publickey = ed25519_dalek::VerifyingKey::try_from(pubkey_bytes)?;
-        let signature = self.0.as_slice().try_into()?;
-        publickey.verify_strict(message_bytes, &signature)
+    ) -> Result<(), curve25519::ed_sigs::Error> {
```
joncinque: I don't know why we did this in the code, but this is a breaking change.

We might be a bit lucky though -- dalek's `SignatureError` is actually a type alias for https://docs.rs/signature/2.2.0/signature/struct.Error.html. Is there a possibility to use one of these generic error types within the solana cryptography crates? That way there's no breakage.

Alternatively, if we do want to break this, let's use the error type from this crate instead

signature/src/lib.rs
```diff
@@ -139,24 +84,55 @@ impl Signature {
...
+        publickey.verify_dalek(&signature, message_bytes)
+    }
+
+    // TODO: remove since SIMD-0376 is merged
+    /// This will be in a follow-up PR.
+    pub(self) fn verify_non_strict_verbose(
+        &self,
+        pubkey_bytes: &[u8],
+        message_bytes: &[u8],
+    ) -> Result<(), curve25519::ed_sigs::Error> {
```
joncinque: Same with this one, let's either avoid breakage by reusing the same error type as dalek, or create a more maintainable break by instead using the error in this crate

joncinque (commented):
Looks good overall! Just a few bits on maintaining this.

Is there a reason to name rename the crate to `curve25519` at https://github.com/anza-xyz/cryptography/blob/91c81f4dddbfe630a96ba139c310de956c536e1a/curve25519/solana-ed25519/Cargo.toml#L23? I find that a bit confusing, especially because the base crate name contains `ed25519`.

We can lean into the solana branding, it's not hard to update callsites. But I might be missing an obvious use-case.

As we've discussed, since this is an immediate change to the cryptography backend that can get picked up by Agave, let's run a validator against mainnet with it, and make sure it behaves as expected.

Thinking a bit more, I'm leaning towards making breaking changes with this release. That'll just entail a lot of crate publishes, but it's doable.

zz-sol:
> Looks good overall! Just a few bits on maintaining this.
> 
> Is there a reason to name rename the crate to `curve25519` at https://github.com/anza-xyz/cryptography/blob/91c81f4dddbfe630a96ba139c310de956c536e1a/curve25519/solana-ed25519/Cargo.toml#L23? I find that a bit confusing, especially because the base crate name contains `ed25519`.


https://github.com/anza-xyz/cryptography/blob/master/curve25519/solana-ed25519/Cargo.toml#L20-L21

the idea was to have a smaller diff from the ed25519-dalek for the ease of code auditing. But i guess now the auditing is done, we can actually remove this and change it back to solana-ed25519

signature/Cargo.toml
zz-sol: we are waiting for the new release of ff/group v0.14.0.
at the moment cryptography uses ff/group v0.13.0 and the rand there is pretty old.

see

https://github.com/anza-xyz/cryptography/issues/17
https://github.com/anza-xyz/cryptography/pull/21

---

updates:

ff and group v0.14.0 are released 1 month ago. I will bump up rand in solana-ed25519

signature/src/lib.rs
zz-sol: Just double checking this:

- `verify_verbose` / `verify_non_strict_verbose` are currently `pub(self)`
- and the public APIs `verify`/`verify_non_strict` return bool. 

Do you think we still need to internalize the error type?

joncinque: I completely missed the `pub(self)` part :disappointed: all good then!

joncinque (commented):
Looks great to me! As mentioned earlier, let's wait until the validator tests are done before merging

samkim-crypto (approved):
LGTM! Let's change the title before merging.

PR #727, opened 2026-05-20
packet: add wincode support

efagerho:
End-to-end support for parsing and serializing packets for gossip in Agave requires
that the solana-packet crate has a wincode serialization path.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -346,7 +346,7 @@ tiny-bip39 = "2.0.0"
 toml = "0.8.23"
 uriparse = "0.6.4"
 wasm-bindgen = "0.2.100"
-wincode = { version = "0.5.0", features = ["derive"], default-features = false }
+wincode = { version = "0.5.4", features = ["derive"], default-features = false }
 zeroize = "1.8.2"
 
 [profile.release]
--- a/packet/src/lib.rs
+++ b/packet/src/lib.rs
@@ -4,6 +4,12 @@
 
 #[cfg(feature = "frozen-abi")]
 use solana_frozen_abi_macro::AbiExample;
+#[cfg(feature = "wincode")]
+use wincode::{
+    config::{self, Configuration},
+    io::Cursor,
+    SchemaRead, SchemaWrite,
+};
 #[cfg(feature = "bincode")]
 use {
     bincode::{Options, Result},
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -346,7 +346,7 @@ tiny-bip39 = "2.0.0"
 toml = "0.8.23"
 uriparse = "0.6.4"
 wasm-bindgen = "0.2.100"
-wincode = { version = "0.5.0", features = ["derive"], default-features = false }
+wincode = { version = "0.5.4", features = ["derive"], default-features = false }
```
kskalski, line 349: could you rebase? I think this is already upstream

packet/src/lib.rs
```diff
@@ -212,6 +226,55 @@ impl Packet {
...
+    {
+        let bytes = self
+            .data(index)
+            .ok_or(wincode::ReadError::Custom("invalid packet slice"))?;
+        let mut reader = bytes;
+        let value = <T as SchemaRead<'de, PacketWincodeConfig>>::get(reader.by_ref())?;
+        if reader.is_empty() {
+            Ok(value)
+        } else {
+            Err(wincode::ReadError::Custom(
```
kskalski: we could use `wincode::config::deserialize_exact` here

Cargo.toml
efagerho: Latest version still has 0.5.0: https://github.com/anza-xyz/solana-sdk/blob/master/Cargo.toml#L349

packet/src/lib.rs
efagerho: Sure. This seems to require 0.5.4 too.

packet/src/lib.rs
```diff
@@ -32,6 +38,14 @@ static_assertions::const_assert_eq!(PACKET_DATA_SIZE, 1232);
 ///   8 bytes is the size of the fragment header
 pub const PACKET_DATA_SIZE: usize = 1280 - 40 - 8;
 
+// Remaining generics default to BincodeLen / LittleEndian / FixInt / u32 tags,
+// matching bincode's wire format so wincode output is byte-identical.
+#[cfg(feature = "wincode")]
+type PacketWincodeConfig = Configuration<true, PACKET_DATA_SIZE>;
+#[cfg(feature = "wincode")]
+const PACKET_WINCODE_CONFIG: PacketWincodeConfig =
+    Configuration::default().with_preallocation_size_limit::<PACKET_DATA_SIZE>();
```
kskalski, line 47: would `.. PacketWincodeConfig: = Configuraton::new();` work here?

kskalski (commented):
So this change only adds serialization support for populating and reading data from packet, not the serialization of the whole packet itself. I would change PR title and description to clarify that.

packet/src/lib.rs
efagerho: It does now.

kskalski: is that in a commit I'm not seeing here?

efagerho:
Decided to just short circuit this package directly in the agave code, so we can drop an external dependency there.

PR #730, opened 2026-05-20
frozen-abi: add collections support thru context in StableAbi

puhtaytow:
### Description

Collections in `StableAbi` were supported thru sampling expression using `stable_abi_sample(with = "<code>"` and thru manual implementation of rand distribution, which demanded more verbose treatment even in cases where strict control over collection wasn't needed.

### Summary of changes
- added implementations for Vec, VecDeque, BTreeMap, HashMap, HashSet


frozen-abi/src/stable_abi.rs
```diff
@@ -93,25 +101,99 @@ impl<T, const N: usize> StableAbi for [T; N]
...
+}
+
+impl<K, V> StableAbi<Len> for BTreeMap<K, V>
+where
+    K: StableAbi<Len> + Ord,
+    V: StableAbi<Len>,
+{
+    fn random(rng: &mut (impl RngCore + ?Sized), ctx: Len) -> Self {
+        (0..ctx.0)
+            .map(|_| (K::random(rng, ctx), V::random(rng, ctx)))
```
puhtaytow: not sure if the ctx should be passed here as well

puhtaytow:
@kskalski if you can have a look whenever you find time :pray:

PR #731, opened 2026-05-20
frozen-abi: replace linked_stable_abi_pair macro  by constants

puhtaytow:
### Issue

Since frozen-abi macros moved to expression in atributes, now we could use constants instead of macro

### Summary of changes
- replaced `linked_stable_abi_pair` macro by constants

### DIFF
--- a/frozen-abi/src/stable_abi.rs
+++ b/frozen-abi/src/stable_abi.rs
@@ -114,89 +114,70 @@ mod tests {
         std::collections::{BTreeMap, VecDeque},
     };
 
-    // Keep the bincode and wincode test fixtures structurally identical so their
-    // derived `test_abi_digest` checks enforce one shared ABI digest across serializers.
-    #[rustfmt::skip]
-    macro_rules! linked_stable_abi_pair {
-        (
-            api_digest_wincode = $api_wincode:literal,
-            api_digest_bincode = $api_bincode:literal,
-            abi_digest = $abi:literal,
-        ) => {
-            #[derive(Debug, serde_derive::Serialize, wincode::SchemaWrite)]
-            #[cfg_attr(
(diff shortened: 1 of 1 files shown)


frozen-abi/src/stable_abi.rs
```diff
@@ -114,90 +114,67 @@ mod tests {
...
-            #[cfg_attr(
+    #[derive(Debug, serde_derive::Serialize)]
+    #[cfg_attr(
                 feature = "frozen-abi",
                 derive(
                     solana_frozen_abi_macro::AbiExample,
                     solana_frozen_abi_macro::StableAbi
                 ),
                 solana_frozen_abi_macro::frozen_abi(
-                    api_digest = $api_bincode,
```
kskalski, line 167: should we still keep the api_digest tested?

puhtaytow: so previously it was there, because api_digest was mandatory, but those tests are about stableabi, not api stability; i could bring it back but it kinda makes to me logical sense to ditch those using this occasion

kskalski: I see, though I think it's still useful here to demonstrate how abi_digest stays the same while api_digest differs - actually I do wonder why this is the case, but I suppose api_digest for wincode is some custom impl that doesn't use serde?

kskalski (dismissed):

frozen-abi/src/stable_abi.rs
puhtaytow: Good point. The api compatibility tests work differently to StableAbi. The api ones hash together canonical representation of types and its implementations, so serde::Serialize derive reflect in different impl set than wincode::SchemaWrite.

--edit 
actually im talking partially nonsense, because there is only special treating (seem like) only for `__SerializeWith<T>` and other than that its about type/field naming etc which in this case would be only type name by looks.

puhtaytow: It's also good moment to mention that AbiExample/Digester should be ported to use Wincode.

kskalski (approved):

PR #732, opened 2026-05-20
expose sec-p params

zz-sol:
as discussed here https://github.com/solana-program/secp256k1/pull/1#discussion_r3270320618

we want to expose `SecpSignatureOffsets` via `#[cfg(target_os = "solana")]` so downstream can reuse the same struct without new definitions.

this will help to ensure BFP and the precompile share an identical API

### DIFF
--- a/secp256k1-program/src/lib.rs
+++ b/secp256k1-program/src/lib.rs
@@ -799,7 +799,7 @@ use serde_derive::{Deserialize, Serialize};
 ))]
 use solana_instruction::Instruction;
 #[cfg(not(any(target_os = "solana", target_arch = "bpf")))]
-use {digest::Digest, solana_signature::error::Error};
+use solana_signature::error::Error;
 
 pub const SECP256K1_PUBKEY_SIZE: usize = 64;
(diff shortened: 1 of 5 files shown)


secp256k1-program/src/lib.rs
```diff
@@ -893,6 +906,7 @@ pub fn new_secp256k1_instruction_with_signature(
 }
 
 /// Creates an Ethereum address from a secp256k1 public key.
+#[cfg(not(any(target_os = "solana", target_arch = "bpf")))]
 pub fn eth_address_from_pubkey(
```
samkim-crypto, line 905: What about exposing this function too? For the keccak hash, I think we can use [keccak-hasher](https://github.com/anza-xyz/solana-sdk/blob/master/keccak-hasher/src/lib.rs#L34), which will either invoke `sha3::Keccak256` or the `sol_keccak256` syscall depending on the target.

samkim-crypto (commented):
Looks good to me! I have not checked, but I think we should be able to export out `eth_address_from_pubkey` for bpf target as well using the keccak syscall.

samkim-crypto (approved):
Looks good to me!

PR #734, opened 2026-05-21
zero-copy: Remove custom PartialEq and ParialOrd impl

### LINKED ISSUE #733: [zero-copy] New arithmetic impls cause type inference failures (breaking change)
#718 introduced a `impl PartialOrd<I64> for i64`. This means crates relying on type inference stopped compiling, e.g.
```
let value: i64 = ...;
if value >= something.into()  // previously inferred to be `i64`, now could be `I64` or `i64`
```

This affects at least `spl-token-2022-interface`; creating a new project with that crate will no longer compile (existing projects with a lockfile are of course unaffected).

```rs
    Checking spl-token-2022-interface v3.0.1
error[E0283]: type annotations needed
  --> ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/spl-token-2022-interface-3.0.1/src/extension/scaled_ui_amount/mod.rs:66:70
   |
66 |         if unix_timestamp >= self.new_multiplier_effective_timestamp.into() {
   |                           -- type must be known at this point        ^^^^
   |
   = note: multiple `impl`s satisfying `i64: PartialOrd<_>` found in the following crates: `core`, `solana_zero_copy`:
           - impl PartialOrd for i64;
           - impl PartialOrd<I64> for i64;
help: try using a fully qualified path to specify the expected types
   |
66 -         if unix_timestamp >= self.new_multiplier_effective_timestamp.into() {
66 +         if unix_timestamp >= <I64 as Into<T>>::into(self.new_multiplier_effective_timestamp) {
   |

```

cc @febo

febo:
### Problem

PR #718 added custom implementations for `PartialEq` and `PartialOrd` to work with both primitive and unaligned types, but this creates problems with existing code that is using `into()` to do the conversion. E.g.:
```rust
if epoch >= self.newer_transfer_fee.epoch.into() {
    &self.newer_transfer_fee
} else {
    &self.older_transfer_fee
}
```
The code above does not compile.

### Solution

Remove the custom implementations, reverting to the previous behaviour.

Fixes #733

### DIFF
--- a/zero-copy/src/unaligned.rs
+++ b/zero-copy/src/unaligned.rs
@@ -386,42 +386,6 @@ macro_rules! impl_int_conversion {
                 *self = *self - rhs;
             }
         }
-        impl core::cmp::PartialEq<$I> for $P {
-            #[inline(always)]
-            fn eq(&self, other: &$I) -> bool {
-                let s: $I = (*self).into();
-                s == *other
-            }
-        }
-        impl core::cmp::PartialEq<$P> for $I {
-            #[inline(always)]
-            fn eq(&self, other: &$P) -> bool {
-                let o: $I = (*other).into();
-                *self == o
-            }
-        }
-        impl core::cmp::PartialOrd<$I> for $P {
-            #[inline(always)]
-            fn partial_cmp(&self, other: &$I) -> Option<core::cmp::Ordering> {
-                let s: $I = (*self).into();
-                s.partial_cmp(other)
-            }
-        }
-        impl core::cmp::PartialOrd<$P> for $I {
-            #[inline(always)]
-            fn partial_cmp(&self, other: &$P) -> Option<core::cmp::Ordering> {
-                let o: $I = (*other).into();
-                self.partial_cmp(&o)
-            }
-        }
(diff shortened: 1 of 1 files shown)


zero-copy/src/unaligned.rs
```diff
@@ -386,42 +386,6 @@ macro_rules! impl_int_conversion {
...
-            }
-        }
-        impl core::cmp::PartialOrd<$P> for $P {
-            #[inline(always)]
-            fn partial_cmp(&self, other: &$P) -> Option<core::cmp::Ordering> {
-                let s: $I = (*self).into();
-                let o: $I = (*other).into();
-                s.partial_cmp(&o)
-            }
-        }
```
joncinque, line 424: We could leave this one if you want

joncinque (dismissed):

zero-copy/src/unaligned.rs
febo: I think it will be useful – added it back in https://github.com/anza-xyz/solana-sdk/pull/734/commits/32a1a9b58f4b416f2dc4680458dce16e3d46c7e1

joncinque (approved):

PR #735, opened 2026-05-25
program-entrypoint: document BumpAllocator arena alignment

### LINKED ISSUE #712: BumpAllocator::new and BumpAllocator::with_fixed_address_range do not check or mention required alignment
The `BumpAllocator::new(arena: &mut [u8])` doesn't check whether arena  is 8-byte-aligned before it's dereferenced as a `*mut usize` , which could lead to UB:

https://github.com/anza-xyz/solana-sdk/blob/e1e1f4bf31f9c377001f909e322edf85cc989ee5/program-entrypoint/src/lib.rs#L328

This should either be mentioned in the safety contract of the function, or it should be checked by the function itself.
Similarly, the documentation comment `BumpAllocator::with_fixed_address_range(start: usize, len: usize)` does not mention the alignment requirement of `start`:

https://github.com/anza-xyz/solana-sdk/blob/e1e1f4bf31f9c377001f909e322edf85cc989ee5/program-entrypoint/src/lib.rs#L348

dotsvm:
`BumpAllocator::new` casts `arena` to `*mut usize` and dereferences it, which is UB if the arena isn't `usize`-aligned. Same requirement applies to `with_fixed_address_range`'s `start`. Document both in the safety contract and add a debug_assert in `new` matching the existing length check.

Fixes #712.

### DIFF
--- a/program-entrypoint/src/lib.rs
+++ b/program-entrypoint/src/lib.rs
@@ -645,7 +645,9 @@ mod test {
         };
         let unaligned = &mut bytes[1..];
         assert_ne!(
-            unaligned.as_ptr().align_offset(core::mem::align_of::<usize>()),
+            unaligned
+                .as_ptr()
+                .align_offset(core::mem::align_of::<usize>()),
             0,
             "test setup: slice must be misaligned"
         );
--- a/program-entrypoint/src/lib.rs
+++ b/program-entrypoint/src/lib.rs
@@ -312,6 +312,11 @@ impl BumpAllocator {
     /// As long as BumpAllocator or any of its allocations are alive,
     /// writing into or deallocating the arena will cause UB.
     ///
+    /// The start of `arena` must be aligned to `align_of::<usize>()`, since
+    /// the first `size_of::<usize>()` bytes are dereferenced as a `*mut usize`
+    /// to store the bump pointer. An unaligned arena results in undefined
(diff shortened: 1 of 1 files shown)


program-entrypoint/src/lib.rs
```diff
@@ -619,4 +632,23 @@ mod test {
...
+                backing.as_mut_ptr() as *mut u8,
+                backing.len() * size_of::<usize>(),
+            )
+        };
+        let unaligned = &mut bytes[1..];
+        debug_assert_ne!(
+            unaligned.as_ptr().align_offset(core::mem::align_of::<usize>()),
+            0,
+            "test setup: slice must be misaligned"
+        );
```
joncinque: nit: since this is a test, how about just doing `assert_ne!`?

joncinque (commented):
Sorry for the lateness on this review, but the change looks good! For some reason, CI didn't run, so could you rebase along with applying the change I suggested?

dotsvm:
> Sorry for the lateness on this review, but the change looks good! For some reason, CI didn't run, so could you rebase along with applying the change I suggested?

thanks for the review, ser. I've applied the suggested changes.

joncinque (commented):
Looks like this just needs a formatting pass, can you run `./cargo nightly fmt --all` from the root of the repo and commit those changes?

joncinque (approved):
Thanks!

PR #736, opened 2026-05-25
hash: derive StableAbi for hash

puhtaytow:
### Issue

Hash as commonly used type doesn't implement StableAbi and as consequence blocking derivation on other types including it.

### Summary of changes
- moved to block frozen-abi macro imports
- derived StableAbi and StableAbiSample on Hash type

### DIFF
--- a/hash/src/lib.rs
+++ b/hash/src/lib.rs
@@ -20,6 +20,8 @@ use solana_sanitize::Sanitize;
 extern crate alloc;
 #[cfg(feature = "borsh")]
 use alloc::string::ToString;
+#[cfg(feature = "frozen-abi")]
+use solana_frozen_abi_macro::{AbiExample, StableAbi, StableAbiSample};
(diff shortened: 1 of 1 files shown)


hash/src/lib.rs
```diff
@@ -35,7 +37,7 @@ pub const MAX_BASE58_LEN: usize = 44;
 ///
 /// [SHA-256]: https://en.wikipedia.org/wiki/SHA-2
 /// [blake3]: https://github.com/BLAKE3-team/BLAKE3
-#[cfg_attr(feature = "frozen-abi", derive(solana_frozen_abi_macro::AbiExample))]
+#[cfg_attr(feature = "frozen-abi", derive(AbiExample, StableAbiSample, StableAbi))]
```
kskalski, line 40: I think I will always be confused by the difference between those 2 derives, we should work towards unifying them or giving better names

kskalski (approved):

PR #738, opened 2026-05-26
frozen-abi: introduce context to StableAbi

puhtaytow:
### Description

This PR improves the StableAbi trait by extending it with a generic context, which can be used to provide additional data used in the generation process.

### Summary of changes
- extended StableAbi trait to operate on generic context
- added random_with_context() to StableAbi trait and delegate random()
- fixed dependent macros; impl_stable_abi_via_standard_uniform, impl_stable_abi_via_size_of_from_bytes

---
https://github.com/anza-xyz/solana-sdk/pull/739 PR with collections

### DIFF
--- a/frozen-abi/src/lib.rs
+++ b/frozen-abi/src/lib.rs
@@ -14,7 +14,7 @@
 //!
 //! The macro would generate the `test_abi_digest` test that verifies binary layout stability:
 //! - Initializes a deterministic random number generator with fixed seed
-//! - Generates 10_000 instances of the type via `StableAbi::random()`
+//! - Generates 10_000 instances of the type
 //! - Serializes each instance
 //! - Hashes all serialized bytes together
 //! - Compares the resulting hash against the provided in `abi_digest` attribute
@@ -30,7 +30,10 @@
 //!
 //! ```rust,ignore
 //! impl ::solana_frozen_abi::stable_abi::StableAbi for MyType {
-//!     fn random(rng: &mut (impl ::solana_frozen_abi::rand::RngCore + ?Sized)) -> Self {
+//!     fn random_with_context(
+//!         rng: &mut (impl ::solana_frozen_abi::rand::RngCore + ?Sized),
+//!         _ctx: (),
+//!     ) -> Self {
 //!         ::solana_frozen_abi::rand::Rng::random::<Self>(rng)
 //!     }
 //! }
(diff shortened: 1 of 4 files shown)


frozen-abi/src/lib.rs
```diff
@@ -60,6 +63,25 @@
...
 //!
+//! `StableAbiSample` also supports passing an explicit context to context-aware types:
+//!
+//! ```rust,ignore
+//! #[derive(StableAbi, StableAbiSample)]
+//! #[frozen_abi(abi_digest = "...")]
+//! struct MyTypeWithCtx {
+//!     #[stable_abi_sample(ctx = SequenceLenMax(32))]
+//!     a: Vec<u8>,
+//!     #[stable_abi_sample(ctx = SequenceLenMax(8))]
```
kskalski: this should be part of the PR adding collections support, here we shouldn't include the context structs related to such change or helper macros allowing to delegate from one collection impl to another

frozen-abi/src/stable_abi/context.rs
```diff
@@ -0,0 +1,21 @@
+/// Context for `StableAbi<...>` impls on sequence-like collections
```
kskalski: let's skip this file too in this PR

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -33,16 +33,47 @@ macro_rules! impl_stable_abi_for_tuples {
...
-                fn random(rng: &mut (impl RngCore + ?Sized)) -> Self {
+                fn random_with_context(rng: &mut (impl RngCore + ?Sized), _ctx: ()) -> Self {
                     ($($t::random(rng),)+)
                 }
             }
         )*
     };
 }
 
+macro_rules! impl_stable_abi_via_default_ctx {
```
kskalski: and those macros

kskalski:
LGTM,
@cpubot could you take a look here? We took the inspiration from original wincode Context design, which should work better here on a simpler trait. The primary intention is to parametrize the way sample collections are created for wincode vs bincode / stability digests of wire format (https://github.com/anza-xyz/solana-sdk/pull/739)

I have some ideas for additional context implementations:
* `SequenceSorted` -> this could be used to define custom sorting of input to create sample collections, such that switching BTreeSet vs Vec would yield the same wire data
* `Nest(outer_ctx, t_ctx)` -> allow using separate context for the immediate collection (or any type) in a field and its nested `T` type, e.g. `Vec<Vec<u8>>` could use different sampling strategy for outer and inner Vec - initial implementations of context will be shallow, i.e. only affect immediate outer type

I wonder if you can suggest other ideas for this design or if you see any downsides here. Separate trait could work here too, but seems unnecessary and adding complexity.

kskalski (approved):

kskalski:
@joncinque wanted to get your stamp too before merging, especially that I'm marked co-author and this is community PR :)

joncinque (approved):
Makes sense to me!

puhtaytow:
Thank you guys! :)

PR #739, opened 2026-05-26
frozen-abi: extend support for collections in StableAbi

puhtaytow:
### Description

Collections in StableAbi were supported thru sampling expression using `stable_abi_sample(with = "<code>")` and thru manual implementation of rand distribution, which demanded highly verbose treatment in every case.

### Summary of changes
- added support for Vec, VecDeque, HashMap, HashSet, BtreeMap, BtreeSet
- added two context types SequenceLenRange and SequenceLenMax
- added new macro impl_with_context_via
- added test macros, to generate equivalent test types for collections, with split for multiple and single element
- updated docs

### DIFF
--- a/frozen-abi/src/stable_abi/context.rs
+++ b/frozen-abi/src/stable_abi/context.rs
@@ -1,0 +1,68 @@
+use core::ops::{Bound, RangeBounds};
+
+/// Context for `StableAbi<...>` impls on sequence-like collections
+/// (`Vec`, `VecDeque`, `HashMap`, `BTreeMap`) that bounds the sampled length
+/// to an inclusive range `min..=max`.
+///
+/// The collection's `random_with_context` draws a length uniformly from
+/// `min..=max` and produces that many elements. `min == max` pins the length
+/// exactly; `min == 0` means "up to `max` elements" (equivalent to
+/// `SequenceLenMax(max)`).
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub struct SequenceLenRange {
+    pub min: usize,
+    pub max: usize,
+}
+
+impl SequenceLenRange {
+    pub fn new(range: impl RangeBounds<usize>) -> Self {
+        let min = match range.start_bound() {
+            Bound::Included(&min) => min,
+            Bound::Excluded(&min) => min.checked_add(1).expect("invalid min in range"),
+            Bound::Unbounded => 0,
+        };
+        let max = match range.end_bound() {
+            Bound::Included(&max) => max,
+            Bound::Excluded(&max) => max.checked_sub(1).expect("invalid max in range"),
+            Bound::Unbounded => usize::MAX,
+        };
+        assert!(min <= max, "invalid sequence length range");
+
+        Self { min, max }
+    }
+}
+
+/// Context for `StableAbi<...>` impls on sequence-like collections
+/// that caps the sampled length at `0..=N`.
+///
+/// Equivalent to `SequenceLenRange { min: 0, max: N }`; the per-collection
+/// impl delegates to the `SequenceLenRange` impl.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct SequenceLenMax(pub usize);
+
+impl From<SequenceLenMax> for SequenceLenRange {
+    fn from(max: SequenceLenMax) -> Self {
+        Self::new(0..=max.0)
+    }
+}
+
+macro_rules! impl_with_context_via {
+    (
+        impl<$($g:tt),+ $(,)?> StableAbi<$ctx:ty> for $self_ty:ty
+        where { $($wc:tt)* },
+        |$bind:pat_param| $convert:expr $(,)?
+    ) => {
+        impl<$($g),+> StableAbi<$ctx> for $self_ty
+        where $($wc)*
+        {
+            fn random_with_context(
+                rng: &mut (impl RngCore + ?Sized),
+                $bind: $ctx,
+            ) -> Self {
+                Self::random_with_context(rng, $convert)
+            }
+        }
+    };
+}
+
+pub(crate) use impl_with_context_via;
--- a/frozen-abi/src/stable_abi/impls.rs
+++ b/frozen-abi/src/stable_abi/impls.rs
@@ -1,8 +1,18 @@
 use {
-    crate::stable_abi::StableAbi,
+    crate::stable_abi::{
+        context::{impl_with_context_via, SequenceLenMax, SequenceLenRange},
+        StableAbi,
+    },
     core::{array, num::NonZero},
     rand::{Rng, RngCore},
+    std::{
+        collections::{BTreeMap, BTreeSet, HashMap, HashSet, VecDeque},
+        hash::{BuildHasher, Hash},
+    },
 };
+
+const DEFAULT_COLLECTION_MAX_SAMPLE_LEN: usize = 5;
+const DEFAULT_COLLECTION_MAX_SAMPLE_LEN_NON_DETERMINISTIC_ORDER: usize = 1;
 
 macro_rules! impl_stable_abi_via_standard_uniform {
     ($($t:ty),* $(,)?) => {
@@ -104,11 +114,154 @@ where
     }
 }
 
+// keep at most one element to mitigate iteration order differences
+impl_with_context_via! {
+    impl<K, V, S> StableAbi<()> for HashMap<K, V, S>
+    where { K: StableAbi + Eq + Hash, V: StableAbi, S: BuildHasher + Default },
+    |_| SequenceLenRange::new(0..=DEFAULT_COLLECTION_MAX_SAMPLE_LEN_NON_DETERMINISTIC_ORDER),
+}
+
+impl<K, V, S> StableAbi<SequenceLenRange> for HashMap<K, V, S>
+where
+    K: StableAbi + Eq + Hash,
+    V: StableAbi,
+    S: BuildHasher + Default,
+{
+    fn random_with_context(rng: &mut (impl RngCore + ?Sized), ctx: SequenceLenRange) -> Self {
+        let len = rng.random_range(ctx.min..=ctx.max);
+        (0..len).map(|_| (K::random(rng), V::random(rng))).collect()
+    }
+}
+
+impl_with_context_via! {
+    impl<K, V, S> StableAbi<SequenceLenMax> for HashMap<K, V, S>
+    where { K: StableAbi + Eq + Hash, V: StableAbi, S: BuildHasher + Default },
+    |ctx| SequenceLenRange::from(ctx),
+}
+
+// keep at most one element to mitigate iteration order differences
+impl_with_context_via! {
+    impl<T, S> StableAbi<()> for HashSet<T, S>
+    where { T: StableAbi + Eq + Hash, S: BuildHasher + Default },
+    |_| SequenceLenRange::new(0..=DEFAULT_COLLECTION_MAX_SAMPLE_LEN_NON_DETERMINISTIC_ORDER),
+}
+
+impl<T, S> StableAbi<SequenceLenRange> for HashSet<T, S>
+where
+    T: StableAbi + Eq + Hash,
+    S: BuildHasher + Default,
+{
+    fn random_with_context(rng: &mut (impl RngCore + ?Sized), ctx: SequenceLenRange) -> Self {
+        let len = rng.random_range(ctx.min..=ctx.max);
+        (0..len).map(|_| T::random(rng)).collect()
+    }
+}
+
+impl_with_context_via! {
+    impl<T, S> StableAbi<SequenceLenMax> for HashSet<T, S>
+    where { T: StableAbi + Eq + Hash, S: BuildHasher + Default },
+    |ctx| SequenceLenRange::from(ctx),
+}
+
+impl_with_context_via! {
+    impl<T> StableAbi<()> for Vec<T>
+    where { T: StableAbi },
+    |_| SequenceLenRange::new(0..=DEFAULT_COLLECTION_MAX_SAMPLE_LEN),
+}
+
+impl<T> StableAbi<SequenceLenRange> for Vec<T>
+where
+    T: StableAbi,
+{
+    fn random_with_context(rng: &mut (impl RngCore + ?Sized), ctx: SequenceLenRange) -> Self {
+        let len = rng.random_range(ctx.min..=ctx.max);
+        (0..len).map(|_| T::random(rng)).collect()
+    }
+}
+
+impl_with_context_via! {
+    impl<T> StableAbi<SequenceLenMax> for Vec<T>
+    where { T: StableAbi },
+    |ctx| SequenceLenRange::from(ctx),
+}
+
+impl_with_context_via! {
+    impl<T> StableAbi<()> for VecDeque<T>
+    where { T: StableAbi },
+    |_| SequenceLenRange::new(0..=DEFAULT_COLLECTION_MAX_SAMPLE_LEN),
+}
+
+impl<T> StableAbi<SequenceLenRange> for VecDeque<T>
+where
+    T: StableAbi,
+{
+    fn random_with_context(rng: &mut (impl RngCore + ?Sized), ctx: SequenceLenRange) -> Self {
+        let len = rng.random_range(ctx.min..=ctx.max);
+        (0..len).map(|_| T::random(rng)).collect()
+    }
+}
+
+impl_with_context_via! {
+    impl<T> StableAbi<SequenceLenMax> for VecDeque<T>
+    where { T: StableAbi },
+    |ctx| SequenceLenRange::from(ctx),
+}
+
+impl_with_context_via! {
+    impl<K, V> StableAbi<()> for BTreeMap<K, V>
+    where { K: StableAbi + Ord, V: StableAbi },
+    |_| SequenceLenRange::new(0..DEFAULT_COLLECTION_MAX_SAMPLE_LEN),
+}
+
+impl<K, V> StableAbi<SequenceLenRange> for BTreeMap<K, V>
+where
+    K: StableAbi + Ord,
+    V: StableAbi,
+{
+    fn random_with_context(rng: &mut (impl RngCore + ?Sized), ctx: SequenceLenRange) -> Self {
+        let len = rng.random_range(ctx.min..=ctx.max);
+        (0..len).map(|_| (K::random(rng), V::random(rng))).collect()
+    }
+}
+
+impl_with_context_via! {
+    impl<K, V> StableAbi<SequenceLenMax> for BTreeMap<K, V>
+    where { K: StableAbi + Ord, V: StableAbi },
+    |ctx| SequenceLenRange::from(ctx),
+}
+
+impl_with_context_via! {
+    impl<T> StableAbi<()> for BTreeSet<T>
+    where { T: StableAbi + Ord },
+    |_| SequenceLenRange::new(0..=DEFAULT_COLLECTION_MAX_SAMPLE_LEN),
+}
+
+impl<T> StableAbi<SequenceLenRange> for BTreeSet<T>
+where
+    T: StableAbi + Ord,
+{
+    fn random_with_context(rng: &mut (impl RngCore + ?Sized), ctx: SequenceLenRange) -> Self {
+        let len = rng.random_range(ctx.min..=ctx.max);
+        (0..len).map(|_| T::random(rng)).collect()
+    }
+}
+
+impl_with_context_via! {
+    impl<T> StableAbi<SequenceLenMax> for BTreeSet<T>
+    where { T: StableAbi + Ord },
+    |ctx| SequenceLenRange::from(ctx),
+}
+
+impl StableAbi for () {
+    fn random_with_context(_rng: &mut (impl RngCore + ?Sized), _ctx: ()) -> Self {}
+}
+
 #[cfg(all(test, feature = "frozen-abi"))]
 mod tests {
     use {
+        crate::stable_abi::context::{SequenceLenMax, SequenceLenRange},
         core::num::NonZero,
-        std::collections::{BTreeMap, VecDeque},
+        std::collections::{BTreeMap, BTreeSet, HashMap, HashSet, VecDeque},
     };
 
     const ABI_SHARED_WINCODE_VS_BINCODE: &str = "AgNkEpErnFBuy7iTAEUUAC1fbvokEkhbsfFnx4DtXAvY";
@@ -281,9 +434,7 @@ mod tests {
         )]
         a: Vec<bool>,
         // Keep a single entry so HashMap iteration order cannot affect the digest.
-        #[stable_abi_sample(
-            with = "std::collections::HashMap::from_iter([(rng.random(), rng.random())])"
-        )]
+        #[stable_abi_sample(with = "HashMap::from_iter([(rng.random(), rng.random())])")]
         b: std::collections::HashMap<u64, bool>,
         #[stable_abi_sample(with = "rng.random::<u64>()")]
         c: u64,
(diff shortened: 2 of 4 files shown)


frozen-abi/src/stable_abi/impls.rs
```diff
@@ -43,6 +50,37 @@ macro_rules! impl_stable_abi_for_tuples {
     };
 }
 
+macro_rules! impl_stable_abi_via_default_ctx {
```
kskalski: I suspect we could remove this macro and support default case with impl_with_context_via (even now with something like `() as _def_ctx => MyDefaultContext`, but probably it could be made more ergonomic by allowing extra syntax like `() => MyDefaultContext` or `default = MyDefaultContext`

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -43,6 +50,37 @@ macro_rules! impl_stable_abi_for_tuples {
...
+    ) => {
+        impl<$($g)*> StableAbi for $self_ty where $($wc)* {
+            fn random_with_context(rng: &mut (impl RngCore + ?Sized), _ctx: ()) -> Self {
+                Self::random_with_context(rng, $default)
+            }
+        }
+    };
+}
+
+macro_rules! impl_with_context_via {
```
kskalski: let's put the macro into context.rs

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -504,4 +681,70 @@ mod tests {
...
+        derive(
+            solana_frozen_abi_macro::StableAbi,
+            solana_frozen_abi_macro::StableAbiSample
+        ),
+        solana_frozen_abi_macro::frozen_abi(
+            abi_digest = "9SaZmoaudvkdtR62vkPuf1RtzEJ8sbqN4LkUJcXGesDL",
+            abi_serializer = "wincode",
+        )
+    )]
+    struct TestStableAbiSampleSizeAttribute {
```
kskalski: instead of adding this one kitchen sink struct, which only tests support for syntax, let's add actual compatibility tests for supported equivalences, e.g.
```rust
struct MultiElementsWith {
   #[stable_abi_sample(with = "..")
   elements: Vec<i32>
}
struct MultiElementsVec {
   #[stable_abi_sample(ctx = SequenceLenMax(5))
   elements: Vec<i32>
}
struct MultiElementsVecDefault {
   elements: Vec<i32>
}
struct MultiElementsVecDeque {
   #[stable_abi_sample(ctx = SequenceLenMax(5))
   elements: VecDeque<i32>
}


struct UpToOneElementWith {
   #[stable_abi_sample(with = "..")
   elements: Vec<i32>
}
struct UpToOneElementVec {
   #[stable_abi_sample(ctx = SequenceLenMax(1))
   elements: Vec<i32>
}
struct UpToOneElementBTreeSet {
   #[stable_abi_sample(ctx = SequenceLenMax(1))
   elements: BTreeSet<i32>
}
```

each kind will have the same ABI digest and I suppose we will have many such structs, might make sesne to generate them with a macro

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -104,11 +142,150 @@ where
...
+
+impl_with_context_via! {
+    impl[T, S] for HashSet<T, S>,
+    SequenceLenMax as ctx => SequenceLenRange { min: 0, max: ctx.0 },
+    where { T: StableAbi + Eq + Hash, S: BuildHasher + Default },
+}
+
+impl_stable_abi_via_default_ctx! {
+    impl[T] for Vec<T>,
+    default = SequenceLenRange { min: 0, max: 3 },
```
kskalski: 3 seems fine, though maybe let's use slightly larger value and keep it non-power-of-2, maybe 5?

frozen-abi/src/stable_abi/mod.rs
```diff
@@ -1,5 +1,6 @@
 use rand::RngCore;
 
+mod context;
```
kskalski: should be `pub`

frozen-abi/src/stable_abi/impls.rs
puhtaytow: Should be ok now*, do we need more granularity, cases?

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -504,4 +652,231 @@ mod tests {
...
+        multiple_elements = "9QBF2ReRkBJ4ssnkrAcWq6Cb1tghYt55MLZzZQnuo7Zz",
+        up_to_one_element = "3cFop5HjPdJhyL23aaE6tzdnXiintAmgbBHgHLoGvNS9",
+        fields = [a: u8, b: bool, c: char, d: i16, e: u128,]
+    );
+    impl_equivalent_sequence_sample_structs!(
+        module = equivalent_sequence_sample_single,
+        multiple_elements = "762V1KB6NRDroyHo31bjg7gwX9nx8QqHCP9EJ3Y7nLGE",
+        up_to_one_element = "7dXyMka1Z72sENE9vva8vmE65F7y6kXZKQ2DHu9aTWyz",
+        fields = [a: u8]
+    );
```
kskalski, line 841: I was thinking more about something like
```rust
impl_sequence_sample_structs!(
    module = equivalent_sequence_sample_single,
    digests = [
        "762V1KB6NRDroyHo31bjg7gwX9nx8QqHCP9EJ3Y7nLGE" => [
            struct MultiElementsWith {
                #[stable_abi_sample(with = "(0..rng.random_range(0..=5)).map(|_| rng.random()).collect()")]
                a: Vec<u8>,
            },
            struct MultiElementsVec {
                #[stable_abi_sample(ctx = SequenceLenMax(5))]
                a: Vec<u8>,
            },
            struct MultiElementsVecDefault {
                a: Vec<u8>,
            },
            struct MultiElementsVecDeque {
                #[stable_abi_sample(ctx = SequenceLenMax(5))]
                a: VecDeque<u8>,
            },
        ],
        "7dXyMka1Z72sENE9vva8vmE65F7y6kXZKQ2DHu9aTWyz" => [
            struct UpToOneElementWith {
                #[stable_abi_sample(with = "(0..rng.random_range(0..=1)).map(|_| rng.random()).collect()")]
                a: Vec<u8>,
            },
            struct UpToOneElementVec {
                #[stable_abi_sample(ctx = SequenceLenMax(1))]
                a: Vec<u8>,
            },
            struct UpToOneElementBTreeSet {
                #[stable_abi_sample(ctx = SequenceLenMax(1))]
                a: BTreeSet<u8>,
            },
        ]
    ]
);
```

impl of the macro:
```rust
macro_rules! impl_sequence_sample_struct {
    (
        derives = ($($derive:path),+),
        serializer = $abi_serializer:literal,
        abi_digest = $abi_digest:expr,
        struct $struct_name:ident { $($body:tt)* }
    ) => {
        #[derive($($derive),+)]
        #[cfg_attr(
            feature = "frozen-abi",
            derive(
                StableAbi, StableAbiSample
            ),
                abi_digest = $abi_digest,
                abi_serializer = $abi_serializer,
            )
        )]
        struct $struct_name { $($body)* }
    };
}

macro_rules! impl_sequence_sample_structs_for_serializer {
    (
        module = $module_name:ident,
        serializer = $abi_serializer:literal,
        derives = ($($derive:path),+ $(,)?),
        digests = [$(
            $abi_digest:expr => [$(
                struct $struct_name:ident { $($body:tt)* }
            ),+ $(,)?]
        ),+ $(,)?]
    ) => {
        mod $module_name {
            $($(
                impl_sequence_sample_struct!(
                    derives = ($($derive),+),
                    serializer = $abi_serializer,
                    abi_digest = $abi_digest,
                    struct $struct_name { $($body)* }
                );
            )+)+
        }
    };
}

macro_rules! impl_sequence_sample_structs {
    (
        module = $module_name:ident,
        digests = [$(
            $abi_digest:expr => [$(
                struct $struct_name:ident { $($body:tt)* }
            ),+ $(,)?]
        ),+ $(,)?]
    ) => {
        mod $module_name {
            impl_sequence_sample_structs_for_serializer!(
                module = wincode,
                serializer = "wincode",
                derives = (SchemaWrite),
                digests = [$(
                    $abi_digest => [$(
                        struct $struct_name { $($body)* }
                    ),+]
                ),+]
            );
            impl_sequence_sample_structs_for_serializer!(
                module = bincode,
                serializer = "bincode",
                derives = (Serialize),
                digests = [$(
                    $abi_digest => [$(
                        struct $struct_name { $($body)* }
                    ),+]
                ),+]
            );
        }
    };
}
```
```

Comment (kskalski) [reply frozen-abi/src/stable_abi/impls.rs:841]: this way we can add a few more tests:
* similar as above, but with different item type, e.g. `(u8, i32)`
* nested collection `Vec<Vec<>>` 
* map type - equivalence of BTreeMap<K, V> with Vec<(K, V)> and Vec<KeyValue>
* context where min len is set

File: frozen-abi/src/stable_abi/impls.rs [hunk 7]
```diff
@@ -504,4 +652,231 @@ mod tests {
...
+                        )]
+                        $field_name: Vec<$field_ty>,
+                    )+
+                }
+
+                #[derive($($derive),+)]
+                #[cfg_attr(
+                    feature = "frozen-abi",
+                    derive(
+                        solana_frozen_abi_macro::StableAbi,
```
kskalski: let's just add all necessary imports in the tests module, so all that code is shorter

frozen-abi/src/stable_abi/impls.rs
kskalski: for 1-element collections, we can actually also add hash-based collections

puhtaytow: Yeah, looks better thanks 👍

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -504,4 +653,189 @@ mod tests {
...
+    macro_rules! impl_sequence_sample_structs {
+    (
+        digests = [$(
+            $abi_digest:expr => [$(
+                struct $struct_name:ident { $($body:tt)* }
+            ),+ $(,)?]
+        ),+ $(,)?]
+    ) => {
+        mod sequence_sample_test_types {
+            impl_sequence_sample_structs_for_serializer!(
```
kskalski: looking at it again, I think it's actually cleaner to inline this macro, so it would just be:
```rust
macro_rules! impl_sequence_sample_structs {
    (
        module = $module_name:ident,
        digests = [$(
            $abi_digest:expr => [$(
                struct $struct_name:ident { $($body:tt)* }
            ),+ $(,)?]
        ),+ $(,)?]
    ) => {
        mod $module_name {
            mod wincode {
                $($(impl_sequence_sample_struct!(
                    derives = (wincode::SchemaWrite),
                    serializer = "wincode",
                    abi_digest = $abi_digest,
                    struct $struct_name { $($body)* }
                );)+)+
            }
            mod bincode {
                $($(impl_sequence_sample_struct!(
                    derives = (serde_derive::Serialize),
                    serializer = "bincode",
                    abi_digest = $abi_digest,
                    struct $struct_name { $($body)* }
                );)+)+
            }
        }
    };
}
```

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -398,7 +546,7 @@ mod tests {
     );
 
     const ABI_DIGEST_EQUIVALENT_KEY_VALUE_SEQUENCES: &str =
-        "9pGP5GGD2HxDRCQeDv3rPGTfVH9SzkZCXMSGHpZnzz4G";
+        "ATEDuwPHiE6J9TtzEuSvS7tMvtAR6pZuKFtzGioozixd";
```
puhtaytow: the digest changed along with manual `with = ""` implementation, the previous one triggered clippy unsafe arithmetics

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -410,7 +558,7 @@ mod tests {
     )]
     struct TestEquivalentBTreeMapVsVecWincode {
         #[stable_abi_sample(
-            with = "(0..rng.random::<u16>() % 4).map(|i| (((i << 8) + rng.random::<u8>() as u16), rng.random())).collect()"
+            with = "(0..rng.random_range(0..4)).map(|i| (u16::from_be_bytes([i as u8, rng.random()]), rng.random())).collect()"
```
kskalski: maybe let's do it as separate PR

puhtaytow: https://github.com/anza-xyz/solana-sdk/pull/741

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -504,4 +652,148 @@ mod tests {
...
+                    #[stable_abi_sample(ctx = SequenceLenMax(1))]
+                    a: Vec<u8>,
+                },
+                struct UpToOneElementBTreeSet {
+                    #[stable_abi_sample(ctx = SequenceLenMax(1))]
+                    a: BTreeSet<u8>,
+                },
+                struct UpToOneElementHashMap {
+                    #[stable_abi_sample(ctx = SequenceLenMax(1))]
+                    a: HashMap<[u8; 0], u8>,
```
kskalski: I think I prefer to flip the zero-sized array from keys to values, maybe better to actually use `()`, it's somewhat more obvious, since using `[u8; 0]` looks like a typo

and we should add (under separate digest) a test for actual non-zero key-value collecitons

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -504,4 +652,148 @@ mod tests {
...
+                    $($(impl_sequence_sample_struct!(
+                        derives = (wincode::SchemaWrite),
+                        serializer = "wincode",
+                        abi_digest = $abi_digest,
+                        struct $struct_name { $($body)* }
+                    );)+)+
+                }
+                mod bincode {
+                    use crate::stable_abi::context::SequenceLenMax;
+                    use std::collections::{HashSet, HashMap, VecDeque, BTreeSet, BTreeMap};
```
kskalski: could move those imports to top level - `mod tests {` ?

puhtaytow:
Should be good now 👍

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -504,4 +654,162 @@ mod tests {
...
+                a: HashMap<u8, u16>,
+            },
+            struct NonZeroKeyValueHashMap {
+                #[stable_abi_sample(ctx = SequenceLenMax(1))]
+                a: HashMap<u8, u16>,
+            },
+            struct NonZeroKeyValueBTreeMap {
+                #[stable_abi_sample(ctx = SequenceLenMax(1))]
+                a: BTreeMap<u8, u16>,
+            },
```
kskalski: let's add `Vec<(u8, u16)>` too

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -504,4 +654,162 @@ mod tests {
...
+        derives = $derives:tt,
+        serializer = $abi_serializer:literal,
+        abi_digest = $abi_digest:expr,
+        struct $struct_name:ident { $($body:tt)* }
+    ) => {
+        #[derive $derives]
+        #[cfg_attr(
+            feature = "frozen-abi",
+            derive(
+                solana_frozen_abi_macro::StableAbi,
```
kskalski: if we import StableAbi, StableAbitSample, frozen_abi, maybe inlining `impl_sequence_sample_struct` would also make sense

frozen-abi-macro/src/lib.rs
```diff
@@ -148,6 +148,28 @@ struct StableAbiSampleOptions {
     ctx_expr: Option<TokenStream2>,
 }
 
+#[cfg(feature = "frozen-abi")]
+// `with = "...rng..."` expressions can come from inside `macro_rules!`; rewrite
+// `rng` to this proc macros call-site span so it resolves to the generated
+// `sample(..., rng)` parameter
+fn normalize_rng_ident(tokens: TokenStream2) -> TokenStream2 {
```
kskalski: a bit unfortunate we need this rewrite... is this because we nest macros or any rng being a macro parameter gets wrong binding?
actually, what does `rng` bind to or what is the error without this normalization?

kskalski:
Another test we should have (another digest) is the `SequenceLenRange` with `min:1`

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -504,4 +654,166 @@ mod tests {
...
+                struct UpToOneElementHashSetDefault {
+                    a: HashSet<u8>,
+                },
+                struct UpToOneElementBTreeMap {
+                    #[stable_abi_sample(ctx = SequenceLenMax(1))]
+                    a: BTreeMap<u8, ()>,
+                },
+            ],
+            "Du8dTBApdeSxYTQVprkbMGvc5dLgCfUKKZ2z63qqr5Bj" => [
+            struct NonZeroVecTuple {
```
kskalski: * move it below the "canonical" struct that has `with =`
* rename to UpToOneKeyValueVec (also rename other key value types to UpToOneKeyValue*)
* fix indentation

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -104,11 +111,154 @@ where
...
+
+impl_with_context_via! {
+    impl[T] for VecDeque<T>,
+    SequenceLenMax as ctx => SequenceLenRange { min: 0, max: ctx.0 },
+    where { T: StableAbi },
+}
+
+impl_with_context_via! {
+    impl[K, V] for BTreeMap<K, V>,
+    () as _ctx => SequenceLenRange { min: 0, max: 3 },
```
kskalski: let's use consistent default max for all collections (except for hash-based of course) and best put it as a constant like `DEFAULT_COLLECTION_MAX_SAMPLE_LEN`

once we support `SequenceSortedInput` we could then match vec and btree with just that attribute

frozen-abi-macro/src/lib.rs
puhtaytow: yes this is caused by macro nesting

```
error[E0425]: cannot find value `rng` in this scope
   --> frozen-abi/src/stable_abi/impls.rs:727:32
    |
727 |                         with = "(0..rng.random_range(0..=5)).map(|_| rng.random()).collect()"
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
```
to my understanding this happens because generated scope thru macro rules is different than directly from type field sampling options

if you know how to fix it :rocket: :rocket:

puhtaytow: I might also try harder with clanker if thats blocker

kskalski: do we actually support `with = expr`? that would solve the issue

frozen-abi-macro/src/lib.rs
```diff
@@ -166,7 +188,7 @@ fn parse_stable_abi_sample_options(field: &syn::Field) -> Result<StableAbiSample
                 let expr = value.parse::<Expr>().map_err(|err| {
                     Error::new(value.span(), format!("invalid `with` expression: {err}"))
                 })?;
-                with_expr = Some(quote! { #expr });
+                with_expr = Some(normalize_rng_ident(quote! { #expr }));
```
kskalski: Clankers for the rescue:

**Repro** (after removing `normalize_rng_ident`): tests fail with `cannot find value 'rng' in this scope` for every `with = "...rng..."` attribute that came out of a `macro_rules!` expansion (`MultiElementsWith`, `UpToOneElementWith`, `UpToOneKeyValueHashMapWith`, etc.).

**Why:** `LitStr::parse::<Expr>()` (in syn) tokenizes the string and then re-spans every resulting token to the `LitStr`'s own span. When the `LitStr` came from a `macro_rules!` body, that span carries the macro's *definition* hygiene context. The generated `fn sample(..., rng)` parameter is emitted via `quote!` with the proc-macro's `Span::call_site()` hygiene. Two different syntax contexts → name lookup fails.

The old `normalize_rng_ident` patched it by walking the token tree and rewriting only the `rng` ident to `Span::call_site()`. That's fragile (hard-codes one name, won't help any other identifier the user reaches for) and obscures what's actually wrong.

**Better fix** (frozen-abi-macro/src/lib.rs:186-195): swap `value.parse::<Expr>()` for `syn::parse_str::<Expr>(&value.value())`. `parse_str` goes through `proc_macro2::TokenStream::from_str`, which yields tokens at the proc macro's `Span::call_site()` — no re-spanning to the `LitStr`. Every identifier in the user's expression (including `rng`, `std::…`, anything else) now resolves in the same context as the generated parameter and the surrounding module.

All 94 frozen-abi tests pass, clippy is clean.


Let's do that as a separate PR, though it would be nice for the fix to actually tigger the error scenario.
Supposedly this is the minimum repro:
```rust
macro_rules! mk {
    ({ $($body:tt)* }) => {
        #[derive(StableAbi, StableAbiSample)]
        #[frozen_abi(abi_digest = "...", abi_serializer = "wincode")]
        struct S { $($body)* }
    };
}
mk!({ #[stable_abi_sample(with = "rng.random::<u8>()")] a: u8, });
```

frozen-abi/src/stable_abi/context.rs
```diff
@@ -0,0 +1,43 @@
...
+/// (`Vec`, `VecDeque`, `HashMap`, `BTreeMap`) that bounds the sampled length
+/// to an inclusive range `min..=max`.
+///
+/// The collection's `random_with_context` draws a length uniformly from
+/// `min..=max` and produces that many elements. `min == max` pins the length
+/// exactly; `min == 0` means "up to `max` elements" (equivalent to
+/// `SequenceLenMax(max)`).
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub struct SequenceLenRange {
+    pub(super) min: usize,
```
kskalski: it's a public struct that users are supposed to use, so those shoudl be public fields. we could also add a `const fn new(impl RangeBounds<usize>)` constructor

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -104,11 +113,154 @@ where
...
+{
+    fn random_with_context(rng: &mut (impl RngCore + ?Sized), ctx: SequenceLenRange) -> Self {
+        let len = rng.random_range(ctx.min..=ctx.max);
+        (0..len).map(|_| (K::random(rng), V::random(rng))).collect()
+    }
+}
+
+impl_with_context_via! {
+    impl[K, V, S] for HashMap<K, V, S>,
+    SequenceLenMax as ctx => SequenceLenRange { min: 0, max: ctx.0 },
```
kskalski: how about adding `.into_range_context()` to `SequenceLenMax` ?

frozen-abi/src/stable_abi/context.rs
```diff
@@ -0,0 +1,62 @@
...
+/// exactly; `min == 0` means "up to `max` elements" (equivalent to
+/// `SequenceLenMax(max)`).
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub struct SequenceLenRange {
+    pub min: usize,
+    pub max: usize,
+}
+
+impl SequenceLenRange {
+    pub const fn new(min: usize, max: usize) -> Self {
```
kskalski: I think the range syntax would be nice to support, maybe just drop `const` - I guess we don't actually need it, since the we always evaluate the `ctx` or `convert` in body function, right?

puhtaytow: right. with const it wasnt possible, fixed now :+1:

frozen-abi/src/stable_abi/context.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+impl From<SequenceLenMax> for SequenceLenRange {
+    fn from(max: SequenceLenMax) -> Self {
+        Self::new(0..=max.0)
+    }
+}
+
+impl From<usize> for SequenceLenRange {
```
kskalski: This kind of conversion is not really obvious or very useful, converting single number into range?

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -528,4 +679,164 @@ mod tests {
...
+            ],
+            "57dywdByMU7XcWbsfnXY5ChZdqne57zjJCn7uEjDKGNc" => [
+                struct UpToOneKeyValueHashMapWithRange {
+                    #[stable_abi_sample(
+                        with = "(0..rng.random_range(1..=1)).map(|_| (rng.random(), rng.random())).collect()"
+                    )]
+                    a: HashMap<u8, u16>,
+                },
+                struct UpToOneKeyValueVecRange {
+                    #[stable_abi_sample(ctx = SequenceLenRange{min: 1, max: 1})]
```
kskalski: could this use the range syntax `SequenceLenRange::new(1..=1)` ?

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -104,11 +114,154 @@ where
...
+
+impl_with_context_via! {
+    impl[T] for VecDeque<T>,
+    SequenceLenMax as ctx => SequenceLenRange::from(ctx),
+    where { T: StableAbi },
+}
+
+impl_with_context_via! {
+    impl[K, V] for BTreeMap<K, V>,
+    () as _ctx => SequenceLenRange::from(DEFAULT_COLLECTION_MAX_SAMPLE_LEN),
```
kskalski: I would rather use the range syntax `SequenceLenRange::new(0..=DEFAULT_COLLECTION_MAX_SAMPLE_LEN)`

frozen-abi/src/stable_abi/context.rs
puhtaytow: agreed :)

frozen-abi/src/stable_abi/context.rs
```diff
@@ -0,0 +1,75 @@
...
+/// Context for `StableAbi<...>` impls on sequence-like collections
+/// that caps the sampled length at `0..=N`.
+///
+/// Equivalent to `SequenceLenRange { min: 0, max: N }`; the per-collection
+/// impl delegates to the `SequenceLenRange` impl.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct SequenceLenMax(pub usize);
+
+impl SequenceLenMax {
+    pub const fn new(max: usize) -> Self {
```
kskalski: this new seem redundant

kskalski (commented):
LGTM, let's have some more eyes on that.

alexpyattaev (dismissed):
LGTM

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -104,11 +114,154 @@ where
     }
 }
 
+// keep at most one element to mitigate iteration order differences
+impl_with_context_via! {
+    impl[K, V, S] for HashMap<K, V, S>,
```
joncinque: Why does the macro use square brackets instead of angle brackets for the generic types? Wouldn't this be more expected for Rust users? Or is the idea for the macro to only be used internally, so we don't really care?

```suggestion
    impl<K, V, S> for HashMap<K, V, S>,
```

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -104,11 +114,154 @@ where
...
+    () as _ctx => SequenceLenRange::new(0..=DEFAULT_COLLECTION_MAX_SAMPLE_LEN_NON_DETERMINISTIC_ORDER),
```
joncinque: The syntax here is a bit rough, and makes me wonder if we can use a closure instead of the macro and context type.

No need to change anything here, but would something like a closure potentially simplify this?

joncinque (commented):
Just a small comment on the macro, looks great overall!

frozen-abi/src/stable_abi/impls.rs
kskalski: This macro might become public in the future, as it's quite a convenient helper, so definitely worth trying.

Seems like we can improve syntax quite a bit, let's try
```rust
macro_rules! impl_with_context_via {
    (
        impl<$($g:tt)*> StableAbi<$ctx:ty> for $self_ty:ty
        where { $($wc:tt)* },
        |$bind:ident| $convert:expr $(,)?
    ) => {
        impl<$($g)*> StableAbi<$ctx> for $self_ty
        where $($wc)*
        {
            fn random_with_context(
                rng: &mut (impl RngCore + ?Sized),
                $bind: $ctx,
            ) -> Self {
                Self::random_with_context(rng, $convert)
            }
        }
    };
}

impl_with_context_via! {
    impl<K, V, S> StableAbi<SequenceLenMax> for HashMap<K, V, S>
    where { K: StableAbi + Eq + Hash, V: StableAbi, S: BuildHasher + Default },
    |ctx| SequenceLenRange::from(ctx),
}

impl_with_context_via! {
    impl<T, S> StableAbi<()> for HashSet<T, S>
    where { T: StableAbi + Eq + Hash, S: BuildHasher + Default },
    |_| SequenceLenRange::new(0..=DEFAULT_COLLECTION_MAX_SAMPLE_LEN_NON_DETERMINISTIC_ORDER),
}
```

joncinque: That's definitely a much nicer option already!

frozen-abi/src/stable_abi/impls.rs
puhtaytow: if im not mistaken it was due to local ambiguity when calling macro `impl_with_context_via`: multiple parsing options: built-in NTs tt ('g') 

but the new context_via syntax seem to work with <> now, thanks!

frozen-abi/src/stable_abi/impls.rs
puhtaytow: should be good now :+1: thank you guys :pray: :pray:

kskalski (approved):

joncinque (approved):
Looks great to me, thanks!

Issue #615, opened 2026-03-09
Potential Unsoundness in `StableVec` via public `addr`, `cap`, and `len` fields


DiuDiu777:
**Description:**

Hello! Thanks for your repo, which helps me a lot. Recently, I've identified a potential soundness issue in the `stable-layout` crate.

The `StableVec` struct exposes its internal state fields (`addr`, `cap`, and `len`) as **`pub`**. However, the crate also provides a safe `Drop` implementation for `StableVec` that reconstructs a `Vec` from these raw parts to deallocate memory:

```rust
// From src/stable_vec.rs
impl<T> Drop for StableVec<T> {
    fn drop(&mut self) {
        let _vec = unsafe {
            Vec::from_raw_parts(
                self.addr as usize as *mut T,
                self.len as usize,
                self.cap as usize,
            )
        };
    }
}

```

Because these fields are public, entirely safe Rust code can mutate them to arbitrary values. When the `StableVec` instance is dropped, it will attempt to free an unowned or invalid pointer with incorrect layout information, leading to immediate Undefined Behavior (UB), such as double-free, invalid-pointer-free, or segmentation faults.

**Proof of Concept:**

The following strictly safe code triggers a memory corruption crash during `Drop`:

```rust
use solana_stable_layout::stable_vec::StableVec;

fn main() {
    // 1. Create a valid StableVec
    let mut stable_vec = StableVec::from(vec![1u64, 2, 3]);

    // 2. Maliciously corrupt internal state via safe public fields
    stable_vec.addr = 0x12345678 as u64; // Arbitrary address
    stable_vec.cap = 999999;             // Arbitrary capacity
    stable_vec.len = 999999;             // Arbitrary length

    // 3. Drop occurs here, triggering unsafe Vec::from_raw_parts with corrupted data.
    // This results in a crash (e.g., free(): invalid pointer).
}

```

Can the visibility of `addr`, `cap`, and `len` fields be changed from `pub` to private? Since the crate already provides safe getter methods like `as_vaddr()` and `len()`, users can still access the information without being able to break the struct's safety invariants.

Thank you for your attention to this issue!

PR #740, opened 2026-05-26
stable-layout: change fields visibility of StableVec<T> from public to private

yilin0518:
This pull request solves the issue #615 which mentions that public fields may have potential risks. This PR makes improvements to the `StableVec` struct in the `stable-layout` crate, focusing on encapsulation and API usability. The most important changes are:

Encapsulation improvements:

* Changed the fields `addr`, `cap`, and `len` in the `StableVec` struct from public to private, preventing direct modification from outside the struct and improving safety.

API enhancements:

* Added a new public method `cap()` to the `StableVec` implementation, allowing users to access the capacity of the vector in a controlled manner.

### DIFF
--- a/stable-layout/src/stable_vec.rs
+++ b/stable-layout/src/stable_vec.rs
@@ -28,9 +28,9 @@ use std::{
 /// ```
 #[repr(C)]
 pub struct StableVec<T> {
-    pub addr: u64,
-    pub cap: u64,
-    pub len: u64,
+    addr: u64,
+    cap: u64,
+    len: u64,
     _marker: PhantomData<T>,
 }
 


stable-layout/src/stable_vec.rs
```diff
@@ -47,6 +47,11 @@ impl<T> StableVec<T> {
         self.len
     }
 
+    #[inline]
+    pub fn cap(&self) -> u64 {
+        self.cap
+    }
```
joncinque: Let's just remove this -- if it's needed, we can add it later

joncinque (commented):
Thanks for your contribution! This change makes sense, and even though it's a breaking change, it won't have big ramifications for other users since everyone uses `StableInstruction`.

To be safe, we'll do a breaking release, but we won't have to do major bumps anywhere else (except for solana-program, but that's a special case).

Can you rebase your branch? It seems like CI didn't run for some reason

yilin0518:
Thank you for your reply! I will do it right now.

yilin0518:
@joncinque I rebase my branch to track the newest commit, and remove the `pub fn cap`. Waiting for your next review and  hoping that I don't get something wrong.

joncinque (approved):
Looks great, thanks!

joncinque:
Thanks for your patience -- since we're doing some breaking releases soon, it's a good time to land this change

PR #741, opened 2026-05-28
frozen-abi: resolve unsafe arithmetics in StableAbi test types

puhtaytow:
#### Description

Clippy was complaining about unsafe arithmetics.

### Summary of changes
- changed formula passed within 'with = ""' field sampling option to safe one.

### DIFF
--- a/frozen-abi/src/stable_abi/impls.rs
+++ b/frozen-abi/src/stable_abi/impls.rs
@@ -410,7 +410,7 @@ mod tests {
     )]
     struct TestEquivalentBTreeMapVsVecWincode {
         #[stable_abi_sample(
-            with = "(0..rng.random::<u16>() % 4).map(|i| (((i << 8) + rng.random::<u8>() as u16), rng.random())).collect()"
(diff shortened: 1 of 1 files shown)


frozen-abi/src/stable_abi/impls.rs
```diff
@@ -410,7 +410,7 @@ mod tests {
     )]
     struct TestEquivalentBTreeMapVsVecWincode {
         #[stable_abi_sample(
-            with = "(0..rng.random::<u16>() % 4).map(|i| (((i << 8) + rng.random::<u8>() as u16), rng.random())).collect()"
+            with = "(0..rng.random_range(0..4)).map(|i| (u16::from_be_bytes([i as u8, rng.random()]), rng.random())).collect()"
```
kskalski: does the digest remain the same if the range still uses `random::<u16>() % 4`? not saying we shouldn't change, just wanted to make sure

puhtaytow: yup it stays the same for `with = "(0..rng.random::<u16>() % 4).map(|i| (u16::from_be_bytes([i as u8, rng.random()]), rng.random())).collect()"`

so i reverted to use rng.random::<u16>() % 4 instead of range

kskalski (approved):

PR #742, opened 2026-05-29
frozen-abi: fix value parsing in sampling option 'with=' 

puhtaytow:
### Description

StableAbi sampling `with = "...rng..."` expressions failed when attribute was emitted from a macro_rules!.
The issue was caused by a span hygiene difference between parsing paths.

`LitStr::parse::<Expr>()` parses string literal and assigns the resulting tokens span of the original `LitStr`. When the literal comes from a macro_rules! body, that span carries the macro definition context.

### Summary of changes
The change doesn't break compatibility with current usage.

- instead of parsing the expression thru `LitStr::parse`, we now parse the raw string contents with `syn::parse_str::<Expr>(&value.value())`


kskalski:
Let's add a bit more context to the problem description - the different between those parsing ways is subtle, so we could mention the span which both methods use.

kskalski (approved):

PR #743, opened 2026-06-01
short-vec: no-std

grod220:
Converts `solana-short-vec` crate to no-std + alloc.

Required for getting tx parsing in pinocchio programs (e.g. [solana-program/nonce](https://github.com/solana-program/nonce)). Next needing to convert:

- [ ] https://github.com/anza-xyz/solana-sdk/pull/743
- message
- transaction

PR #744, opened 2026-06-01
transaction-error: no-std

grod220:
Converts `solana-transaction-error` crate to no-std. Continued work from: https://github.com/anza-xyz/solana-sdk/pull/743.

### DIFF
--- a/transaction-error/src/lib.rs
+++ b/transaction-error/src/lib.rs
@@ -1,9 +1,15 @@
 #![cfg_attr(feature = "frozen-abi", feature(min_specialization))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![no_std]
 #[cfg(feature = "serde")]
 use serde_derive::{Deserialize, Serialize};
 #[cfg(feature = "frozen-abi")]
 use solana_frozen_abi_macro::{AbiEnumVisitor, AbiExample};
+#[cfg(any(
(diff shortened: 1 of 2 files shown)


febo (approved):
Looks great!

transaction-error/src/lib.rs
```diff
@@ -357,15 +363,15 @@ impl From<SanitizeError> for SanitizeMessageError {
...
 }
 
-#[cfg(not(target_os = "solana"))]
+#[cfg(not(any(target_os = "solana", target_arch = "bpf")))]
 #[derive(Debug)]
 pub enum TransportError {
     IoError(std::io::Error),
     TransactionError(TransactionError),
-    Custom(String),
+    Custom(std::string::String),
```
joncinque, line 371: Why don't we use the `alloc` version and remove std entirely in this crate? https://doc.rust-lang.org/alloc/string/index.html

joncinque (commented):
Just one comment

transaction-error/src/lib.rs
grod220: Figured one of the variants in this enum couldn't be replaced anyway: `IoError(std::io::Error)`

joncinque: Ah gotcha, makes sense, thanks!

PR #745, opened 2026-06-01
port over solana-svm-transaction from agave

buffalojoec:


apfitzge:
@joncinque some context on this move: https://github.com/anza-xyz/agave/pull/12486#issuecomment-4588878128

Could you review wrt how the sdk repo handles CI, versioning & publishing? Want to make sure everything around that is done correctly so we can publish

svm-transaction/src/svm_message.rs
```diff
@@ -0,0 +1,160 @@
...
+            )
+            .filter(|ix| {
+                /// Serialized value of [`SystemInstruction::AdvanceNonceAccount`].
+                const SERIALIZED_ADVANCE_NONCE_ACCOUNT: [u8; 4] = 4u32.to_le_bytes();
+                const SERIALIZED_SIZE: usize = SERIALIZED_ADVANCE_NONCE_ACCOUNT.len();
+
+                ix.data
+                    .get(..SERIALIZED_SIZE)
+                    .map(|data| data == SERIALIZED_ADVANCE_NONCE_ACCOUNT)
+                    .unwrap_or(false)
```
apfitzge, line 121: This is what exists in the monorepo's version, so let's keep it for this simple port.

There already does exist an `inline_nonce::is_advance_nonce_instruction_data` in sdk that does this. As follow-up to the port, seems reasonable we just use that.

apfitzge (approved):

apfitzge:
Let's wait for @joncinque's input before merging.

svm-transaction/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "solana-svm-transaction"
+description = "Solana SVM Transaction"
+documentation = "https://docs.rs/solana-svm-transaction"
+version = "4.1.0"
```
joncinque, line 5: The publish job bumps the crate too, so we can leave this at 4.0.1. Unless we do a publish at the current version -- whatever you prefer

svm-transaction/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+rust-version = { workspace = true }
+authors = { workspace = true }
+repository = { workspace = true }
+homepage = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
```
joncinque, line 11: We're lucky here -- we haven't bumped to edition 2024 yet so we don't have to downgrade this crate 🥳

joncinque (approved):
Looks good to me! We'll just have to be careful on the publish job if we keep the current version

svm-transaction/Cargo.toml
apfitzge: I think we need a publication in order to delete the one from the monorepo - if we swtich monorepo to pull in 4.0.1 then we get a bunch of failures because of the solana-hash version being mismatched.

buffalojoec: I can drop https://github.com/anza-xyz/solana-sdk/pull/745/commits/d2a8a06666ebe1be71b13a2ff871ebba532059d1 and do the version bump through the automation?

joncinque: The only difference is that we won't have a separate commit with the version bump -- since this PR is adding the whole crate at once, it isn't such a big deal. Let's just land it

PR #746, opened 2026-06-01
message: no-std

grod220:
Converts `solana-message` crate to no-std + alloc. Allows no-std programs to parse messages now (relevant to [solana-program/nonce](https://github.com/solana-program/nonce)).

Continues the work of https://github.com/anza-xyz/solana-sdk/pull/743 https://github.com/anza-xyz/solana-sdk/pull/744

Crate `solana-transaction` will be last.


message/src/versions/v1/message.rs
```diff
@@ -350,6 +352,7 @@ impl Message {
     ///
     /// This method should not be used directly.
     #[inline(always)]
+    #[cfg(not(any(target_os = "solana", target_arch = "bpf")))]
     pub(crate) fn is_writable_index(&self, i: usize) -> bool {
```
grod220: Technically no-std compatible, but used by std parts of this crate. Removing the cfg attr results in dead code warnings. This is the case in a few places in this PR.

message/src/versions/mod.rs
```diff
@@ -48,7 +51,7 @@ pub const MESSAGE_VERSION_PREFIX: u8 = 0x80;
 /// format.
 #[cfg_attr(
     feature = "frozen-abi",
-    frozen_abi(digest = "6CoVPUxkUvDrAvAkfyVXwVDHCSf77aufm7DEZy5mBVeX"),
+    frozen_abi(digest = "9xQQLkQntX2QKgwxbbpeuNrs5V2WopsBa11su46WWCro"),
```
grod220, line 54: Now that versions & account_keys compiles in programs, moved this out of the private module and at the root level. The import is the same, so no API changes, but this triggers a digest update. 

```diff
+ #[cfg(not(any(target_os = "solana", target_arch = "bpf")))]
mod non_bpf_modules {
-   mod account_keys;
    mod address_loader;
    mod sanitized;
    mod versions;

-   pub use {account_keys::*, address_loader::*, sanitized::*, versions::*};
+   pub use {address_loader::*, sanitized::*};
}

pub use non_bpf_modules::*;

+ mod versions;
+ mod account_keys;
+ pub use {
+    account_keys::*,
+    versions::*,
+};
```

Suppose the internal path of `solana_message::non_bpf_modules::versions` to `solana_message::versions` is what did it.

message/src/versions/v0/loaded.rs
```diff
@@ -1,13 +1,15 @@
 #[cfg(feature = "serde")]
 use serde_derive::{Deserialize, Serialize};
+#[cfg(not(any(target_os = "solana", target_arch = "bpf")))]
 use {
     crate::{v0, AccountKeys},
-    solana_address::Address,
     solana_sdk_ids::bpf_loader_upgradeable,
     std::{borrow::Cow, collections::HashSet},
```
joncinque: `Cow` does exist in `alloc`, so we could use that for `LoadedMessage` https://doc.rust-lang.org/stable/alloc/borrow/enum.Cow.html

It wouldn't be massively useful since the constructors all still require a `HashSet`, but we could add a `BTreeSet` variant.

Or to be honest, since this is on its way out, we could just say "who cares" 😅

message/src/versions/v0/mod.rs
```diff
@@ -14,16 +14,20 @@ pub use loaded::*;
...
 #[cfg(feature = "frozen-abi")]
 use solana_frozen_abi_macro::AbiExample;
+use {
+    crate::{compiled_instruction::CompiledInstruction, MessageHeader},
+    alloc::vec::Vec,
+    solana_address::Address,
+    solana_hash::Hash,
+    solana_sanitize::SanitizeError,
+};
+#[cfg(not(any(target_os = "solana", target_arch = "bpf")))]
```
joncinque: I know the rest of the crate is doing this already, but maybe we should consider adding an `std` feature to the crate. That'll give more flexibility and clarity, so anyone can potentially use it (or not).

Although it's technically a breaking change, we could enable the std feature by default. What do you think?

message/src/versions/v1/message.rs
```diff
@@ -456,7 +460,7 @@ impl Message {
         }
 
         // no duplicate addresses
-        let unique_keys: HashSet<_> = self.account_keys.iter().collect();
+        let unique_keys: BTreeSet<_> = self.account_keys.iter().collect();
```
joncinque, line 465: This is maybe overly cautious, but are we sure this won't cause any issues?

joncinque (commented):
Looks great overall! Just a few questions

message/src/versions/v0/loaded.rs
grod220: I figured `VersionedMessage` is what folks really need and this one is effectively deprecated. We'd need a no-std version of new(), new_borrowed(), and set_is_writable_account_cache(). I can't see a usecase for it, so probably not worth it?

message/src/versions/v0/mod.rs
grod220: Agree. I found this cfg usage a bit confusing at first versus just looking for the std flag. Will swap.

message/src/versions/v1/message.rs
grod220: Ran a bunch of equivalence tests locally to be sure. Here we are only comparing lengths. We don't iterate or depend on the ordering and both set types treat duplicates the same.

message/src/compiled_keys.rs
```diff
@@ -139,7 +137,6 @@ impl CompiledKeys {
         Ok((header, static_account_keys))
     }
 
-    #[cfg(not(target_os = "solana"))]
```
grod220, line 142: With the transition to the `std` feature, noticed a few of these are no-std and would compile in a solana-program

message/src/legacy.rs
```diff
@@ -472,14 +475,14 @@ impl Message {
     }
 
     /// Compute the blake3 hash of this transaction's message.
-    #[cfg(all(not(target_os = "solana"), feature = "wincode", feature = "blake3"))]
```
grod220, line 475: In `VersionedMessage` we don't exclude solana os in the hash fns. And after blake3 was moved to no-std, don't see a reason why we can't use this in a solana env.

scripts/check-no-std.sh
```diff
@@ -51,6 +51,7 @@ no_std_alloc_crates=(
   -p solana-curve25519
   -p solana-instruction
   -p solana-instructions-sysvar
+  -p solana-message
```
grod220, line 54: Default feature is `std` so need to exclude now

Cargo.toml
joncinque: Can we add `default-features = false` so that individual crates enable the std feature if they need it? That way all other users get no-std support more easily

scripts/build-sbf.sh
```diff
@@ -12,6 +12,7 @@ build_sbf_excludes=(
   --exclude solana-file-download
   --exclude solana-genesis-config
   --exclude solana-keypair
+  --exclude solana-message
```
joncinque, line 15: Should be fine to exclude since we have the no-std check. In the future, it would be nice to have `cargo-build-sbf` work through `cargo-hack` so we can just exclude the std feature on all crates that have it, but that's a problem for another day 😄

message/Cargo.toml
```diff
@@ -35,6 +36,7 @@ serde = [
     "solana-hash/serde",
     "solana-short-vec/serde",
 ]
+std = ["solana-instruction/std", "wincode?/std"]
```
joncinque: Are any of these extra feature enablements needed? I ran a `cargo hack check --feature-powerset` and it seemed fine without them

message/Cargo.toml
```diff
@@ -44,19 +46,18 @@ wincode = [
...
 blake3 = { workspace = true, features = ["traits-preview"], optional = true }
-lazy_static = { workspace = true }
 serde = { workspace = true, optional = true }
 serde_derive = { workspace = true, optional = true }
 solana-address = { workspace = true }
 solana-frozen-abi = { workspace = true, optional = true }
 solana-frozen-abi-macro = { workspace = true, optional = true }
 solana-hash = { workspace = true, features = ["decode", "sanitize"] }
-solana-instruction = { workspace = true, features = ["std"] }
+solana-instruction = { workspace = true, default-features = false }
```
joncinque: nit: since it's already `default-features = false` at the workspace level, we don't need to specify it again here

message/src/lib.rs
```diff
@@ -39,39 +40,47 @@
...
-    mod address_loader;
-    mod sanitized;
-    mod versions;
+use {
+    crate::compiled_instruction::CompiledInstruction, alloc::vec::Vec,
+    solana_sdk_ids::bpf_loader_upgradeable,
+};
 
-    pub use {account_keys::*, address_loader::*, sanitized::*, versions::*};
-}
```
joncinque, line 64: much cleaner now 🥳

message/src/legacy.rs
joncinque: Should be fine -- I don't know enough about the blake3 hash implementation to know if it'll complete in time, but it definitely can't hurt!

joncinque (commented):
Just a few last things.

I also did run `cargo build-sbf --no-default-features` just to be safe, and it's all good

message/Cargo.toml
grod220: True! Removed.

joncinque (approved):
Looks great!

grod220:
@apfitzge or @tao-stones, as members of the tx-metadata group requiring approval, can you review?

apfitzge (approved):

PR #748, opened 2026-06-03
feat(vote-interface): support wincode

kskalski:
Adds a `wincode` feature mirroring the existing `serde`/`bincode` support, deriving `wincode::SchemaWrite`/`SchemaRead` on every type that derives serde `Serialize`/`Deserialize`. The schemas are bincode-wire-compatible - this is verified by switching `abi_serializer` to `wincode`, which maintains the `abi_digest` unchanged.

**The bulk of the change is the compact encoding** (`CompactUpdateVoteState*` / `TowerSync*`), which serde implements via custom `serde(with = …)` modules that delta-encode lockout slots as varint offsets in a `short_vec`. To support this without duplicating the wire layout:
- A new private `compact` module defines the wire-format repr structs **once**, with per-backend field annotations (`serde(with)` / `wincode(with = "Leb128Int…/LockoutOffsetShortVec")`), and owns the shared `VoteStateUpdate`/`TowerSync` ⇄ repr conversions.
- `serde_compact_*` and a new `wincode_compact` module are thin bridges over those conversions; the wincode schemas are tiny marker types referenced from `VoteInstruction` via `wincode(with = …)` (a type can only carry one `SchemaWrite`/`SchemaRead`, so the non-compact derived schema and the compact one must live on different types — same reason serde needs `with`-modules).

**Worth a closer review:** factoring the conversions out re-implements the slot-delta calculation. The serde code used `iter().scan(…).collect()`; the shared `lockout_offsets`/`lockouts_from_offsets` use explicit loops over `Vec::with_capacity`/`VecDeque::with_capacity`, so the target collections are now correctly pre-allocated (the `scan`+`collect` path didn't size-hint). Logic is otherwise equivalent.

### DIFF
--- a/vote-interface/src/instruction.rs
+++ b/vote-interface/src/instruction.rs
@@ -16,6 +16,11 @@ use {
     solana_instruction::{AccountMeta, Instruction},
     solana_sdk_ids::sysvar,
 };
+#[cfg(feature = "wincode")]
+use {
+    crate::state::wincode_compact::{CompactTowerSync, CompactVoteStateUpdate},
+    wincode::{SchemaRead, SchemaWrite},
+};
 #[cfg(feature = "serde")]
 use {
     crate::state::{serde_compact_vote_state_update, serde_tower_sync},
@@ -24,6 +29,7 @@ use {
 
 #[repr(u8)]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+#[cfg_attr(feature = "wincode", derive(SchemaWrite, SchemaRead))]
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum CommissionKind {
     InflationRewards = 0,
@@ -31,6 +37,7 @@ pub enum CommissionKind {
 }
 
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+#[cfg_attr(feature = "wincode", derive(SchemaWrite, SchemaRead))]
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum VoteInstruction {
     /// Initialize a vote account
@@ -161,7 +168,9 @@ pub enum VoteInstruction {
     ///   0. `[Write]` Vote account to vote with
     ///   1. `[SIGNER]` Vote authority
     #[cfg_attr(feature = "serde", serde(with = "serde_compact_vote_state_update"))]
-    CompactUpdateVoteState(VoteStateUpdate),
+    CompactUpdateVoteState(
+        #[cfg_attr(feature = "wincode", wincode(with = "CompactVoteStateUpdate"))] VoteStateUpdate,
+    ),
 
     /// Update the onchain vote state for the signer along with a switching proof.
     ///
@@ -170,6 +179,7 @@ pub enum VoteInstruction {
     ///   1. `[SIGNER]` Vote authority
     CompactUpdateVoteStateSwitch(
         #[cfg_attr(feature = "serde", serde(with = "serde_compact_vote_state_update"))]
+        #[cfg_attr(feature = "wincode", wincode(with = "CompactVoteStateUpdate"))]
         VoteStateUpdate,
         Hash,
     ),
@@ -180,7 +190,7 @@ pub enum VoteInstruction {
     ///   0. `[Write]` Vote account to vote with
     ///   1. `[SIGNER]` Vote authority
     #[cfg_attr(feature = "serde", serde(with = "serde_tower_sync"))]
-    TowerSync(TowerSync),
+    TowerSync(#[cfg_attr(feature = "wincode", wincode(with = "CompactTowerSync"))] TowerSync),
 
     /// Sync the onchain vote state with local tower along with a switching proof
     ///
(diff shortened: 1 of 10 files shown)


kskalski:
Rebased, since we now have `StableAbi` derives there and in this PR we can now test that switching serializer to wincode preserves the `abi_digests` ([1266fdb](https://github.com/anza-xyz/solana-sdk/pull/748/commits/1266fdbd54c046378ccab9b76664b71184e5e1c3)).

@alexpyattaev since you just reviewed the `StableAbi` additions there today, you will have good context on reviwing this one now.

vote-interface/src/instruction.rs
```diff
@@ -161,7 +168,9 @@ pub enum VoteInstruction {
     ///   0. `[Write]` Vote account to vote with
     ///   1. `[SIGNER]` Vote authority
     #[cfg_attr(feature = "serde", serde(with = "serde_compact_vote_state_update"))]
-    CompactUpdateVoteState(VoteStateUpdate),
+    CompactUpdateVoteState(
+        #[cfg_attr(feature = "wincode", wincode(with = "CompactVoteStateUpdate"))] VoteStateUpdate,
+    ),
```
buffalojoec, line 173: nit: shouldn't the wincode attributes go in the exact same place as their serde counterparts?

vote-interface/src/state/vote_instruction_data.rs
```diff
@@ -20,11 +20,13 @@ use {
     feature = "frozen-abi",
     frozen_abi(
         api_digest = "GvUzgtcxhKVVxPAjSntXGPqjLZK5ovgZzCiUP1tDpB9q",
-        abi_digest = "6kXER3mQxF3R1Th74tQFGfYUWJfBakNZZVyU4rTrNEek"
+        abi_digest = "6kXER3mQxF3R1Th74tQFGfYUWJfBakNZZVyU4rTrNEek",
+        abi_serializer = "wincode"
```
buffalojoec: It would be great if `frozen-abi` was outfitted to support multiple `abi_serializer` arguments here. Technically this creates a footgun if a future change ever lands that _does not_ affect the wincode digest but _would have_ affected the bincode digest.

Agave uses the bincode path, and I'm wary of losing protection on the bincode ABI digest until the Vote program is migrated over to wincode. Can we keep this using `bincode` for now?

vote-interface/src/state/mod.rs
```diff
@@ -200,212 +204,311 @@ impl<I> CircBuf<I> {
...
+            let confirmation_count = u8::try_from(lockout.confirmation_count())
+                .map_err(|_| "Invalid confirmation count")?;
+            offsets.push(LockoutOffset {
+                offset,
+                confirmation_count,
+            });
+            slot = lockout.slot();
+        }
+        Ok(offsets)
+    }
```
buffalojoec, line 296: nit: this was an unrelated change and should have been done separately. We're trying to verify code movement here, and unnecessary implementation changes muddy the diff.

vote-interface/src/instruction.rs
kskalski: not supported yet, though I'm adding https://github.com/anza-xyz/wincode/pull/325

vote-interface/src/state/vote_instruction_data.rs
kskalski: adding support through https://github.com/anza-xyz/solana-sdk/pull/760 though I suppose it will be a bit tricky to gate it with `cfg(feature = wincode)` here too

for now I will change to bincode and see if we can extend after above PR is in

vote-interface/src/state/mod.rs
kskalski: fair point, I created separate https://github.com/anza-xyz/solana-sdk/pull/761 to include the perf improvement separately and reduce the diff

buffalojoec:
@kskalski we can probably rebase onto #760 and #761, then just use the "both serializers" option in this PR. Wdyt?

buffalojoec (approved):

Rchie121:
> Adds a `wincode` feature mirroring the existing `serde`/`bincode` support, deriving `wincode::SchemaWrite`/`SchemaRead` on every type that derives serde `Serialize`/`Deserialize`. The schemas are bincode-wire-compatible - this is verified by switching `abi_serializer` to `wincode`, which maintains the `abi_digest` unchanged.
> 
> 
> 
> **The bulk of the change is the compact encoding** (`CompactUpdateVoteState*` / `TowerSync*`), which serde implements via custom `serde(with = …)` modules that delta-encode lockout slots as varint offsets in a `short_vec`. To support this without duplicating the wire layout:
> 
> - A new private `compact` module defines the wire-format repr structs **once**, with per-backend field annotations (`serde(with)` / `wincode(with = "Leb128Int…/LockoutOffsetShortVec")`), and owns the shared `VoteStateUpdate`/`TowerSync` ⇄ repr conversions.
> 
> - `serde_compact_*` and a new `wincode_compact` module are thin bridges over those conversions; the wincode schemas are tiny marker types referenced from `VoteInstruction` via `wincode(with = …)` (a type can only carry one `SchemaWrite`/`SchemaRead`, so the non-compact derived schema and the compact one must live on different types — same reason serde needs `with`-modules).
> 
> 
> 
> **Worth a closer review:** factoring the conversions out re-implements the slot-delta calculation. The serde code used `iter().scan(…).collect()`; the shared `lockout_offsets`/`lockouts_from_offsets` use explicit loops over `Vec::with_capacity`/`VecDeque::with_capacity`, so the target collections are now correctly pre-allocated (the `scan`+`collect` path didn't size-hint). Logic is otherwise equivalent.
> 
>

PR #749, opened 2026-06-03
transaction: no-std

grod220:
Continues work of https://github.com/anza-xyz/solana-sdk/pull/746 https://github.com/anza-xyz/solana-sdk/pull/744  https://github.com/anza-xyz/solana-sdk/pull/743 that will enable pinnochio programs to parse tx data.

The PR makes the `solana-transaction` crate no-std with default enabled `std` feature for backwards compatibility.

### DIFF
--- a/transaction/src/versioned/mod.rs
+++ b/transaction/src/versioned/mod.rs
@@ -2,18 +2,23 @@
 
 use {
     crate::Transaction,
+    alloc::vec::Vec,
+    core::cmp::Ordering,
     solana_message::{inline_nonce::is_advance_nonce_instruction_data, VersionedMessage},
     solana_sanitize::SanitizeError,
     solana_sdk_ids::system_program,
     solana_signature::Signature,
-    std::cmp::Ordering,
+};
+#[cfg(all(feature = "wincode", feature = "std"))]
+use {
+    alloc::string::ToString,
+    solana_signer::{signers::Signers, SignerError},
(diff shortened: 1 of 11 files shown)


transaction/src/versioned/mod.rs
```diff
@@ -86,11 +91,11 @@ impl From<Transaction> for VersionedTransaction {
 impl VersionedTransaction {
     /// Signs a versioned message and if successful, returns a signed
     /// transaction.
-    #[cfg(feature = "wincode")]
+    #[cfg(all(feature = "wincode", feature = "std"))]
```
grod220, line 94: Note that the `all` config like so allows no-std consumers to still get wincode deserialization with 
```toml
solana-transaction = { default-features = false, features = ["wincode"] }
```

joncinque (approved):
Looking at solana-signer, it would be nice to make that no-std too so that we don't need to as much strange gating in this PR.

We just need to gate the `EncodablePubkey` and `EncodableKeypair` traits on `std` over there.

We can move forward with this as is though, since it would only add all of the signing functionality without the `std` feature required

grod220:
Good call, can jump on that next 👍

grod220:
@apfitzge or @tao-stones can you review this as well (requires tx-metadata approval)?

tao-stones (approved):
:shipit:

PR #750, opened 2026-06-04
frozen-abi: fix propagate generic bounds in stableabi derives

puhtaytow:
### Description

Derive macros did not propagate generic bounds, which rendered those unusable on types such e.g. `CircBuf` which originally exposed the issue.

### Summary

- fixed StableAbi derives for generic types by adding bounds before generating impls
- added new test type to cover the case

### DIFF
--- a/frozen-abi-macro/src/lib.rs
+++ b/frozen-abi-macro/src/lib.rs
@@ -33,11 +33,11 @@ pub fn derive_stable_abi(_item: TokenStream) -> TokenStream {
 #[proc_macro_derive(StableAbi)]
 pub fn derive_stable_abi(item: TokenStream) -> TokenStream {
     let item = parse_macro_input!(item as Item);
-    let (ident, generics) = match &item {
-        Item::Struct(s) => (&s.ident, &s.generics),
-        Item::Enum(e) => (&e.ident, &e.generics),
-        Item::Type(t) => (&t.ident, &t.generics),
-        _ => {
+    let (ident, generics) = match item {
+        Item::Struct(s) => (s.ident, s.generics),
+        Item::Enum(e) => (e.ident, e.generics),
+        Item::Type(t) => (t.ident, t.generics),
+        item => {
             return Error::new_spanned(
                 item,
                 "StableAbi can only be derived for struct, enum, or type alias",
@@ -46,6 +46,7 @@ pub fn derive_stable_abi(item: TokenStream) -> TokenStream {
             .into();
         }
     };
+    let generics = add_stable_abi_type_param_bounds(generics);
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
 
     let expanded = quote! {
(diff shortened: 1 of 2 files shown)


frozen-abi-macro/src/lib.rs
```diff
@@ -46,6 +46,8 @@ pub fn derive_stable_abi(item: TokenStream) -> TokenStream {
             .into();
         }
     };
+    let mut generics = generics.clone();
```
kskalski: looks like we can operate on owned values:
```
--- a/frozen-abi-macro/src/lib.rs
+++ b/frozen-abi-macro/src/lib.rs
@@ -33,10 +33,10 @@ pub fn derive_stable_abi(_item: TokenStream) -> TokenStream {
 #[proc_macro_derive(StableAbi)]
 pub fn derive_stable_abi(item: TokenStream) -> TokenStream {
     let item = parse_macro_input!(item as Item);
-    let (ident, generics) = match &item {
-        Item::Struct(s) => (&s.ident, &s.generics),
-        Item::Enum(e) => (&e.ident, &e.generics),
-        Item::Type(t) => (&t.ident, &t.generics),
+    let (ident, mut generics) = match item {
+        Item::Struct(s) => (s.ident, s.generics),
+        Item::Enum(e) => (e.ident, e.generics),
+        Item::Type(t) => (t.ident, t.generics),
         _ => {
             return Error::new_spanned(
                 item,
@@ -46,7 +46,6 @@ pub fn derive_stable_abi(item: TokenStream) -> TokenStream {
             .into();
         }
     };
-    let mut generics = generics.clone();
```

frozen-abi-macro/src/lib.rs
```diff
@@ -216,10 +218,22 @@ fn stable_abi_sample_field_expr(field: &syn::Field) -> Result<TokenStream2, Erro
...
+            ::solana_frozen_abi::stable_abi::StableAbi
+        ));
+    }
+}
+
 #[cfg(feature = "frozen-abi")]
 fn derive_stable_abi_sample_struct_type(input: ItemStruct) -> Result<TokenStream2, Error> {
     let type_name = &input.ident;
-    let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
+    let mut generics = input.generics.clone();
```
kskalski: remove clone, just assign the owned field

frozen-abi-macro/src/lib.rs
```diff
@@ -297,7 +311,10 @@ fn stable_abi_sample_enum_variant_expr(
 fn derive_stable_abi_sample_enum_type(input: ItemEnum) -> Result<TokenStream2, Error> {
     let type_name = &input.ident;
     let variants = &input.variants;
-    let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
+    let mut generics = input.generics.clone();
```
kskalski: assign owned field instead of clone

kskalski (commented):
just small nits

frozen-abi-macro/src/lib.rs
```diff
@@ -216,10 +217,22 @@ fn stable_abi_sample_field_expr(field: &syn::Field) -> Result<TokenStream2, Erro
     })
 }
 
+#[cfg(feature = "frozen-abi")]
+fn add_stable_abi_type_param_bounds(generics: &mut syn::Generics) {
```
kskalski: one last thing, since it's always called for the owned value now, how about taking owned syn::Generics and returning it back after modification

kskalski (approved):
Looks good, thanks!

puhtaytow:
Thank you! :)

PR #752, opened 2026-06-04
feat(frozen-abi): add implementations for smart pointers and net

kskalski:
New impls for `StableAbi`:
- `PhantomData<T>`
- `Box<T>`, `Rc<T>`, `Arc<T>` — forward the sampling context to the inner type
- `Result<T, E>`
- `Ipv4Addr`, `Ipv6Addr`, `IpAddr`
- `SocketAddrV4`, `SocketAddrV6`, `SocketAddr` (note: wincode/bincode skip V6 `flowinfo`/`scope_id` (deserialize them to 0), so the samples generated there wouldn't usually round-trip, but this trait is used right now just for digest of wire format)

### DIFF
--- a/frozen-abi/src/stable_abi/impls.rs
+++ b/frozen-abi/src/stable_abi/impls.rs
@@ -3,11 +3,14 @@ use {
         context::{impl_with_context_via, SequenceLenMax, SequenceLenRange},
         StableAbi,
     },
-    core::{array, num::NonZero},
+    core::{array, marker::PhantomData, num::NonZero},
     rand::{Rng, RngCore},
     std::{
         collections::{BTreeMap, BTreeSet, HashMap, HashSet, VecDeque},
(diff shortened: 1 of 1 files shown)


joncinque (approved):
Looks great!

I tried to find a way to implement `StableAbi` for all container types, like we did for `Signer`: https://github.com/anza-xyz/solana-sdk/blob/ed786f554ebec48e2edfc225ad536fdb0d38fd06/signer/src/lib.rs#L126

But there aren't any trait bounds to make that work, so we'll just need to implement for all container types.

kskalski:
I think it's a common problem in rust and the only solution I've seen is through macros like https://crates.io/crates/auto_impl

The other thing, I would prefer to avoid adding blanket impl unless we have a very good reason / fit for it, another limitation of (stable) rust - once you add blanket impl even when constrained on some trait, you can't have others.

PR #753, opened 2026-06-05
feat(frozen-abi): add helper for creating sample FromIterator collections

kskalski:
It's unlikely we would be able (or want to) implement relevant traits (`StableAbi` or random distribution) for all external crates we might need.

In particular a common use-case are purpose-optimized collections.  Several such types are used in the codebase (e.g. `BitVec` or `imbl::HashMap`), we should be able to support stable abi for collections in such a way that:
* it's convenient to use in the types containing them
* the constants and algorithm used for generating them conforms to the way other collections are created and converting from custom collection to standard one (or vice-verse) doesn't cause `abi_digest` differences

Currently providing sampling capability for most of such collections is possible using `frozen_abi(with = "(0..rng.random_range(0..=5)).map(|_| rng.random()).collect()")`, but it's not necessarily meeting above criteria.

This PR adds a helper function that relies on `FromIterator` trait implemented by most collections to conveniently generate sample instances using standard algorithm from `StablAbi` impls.

### DIFF
--- a/frozen-abi/src/stable_abi/impls.rs
+++ b/frozen-abi/src/stable_abi/impls.rs
@@ -14,7 +14,7 @@ use {
     },
 };
 
-const DEFAULT_COLLECTION_MAX_SAMPLE_LEN: usize = 5;
+pub(crate) const DEFAULT_COLLECTION_MAX_SAMPLE_LEN: usize = 5;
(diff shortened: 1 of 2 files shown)


joncinque (approved):
This is a great idea!

PR #754, opened 2026-06-05
frozen-abi: extend stableabi by adding roundtrip

puhtaytow:
### Description

Currently tests generated by `StableAbi` do not perform round trip check, which was mentioned as useful feature number of times.

### Summary
This PR adds the new functionality, without breaking existing implementations.

- added new functionality, to perform roundtrip serialization by default
- added new frozen_abi option `test_roundtrip` with 3 variants; `no`, `wire_only` and `eq_and_wire`
- added a test types to cover new scenarios
- added missing decoding and eq derives to test types
- updated docs

### DIFF
--- a/frozen-abi-macro/src/lib.rs
+++ b/frozen-abi-macro/src/lib.rs
@@ -125,12 +125,8 @@ impl AbiSerializer {
 
     fn serialize_expr(self) -> TokenStream2 {
         match self {
-            Self::Bincode => {
-                quote! { ::solana_frozen_abi::bincode::serialize(&val).unwrap() }
-            }
-            Self::Wincode => {
-                quote! { ::solana_frozen_abi::wincode::serialize(&val).unwrap() }
-            }
+            Self::Bincode => quote! { ::solana_frozen_abi::bincode },
+            Self::Wincode => quote! { ::solana_frozen_abi::wincode },
         }
     }
 }
@@ -170,6 +166,26 @@ fn parse_abi_serializers(expr: &Expr) -> Result<Vec<AbiSerializer>, Error> {
                 .collect()
         }
         expr => Ok(vec![AbiSerializer::from_lit_str(lit_str(expr)?)?]),
+    }
+}
+
+#[cfg(feature = "frozen-abi")]
+enum RoundtripTest {
+    No,
+    WireOnly,
+    EqAndWire,
+}
+
+#[cfg(feature = "frozen-abi")]
+fn parse_roundtrip_test(value: Option<&LitStr>) -> Result<RoundtripTest, Error> {
+    match value {
+        None => Ok(RoundtripTest::WireOnly),
+        Some(value) => match value.value().as_str() {
+            "no" => Ok(RoundtripTest::No),
+            "wire_only" => Ok(RoundtripTest::WireOnly),
+            "eq_and_wire" => Ok(RoundtripTest::EqAndWire),
+            _ => Err(Error::new_spanned(value, "unsupported `test_roundtrip` value; expected \"no\", \"wire_only\", or \"eq_and_wire\"")),
+        },
     }
 }
 
@@ -616,6 +632,7 @@ fn quote_for_test(
     expected_api_digest: Option<&Expr>,
     expected_abi_digest: Option<&Expr>,
     abi_serializers: &[AbiSerializer],
+    roundtrip_test: RoundtripTest,
 ) -> TokenStream2 {
     let test_api = if let Some(expected_api_digest) = expected_api_digest {
         quote! {
@@ -660,6 +677,46 @@ fn quote_for_test(
                 Span::call_site(),
             );
             let abi_serialize_expr = abi_serializer.serialize_expr();
+            let test_roundtrip = match roundtrip_test {
+                RoundtripTest::No => TokenStream2::new(),
+                RoundtripTest::WireOnly | RoundtripTest::EqAndWire => {
+                    let test_roundtrip_eq = match roundtrip_test {
+                        RoundtripTest::EqAndWire => quote! {
+                            assert!(
+                                val == roundtrip_val,
+                                "deserializing serialized {} should preserve value",
+                                roundtrip_type_name
+                            );
+                        },
+                        _ => TokenStream2::new(),
+                    };
+                    quote! {
+                        let roundtrip_type_name = ::std::any::type_name::<#type_name>();
+                        let roundtrip_val: #type_name =
+                            #abi_serialize_expr::deserialize::<#type_name>(&bytes).expect(
+                                ::std::concat!(
+                                    "must deserialize serialized ",
+                                    ::std::stringify!(#type_name)
+                                )
+                            );
+
+                        #test_roundtrip_eq
+
+                        let roundtrip_bytes = #abi_serialize_expr::serialize(&roundtrip_val).expect(
+                            ::std::concat!(
+                                "must re-serialize ",
+                                ::std::stringify!(#type_name)
+                            )
+                        );
+                        assert_eq!(
+                            bytes,
+                            roundtrip_bytes,
+                            "re-serializing deserialized {} should match bytes",
+                            roundtrip_type_name
+                        );
+                    }
+                }
+            };
             quote! {
                 #[test]
                 fn #test_fn_name() {
@@ -672,7 +729,12 @@ fn quote_for_test(
 
                     for _ in 0..10_000 {
                         let val = <#type_name>::random(&mut rng);
-                        digester.hash(&#abi_serialize_expr);
+                        let bytes = #abi_serialize_expr::serialize(&val)
+                            .expect("must serialize");
+
+                        #test_roundtrip
+
+                        digester.hash(&bytes);
                     }
                     assert_eq!(#expected_abi_digest, ::std::format!("{}", digester.result()), "ABI layout has changed!");
                 }
@@ -704,6 +766,7 @@ fn frozen_abi_type_alias(
     expected_api_digest: Option<&Expr>,
     expected_abi_digest: Option<&Expr>,
     abi_serializers: &[AbiSerializer],
+    roundtrip_test: RoundtripTest,
 ) -> TokenStream {
     let type_name = &input.ident;
     let test = quote_for_test(
@@ -712,6 +775,7 @@ fn frozen_abi_type_alias(
         expected_api_digest,
         expected_abi_digest,
         abi_serializers,
+        roundtrip_test,
     );
     let result = quote! {
         #input
@@ -726,6 +790,7 @@ fn frozen_abi_struct_type(
     expected_api_digest: Option<&Expr>,
     expected_abi_digest: Option<&Expr>,
     abi_serializers: &[AbiSerializer],
+    roundtrip_test: RoundtripTest,
 ) -> TokenStream {
     let type_name = &input.ident;
     let test = quote_for_test(
@@ -734,6 +799,7 @@ fn frozen_abi_struct_type(
         expected_api_digest,
         expected_abi_digest,
         abi_serializers,
+        roundtrip_test,
     );
     let result = quote! {
         #input
@@ -794,6 +860,7 @@ fn frozen_abi_enum_type(
     expected_api_digest: Option<&Expr>,
     expected_abi_digest: Option<&Expr>,
     abi_serializers: &[AbiSerializer],
+    roundtrip_test: RoundtripTest,
 ) -> TokenStream {
     let type_name = &input.ident;
     let test = quote_for_test(
@@ -802,6 +869,7 @@ fn frozen_abi_enum_type(
         expected_api_digest,
         expected_abi_digest,
         abi_serializers,
+        roundtrip_test,
     );
     let result = quote! {
         #input
@@ -816,6 +884,7 @@ pub fn frozen_abi(attrs: TokenStream, item: TokenStream) -> TokenStream {
     let mut api_expected_digest: Option<Expr> = None;
     let mut abi_expected_digest: Option<Expr> = None;
     let mut abi_serializers = vec![AbiSerializer::Bincode];
+    let mut test_roundtrip: Option<LitStr> = None;
 
     let attrs_parser = syn::meta::parser(|meta| {
         if meta.path.is_ident("digest") || meta.path.is_ident("api_digest") {
@@ -826,6 +895,9 @@ pub fn frozen_abi(attrs: TokenStream, item: TokenStream) -> TokenStream {
             Ok(())
         } else if meta.path.is_ident("abi_serializer") {
             abi_serializers = parse_abi_serializers(&meta.value()?.parse::<Expr>()?)?;
+            Ok(())
+        } else if meta.path.is_ident("test_roundtrip") {
+            test_roundtrip = Some(meta.value()?.parse::<LitStr>()?);
             Ok(())
         } else {
             Err(meta.error("unsupported \"frozen_abi\" property"))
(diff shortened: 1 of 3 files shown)


puhtaytow:
@alexpyattaev @kskalski whenever you find a moment :pray:

frozen-abi-macro/src/lib.rs
```diff
@@ -577,13 +578,18 @@ fn quote_for_test(
...
-        }
-        AbiSerializer::Wincode => {
-            quote! { ::solana_frozen_abi::wincode::serialize(&val).unwrap() }
+    let abi_serializer = match abi_serializer {
+        AbiSerializer::Bincode => quote! { ::solana_frozen_abi::bincode },
+        AbiSerializer::Wincode => quote! { ::solana_frozen_abi::wincode },
+    };
+    let abi_roundtrip = if abi_roundtrip {
+        quote! {
+            let roundtrip_val: #type_name = #abi_serializer::deserialize::<#type_name>(&bytes).unwrap();
```
kskalski: I would use `.expect()` with some relevant info which part of the checking process is being done

frozen-abi-macro/src/lib.rs
```diff
@@ -577,13 +578,18 @@ fn quote_for_test(
...
-        }
-        AbiSerializer::Wincode => {
-            quote! { ::solana_frozen_abi::wincode::serialize(&val).unwrap() }
+    let abi_serializer = match abi_serializer {
+        AbiSerializer::Bincode => quote! { ::solana_frozen_abi::bincode },
+        AbiSerializer::Wincode => quote! { ::solana_frozen_abi::wincode },
+    };
+    let abi_roundtrip = if abi_roundtrip {
+        quote! {
+            let roundtrip_val: #type_name = #abi_serializer::deserialize::<#type_name>(&bytes).expect("ABI roundtrip could not deserialize!");
```
kskalski: convention for `expect` is something like `"should do X"` or `"must .."`
```suggestion
            let roundtrip_val: #type_name = #abi_serializer::deserialize::<#type_name>(&bytes).expect("must deserialize serialized #type_name");
```
(not sure the proper way to include type name in the expect, but it would be cool to have it)

frozen-abi-macro/src/lib.rs
```diff
@@ -577,13 +578,18 @@ fn quote_for_test(
...
+            let roundtrip_bytes = #abi_serializer::serialize(&roundtrip_val).expect("ABI roundtrip could not re-serialize!");
+            assert_eq!(bytes, roundtrip_bytes, "ABI roundtrip changed bytes!");
```
kskalski:
```suggestion
            assert_eq!(bytes, roundtrip_bytes, "Re-serializing deserialized #type_name should match bytes");
```

kskalski (commented):
I don't think `abi_roundtrip` is a good name, this test isn't that much about ABI (it actually doesn't care what it is and whether it's stable), but about re-serialization of serialized values to check serialization isn't changing data...

I guess `test_roundtrip` would be better, maybe `serialization_roundtrips` (could allow both `serialization_roundtrip` and `serialization_roundstrips`)

frozen-abi-macro/src/lib.rs
```diff
@@ -599,7 +612,9 @@ fn quote_for_test(
 
                 for _ in 0..10_000 {
                     let val = <#type_name>::random(&mut rng);
-                    digester.hash(&#abi_serialize_expr);
+                    let bytes = #abi_serialize_expr::serialize(&val).expect("ABI could not serialize!");
```
kskalski:
```suggestion
                    let bytes = #abi_serialize_expr::serialize(&val).expect("must serialize for ABI digest");
```

kskalski (commented):
Alex, Jon, any preference for the attribute name or different way to trigger such tests?

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -1047,4 +1018,23 @@ mod tests {
...
+            abi_digest = "ErVp1LhW4wAyXr8KudiFF9DpT3ZuGx8mWBeT5mXnCn8m",
+            abi_serializer = "wincode",
+            test_roundtrip
+        )
+    )]
+    struct TestRoundtrip {
+        a: u8,
+        b: bool,
+        c: Option<Vec<u16>>,
+    }
```
kskalski, line 1070: do we have a way to add negative test, i.e. create something with incompatible roundtrip serialization (say #[serde(skip)] + #[wincode(skip)])?

and actually, maybe the other useful feature here will be to have `frozen_abi(skip)` such that we can prevent some fields to be randomly sampled

joncinque (commented):
This looks good to me!

My only question is whether we should make this optional at all. Are there any situations in which a type would implement stable abi but _not_ have roundtrip serialization consistency?

kskalski:
> My only question is whether we should make this optional at all. Are there any situations in which a type would implement stable abi but _not_ have roundtrip serialization consistency?

This might happen if random sampling is creating values that are not normally used in serialization scenarios. I actually introduced one recently (`SocketAddrV6` - it has some fields, which get dropped by `serialize`), but I think it deserves a fix - `StableAbi` should generate values that do roundtrip properly.

Enabling it by default seems like not that bad idea, though:
* we need `#[skip()]` support first - if `#[serde(skip)` or `#[wincode(skip)` is used, derived samples generator should also fill the field with defaults (in wincode it's possible to provide default value / expression too to fill in the value upon deserialization)
* the roundtrip should be generated only if `Deserialize` / `SchemaRead` is actually implemented, not sure how we can detect / gate it automatically
*  we should have an escape hatch anyway, there could be types that have some ambiguity or multiple possible representations for the same serialized format - I know one example is `bv::BitVec` where AFAIR there is inner `Option<Vec<bits>>` and `None` or `Some(vec![])` is semantically the same, but depending on impl could generate different serialized values, while deserialization will create canonical object, so it won't serialize again using the same format

The other thing about current design: we decided to use `serialize` -> `deserialize` -> `serialize` comparing the results of both `serialize` such that we don't require `PartialEq` on the type to compare. It somewhat relates to the possible scope of problems above (eliminates some, introduces different ;-)).

joncinque:
So if I understand correctly, is the proposed approach to move forward with this PR and enable it by default in the future? That's fine with me!

kskalski:
I chatted with @puhtaytow, I guess we can enable by default, but add support for disabling, since the cases not supporting roundtrip should be rare. It should be something like:
* `#[frozen_abi()]`, `#[frozen_abi(test_roundtrip)]` or `#[frozen_abi(test_roundtrip = true)]` -> enable roundtrip if `StableAbi` is derived
* `#[frozen_abi(test_roundtrip = false)]` -> do not test roundtrip

joncinque:
That's even better!

Edit: if we need to keep it simple, we could also _just_ add `skip_roundtrip` as an input to the macro

puhtaytow:
Sorry for not being active during the exchange here. 

The roundtrip now is active by default and could be overriden by `test_roundtrip_skip`. SDK/Agave frozen-abi tests are passing, had to add a few missing derives for test types.

Should be good now :+1:

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -1040,11 +1017,32 @@ mod tests {
         ),
         solana_frozen_abi_macro::frozen_abi(
             abi_digest = "z3Tk2mbrbe6j266gAp6Eo9D9Z5NgvxFQRVHvyJEubrv",
-            abi_serializer = "wincode"
+            abi_serializer = "wincode",
+            test_roundtrip_skip
```
kskalski: the SchemaRead is added on this type, so why skip here?

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -788,9 +794,10 @@ mod tests {
         e: std::net::SocketAddrV6,
```
kskalski, line 785: I imagine this shouldn't pass the roundtrip, can you confirm? will need a skip then, it's a good test actually and we will remove it in subsequent PR

kskalski (commented):
I think `skip_roundtrip_test` is a better sounding and gramatically correct name

frozen-abi/src/stable_abi/impls.rs
puhtaytow: good catch

frozen-abi/src/stable_abi/impls.rs
puhtaytow: It does pass without the skip, so looks like flowinfo and scope_id arent included in the serialized instances. Seem correct no?

puhtaytow: I found this in wincode schame implementations

```
unsafe impl<'de, C: ConfigCore> SchemaRead<'de, C> for SocketAddrV6 {
    type Dst = Self;

    const TYPE_META: TypeMeta = TypeMeta::join_types([
        <Ipv6Addr as SchemaRead<C>>::TYPE_META,
        <u16 as SchemaRead<C>>::TYPE_META,
    ]);

    #[inline]
    fn read(mut reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
        let ip = <Ipv6Addr as SchemaRead<'de, C>>::get(reader.by_ref())?;
        let port = <u16 as SchemaRead<'de, C>>::get(reader)?;
        // serde and thus bincode doesn't encode flowinfo and scope_id
        dst.write(SocketAddrV6::new(ip, port, 0, 0));
        Ok(())
    }
}
```

kskalski: ahh, right, since we don't really compare to the original, it does pass based on purely serialized info roundtrip

This makes me think, that maybe we should still have a mode of comparing deserialized instance to the original instance (applicable if `PartialEq` is available). I wonder how we would configure it then, maybe the attribute should actually be:
* test_roundtrip = no
* test_roundtrip = wire_only (default)
* test_roundtrip = eq_and_wire

puhtaytow: good idea, thanks :pray:

frozen-abi/src/stable_abi/impls.rs
kskalski, line 741: I suppose some (or all?) structs can actually support `eq_and_wire`, let's add them and have `wire_only` (also through default) only for selected types that do not support PartialEq

kskalski (dismissed):
LGTM

alexpyattaev:
> Alex, Jon, any preference for the attribute name or different way to trigger such tests?

Technically as it is this PR is a breaking change. Every consumer of stable_abi gets roundtrip tests by default, whether they want them or not. This is probably a good default, but we must acknowledge that and bump versions accordingly. As for flags, I would go for `no_roundtrip`  to opt-out in cases where deserialization is not expected to be available for the type.

frozen-abi-macro/src/lib.rs
```diff
@@ -578,11 +607,36 @@ fn quote_for_test(
...
+    let test_roundtrip = match roundtrip_test {
+        RoundtripTest::No => TokenStream2::new(),
+        RoundtripTest::WireOnly | RoundtripTest::EqAndWire => {
+            let test_roundtrip_eq = match roundtrip_test {
+                RoundtripTest::EqAndWire => quote! {
+                    assert!(
+                        val == roundtrip_val,
+                        "deserializing serialized {} should preserve value",
+                        roundtrip_type_name
+                    );
```
alexpyattaev: assert_eq! is nicer here as it will also print what the values are.

frozen-abi-macro/src/lib.rs
```diff
@@ -578,11 +607,36 @@ fn quote_for_test(
...
+                },
+                RoundtripTest::No | RoundtripTest::WireOnly => TokenStream2::new(),
+            };
+            quote! {
+                let roundtrip_type_name = ::std::any::type_name::<#type_name>();
+                let roundtrip_val: #type_name = #abi_serialize_expr::deserialize::<#type_name>(&bytes)
+                    .expect(&::std::format!("must deserialize serialized {}", roundtrip_type_name));
```
alexpyattaev: This will run format for every iteration before calling expect and discarding the string pretty much 100% of the time. Not ideal (but not a disaster either).

alexpyattaev (dismissed):
Few minor nits, but generally this looks good.

frozen-abi-macro/src/lib.rs
kskalski: this would require type to implement `Debug` if I recall correctly

frozen-abi/src/lib.rs
```diff
@@ -128,6 +129,37 @@
...
+//!     b: bool,
+//! }
+//! ```
+//!
+//! ```rust,ignore
+//! #[derive(PartialEq, StableAbi, StableAbiSample, wincode::SchemaWrite, wincode::SchemaRead)]
+//! #[frozen_abi(
+//!     abi_digest = "...",
+//!     abi_serializer = "wincode",
+//!     test_roundtrip = eq_and_wire,
```
joncinque: nit: for consistency with all of the other attributes, I would vote to have the test type in quotes too, ie
```suggestion
//!     test_roundtrip = "eq_and_wire",
```

joncinque (commented):
Just one little nit from me, take it or leave it

frozen-abi-macro/src/lib.rs
kskalski: might make sense to construct the string using compile time constructs:
```rust
.expect(::std::concat!(
            "must deserialize serialized ",
            ::std::stringify!(#type_name)
        ));
```

kskalski:
> Technically as it is this PR is a breaking change. Every consumer of stable_abi gets roundtrip tests by default, 

Yeah, I wonder... This is a breaking change with an easy escape switch, we were treating frozen-abi::StableAbi as somewhat experimental so far and relying on our ability to fix the only codebase that uses it (solana-sdk / agave). The issue is that major version change will cause pain for all the other crate updates (I think), so I vote on passing this one as minor still.

> accordingly. As for flags, I would go for `no_roundtrip` to opt-out in cases where deserialization is not expected to be available for the type.

Since now we need a way to pick one of 3 modes of operation (I can imagine we could add more or make them additive based on a list of "features", like enabling eq check, wire test, debug print example for failed checks), I think we should keep the `test_roundtrip = ..` specifier. 
`no_roundtrip` seems like a nice quick "tag" on the type to say it doesn't support any roundtrip operation, if we really need this kind of shorthand, it could be added in addition (not necessarily in this PR) to `test_roundtrip = `, I guess.

frozen-abi-macro/src/lib.rs
puhtaytow: yeah that would require Debug

frozen-abi-macro/src/lib.rs
```diff
@@ -578,11 +600,41 @@ fn quote_for_test(
...
+                    );
+                },
+                RoundtripTest::No | RoundtripTest::WireOnly => TokenStream2::new(),
+            };
+            quote! {
+                    let roundtrip_type_name = ::std::any::type_name::<#type_name>();
+                    let roundtrip_val: #type_name = #abi_serialize_expr::deserialize::<#type_name>(&bytes).expect(::std::concat!(
+                "must deserialize serialized ",
+                ::std::stringify!(#type_name)
+            ));
```
kskalski: something looks off with formatting, could it be improved?

puhtaytow: strangely fmt check script seem to not be able to catch these

kskalski (approved):
LGTM, I think all the issues are addressed.

joncinque (approved):
Looks great to me!

puhtaytow:
Thank you guys! :sunglasses:

PR #755, opened 2026-06-05
signer: no-std

grod220:
Follow up from https://github.com/anza-xyz/solana-sdk/pull/749

Makes the signer crate no-std compatible and creates a new default `std` feature that is default on.

### DIFF
--- a/signer/src/lib.rs
+++ b/signer/src/lib.rs
@@ -1,17 +1,26 @@
 //! Abstractions and implementations for transaction signers.
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![no_std]
+extern crate alloc;
+#[cfg(feature = "std")]
+extern crate std;
+
+#[cfg(feature = "std")]
 use {
-    core::fmt,
-    solana_pubkey::Pubkey,
-    solana_signature::Signature,
-    solana_transaction_error::TransactionError,
+    alloc::{boxed::Box, string::ToString},
     std::{
         error,
         fs::{self, File, OpenOptions},
         io::{Read, Write},
-        ops::Deref,
         path::Path,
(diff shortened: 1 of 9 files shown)


signer/src/lib.rs
```diff
@@ -154,21 +163,21 @@ impl PartialEq for dyn Signer {
...
 
-impl std::fmt::Debug for dyn Signer {
-    fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
+impl fmt::Debug for dyn Signer {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         write!(fmt, "Signer: {:?}", self.pubkey())
     }
 }
 
-/// Removes duplicate signers while preserving order. O(n²)
```
grod220, line 163: Don't think this was actually `n²`

signer/src/lib.rs
```diff
@@ -154,21 +163,21 @@ impl PartialEq for dyn Signer {
...
+    let mut seen = BTreeSet::new();
     for signer in signers {
         let pubkey = signer.pubkey();
-        if !seen.contains(&pubkey) {
-            seen.insert(pubkey);
+        let is_unique = seen.insert(pubkey);
+        if is_unique {
             out.push(signer);
         }
     }
```
grod220, line 183: We could keep the original and put it behind a `std` flag, but figured swapping for `BTreeSet` is low effort

tao-stones (approved):
lgtm

grod220:
Additionally requesting @joncinque as he requested the follow up!

joncinque (approved):
Looks great, thanks!

PR #758, opened 2026-06-08
frozen-abi: skip field sampling

puhtaytow:
### Description

StableAbi sampling does not include option to skip field sampling path, which might be problematic in some scenarios.

### Summary of changes
- added sampling option `skip` to redirect from field sampling path to defaults
- updated docs
- added new test type

ref #754

### DIFF
--- a/frozen-abi/src/stable_abi/impls.rs
+++ b/frozen-abi/src/stable_abi/impls.rs
@@ -1102,4 +1102,45 @@ mod tests {
         b: bool,
         c: Option<Vec<u16>>,
     }
+
+    #[derive(Default, wincode::SchemaWrite, wincode::SchemaRead)]
+    struct TestSkippedFieldWithoutStableAbi;
+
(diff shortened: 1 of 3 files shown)


puhtaytow:
@kskalski :pray:

frozen-abi/src/stable_abi/impls.rs
```diff
@@ -1102,4 +1102,49 @@ mod tests {
...
+            abi_digest = SKIP_VS_DEFAULT_DIGEST,
+            abi_serializer = "wincode",
+            test_roundtrip = "no",
+        )
+    )]
+    struct TestSkipFieldSampling {
+        #[stable_abi_sample(skip)]
+        a: u64,
+        #[stable_abi_sample(skip)]
+        b: Vec<u64>,
```
kskalski: the best set will be to include some not-skipped field and field that is not supporting StableAbi, then we could also control that against a type where skipped field is replaced by (), which shouldn't change the digest

kskalski (approved):
It's more or less a syntax sugar for `with = Default::default()`, but I think it's useful, as it would match wincode / serde skip attribute

PR #759, opened 2026-06-08
nullable: Add int primitives impl

febo:
### Problem

Currently it is not possible to use integer primitives with `MaybeNull` since they do not have a `Nullable` impl.

### Solution

Add a `Nullable` impl that reserves `0` as the "null" value.


joncinque (approved):
Great idea!

grod220 (approved):
LGTM

PR #760, opened 2026-06-09
feat(frozen-abi-macro): support multiple serializers

kskalski:
### Problem

`#[frozen_abi(...)]`'s `abi_serializer` only accepts a single serializer name (`"bincode"` or `"wincode"`). When a type is serialized with **both** `bincode` and `wincode` and we want to assert they produce the same ABI digest, there is no way to express that on one type — it requirs two near-identical duplicate structs (`TestStructWincode` / `TestStructBincode`) differing only in name and serializer.

### Change
`abi_serializer` now accepts either a single string or a list:

```rust
#[frozen_abi(abi_digest = "...", abi_serializer = "wincode")]              // single (unchanged)
#[frozen_abi(abi_digest = "...", abi_serializer = ["bincode"])]           // list of one
#[frozen_abi(abi_digest = "...", abi_serializer = ["bincode", "wincode"])] // list of many
```

For each serializer, a **separate** ABI test is generated with the serializer name in the test name (`test_abi_digest_bincode`, `test_abi_digest_wincode`), all asserting against the **shared** `abi_digest`. The tests run independently/in sequence — no cross-checking between serializers. The API digest test (`test_api_digest`) is serializer-independent and is still emitted once.

### Backward compatibility
Fully backward compatible — the single-string form behaves exactly as before.

### DIFF
--- a/frozen-abi-macro/src/lib.rs
+++ b/frozen-abi-macro/src/lib.rs
@@ -95,7 +95,7 @@ use syn::{
 };
 
 #[cfg(feature = "frozen-abi")]
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 enum AbiSerializer {
     Bincode,
     Wincode,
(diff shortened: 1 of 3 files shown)


frozen-abi-macro/src/lib.rs
```diff
@@ -90,16 +90,89 @@ use proc_macro2::{Span, TokenStream as TokenStream2, TokenTree};
...
+            }
+        }
+    }
+}
+
+/// Parse the `abi_serializer` attribute value, which may be either a single
+/// string literal (e.g. `"wincode"`) or a list of string literals
+/// (e.g. `["bincode", "wincode"]`).
+#[cfg(feature = "frozen-abi")]
+fn parse_abi_serializers(expr: &Expr) -> Result<Vec<AbiSerializer>, Error> {
```
buffalojoec, line 142: Can you add some unit tests for this parser? I can't believe we don't have any in here already.

buffalojoec (commented):
Awesome! Thanks for adding this. I think we should use this everywhere.

frozen-abi-macro/src/lib.rs
kskalski: done 1fedb0d6

buffalojoec (approved):
Lgtm, but should we update `scripts/test-stable.sh` or `scripts/test-frozen-abi.sh` to run these tests? I believe we're covered by the [coverage tests](https://github.com/anza-xyz/solana-sdk/actions/runs/27262803465/job/80512720373).

kskalski:
Hah, right, those tests won't run in stable because we don't enable frozen-abi due to it requiring nightly... I think this needs a bit of poking to see how to enable tests - I'm not convinced froze-abi-macro actually needs to gate with nightly builds/frozen-abi feature, but if so we might stick the test into `test-frozen-abi.sh`.

Since we are indeed covered in coverages tests, I think I will follow up on separate PR.

kskalski:
I think the best way will be to cleanup this weirdness where `froze-abi-macros` is not doing anything by default and requires enabling `frozen-abi` feature to actually generate the expected tests when derives are placed on structs. For sure would put someone in a false conviction that they are safe. ;-)

This would be a breaking change, but AFAICT all actual users of the derive do gate derive attributes on frozen-abi, so seems like it's worth cleaning up and it will also bring better consistency to the testing of this crate in solana-sdk itself.

PR #762, opened 2026-06-09
ci: adding gh app for release flow

nbelenkov:
Similar to https://github.com/anza-xyz/pinocchio/pull/424, we have a PAT that makes the tag + bump version and we should be using a github app here. 

## What we need to do still
- Create the release github app
- Needs `vars.CLIENT_ID`, `vars.APP_ID`, and `secrets.PRIVATE_KEY` set in the `prod` environment.
- The App has to be in the branch-protection bypass list, since publish pushes the version commit + tag to a protected branch

### DIFF
--- a/.github/workflows/publish-rust.yml
+++ b/.github/workflows/publish-rust.yml
@@ -218,10 +218,18 @@ jobs:
     permissions:
       contents: write
     steps:
+      - name: Generate app token
+        if: github.event.inputs.dry_run != 'true'
+        id: app-token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1  # v3.2.0
+        with:
+          client-id: ${{ vars.CLIENT_ID }}
+          private-key: ${{ secrets.PRIVATE_KEY }}
+
       - name: Git Checkout
         uses: actions/checkout@v4
         with:
-          token: ${{ secrets.ANZA_TEAM_PAT }}
+          token: ${{ inputs.dry_run && github.token || steps.app-token.outputs.token }}
           fetch-depth: 0 # get the whole history for git-cliff
 
       - name: Setup Environment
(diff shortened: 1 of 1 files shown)


nbelenkov:
cc @yihau for github app settings

yihau:
looks like we're missing the `environment: prod` here.

.github/workflows/publish-rust.yml
```diff
@@ -247,8 +255,16 @@ jobs:
...
         run: |
-          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
-          git config --global user.name "github-actions[bot]"
+          if [ "${{ inputs.dry_run }}" == "true" ]; then
+            git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+            git config --global user.name "github-actions[bot]"
+          else
+            git config --global user.email "${APP_ID}+${APP_SLUG}[bot]@users.noreply.github.com"
+            git config --global user.name "${APP_SLUG}[bot]"
+          fi
```
yihau: nig: we can just use `github-actions[bot]` to keep things simple.

nbelenkov: simplified it in the same way as pinocchio here [091cc0d](https://github.com/anza-xyz/solana-sdk/pull/762/commits/091cc0dc707cb25be495fba650d10e23b7a4e9f1). I am keen to keep it so that it matches our other repos

joncinque (commented):
Looks good to me! But I'll let @yihau give the final approval

yihau (approved):

kskalski:
I guess we need to remove the "upgrade" option from Publish Crate action now, or should it be able to update Cargo.toml with new version? It failed on me during last step of publish (https://github.com/anza-xyz/solana-sdk/actions/runs/27369584175/job/80878271722)

nbelenkov:
Hmm, yeah, the easy fix is to allow the GitHub app to push to master, so add to the bypass list. The better long-term solution is to bump before the release flow, similar to how we do it in agave. We would be it manually or via an action

joncinque:
Let's add the app to the bypass list for now -- if we add a separate version bump flow, it'll also need the privileges anyway, right?

yihau:
sorry, I totally forgot sdk will push the bump version directly. let's use this to add the app to the bypass list: https://github.com/anza-xyz/devops/pull/1772 + https://github.com/anza-xyz/devops/pull/1773

nbelenkov:
> separate version bump flow, it'll also need the privileges anyway, right

No, the current permissions would be fine, because it will only open a PR, which anyone can do.

joncinque:
Oh sorry, I meant a job that would actually commit the version bump, not just create a PR.

PR #763, opened 2026-06-10
feat(frozen-abi-macro): enable frozen-abi feature by default

kskalski:
#### Problem
`solana-frozen-abi-macro` defaults to `default = []`, exposing dummy no-op proc-macros unless `frozen-abi` is explicitly enabled. The justifying comments ("requires nightly Rust" / "dummy macros for stable rustc") are obsolete / wrong — the real macros compile fine on stable, and the dummy path is not exercised in any reasonable scenario:
- crates in agave typically depend on macros with the `frozen-abi` feature on — directly or transitively via `solana-frozen-abi/frozen-abi` — so feature unification selects the real macros in any frozen-abi build - if they do not (as in some examples in solana-sdk itself), it's a bug
- **Consumers gate the derives at the call site behind their own `frozen-abi` feature** (e.g. `#[cfg_attr(feature = "frozen-abi", derive(AbiExample, StableAbi))]`), so the macro crate's toggle only matters when that gate is active — at which point it's always on.

The bad side-effect of this feature gating is that:
* `scripts/test-stable.sh` skips running any tests for the `frozen-abi-macro` crate itself (which are now present).
* if feature enablement is forgotten, we will silently skip generating digest tests (as in genesis-config)

#### Change
* Enable `frozen-abi` by default and mark it deprecated. The `#[cfg]` gates and dummy macros are left in place — this is deprecation, not removal — so `features = ["frozen-abi"]` and `default-features = false` both keep working.
* Fix some sdk crates to enable the feature on `solana-frozen-abi` (the failure mode is that they declare frozen-abi derives and generate tests, but the library's actual content isn't compiled)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -260,7 +260,7 @@ solana-fee-calculator = { path = "fee-calculator", version = "3.2.1" }
 solana-fee-structure = { path = "fee-structure", version = "3.0.0" }
 solana-file-download = { path = "file-download", version = "3.0.0" }
 solana-frozen-abi = { path = "frozen-abi", version = "3.5.0" }
-solana-frozen-abi-macro = { path = "frozen-abi-macro", version = "3.5.0" }
+solana-frozen-abi-macro = { path = "frozen-abi-macro", version = "3.5.0", default-features = false }
 solana-genesis-config = { path = "genesis-config", version = "4.0.0" }
 solana-hard-forks = { path = "hard-forks", version = "3.1.0", default-features = false }
 solana-hash = { path = "hash", version = "4.4.0", default-features = false }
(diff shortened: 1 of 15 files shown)


kskalski:
Maybe we should do the same to `frozen-abi` crate? The issue is that it bring in nightly-only feature compilations, so maybe not (would require careful gating of dependency as optional and only enabled when nightly toolchain is used)

apfitzge (commented):
tx-metadata: no issue with the changes themselves. Will let Jon weigh in on if this is the direction we want to go for default features before I give approval.

joncinque (commented):
Yeah this makes sense, thanks!

I can't totally remember why we did it this way -- maybe we didn't gate all usage with the `frozen-abi` feature in individual crates?

Can you just add `default-features = false` at the workspace-level definition?

https://github.com/anza-xyz/solana-sdk/blob/914140ed54b993326cd95077cf0161621208bacf/Cargo.toml#L263

Since we're enabling the feature explicitly in individual crates, we can disable it explicitly at the top-level

joncinque (approved):
Looks great!

apfitzge (approved):

PR #764, opened 2026-06-11
fix(frozen-abi): make SocketAddrV6 compatible with eq_and_wire

kskalski:
Since we made `test_roundtrip` default and will encourage using `test_roundtrip = "eq_and_wire"` whenever types support `PartialEq`, the previous choice to generate all possible `SocketAddrV6`, even those where serialization strips part of the state, seem wrong.

Make all `StableAbi` impls provided by the library conform to `eq_and_wire` by only generating samples that do serialize and deserialize back into the same object.


joncinque (approved):
Makes sense!

PR #765, opened 2026-06-12
instructions-sysvar: handle overflow

apfitzge:
### Problem
- txv1's larger size introduces the possibility of overflowing the instructions sysvar format

### Solution
- handle overflow in instruction sysvar construction
- add transaction error variant for the overflow

### DIFF
--- a/instructions-sysvar/src/lib.rs
+++ b/instructions-sysvar/src/lib.rs
@@ -67,15 +67,16 @@ pub struct Instructions();
 solana_sysvar_id::impl_sysvar_id!(Instructions);
 
 /// Construct the account data for the instructions sysvar.
+/// Returns None if the instructions cannot be serialized correctly.
 ///
 /// This function is used by the runtime and not available to Solana programs.
 #[cfg(not(target_os = "solana"))]
-pub fn construct_instructions_data(instructions: &[BorrowedInstruction]) -> Vec<u8> {
-    let mut data = serialize_instructions(instructions);
+pub fn construct_instructions_data(instructions: &[BorrowedInstruction]) -> Option<Vec<u8>> {
+    let mut data = serialize_instructions(instructions)?;
     // add room for current instruction index.
     data.resize(data.len() + 2, 0);
 
-    data
+    Some(data)
 }
 
 #[cfg(not(target_os = "solana"))]
@@ -110,9 +111,11 @@ bitflags! {
 // - N = num_instructions
 // - A = number of accounts in a particular instruction
 // - D = data_len
+//
+// Returns None if the instructions cannot be serialized correctly.
 #[cfg(not(target_os = "solana"))]
 #[cfg_attr(feature = "dev-context-only-utils", qualifiers(pub))]
-fn serialize_instructions(instructions: &[BorrowedInstruction]) -> Vec<u8> {
+fn serialize_instructions(instructions: &[BorrowedInstruction]) -> Option<Vec<u8>> {
     // 64 bytes is a reasonable guess, calculating exactly is slower in benchmarks
     let mut data = Vec::with_capacity(instructions.len() * (32 * 2));
     append_u16(&mut data, instructions.len() as u16);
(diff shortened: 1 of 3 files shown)


instructions-sysvar/src/lib.rs
```diff
@@ -110,9 +111,11 @@ bitflags! {
 // - N = num_instructions
 // - A = number of accounts in a particular instruction
 // - D = data_len
+//
+// Returns None if the instructions cannot be serialized correctly.
 #[cfg(not(target_os = "solana"))]
 #[cfg_attr(feature = "dev-context-only-utils", qualifiers(pub))]
-fn serialize_instructions(instructions: &[BorrowedInstruction]) -> Vec<u8> {
+fn serialize_instructions(instructions: &[BorrowedInstruction]) -> Option<Vec<u8>> {
```
t-nelson: `Result<T>` plz

apfitzge: 👍 sure was on fence here. No current appropriate error in scope so I went to Option...but yeah Result is better

instructions-sysvar/src/lib.rs
```diff
@@ -67,15 +67,18 @@ pub struct Instructions();
 solana_sysvar_id::impl_sysvar_id!(Instructions);
 
 /// Construct the account data for the instructions sysvar.
+/// Returns None if the instructions cannot be serialized correctly.
```
joncinque: nit: since it's a `Result`, probably don't need to specify anything here

instructions-sysvar/src/lib.rs
```diff
@@ -110,9 +119,13 @@ bitflags! {
 // - N = num_instructions
 // - A = number of accounts in a particular instruction
 // - D = data_len
+//
+// Returns None if the instructions cannot be serialized correctly.
```
joncinque: nit: same here, can probably remove

transaction-error/src/lib.rs
```diff
@@ -147,6 +147,9 @@ pub enum TransactionError {
 
     /// Commit cancelled internally.
     CommitCancelled,
+
+    /// Instruction sysvar overflow.
+    InstructionsSysvarOverflow,
```
joncinque, line 160: This is a breaking change, but *shockingly*, we don't need to bump majors everywhere.

We could just publish this as a minor and go on with our lives, but if we want to publish a major, we'd also need to publish the following, minor bumps will be fine:

* solana-message
* solana-client-traits
* solana-sdk
* solana-transaction

solana-signer is the diciest one, and should technically be a a major since it wraps `TransactionError` for some reason that I don't understand, but I think we can get away with a minor.

joncinque (commented):
Looks great to me! Just the comment nits and the part about publishing, let me know how you want to proceed

instructions-sysvar/src/lib.rs
apfitzge: 8186be7e

instructions-sysvar/src/lib.rs
apfitzge: 8186be7e

transaction-error/src/lib.rs
apfitzge: If it's breaking, any reason to not just have it be a major? Simply annoyance on our (and devs) side?

joncinque: Yep annoyance to everyone is the main factor. Just note that you'll need to disable semver checks when publishing.

And actually, I was overly optimistic -- we'd need a new major on solana-transaction, since it re-exports `TransactionError` and `TransactionResult`, and also includes the type in its public API.

joncinque (approved):

transaction-error/src/lib.rs
jamie-osec: This also broke `solana-runtime@3`, which matches exhaustively on `TransactionError`

joncinque: Ah darn, I thought runtime pinned those dependencies, we might have to do the right thing then -- yank and re-publish a bunch

joncinque: We ended up reverting the new error type, so please use the newest solana-transaction-error

PR #766, opened 2026-06-12
stake-history: extract crate from `solana-stake-interface`

grod220:
> [!IMPORTANT]
> Do not squash-merge this PR. It carries commits of grafted history and must be rebased-merged.

### Problem

`StakeHistory` is the only sysvar whose type lives outside this repo. Two motivations with this PR:
  - Co-locate all sysvar crates together
  - Unblock sizing work from anza-xyz/agave#12245 as the plan is for every sysvar crate to own its serialized account size and it's likely that we'll create a shared sizing crate here making impl easier if all sysvars were co-located

### Change summary

The two source files were grafted from solana-program/stake with their full history then rebased onto `master`:

```
git filter-repo \
  --path interface/src/stake_history.rs \
  --path interface/src/sysvar/stake_history.rs \
  --path-rename interface/src/stake_history.rs:stake-history/src/lib.rs \
  --path-rename interface/src/sysvar/stake_history.rs:sysvar/src/stake_history.rs
```

Also had a bot rewrite the PR refs in the commit messages to the qualified form (`(#343)` → `(solana-program/stake#343)`)  so links resolve to their actual repo.

There were additional commits on top of all the migrated work:
- Scaffolding crates
- `declare_sysvar_id!` is replaced by `impl_sysvar_id!` in the type crate
- Making crate no-std (small import change)

### Follow-ups

- [x] `SIZE` consts across all sysvar crates + patch releases
- [x] Publish `solana-stake-history` 1.0.0, then `solana-sysvar` 4.1.0 (dependent publishes second)
- [x] `solana-stake-interface` to depend on `solana-stake-history` + re-export
- [x] Stake program: consume `solana_sysvar::stake_history::StakeHistorySysvar` + `solana-stake-history` types directly
- [x] js updates ([bot](https://github.com/anza-xyz/solana-sdk/pull/766#issuecomment-4689519624))
- [x] https://github.com/anza-xyz/agave/pull/13550
- [x] Resume anza-xyz/agave#12245, closing anza-xyz/agave#10672
- [x] https://github.com/anza-xyz/mollusk/pull/271


stake-history/src/lib.rs
```diff
@@ -0,0 +1,189 @@
+//! A type to hold data for the [`StakeHistory` sysvar][sv].
```
grod220, line 1: File migrated from https://github.com/solana-program/stake/blob/0fa06c20f74929d1781e41807d80c569ac32a3b9/interface/src/stake_history.rs

sysvar/src/stake_history.rs
```diff
@@ -0,0 +1,248 @@
...
+};
+
+impl Sysvar for StakeHistory {}
+
+// we do not provide Default because this requires the real current epoch
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct StakeHistorySysvar(pub Epoch);
+
+impl StakeHistoryGetEntry for StakeHistorySysvar {
+    fn get_entry(&self, target_epoch: Epoch) -> Option<StakeHistoryEntry> {
```
joncinque: Since we want to lighten up `solana-sysvar` and eventually make it a `solana-sdk`-like crate which just re-exports other types, how about moving this type and implementation into `sysvar.rs` in `solana-stake-history`?

It just needs `get_sysvar` from the root of this repo, but we can inline that in `solana-stake-history`, or maybe even move it into some lower-level crate that other sysvar crates can use.

stake-history/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name = "solana-stake-history"
+description = "Types and utilities for the Solana StakeHistory sysvar."
+documentation = "https://docs.rs/solana-stake-history"
+version = "1.0.0"
```
joncinque: Just be careful on the first publish to set the version to 1.0.0 explicitly, otherwise we'll get a strange first version.

Alternatively, we can set this to `0.0.0` so that the first major publish becomes `1.0.0`, I'm happy either way

joncinque (commented):
Looks good to me overall! Just a point on the solana-sysvar bit

stake-history/Cargo.toml
grod220: True, will put it to 0.0.0

sysvar/src/stake_history.rs
grod220: I see what you mean. Yes, can move to the stake-history crate, but we indeed need to figure out the `get_sysvar` first. I tried inlining a duplicate version, but it seems mocks go through the sysvar crate's private `SYSCALL_STUBS`.

The second option with an external crate seems the way to go, but should go in before this PR. 

Question: did we want to migrate `program_stubs.rs` as well to this new crate? It would come with additional deps though. If not, we'd have to have some forwarder for the stubs calls.

joncinque: Eh, don't worry about `program_stubs`, we should really just deprecate them

stake-history/src/sysvar.rs
```diff
@@ -0,0 +1,219 @@
+//! History of stake activations and de-activations.
```
grod220, line 1: Migrated from: https://github.com/solana-program/stake/blob/0fa06c20f74929d1781e41807d80c569ac32a3b9/interface/src/sysvar/stake_history.rs

README.md
joncinque, line 1: Can we revert these changes? They're technically still correct for v3 😅

joncinque (commented):
Looks great to me! Just the bit on the README

stake-history/src/lib.rs
buffalojoec: Looks like you also picked up this one? https://github.com/solana-program/stake/commit/26f219306def9e5a7390b0c1e69a0337f0b328e5

stake-history/src/sysvar.rs
```diff
@@ -115,40 +116,10 @@ impl StakeHistoryGetEntry for StakeHistorySysvar {
...
-            let slice = unsafe { std::slice::from_raw_parts_mut(var_addr, length as usize) };
-            slice.copy_from_slice(&self.data[offset as usize..(offset + length) as usize]);
-            0 // SUCCESS
-        }
-    }
-    pub fn mock_get_sysvar_syscall(data: &[u8]) {
-        set_syscall_stubs(Box::new(MockGetSysvarSyscall {
-            data: data.to_vec(),
-        }));
-    }
```
buffalojoec: I saw in the previous PR we intentionally dropped use of stubs, but aren't we losing regression coverage on `StakeHistorySysvar`?

README.md
grod220: Oh true. Also, my ide autoformatter totally mangled the format too. Reverted.

stake-history/src/lib.rs
grod220: Yep! Migrated that commit.

stake-history/src/sysvar.rs
grod220: I can't replicate the get_entry() call, but this sort of relates to this comment from the last PR https://github.com/anza-xyz/solana-sdk/pull/770#discussion_r3423676218. Think we can extract a private helper that does the offset calculation/decoding and then we can test that, restoring coverage. Have a look.

joncinque: Good shout, sorry for missing that on this PR. I like the refactor to make it testable :+1:

buffalojoec (commented):
LGTM, deferring final stamp to Jon.

grod220:
@joncinque if things look good, would you be able to use admin privileges to do a rebase-merge and not a squash merge on master?

joncinque (approved):

joncinque:
Done!

PR #767, opened 2026-06-13
message: Fix AccountKeys equality for different lengths

fallintoplace:
## Summary
- Compare `AccountKeys` iterators directly so equality checks that both iterators are exhausted.
- Add regression coverage for equal flattened key sequences and same-prefix sequences with extra keys.

## Details
The previous implementation zipped the two flattened key iterators and checked that all paired keys matched. Since `zip` stops at the shorter iterator, a key list could compare equal to another list with the same prefix plus additional keys.

## Tests
- `cargo test -p solana-message --features serde`

### DIFF
--- a/message/src/account_keys.rs
+++ b/message/src/account_keys.rs
@@ -1,7 +1,7 @@
 use {
     crate::{compiled_instruction::CompiledInstruction, v0::LoadedAddresses, CompileError},
     alloc::{collections::BTreeMap, vec::Vec},
-    core::{iter::zip, ops::Index},
+    core::ops::Index,
     solana_address::Address,
     solana_instruction::Instruction,
 };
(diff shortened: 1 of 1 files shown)


apfitzge (commented):
@joncinque similarly here. this PR is correct in implementation, but wonder if the implementation of trait is useful or not.

joncinque:
I completely removed the `Eq` derivation and `PartialEq` implementation, and both the SDK and Agave still passed clippy checks, so this is definitely unused.

It's technically a breaking change to remove it, but I still lean towards killing it. What do you think?

apfitzge:
I think for now we can fix this, create a release, and then consider full removal separately?

apfitzge (approved):

joncinque (approved):
Works for me!

PR #768, opened 2026-06-13
Fix versioned transaction signature verification

fallintoplace:
## Summary
- Reject invalid versioned transaction signature counts before hashing the message.
- Return `false` from `verify_with_results()` for missing required signatures.
- Add regression coverage for missing signatures.

## Details
`VersionedTransaction::_verify_with_results()` zipped signatures with static account keys, so a transaction with fewer signatures than required produced fewer verification results. `verify_and_hash_message()` then called `.all()` on those results, which allowed an empty result set to pass.

This keeps the existing `SignatureFailure` result for verification failure while making the versioned transaction verifier fail closed when the signature count is invalid.

## Tests
- `cargo test -p solana-transaction --features dev-context-only-utils,wincode`

### DIFF
--- a/transaction/src/versioned/mod.rs
+++ b/transaction/src/versioned/mod.rs
@@ -199,6 +199,10 @@ impl VersionedTransaction {
         &self,
     ) -> solana_transaction_error::TransactionResult<solana_hash::Hash> {
         let message_bytes = self.message.serialize();
+        if self.sanitize_signatures().is_err() {
+            return Err(solana_transaction_error::TransactionError::SignatureFailure);
+        }
+
         if !self
             ._verify_with_results(&message_bytes)
             .iter()
@@ -219,10 +223,20 @@ impl VersionedTransaction {
 
     #[cfg(feature = "verify")]
     fn _verify_with_results(&self, message_bytes: &[u8]) -> Vec<bool> {
-        self.signatures
-            .iter()
-            .zip(self.message.static_account_keys().iter())
-            .map(|(signature, pubkey)| signature.verify(pubkey.as_ref(), message_bytes))
+        let num_required_signatures = usize::from(self.message.header().num_required_signatures);
+        let static_account_keys = self.message.static_account_keys();
+
+        (0..num_required_signatures)
+            .map(|index| {
+                let Some(signature) = self.signatures.get(index) else {
+                    return false;
+                };
+                let Some(pubkey) = static_account_keys.get(index) else {
+                    return false;
+                };
+
+                signature.verify(pubkey.as_ref(), message_bytes)
+            })
             .collect()
     }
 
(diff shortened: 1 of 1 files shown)


apfitzge (commented):
This seems to be a correct in that it fixes the issue.

That we did not notice it before is probably because this method is not used anywhere (afaict) in agave.

@joncinque does this method actually have value? What is the point of verifying an *unsanitized* transaction?

joncinque (commented):
This looks worth fixing.

I don't know the full codepaths well, but it's used in agave at https://github.com/anza-xyz/agave/blob/065bdf7c035702af9bd91bfd06236ad891c52083/runtime/src/bank.rs#L5440

And used in the CLI at https://github.com/anza-xyz/agave/blob/065bdf7c035702af9bd91bfd06236ad891c52083/cli-output/src/cli_output.rs#L2975

That one is using `Transaction` instead of `VersionedTransaction`, but it's reasonable to assume we'd port that to support `VersionedTransaction at some point.

transaction/src/versioned/mod.rs
```diff
@@ -219,10 +223,20 @@ impl VersionedTransaction {
 
     #[cfg(feature = "verify")]
     fn _verify_with_results(&self, message_bytes: &[u8]) -> Vec<bool> {
-        self.signatures
-            .iter()
-            .zip(self.message.static_account_keys().iter())
-            .map(|(signature, pubkey)| signature.verify(pubkey.as_ref(), message_bytes))
+        let num_required_signatures = usize::from(self.message.header().num_required_signatures);
```
apfitzge, line 226: IMO it is better for us to put the sanitize signatures check here - and then leave the zip loop as it was.

With the current impl using .get we are doing additional bounds checks on every iteration.

apfitzge:
> This looks worth fixing.
> 
> I don't know the full codepaths well, but it's used in agave at [anza-xyz/agave@`065bdf7`/runtime/src/bank.rs#L5440](https://github.com/anza-xyz/agave/blob/065bdf7c035702af9bd91bfd06236ad891c52083/runtime/src/bank.rs#L5440)
> 
> And used in the CLI at [anza-xyz/agave@`065bdf7`/cli-output/src/cli_output.rs#L2975](https://github.com/anza-xyz/agave/blob/065bdf7c035702af9bd91bfd06236ad891c52083/cli-output/src/cli_output.rs#L2975)
> 
> That one is using `Transaction` instead of `VersionedTransaction`, but it's reasonable to assume we'd port that to support `VersionedTransaction at some point.

@joncinque wrt to the agave call - the fix would allow us to fail earlier on these invalid txns. Right now we could be verifying incorrectly on these mismatched sizes. but then we sanitize immediately after verification so discard there if there was a mismatch.

joncinque:
Definitely sounds like it's worth it then! Also agreed with your comment

PR #770, opened 2026-06-15
get-sysvar: extract from `solana-sysvar`

grod220:
Inspired from https://github.com/anza-xyz/solana-sdk/pull/766#discussion_r3402719323

We want to lighten `solana-sysvar` and let each sysvar's accessor live in its own type crate rather than all of them sitting in `solana-sysvar`. The blocker is `get_sysvar` which needs to be shared across crates. This PR extracts it into a low-level `solana-get-sysvar` crate that type crates can depend on directly.

`solana-sysvar` keeps the existing public shape by re-exporting `GetSysvar as Sysvar`, so downstream code does not break.

### Changes
- new `solana-get-sysvar` crate
  - `get_sysvar` / `get_sysvar_unchecked` copied over
  - new `GetSysvarStub` trait
  - no-std support
- `solana-sysvar` (no public API change)
- Update sysvar type crates to implement new `GetSysvar`


get-sysvar/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+#[cfg(all(not(target_os = "solana"), feature = "std"))]
+mod stubs;
+
+#[cfg(all(not(target_os = "solana"), feature = "std"))]
+pub use stubs::{clear_get_sysvar_stub, set_get_sysvar_stub, GetSysvarStub};
+
+/// Syscall success code.
+//
+// Defined in solana-program-entrypoint as [`SUCCESS`](https://github.com/anza-xyz/solana-sdk/blob/program-entrypoint@v2.2.1/program-entrypoint/src/lib.rs#L35).
+const SUCCESS: u64 = 0;
```
grod220, line 17: Locally defined to avoid std import (program-entrypoint)

get-sysvar/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        stubs::sol_get_sysvar(sysvar_id, var_addr, offset, length)
+    }
+
+    // Off-chain no-std builds have neither runtime syscalls nor host stubs.
+    #[cfg(all(not(target_os = "solana"), not(feature = "std")))]
+    {
+        let _ = (sysvar_id, var_addr, offset, length);
+        solana_program_error::UNSUPPORTED_SYSVAR
+    }
```
grod220: Given the no-std feature, we need an explicit host + no-std fallback that raises

get-sysvar/src/lib.rs
```diff
@@ -0,0 +1,102 @@
+//! Access to the `sol_get_sysvar` syscall, used to fetch sysvar data from the runtime.
```
grod220, line 1: Functionality copied over from sysvar/src/lib.rs but with some modifications
- sol_get_sysvar() for better re-use
- fix code comments referencing agave master branch w/ broken links

get-sysvar/src/stubs.rs
```diff
@@ -0,0 +1,46 @@
+use std::sync::{Arc, RwLock};
+
+/// Off-chain stub implementation of the `sol_get_sysvar` syscall.
```
grod220: This file is meant to be a simple replacement for the use of `program_stubs.rs` which seems to be on its way out. Added as there still seems to be need to run tests and inject mocks offchain (sysvar crates own tests & program-test crate).

sysvar/src/program_stubs.rs
```diff
@@ -142,19 +142,6 @@ pub fn sol_invoke_signed(
...
-    sysvar_id_addr: *const u8,
-    var_addr: *mut u8,
-    offset: u64,
-    length: u64,
-) -> u64 {
-    SYSCALL_STUBS
-        .read()
-        .unwrap()
-        .sol_get_sysvar(sysvar_id_addr, var_addr, offset, length)
-}
```
grod220, line 156: Now that `get_sysvar` reads `solana-get-sysvar`'s registry instead of `SYSCALL_STUBS`, the mocks have to install there instead. 

Anyone mocking sysvars (agave's program-test) must install a `GetSysvarStub` via set_get_sysvar_stub instead. This is an important follow up.

joncinque: This is the correct approach, which we did when we moved out the sysvar-specific syscalls

get-sysvar/src/stubs.rs
joncinque: Let's please remove this, it's mostly a footgun. We can test the functions without invoking the actual syscall, or we can just put in a fallback just for testing

get-sysvar/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+//!
+//! On-chain calls go directly to the syscall. Off-chain calls are routed through a
+//! stub which test harnesses can install with [`set_get_sysvar_stub`] to serve mock
+//! sysvar data.
+#![no_std]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+
+#[cfg(all(not(target_os = "solana"), feature = "std"))]
+extern crate std;
```
joncinque: I think we can remove all std usage if we remove the stubs, which is another good reason to remove it

joncinque (commented):
Looks good overall!

I think we should move the `Sysvar` trait down into this new crate, since that's just a convenience for the `get()` function included here.

At that point, maybe we change the trait name to `GetSysvar` and re-export as `Sysvar` in `solana-sysvar`.

Also tagging @febo for his thoughts on this.

clock/src/sysvar.rs
```diff
@@ -1,4 +1,12 @@
...
+    crate::Clock,
+    solana_get_sysvar::{impl_sysvar_get, GetSysvar},
+    solana_sysvar_id::impl_sysvar_id,
+};
 
 impl_sysvar_id!(Clock);
+
+impl GetSysvar for Clock {
+    impl_sysvar_get!(id());
+}
```
grod220, line 12: Moved all `GetSysvar` impls into their respective crates as orphan rule issues with external crates doing impls for external traits.

epoch-schedule/src/sysvar.rs
```diff
@@ -1,4 +1,109 @@
...
-use {crate::EpochSchedule, solana_sysvar_id::impl_sysvar_id};
+use {
+    crate::EpochSchedule,
+    solana_get_sysvar::{get_sysvar_unchecked, GetSysvar},
+    solana_program_error::ProgramError,
+    solana_sysvar_id::impl_sysvar_id,
+};
 
 impl_sysvar_id!(EpochSchedule);
+
```
grod220, line 10: Changes in this file copied over from `sysvar/src/epoch_schedule.rs` (and re-exported there)

get-sysvar/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+    #[cfg(not(target_os = "solana"))]
+    {
+        let _ = (sysvar_id, var_addr, offset, length); // warning suppression
+        solana_program_error::UNSUPPORTED_SYSVAR
+    }
+}
+
+/// Implements [`GetSysvar::get`] for runtime-backed sysvars.
+#[macro_export]
+macro_rules! impl_sysvar_get {
```
grod220: Copied over from `sysvar/src/lib.rs` except for the fee sysvar specific variant (implemented that separately)

sysvar/src/clock.rs
```diff
@@ -4,7 +4,7 @@
 //! current slot, the current epoch, and the approximate real-world time of the
 //! slot.
 //!
-//! [`Clock`] implements [`Sysvar::get`] and can be loaded efficiently without
+//! [`Clock`] implements [`crate::Sysvar::get`] and can be loaded efficiently without
```
grod220, line 7: These comment links need to route through the re-export now

sysvar/src/clock.rs
```diff
@@ -153,41 +148,4 @@ mod tests {
...
-        let expected = Clock {
-            slot: 1,
-            epoch_start_timestamp: 2,
-            epoch: 3,
-            leader_schedule_epoch: 4,
-            unix_timestamp: 5,
-        };
-
-        let data = to_bytes(&expected);
-        crate::tests::mock_get_sysvar_syscall(&data);
```
grod220, line 169: Lots of tests deleted as they just exercise stub calls (which are now gone in get-sysvar crate)

sysvar/src/fees.rs
```diff
@@ -56,16 +56,36 @@ impl Fees {
...
+                SUCCESS => Ok(fees),
+                _ => Err(solana_program_error::ProgramError::UnsupportedSysvar),
+            }
+        }
+
+        #[cfg(not(target_os = "solana"))]
+        {
+            Err(solana_program_error::ProgramError::UnsupportedSysvar)
+        }
+    }
```
grod220, line 80: Instead of keeping the macro variant for this one deprecated use-case, just implemented it here (1:1 the same [behavior](https://github.com/anza-xyz/solana-sdk/pull/770/changes#diff-5d13454213dfea7a8682d2671d090df7730cd6c7bc65cd5e5e3d57c02becb597L167-L183)).

clock/src/sysvar.rs
joncinque: Even better!

get-sysvar/src/lib.rs
joncinque: nit: can we rename this to `impl_get_sysvar`?

get-sysvar/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+// Defined in the Agave syscalls crate as [`OFFSET_LENGTH_EXCEEDS_SYSVAR`](https://github.com/anza-xyz/agave/blob/v4.0.2/syscalls/src/sysvar.rs#L180).
+const OFFSET_LENGTH_EXCEEDS_SYSVAR: u64 = 1;
+
+/// Return value indicating that the sysvar was not found.
+//
+// Defined in the Agave syscalls crate as [`SYSVAR_NOT_FOUND`](https://github.com/anza-xyz/agave/blob/v4.0.2/syscalls/src/sysvar.rs#L179).
+const SYSVAR_NOT_FOUND: u64 = 2;
+
+/// Interface for loading a sysvar directly from the runtime.
+pub trait GetSysvar: Default + Sized {
```
joncinque: Since we use `MaybeUninit` in the implementation, I think we can remove the `Default` bound here. I don't think that counts as a breaking change

sysvar/src/slot_hashes.rs
```diff
@@ -196,83 +192,4 @@ mod tests {
...
-            slot_hashes.push((
-                i as u64,
-                hash(&[(i >> 24) as u8, (i >> 16) as u8, (i >> 8) as u8, i as u8]),
-            ));
-        }
-
-        let check_slot_hashes = SlotHashes::new(&slot_hashes);
-        mock_slot_hashes(&check_slot_hashes);
-
-        let pod_slot_hashes = PodSlotHashes::fetch().unwrap();
```
joncinque, line 231: This test is exercising a bit more logic to make sure `SlotHashes` and `PodSlotHashes` behave similarly, so can we instead create `PodSlotHashes` from the bincode-serialized bytes of `check_slot_hashes` and have the test behave as before?

joncinque (commented):
This is really great! We've wanted this for a long time, and it's definitely the correct approach to get away from `solana-sysvar`, and should also be a step in the right direction for pinocchio.

Just a few small things on my side

get-sysvar/src/lib.rs
grod220: Needed for SysvarSerialize, but can move this bound there

sysvar/src/slot_hashes.rs
grod220: True, that removal was too blunt. Going to restore. Also think a small private fn in that fetch call will allow the test to access the stuff we are asserting later.

sysvar/src/lib.rs
```diff
@@ -77,15 +77,11 @@
 //!
 //! [sysvardoc]: https://docs.solanalabs.com/runtime/sysvars
 
-// hidden re-exports to make macros work
-pub mod __private {
-    #[cfg(target_os = "solana")]
-    pub use solana_define_syscall::definitions;
-    pub use {solana_program_entrypoint::SUCCESS, solana_program_error::ProgramError};
-}
+pub use solana_get_sysvar::{get_sysvar, impl_get_sysvar, GetSysvar as Sysvar};
```
joncinque: nti: not sure if we care, but the rename is technically a breaking change
```suggestion
pub use solana_get_sysvar::{get_sysvar, impl_get_sysvar as impl_sysvar_get, GetSysvar as Sysvar};
```

joncinque (commented):
One last point, looks great!

sysvar/src/lib.rs
grod220: True, easy enough to do

joncinque (approved):
:shipit:

PR #771, opened 2026-06-15
feat: support and check StableAbi for message and transaction

kskalski:
VersionedTransaction` and `TransactionError` previously had only `AbiExample`, not `StableAbi`. This adds `StableAbi` to both, along with the supporting types in their field trees (the message components and `InstructionError`).

Most types use the `StableAbiSample` derive. `VersionedTransaction` instead uses a manual `Distribution` (like the legacy `Transaction`): the wire format couples fields together — the legacy message's first byte must stay below the versioned-message prefix, and V1 sizes its signature array from the header — so a random sample has to be built consistently to survive a roundtrip. It still reuses the derived component samplers.

Each top-level type gains an `abi_digest` layout test. Types that support both encoders are digested under `["bincode", "wincode"]` (the two agree byte-for-byte); `VersionedTransaction` is wincode-only since V1 isn't bincode-compatible. All tests use `test_roundtrip = "eq_and_wire"`.

A `wincode`-enabling dev-dependency was added to a couple of crates so their wincode digest tests compile.

### DIFF
--- a/transaction/src/versioned/mod.rs
+++ b/transaction/src/versioned/mod.rs
@@ -1,5 +1,7 @@
 //! Defines a transaction which supports multiple versions of messages.
 
+#[cfg(feature = "frozen-abi")]
+use solana_frozen_abi_macro::{frozen_abi, AbiExample, StableAbi};
 use {
     crate::Transaction,
     alloc::vec::Vec,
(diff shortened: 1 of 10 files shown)


tao-stones (dismissed):
ABI usage looks consistent to me, please give @apfitzge a chance to review before merge.

transaction/src/versioned/mod.rs
```diff
@@ -79,6 +89,94 @@ pub struct VersionedTransaction {
...
+                    header,
+                    account_keys: <Vec<Address> as StableAbi>::random(rng),
+                    recent_blockhash: StableAbi::random(rng),
+                    instructions: <Vec<CompiledInstruction> as StableAbi>::random(rng),
+                    address_table_lookups:
+                        <Vec<v0::MessageAddressTableLookup> as StableAbi>::random(rng),
+                }),
+                <Vec<Signature> as StableAbi>::random(rng),
+            ),
+            _ => {
```
apfitzge: could we do 1, and then _ => unreachable!()?

it feels weird to have txv1 be a catchall.

apfitzge (commented):
minor comment otherwise lgtm

transaction/src/versioned/mod.rs
kskalski: it's actually 2 => and then _ => unreachable, done in [97f95d9](https://github.com/anza-xyz/solana-sdk/pull/771/commits/97f95d9b3a33c1f71d6ffd47a207ffdea60e6cd2)

apfitzge (approved):
lgtm

PR #772, opened 2026-06-16
feat: add StableAbi to several crates

kskalski:
Continues the `StableAbi` rollout to the public serialized types in `solana-hard-forks`, `solana-epoch-schedule`, `solana-inflation`, `solana-rent`, and `solana-account` (the previously-uncovered `AccountSharedData`). Each gains the `StableAbi` + `StableAbiSample` derives; all of their fields already implement `StableAbi`, so no other changes are needed.


joncinque (approved):
Looks great! Keep the train going 😄

PR #775, opened 2026-06-17
ci: pinning all gh actions

nbelenkov:
continuing my efforts to pin all of our deps/actions across all repos

- Pinned every external action to a SHA across the workflows and the `setup` composite action.
- Swapped `dtolnay/rust-toolchain` for `actions-rust-lang/setup-rust-toolchain`, so you can actually pin it. 
- Added `github-actions` to `dependabot.yml` (daily, max 3 open PRs, 7-day cooldown), mirroring agave.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -21,6 +21,6 @@ jobs:
         )
       )
     steps:
-      - uses: tibdex/backport@9565281eda0731b1d20c4025c43339fb0a23812e
+      - uses: tibdex/backport@9565281eda0731b1d20c4025c43339fb0a23812e  # v2.0.4
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


.github/workflows/backport.yml
```diff
@@ -21,6 +21,6 @@ jobs:
         )
       )
     steps:
-      - uses: tibdex/backport@9565281eda0731b1d20c4025c43339fb0a23812e
+      - uses: tibdex/backport@9565281eda0731b1d20c4025c43339fb0a23812e  # v2.0.4
```
joncinque, line 24: We could eventually replace this action, it doesn't work most of the time 😅

joncinque (approved):
Looks great, thanks!

.github/workflows/backport.yml
nbelenkov: Yeah there are some actions we can clean up, but let's do it in another PR

PR #776, opened 2026-06-18
AbiEnumVisitor for smart pointers

apfitzge:
### Problem
- Arc/Rc/Box do not implement AbiEnumVisitor
- Ran into an issue in agave when trying to change a type from `Arc<Vec<PacketBatch>>` to `Arc<PacketBatch>`

### Solution
- Implement AbiEnumVisitor for Arc/Rc/Box

### DIFF
--- a/frozen-abi/src/abi_digester.rs
+++ b/frozen-abi/src/abi_digester.rs
@@ -533,7 +533,11 @@ impl SerializeStructVariant for AbiDigester {
 
 #[cfg(test)]
 mod tests {
-    use std::{collections::HashMap, sync::atomic::AtomicIsize};
+    use std::{
+        collections::HashMap,
+        rc::Rc,
+        sync::{atomic::AtomicIsize, Arc},
+    };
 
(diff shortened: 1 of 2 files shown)


joncinque (commented):
I wonder if we can just update the implementation on `&T` to instead work with `AsRef<T>` and make our lives simpler.

What do you think @kskalski ?

kskalski:
I suppose the purpose of `AbiExample` is to print out exactly the types (i.e. changing `Arc<u8>` to `Rc<u8>` should change the api digest, right?), this implementation does exactly that. 

Would the idea of `AsRef<T>` still allow that and also, wouldn't it need a blanket impl to work on trait, i.e. I read what you suggested as
```rust
impl<T: AsRef<AbiEnumVisitor>> AbiEnumVisitor for T {
}
```
and I'm not sure we can (or should) add such blanket impl.

joncinque:
Oh gotcha, that makes sense then, we can go forward with this

joncinque (approved):

PR #777, opened 2026-06-18
perf(account): skip redundant memory init during sysvar account create

kskalski:
#### Problem
When creating sysvar accounts we initialize a `vec![0; spaze]` and then overwrite it serializing the actual data. This came up in profiles and we can skip zeroing by using uninitialized Vec as serialize destination.

#### Changes
1. **New `Account::new_with_data(lamports, data, owner)`** — takes ownership of a pre-built `Vec` without reallocating or zeroing.
2. **`Account::new_data_with_space`** and **`AccountSharedData::new_data_with_space`** — rewritten to allocate once at `space` via `Vec::with_capacity`, `serialize_into` (no realloc), then `resize` to zero **only the trailing padding**. The explicit `serialized_size > space` check preserves the original `SizeLimit` error (which a bare `serialize_into` into a `Vec` would have silently lost by growing past `space`).
3. **`create_account_with_fields`** — delegate to the optimized `new_data_with_space`.

### DIFF
--- a/account/src/lib.rs
+++ b/account/src/lib.rs
@@ -357,6 +357,24 @@ fn shared_deserialize_data<T: serde::de::DeserializeOwned, U: ReadableAccount>(
     bincode::deserialize(account.data())
 }
 
+/// Serialize `state` into `writer`, failing with [`bincode::ErrorKind::SizeLimit`]
+/// once `limit` bytes have been written. The limit is enforced during
+/// serialization, so a `Serialize` impl that under-reports its size (or whose
+/// output varies between calls) cannot write past `limit`. Fixint encoding
+/// matches the default `bincode::serialize` byte format.
+#[cfg(feature = "bincode")]
+fn limited_serialize_into<W: std::io::Write, T: serde::Serialize>(
+    writer: W,
+    state: &T,
+    limit: u64,
+) -> Result<(), bincode::Error> {
+    use bincode::Options;
+    bincode::options()
+        .with_fixint_encoding()
+        .with_limit(limit)
+        .serialize_into(writer, state)
+}
+
 #[cfg(feature = "bincode")]
 fn shared_serialize_data<T: serde::Serialize, U: WritableAccount>(
     account: &mut U,
@@ -422,11 +440,10 @@ impl Account {
         space: usize,
         owner: &Pubkey,
     ) -> Result<Self, bincode::Error> {
-        if bincode::serialized_size(state)? > space as u64 {
-            return Err(Box::new(bincode::ErrorKind::SizeLimit));
-        }
+        // The byte limit prevents a misbehaving `Serialize` impl from growing
+        // `data` past `space` (which the `resize` below would then truncate).
         let mut data = Vec::with_capacity(space);
-        bincode::serialize_into(&mut data, state)?;
+        limited_serialize_into(&mut data, state, space as u64)?;
         data.resize(space, 0);
         Ok(Account::new_with_data(lamports, data, owner))
     }
@@ -583,11 +600,10 @@ impl AccountSharedData {
         space: usize,
         owner: &Pubkey,
     ) -> Result<Self, bincode::Error> {
-        if bincode::serialized_size(state)? > space as u64 {
-            return Err(Box::new(bincode::ErrorKind::SizeLimit));
-        }
+        // The byte limit prevents a misbehaving `Serialize` impl from growing
+        // `data` past `space` (which the `resize` below would then truncate).
         let mut data = Vec::with_capacity(space);
-        bincode::serialize_into(&mut data, state)?;
+        limited_serialize_into(&mut data, state, space as u64)?;
         data.resize(space, 0);
         Ok(Self::create_from_existing_shared_data(
             lamports,
--- a/account/src/lib.rs
+++ b/account/src/lib.rs
@@ -381,6 +381,17 @@ impl Account {
     pub fn new_ref(lamports: u64, space: usize, owner: &Pubkey) -> Rc<RefCell<Self>> {
         Rc::new(RefCell::new(Account::new(lamports, space, owner)))
     }
+    /// Create an `Account` from an already-built data buffer, taking ownership
+    /// of it without reallocating or zeroing.
+    pub fn new_with_data(lamports: u64, data: Vec<u8>, owner: &Pubkey) -> Self {
+        Account {
+            lamports,
+            data,
+            owner: *owner,
+            executable: false,
+            rent_epoch: Epoch::default(),
+        }
+    }
     #[cfg(feature = "bincode")]
     pub fn new_data<T: serde::Serialize>(
         lamports: u64,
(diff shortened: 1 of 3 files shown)


account/src/lib.rs
```diff
@@ -411,9 +422,13 @@ impl Account {
...
     ) -> Result<Self, bincode::Error> {
-        let mut account = Account::new(lamports, space, owner);
-        shared_serialize_data(&mut account, state)?;
-        Ok(account)
+        if bincode::serialized_size(state)? > space as u64 {
+            return Err(Box::new(bincode::ErrorKind::SizeLimit));
+        }
+        let mut data = Vec::with_capacity(space);
+        bincode::serialize_into(&mut data, state)?;
+        data.resize(space, 0);
```
LucasSte, line 447: `serialize_into` returns an error when the buffer is not large enough.

When you do `with_capacity`, although you allocate enough bytes, `vec.len()` will still return zero. Is this construct correctly working? Will serialize correctly grow the vector inside?

I'm wondering if the construct would work better with either `Vec::set_len` of `MaybeUninit`.

LucasSte:
I believe the main user of `struct Account` is still accounts DB, so I would defer the ultimate approval to @brooksprumo.

Although I was the last one to modify these files, runtime is mostly using `AccountSharedData`.

account/src/lib.rs
kskalski: I think that is true for slices, which cannot be grown, here we pass `&mut Vec`:
- `bincode::serialize_into` takes any `W: std::io::Write`.
- `&mut Vec<u8>` implements `Write` by **appending** (`Vec::extend`), growing the vec from `len() == 0` upward and using whatever capacity is already reserved.

I imagine using MaybeUninit and some unsafe code would save... a few comparisons, likely not worth it.

kskalski:
For reference, this is the stack trace that Alessandro mention to motivate this change:
```
solana_core::replay_stage::ReplayStage::new::{{closure}} [/home/sol/src/agave/core/src/replay_stage.rs:1018]
solana_core::replay_stage::ReplayStage::process_active_banks [/home/sol/src/agave/core/src/replay_stage.rs:4303]
solana_core::replay_stage::ReplayStage::process_replay_results [/home/sol/src/agave/core/src/replay_stage.rs:4006]
solana_runtime::bank::Bank::freeze [/home/sol/src/agave/runtime/src/bank.rs:2996]
solana_runtime::bank::Bank::update_slot_history [/home/sol/src/agave/runtime/src/bank.rs:2513]
solana_runtime::bank::Bank::update_sysvar_account [/home/sol/src/agave/runtime/src/bank.rs:2343]
solana_runtime::bank::Bank::update_slot_history::{{closure}} [/home/sol/src/agave/runtime/src/bank.rs:2519]
solana_account::create_account_shared_data_with_fields [solana-account-4.3.0/src/lib.rs:636]
solana_account::create_account_with_fields [solana-account-4.3.0/src/lib.rs:619]
solana_account::Account::new [solana-account-4.3.0/src/lib.rs:366]
alloc::vec::from_elem [library/alloc/src/vec/mod.rs:3677]
<u8 as alloc::vec::spec_from_elem::SpecFromElem>::from_elem [library/alloc/src/vec/spec_from_elem.rs:51]
alloc::raw_vec::RawVec<T,A>::with_capacity_zeroed_in [library/alloc/src/raw_vec/mod.rs:221]
```

account/src/lib.rs
```diff
@@ -568,9 +583,19 @@ impl AccountSharedData {
...
         owner: &Pubkey,
     ) -> Result<Self, bincode::Error> {
-        let mut account = AccountSharedData::new(lamports, space, owner);
-        shared_serialize_data(&mut account, state)?;
-        Ok(account)
+        if bincode::serialized_size(state)? > space as u64 {
+            return Err(Box::new(bincode::ErrorKind::SizeLimit));
+        }
+        let mut data = Vec::with_capacity(space);
+        bincode::serialize_into(&mut data, state)?;
```
brooksprumo: Codex found an issue here. Since `state` is generic and just has to implement `Serialize`, a malicious input could be a type that produces different sizes on the second call to serialize().

The `state` is serialized once to get the size, and check if too large. That could return OK. Then when `state` is serialized a second time for real, it could be a different (larger) value.

We should have a way that limits how much we serialize into `data`. Maybe like so:

```rust
bincode::options()
      .with_limit(space as u64)
      .serialize_into(&mut data, state)?;
```

That or we could implement our own wrapper around Vec that implements `Writer` and prevents writing into it beyond a given size limit.

We'll need to do this in both spots.

brooksprumo (changes requested):

account/src/lib.rs
kskalski: That's possible scenario. I changed to use bincode options for now - we will add code for wincode separately and use MaybeUninit, probably in a way that isn't re-useable for bincode anyway.

The limited serialize into is now extracted into a function, so it could be used in a few places, but not in  `shared_serialize_data` - this one reports a specialized error kind that is checked in `state_traits.rs`:
```rust
    fn set_state(&mut self, state: &T) -> Result<(), InstructionError> {
        self.serialize_data(state).map_err(|err| match *err {
            ErrorKind::SizeLimit => InstructionError::AccountDataTooSmall,
            _ => InstructionError::GenericError,
        })
    }
```

nagisa: Why do we need to care about arbitrary `T`s here? Won't this code only ever be instantiated with well known `T`s that we control (and therefore provably not malicious in any way?)

(`bincode::options().with_limit()` sounds like an okay idea regardless...)

kskalski: Right, I'm not sure how sensitive those checks are, the code could also allow the serialization to take up as much space as it wants and check the `.len()` afterwards.

bincode's `serialized_size` is actually performing the serialization, but providing a custom sink that counts bytes, so it's pretty unusual for those to not match (the `Serialize` impl would need to be malicious and behave differently on subsequent runs). If this API is only called on user's own types, maybe we could keep the checks simple (though is `options().with_limit` faster or slower than `serialized_size` + `serialized_into` is not an easy question with universal answer I suppose).

brooksprumo: Darn, `with_limit().serialize_into()` won't fix the issue either. It's the same problem as the original impl.

---

I think understanding who is allowed to call this code will dictate the appropriate fix.

kskalski: Haha, `with_limit` just does `serialized_size` first...
I guess the best route is to just replace this stuff with wincode, I don't have a clarity what is the path for migrating sysvars / their handling in `solana-account` though.

@grod220 do you have an opinion how this API should look like long term?

grod220: > @grod220 do you have an opinion how this API should look like long term?

1 - Remove usage of SysvarSerialize https://github.com/anza-xyz/agave/issues/10672 in agave (still actively being worked on) to reduce dep version coupling issues between sdk and agave. Eventually it should be deleted.
2. Have Agave construct sysvar accounts directly using newly exported size consts https://github.com/anza-xyz/solana-sdk/pull/786. This addresses the feedback from https://github.com/anza-xyz/agave/pull/12245 which then can be progressed.
3. Swap out bincode w/ wincode

kskalski:
Closing in favor of https://github.com/anza-xyz/solana-sdk/pull/842

PR #781, opened 2026-06-19
fix: reject invalid txv1 config mask on deserialize

apfitzge:
### Problem
- raw bytes containing unknown config bits are not rejected. they are then serialized differently

### Summary of Changes
- Reject txs with unknown bits on deserialization
- Remove misleading sanitization check on mask


joncinque (approved):
This makes sense, confirming that it isn't possible to construct `Message` with incorrect bits otherwise

tao-stones (approved):
lgtm

Issue #774, opened 2026-06-16
[solana-signer-store] malleability issues in `decode` function


robertkiel:
### Problem 

To facilitate the verification of threshold signatures, the verifier is given a bitmap (`signer-store`) of the utilized validator public keys. The encoding of the bitmap has the form `[flag, length[0..2], encoding[..]]` where `flag` refers to the utilized encoding and `length` encodes the length in bits of the encoded bitmap. Depending on the given flag, the bitmap is encoded using base-2 or base-3 encoding.

It turned out that the [decoding](https://github.com/anza-xyz/solana-sdk/blob/496131f4dacc0b23c497cfff0da99a89fd84ff42/signer-store/src/lib.rs#L186) function has a malleability issue when length is not a multiple of 8, e.g. 7. In that case, the adversary may set the most significant bits arbitrarily.

Example for Base-2 encoding with length 1.

```
[Version::Base2, [1,0], 0b0000_0001]
[Version::Base2, [1,0], 0bxxxx_xxx1]

// -> all decode to Decode::Base2(bitvec![u8, Lsb0; 1])
```

As there are multiple ways to express the same input, it becomes more challenging for a validator to fingerprint validator public key bitmaps, e.g. in order to prevent from replays.
Secondly, the adversary may change length to up to the next multiple of 8 so that the decoder decodes the same encoding to a different bitmap. 

### Proposed to solution

Change the `decode` method to require all unused bits set to 0 and fail decoding otherwise.

For base-3 encoding:

At the end of the decoding of a chunk, require remainder to be zero

```rust
    for bit_index in start_bit..end_bit {
        // ...
        // Reject non-canonical chunks: any value not consumed by the 5 symbols
        // (byte >= 243, or unused high symbols in the final partial chunk).
        if block_num != 0 {
            return Err(DecodeError::CorruptDataPayload);
        }

        // ...
    }
```

### PoCs

Malleability in encoding for base-2 and base-3

```rust
/// `decode` validates the *number* of payload bytes exactly, but it does NOT
/// validate the unused padding bits in the final data byte: `decode_impl_base2`
/// reads the whole byte and then `truncate`s away the high bits. As a result,
/// two *different* byte arrays decode to the *same* value — the encoding is
/// malleable / non-canonical.
#[test]
fn test_decode_base2_non_canonical_padding_bits_accepted() {
    // A 3-bit vector [1, 0, 1]. Only the low 3 bits of the single data byte
[...]

PR #782, opened 2026-06-21
[signer-store] Prevent padding malleability in signer-store

samkim-crypto:
#### Problem

https://github.com/anza-xyz/solana-sdk/issues/774

#### Summary of Changes

Fixes malleability vulnerabilities in the base-2 and base-3 decoders by strictly enforcing that all unused padding bits and unconsumed chunk symbols are set to zero, rejecting non-canonical encodings.


akhi3030 (approved):
nice!

PR #785, opened 2026-06-22
frozen-abi: extend wire stability tests; random, fuzzer

puhtaytow:
### Description

This PR brings serialization testing strategies to existing `StableAbi` feature set. It not only keep the original `Random` rng based strategy with fixed seed as before, but brings as well fuzzer over `BoreloFuzzer`.

### Summary of changes
The changes here are backward compatible with existing implementations, and don't demand from users taking any actions.

- Added two testing strategies; `Random` and `BoleroFuzzer`
- Added new test type for bolero fuzzer, which could be used with any fuzzing engine supported by harness
- Added `fuzz-frozen-abi.sh` script to invoke fuzzing over harness
- Extracted roundtrip quote to separate helper function
- Patched existing quote tests body to use the new roundtrip helper
- Previously implemented stability tests over `frozen_abi` proc attribute macro, been redirected to `quote_for_stable_abi_macro_invocation` for consistency reasons
- Added `generate_serialization_test` function-like macro to implement both testing strategies to type without using proc attribute or derive
- Updated docs!
- Added new test types covering newly added functionality
- Added new features `fuzz-bolero` and `fuzz-bolero-repro`, the latter used to gate test type which being successfully crashed by fuzzer

---
Fuzzer output

```
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 2193902655
INFO: Loaded 1 modules   (39920 inline 8-bit counters): 39920 [0x5834277ce330, 0x5834277d7f20), 
INFO: Loaded 1 PC tables (39920 PCs): 39920 [0x5834277d7f20,0x583427873e20), 
INFO:       22 files found in __fuzz__/solana-frozen-abi/stable_abi__impls__tests__fuzzers__TestFuzzerFeed_frozen_abi_fuzzer__test_fuzzer_wincode/corpus
INFO:        0 files found in __fuzz__/solana-frozen-abi/stable_abi__impls__tests__fuzzers__TestFuzzerFeed_frozen_abi_fuzzer__test_fuzzer_wincode/crashes
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
INFO: seed corpus: files: 22 min: 1b max: 171b total: 516b rss: 36Mb
#23     INITED cov: 101 ft: 117 corp: 16/218b exec/s: 0 rss: 37Mb
#1048576        pulse  cov: 101 ft: 117 corp: 16/218b lim: 4096 exec/s: 524288 rss: 99Mb
#1563079        NEW    cov: 103 ft: 119 corp: 17/261b lim: 4096 exec/s: 521026 rss: 100Mb L: 43/45 MS: 1 CopyPart-
#2097152        pulse  cov: 103 ft: 119 corp: 17/261b lim: 4096 exec/s: 419430 rss: 100Mb
#4194304        pulse  cov: 103 ft: 119 corp: 17/261b lim: 4096 exec/s: 419430 rss: 100Mb
#8388608        pulse  cov: 103 ft: 119 corp: 17/261b lim: 4096 exec/s: 399457 rss: 102Mb
#12238750       DONE   cov: 103 ft: 119 corp: 17/261b lim: 4096 exec/s: 394798 rss: 102Mb
Done 12238750 runs in 31 second(s)
fuzzing solana-frozen-abi::stable_abi::impls::tests::fuzzers::TestFuzzerMultiTypeB_frozen_abi_fuzzer::test_fuzzer_wincode for 30s
    Finished `fuzz` profile [optimized] target(s) in 0.05s
  Executable unittests src/lib.rs (target/fuzz/build_1bd7b7ad2d9f6c17/x86_64-unknown-linux-gnu/fuzz/deps/solana_frozen_abi-f594fd0335e52a6c)
  ```
  
  ```
  INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 717350917
INFO: Loaded 1 modules   (40651 inline 8-bit counters): 40651 [0x5f1db67c73b0, 0x5f1db67d127b), 
INFO: Loaded 1 PC tables (40651 PCs): 40651 [0x5f1db67d1280,0x5f1db686ff30), 
INFO:        1 files found in frozen-abi/src/stable_abi/__fuzz__/stable_abi__impls__tests__bolero_repro__TestFuzzerBreakSerializationAboveN_frozen_abi_fuzzer__fuzzer_wincode/corpus
INFO:        0 files found in frozen-abi/src/stable_abi/__fuzz__/stable_abi__impls__tests__bolero_repro__TestFuzzerBreakSerializationAboveN_frozen_abi_fuzzer__fuzzer_wincode/crashes
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
INFO: seed corpus: files: 1 min: 1b max: 1b total: 1b rss: 37Mb
test failed; shrinking input...

======================== Test Failure ========================

Input: 
Length: 1 (0x1) bytes
0000:   11                                                   .

Error: 
panicked at frozen-abi/src/stable_abi/impls.rs:1289:13:
intentional repro panic for values above 16
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

==============================================================

==409386== ERROR: libFuzzer: deadly signal

```


frozen-abi-macro/src/lib.rs
```diff
@@ -970,6 +1122,21 @@ pub fn frozen_abi(attrs: TokenStream, item: TokenStream) -> TokenStream {
     }
 }
 
+#[cfg(feature = "frozen-abi")]
+#[proc_macro]
+pub fn frozen_abi_test(input: TokenStream) -> TokenStream {
```
puhtaytow: probably better name

frozen-abi-macro/src/lib.rs
```diff
@@ -189,6 +190,83 @@ fn parse_roundtrip_test(value: Option<&LitStr>) -> Result<RoundtripTest, Error>
...
+
+#[cfg(feature = "frozen-abi")]
+impl Parse for FuzzerTestArgs {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let target_type = input.parse::<Type>()?;
+        if !input.is_empty() {
+            input.parse::<Token![,]>()?;
+        }
+
+        let mut serializers = vec![AbiSerializer::Bincode, AbiSerializer::Wincode];
```
puhtaytow: should both be out of box?

kskalski: I would just add wincode by default

frozen-abi-macro/src/lib.rs
kskalski: maybe `generate_abi_serialization_test`

frozen-abi-macro/src/lib.rs
```diff
@@ -970,6 +1110,21 @@ pub fn frozen_abi(attrs: TokenStream, item: TokenStream) -> TokenStream {
     }
 }
 
+#[cfg(feature = "frozen-abi")]
+#[proc_macro]
+pub fn frozen_abi_test(input: TokenStream) -> TokenStream {
+    let args = parse_macro_input!(input as FuzzerTestArgs);
+    let generated = match args.strategy {
```
kskalski, line 1284: I think it would be nice to have `Random` as one of the strategies and:
* make old macros just generate appropriate invocation of `generate_abi_serialization_test`
* allow users to manually add such invocations instead of using type derive attribute

frozen-abi-macro/src/lib.rs
puhtaytow: without the abi in name, `generate_serialization_test` sounds better tbh

frozen-abi-macro/src/lib.rs
puhtaytow: good idea yea :)

puhtaytow:
> If this PR represents a change to the sysvar layout, please open a follow-up PR to update the JavaScript client `@solana/sysvars` ([example](https://github.com/anza-xyz/kit/pull/375)) Thank you for keeping the JavaScript clients in sync with the Rust clients.

Ignore it please, i just messed up resolving conflicts at one point and this poped up.

frozen-abi/Cargo.toml
```diff
@@ -18,7 +18,7 @@ rustdoc-args = ["--cfg=docsrs"]
 default = []
 # activate the frozen-abi feature when we actually want to do frozen-abi testing,
 # otherwise leave it off because it requires nightly Rust
-frozen-abi = ["dep:wincode", "solana-frozen-abi-macro/frozen-abi"]
+frozen-abi = ["dep:bolero", "dep:wincode", "solana-frozen-abi-macro/frozen-abi"]
```
kskalski: we will want a separate feature for this

scripts/fuzz-frozen-abi.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+#
```
kskalski, line 2: let's add documentation that this script runs the fuzzing to explore input space, while you can separately do `cargo test --features fuzz-bolero` to replay existing corpus (how would we save one, especially if fuzzing discovers a failing scenario?)

I guess it would be good to test the actual workflow of adding bug (e.g. break serialization if input is >N on some type), discovering it with fuzzing, using the command / repro from fuzz test to persist the case / corpus, replay that with a `cargo test`. Maybe we should include the sample corpus in the repo, not sure.

scripts/fuzz-frozen-abi.sh
```diff
@@ -0,0 +1,59 @@
...
+  work_dir="$fuzz_root/$package/$target_dir"
+
+  bolero_test_args=(
+    "${bolero_args[@]}"
+    --corpus-dir "$work_dir/corpus"
+    --crashes-dir "$work_dir/crashes"
+  )
+
+  if [[ -n "$max_input_len" ]]; then
+    bolero_test_args+=(--max-input-length "$max_input_len")
```
kskalski: let's allow to specify any extra args to cargo bolero test by passing through any args given to the script, I guess it would remove the need to manually read and pass them like here

scripts/fuzz-frozen-abi.sh
```diff
@@ -0,0 +1,59 @@
...
+#
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+export RUST_MIN_STACK="${RUST_MIN_STACK:-16777216}"
+export ASAN_OPTIONS="${ASAN_OPTIONS:-detect_leaks=0}"
+export LSAN_OPTIONS="${LSAN_OPTIONS:-detect_leaks=0}"
+
+duration="${FUZZ_TIME:-30s}"
```
kskalski: would be better to make this a total run time, not per-target time

kskalski (commented):
This looks pretty solid. 

My main concern is how to parametrize the fuzzer macro invocation such that we can flexibly tweak it in future (say if we wanted to switch between bolero raw input fuzzing to making bolero generate sample instances; similarly, `abi_digest` only makes sense for random strategy).

One possible direction is to separate the concept of running the tests (fuzz vs fixed N=10_000 runs) vs how inputs are generated (random / fuzz_raw_input / fuzz_objects), though typically fuzz vs fixed do force specific set of input generation algorithms and are not interchangeable, so it's basically good as is.

scripts/fuzz-frozen-abi.sh
puhtaytow: Good idea yea. I kept total time and rust stack control thru envs with defaults, but everything else now goes as passed to bolero args.

scripts/fuzz-frozen-abi.sh
puhtaytow: Had to introduce another feature to gate the type which comes with broken corpse, not sure if could do that without, but it works nicely as is.

frozen-abi/src/stable_abi/__fuzz__/stable_abi__impls__tests__bolero_repro__TestFuzzerBreakSerializationAboveN_frozen_abi_fuzzer__fuzzer_wincode/corpus/breaks-above-threshold
```diff
@@ -0,0 +1 @@
+
```
puhtaytow: seem like github cant render it ``

puhtaytow:
```bash
xxd -g 1 frozen-abi/src/stable_abi/__fuzz__/stable_abi__impls__tests__bolero_repro__TestFuzzerBreakSerializationAboveN_frozen_abi_fuzzer__fuzzer_wincode/corpus/breaks-above-threshold

00000000: 11     
```

puhtaytow:
Whenever you find time to re-review :pray: @kskalski

frozen-abi/src/stable_abi/__fuzz__/stable_abi__impls__tests__bolero_repro__TestFuzzerBreakSerializationAboveN_frozen_abi_fuzzer__fuzzer_wincode/corpus/breaks-above-threshold
kskalski: I think it should be put into some other dir and we could still gitignore __fuzz__

scripts/fuzz-frozen-abi.sh
```diff
@@ -0,0 +1,145 @@
...
+
+  echo "preparing $target"
+  prepare_command=(cargo bolero test "${bolero_test_args[@]}" --runs 0 "$target")
+  "${prepare_command[@]}"
+
+  echo "fuzzing $target for $duration"
+
+  fuzz_command=(cargo bolero test "${bolero_test_args[@]}" -T "$duration" "$target")
+
+  if timeout "$duration" "${fuzz_command[@]}"; then
```
kskalski: since we are passing -T with the same duration, I suppose we could skip this `timeout` command to avoid weird races (i.e. cargo bolero counting only the test run, but exclusing set-up / tear-down)

scripts/fuzz-frozen-abi.sh
```diff
@@ -0,0 +1,145 @@
...
+#
+# Example:
+#   RUST_MIN_STACK=16777216 TOTAL_FUZZ_TIME=60s scripts/fuzz-frozen-abi.sh \
+#     'stable_abi::impls::tests::bolero_fuzzer::TestFuzzerFeed_frozen_abi_fuzzer::test_fuzzer_wincode' \
+#     -- --package solana-frozen-abi \
+#        --max-input-length 64 \
+#        --corpus-dir __fuzz__/corpus \
+#        --crashes-dir __fuzz__/crashes
+#
+# Reproduce the intentionally skipped serialization bug target:
```
kskalski, line 27: maybe we can split and limit the responsibilities of this script, such that replay of pre-determined corpora can be run with plain `cargo test --features fuzz-bolero` (we could also add it as CI step) and general fuzzing / discovery of new failures could be done with this script?

scripts/fuzz-frozen-abi.sh
kskalski: We might also need per-input timeout: bolero has `--timeout` (default **10s**, distinct from `-T` total time), we should probably set both to properly get rid of shell `timeout` protection

puhtaytow: Fixed.

scripts/fuzz-frozen-abi.sh
puhtaytow: Thru cargo test it wont work unfortunately, as test harness don't support bolero specific functionality, but it is possible thru `cargo bolero test ...`. 

I have updated the script example command to reproduce via naked `cargo bolero test`. If thats ok.

frozen-abi/src/stable_abi/__fuzz__/stable_abi__impls__tests__bolero_repro__TestFuzzerBreakSerializationAboveN_frozen_abi_fuzzer__fuzzer_wincode/corpus/breaks-above-threshold
puhtaytow: To be honest not sure where, in stable_abi/__fuzz__/ seem to be the closest to the test showing failing scenario.

scripts/fuzz-frozen-abi.sh
kskalski: I tried on your branch:
`RUSTC_BOOTSTRAP=1 cargo test -p solana-frozen-abi --features fuzz-bolero-repro 'stable_abi::impls::tests::bolero_repro::TestFuzzerBreakSerializationAboveN_frozen_abi_fuzzer::test_fuzzer_wincode' -- --nocapture 2>&1`
seem to work and produces
```
running 1 test
test stable_abi::impls::tests::bolero_repro::TestFuzzerBreakSerializationAboveN_frozen_abi_fuzzer::test_fuzzer_wincode ...	run time: 117.704µs | iterations/s: 8496 | corpus inputs: 1 | exit reason: test failure

======================== Test Failure ========================

Input: 
Length: 1 (0x1) bytes
0000:   11                                                   .
```

idealy we should make it CI runnable (and the failure above should be an `should_panic` test)

kskalski: I also validated that this really uses the on-disk crash corpus, not re-discovering the faliure from scratch, so we can in fact use the crashes corpora for regression testing that things still crash (though maybe this is not really too useful and making it CI runnable will require extending the macros to generate `should_panic` tests...)

kskalski: to make sure it tests the on-disk crash you need something different than threshold failure, it needs a specific rare stamp:
```rust
const TRIGGER: [u8; 16] = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
struct RareBreak([u8; 16]);
...
 assert!(src.0 != TRIGGER, "rare trigger hit: only the committed corpus reaches this");
```

frozen-abi/src/stable_abi/__fuzz__/stable_abi__impls__tests__bolero_repro__TestFuzzerBreakSerializationAboveN_frozen_abi_fuzzer__fuzzer_wincode/corpus/breaks-above-threshold
kskalski: maybe we could designate `<crate>/fuzz_tests/` (similar to `<crate>/tests/` that some crates have) as root for fixed corpora tests we will run (seems like `test_source_sore/__fuzz__` subdir is hardcoded by bolero to be picked by plain `cargo test`)

and for the scripts/ util, we could think of it as something run maybe by dedicated long running job / worker or maybe your own laptop to discover failures - this could also use auto-dir, which we could git ignore or use a specified directory with --corpus-dir to `bolero test`

PR #786, opened 2026-06-22
Add sysvar size consts

grod220:
Unblocks https://github.com/anza-xyz/agave/pull/12245
Continuation of https://github.com/solana-program/stake/pull/399 & https://github.com/anza-xyz/solana-sdk/pull/766

Adds `SIZE` consts to all sysvars. Considered adding a small `SysvarSize` crate/trait, but found it may be a premature abstraction as it doesn't make https://github.com/anza-xyz/agave/pull/12245 that much simpler and risks re-introducing a dep coupling issue we currently have with `SysvarSerialize`. 

All sysvars have a serializer test to backup the value.

### DIFF
--- a/clock/src/lib.rs
+++ b/clock/src/lib.rs
@@ -171,9 +171,21 @@ pub struct Clock {
     pub unix_timestamp: UnixTimestamp,
 }
 
+/// Serialized size of the `Clock` sysvar account.
+pub const SIZE: usize = size_of::<Clock>();
+const _: () = assert!(SIZE == 40);
+
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn test_size_of() {
+        assert_eq!(
+            wincode::serialized_size(&Clock::default()).unwrap() as usize,
+            SIZE,
+        );
+    }
 
     #[test]
     fn test_clone() {
--- a/epoch-rewards/src/lib.rs
+++ b/epoch-rewards/src/lib.rs
@@ -58,6 +58,16 @@ pub struct EpochRewards {
     pub active: bool,
 }
 
+/// Serialized size of the `EpochRewards` sysvar account.
+pub const SIZE: usize = size_of::<u64>() // distribution_starting_block_height
+    + size_of::<u64>() // num_partitions
+    + size_of::<Hash>() // parent_blockhash
+    + size_of::<u128>() // total_points
+    + size_of::<u64>() // total_rewards
+    + size_of::<u64>() // distributed_rewards
+    + size_of::<bool>(); // active
+const _: () = assert!(SIZE == 81);
+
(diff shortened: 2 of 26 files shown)


clock/src/lib.rs
```diff
@@ -171,10 +171,22 @@ pub struct Clock {
     pub unix_timestamp: UnixTimestamp,
 }
 
+/// Serialized size of the `Clock` sysvar account.
+pub const SIZE: usize = size_of::<Clock>();
+const _: () = assert!(SIZE == 40);
```
grod220, line 176: Case 1: in-memory layout matches serialized layout, so can just do `size_of::<Struct>()`

epoch-rewards/src/lib.rs
```diff
@@ -58,6 +58,16 @@ pub struct EpochRewards {
...
 
+/// Serialized size of the `EpochRewards` sysvar account.
+pub const SIZE: usize = size_of::<u64>() // distribution_starting_block_height
+    + size_of::<u64>() // num_partitions
+    + size_of::<Hash>() // parent_blockhash
+    + size_of::<u128>() // total_points
+    + size_of::<u64>() // total_rewards
+    + size_of::<u64>() // distributed_rewards
+    + size_of::<bool>(); // active
+const _: () = assert!(SIZE == 81);
```
grod220, line 69: Case 2: broke out the fields when `size_of::<Struct>()` would include padding

slot-hashes/src/lib.rs
```diff
@@ -35,6 +35,13 @@ pub fn set_entries_for_tests_only(entries: usize) {
 
 pub type SlotHash = (u64, Hash);
 
+const LEN_PREFIX: usize = size_of::<u64>();
+const SLOT_HASH_SERIALIZED_SIZE: usize = size_of::<u64>() + size_of::<Hash>();
+
+/// Serialized size of the `SlotHashes` sysvar account.
+pub const SIZE: usize = LEN_PREFIX + MAX_ENTRIES * SLOT_HASH_SERIALIZED_SIZE;
```
grod220, line 42: Case 3: formulas for collections

slot-history/src/lib.rs
```diff
@@ -51,6 +51,9 @@ impl std::fmt::Debug for SlotHistory {
 
 pub const MAX_ENTRIES: u64 = 1024 * 1024; // 1 million slots is about 5 days
 
+/// Serialized size of the `SlotHistory` sysvar account.
+pub const SIZE: usize = 131_097;
```
grod220, line 55: Case 4: too complicated to size up `BitVec` with its container internals so just hardcoding size (but backed up via test)

joncinque (approved):
Looks great!

PR #788, opened 2026-06-23
[secp256k1-recover] Make secp256k1-recover no-std

samkim-crypto:
#### Problem

Currently, we can't invoke the secp256-recover syscall using `solana-secp256k1-recover` inside a no-std Solana program because the `solana-secp256k1-recover` crate does not support no-std.

#### Summary of Changes

Added `#![no_std]` support to the crate root, made `std` feature, and included the `std` feature as a default feature.

### DIFF
--- a/secp256k1-recover/Cargo.toml
+++ b/secp256k1-recover/Cargo.toml
@@ -13,8 +13,10 @@ edition = { workspace = true }
 targets = ["x86_64-unknown-linux-gnu"]
 
 [features]
+default = ["std"]
+std = ["borsh?/std", "thiserror/std", "k256/std"]
 borsh = ["dep:borsh"]
-frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro"]
+frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro", "std"]
 
 [dependencies]
 borsh = { workspace = true, optional = true }
--- a/secp256k1-recover/Cargo.toml
+++ b/secp256k1-recover/Cargo.toml
@@ -13,10 +13,8 @@ edition = { workspace = true }
 targets = ["x86_64-unknown-linux-gnu"]
 
 [features]
-default = ["std"]
-std = ["borsh?/std", "thiserror/std", "k256/std"]
 borsh = ["dep:borsh"]
-frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro", "std"]
+frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro"]
 
 [dependencies]
 borsh = { workspace = true, optional = true }
--- a/secp256k1-recover/src/lib.rs
+++ b/secp256k1-recover/src/lib.rs
@@ -29,6 +29,8 @@
 
 #[cfg(feature = "borsh")]
 extern crate alloc;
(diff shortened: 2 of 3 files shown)


secp256k1-recover/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
joncinque: We typically have this always enabled in the sdk
```suggestion
#![no_std]
```

secp256k1-recover/src/lib.rs
```diff
@@ -26,6 +27,10 @@
 //! [sp]: https://docs.rs/solana-program/latest/solana_program/secp256k1_program/
 //! [`ecrecover`]: https://docs.soliditylang.org/en/v0.8.14/units-and-global-variables.html?highlight=ecrecover#mathematical-and-cryptographic-functions
 
+#[cfg(not(feature = "std"))]
+extern crate alloc;
```
joncinque: If we don't actually need std for anything other than frozen-abi, we can keep it more specific
```suggestion
#[cfg(feature = "borsh")]
extern crate alloc;
```

secp256k1-recover/src/lib.rs
```diff
@@ -26,6 +27,10 @@
...
+#[cfg(all(not(feature = "std"), feature = "borsh"))]
```
joncinque: Since only borsh needs this, let's be more specific
```suggestion
#[cfg(feature = "borsh")]
```

secp256k1-recover/src/lib.rs
```diff
@@ -26,6 +27,10 @@
 //! [sp]: https://docs.rs/solana-program/latest/solana_program/secp256k1_program/
 //! [`ecrecover`]: https://docs.soliditylang.org/en/v0.8.14/units-and-global-variables.html?highlight=ecrecover#mathematical-and-cryptographic-functions
```
joncinque, line 29: For frozen-abi, we will need std, so we can do:

```
extern crate std;
```

secp256k1-recover/Cargo.toml
```diff
@@ -13,8 +13,10 @@ edition = { workspace = true }
 targets = ["x86_64-unknown-linux-gnu"]
 
 [features]
+default = ["std"]
+std = ["borsh?/std", "thiserror/std", "k256/std"]
```
joncinque: With the other changes I recommended, we won't need either of these

joncinque (commented):
Can you add the crate to `./scripts/check-no-std.sh` under the no_std_crates? It'll also get tested with alloc and borsh enabled.

I tested out the changes I suggested, so it should make this even neater

samkim-crypto:
Thanks for the feedback! I removed the std feature and just made it no-std by default. I'll make sure to follow this blueprint for the other syscall crates.

secp256k1-recover/Cargo.toml
joncinque: Can you also remove these?

joncinque (commented):
One last bit

secp256k1-recover/Cargo.toml
samkim-crypto: Oh oops, it seems like this was added back in after rebase.

joncinque (approved):
Looks great!

PR #793, opened 2026-06-24
refactor(account): split code into modules and feature-gated sub-mods

kskalski:
#### Problem
`solana-account` crate includes most of its code in `lib.rs` and interleaves feature gated code with common code. This makes it a bit hard to manage the code, especially in face of adding wincode support.

#### Changes
This PR breaks the monolithic `account/src/lib.rs` (~1100 lines) into focused modules, isolating serialization-codec code so bincode (and a future wincode path) can live side-by-side:

```
account/src/
  lib.rs                  # Account/AccountSharedData (kept to keep frozen-abi api_digests stable) 
                          # + crate wiring + re-exports
  account.rs              # all codec-independent code
  account/bincode.rs      # bincode impls + sysvar helpers (create_account_*, from/to_account)
  serde.rs                # Serde Serialize impls
  state_traits.rs         # State/StateMut trait definitions
  state_traits/bincode.rs # bincode StateMut impls
```

- Codec-specific code (`#[cfg(feature = "bincode")]` impl blocks, sysvar free functions, the `serde` impls) moved out of `lib.rs` into their own files.
- Public API is **flat and unchanged**: inherent/trait impls apply crate-wide regardless of file; free functions are re-exported up to the crate root.
- `state_traits` module gate widened from `bincode` to `any(bincode, wincode)` (trait defs are codec-agnostic), leaving a clean slot for wincode impls.

#### Testing
This is refactor without logic changes. The best way to review is look at the "zebra mode" diff and watch for green+red (real changes), other colors mean exact code movements:
```
git show 0d3d5d71 --color-moved=zebra --color-moved-ws=allow-indentation-change
```

#### Further work
The plan is to add bincode-equivalent APIs with wincode schema trait constraints in such a way that wincode sub-module for account and state_traits is enabled for `cfg(not(bincode) && wincode)` which will preserve compatibility (user needs to disable `bincode` feature to observe function signature change) - if there is a different / preferrable way to add wincode support (e.g. add duplicate APIs with `_wincode` suffix), let's settle it now.


account/src/state_traits.rs
```diff
@@ -16,68 +14,3 @@ pub trait State<T> {
     fn state(&self) -> Result<T, InstructionError>;
     fn set_state(&self, state: &T) -> Result<(), InstructionError>;
 }
```
joncinque, line 16: If I understand correctly, the idea is to eventually implement these crates for the wincode traits if the wincode feature is enabled, correct?

To keep a cleaner separation, I would recommend creating totally new traits that can look similar, but explicitly only work with wincode. That way, someone can potentially use both traits if they need to.

Most likely, the only downstream user of these traits is Agave, so we could get by with your suggestion of `#[cfg(all(not(bincode), wincode))]`, but I think a new trait just for wincode will be simpler.

And that way we could eventually deprecate and remove `State` and `StateMut` totally.

Maybe we should get some other opinions though. @brooksprumo what do you think?

account/src/lib.rs
```diff
@@ -2,25 +2,26 @@
...
-#[cfg(feature = "bincode")]
+use {solana_clock::Epoch, solana_pubkey::Pubkey, std::sync::Arc};
+
+mod account;
+pub use account::*;
+
+#[cfg(feature = "serde")]
+mod serde;
+
+#[cfg(any(feature = "bincode", feature = "wincode"))]
```
joncinque: If you want to land this sooner, we can revert this and go forward
```suggestion
#[cfg(feature = "bincode")]
```

joncinque (commented):
Looks great to me overall! Just the question about wincode with the state traits.

account/src/lib.rs
kskalski: sure, this wasn't implemented in this PR anyway

account/src/state_traits.rs
kskalski: Right, because bincode implements them with blanket impl (serde traits specified only in `where` for any `Account`), we cannot also implement them in wincode.

Adding new traits seems less intrusive - though I wonder if we would be still fine to do "exclusive" replacement of the `account.rs` impls when `#[cfg(all(not(bincode), wincode)))]`? I suppose those things could use different approach - the fix is to just rename imported `State*` traits so code using both "serialization APIs" from `Account` and extension traits would be easy to migrate when `bincode` feature is disabled.

joncinque: Yeah that was my thinking.

We certainly *can* do what you're suggesting, and it would probably work out, but it has a higher possibility to cause annoyance during upgrades.

I don't feel too strongly though, so if you want to go with your approach, I won't stop you :smile:

joncinque (approved):
Looks great! I'm find to land this, and we can figure out wincode-compatible traits in another PR

PR #798, opened 2026-06-29
Ensure InstructionHeader repr doesn't change

apfitzge:
### Problem
- tuple layout is technically not stable and future rust versions **could** change the order of fields 
- we take a slice of tuples during wincode deserialization of transaction

### Changes
- Add compile-time asserts on size & offsets for `InstructionHeader`

### DIFF
--- a/message/src/versions/v1/message.rs
+++ b/message/src/versions/v1/message.rs
@@ -23,7 +23,8 @@
 //! │                                                        │
 //! │ * ConfigValues ([[u8; 4] * variable based on mask])    │
 //! │                                                        │
-//! │ * InstructionHeaders [(u8, u8, u16) x NumInstructions] │
+//! │ * WireInstructionHeaders [WireInstructionHeader        │
+//! │                            x NumInstructions]          │
 //! │                                                        │
 //! │ * InstructionPayloads (variable based on headers)      │
 //! │    └─ Per NumInstructions:                             │
@@ -43,7 +44,7 @@ use solana_frozen_abi_macro::AbiExample;
 use std::collections::HashSet;
 #[cfg(feature = "wincode")]
 use {
-    crate::v1::{InstructionHeader, FIXED_HEADER_SIZE},
+    crate::v1::{WireInstructionHeader, FIXED_HEADER_SIZE},
     core::{mem::MaybeUninit, slice::from_raw_parts},
     wincode::{
         config::{Config, ConfigCore},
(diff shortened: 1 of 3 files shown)


message/src/versions/v1/message.rs
kskalski, line 81: to ensure wire format compatibility, we should actually add `StableAbi` derives and digest + roundtrip test here

message/src/versions/v1/message.rs
```diff
@@ -654,13 +647,26 @@ unsafe impl<'de, C: Config> SchemaRead<'de, C> for Message {
...
+        // patterns and alignment 1 here, so the only risk is field reordering or
+        // padding; pin both at compile time so a future layout change fails the build
+        // rather than silently mis-parsing instructions.
+        const _: () = {
+            assert!(size_of::<InstructionHeader>() == 4);
+            assert!(core::mem::offset_of!(InstructionHeader, 0) == 0); // program_id_index
+            assert!(core::mem::offset_of!(InstructionHeader, 1) == 1); // num_accounts
+            assert!(core::mem::offset_of!(InstructionHeader, 2) == 2); // data_len (u16 LE)
+        };
+
```
kskalski: I think a better idea would be to define a private `InstructionHeaderWire` (seems like we can't just change `InstructionHeader`, since it's pubicly exported type, but maybe we can deprecate it) that is a proper struct with fields and `repr(C)` annotation zero-copy enabled assert.

message/src/versions/v1/message.rs
apfitzge: I had thought someone added this on the VersionedMessage type which would include this type - is that not sufficient?

kskalski: it's not there https://github.com/anza-xyz/solana-sdk/blob/6550d4caa13836924fd4b798913d0fc4fd576ee1/message/src/versions/mod.rs#L52-L55

I was adding that to versioned transaction and I think a bit earlier to *some types* in `solana-message`, but not all the way to actual messages... Anyway, looks like a bit larger and unrelated change, I should get around to that later.

message/src/versions/v1/message.rs
apfitzge: LMK if this is what you had in mind: 7cf5b93e

message/src/versions/v1/mod.rs
```diff
@@ -17,8 +17,16 @@ pub use {config::*, error::*, message::*};
 ///  - data_len
 ///
 /// This is used to parse the instruction portion of a V1 message.
+#[deprecated(since = "4.3.0", note = "Use WireInstructionHeader instead")]
 pub type InstructionHeader = (u8, u8, [u8; 2]);
 
+#[repr(C)]
```
kskalski, line 25: let's give it `assert_zero_copy` (https://docs.rs/wincode/latest/wincode/#top-level), since it's implicitly used as such

apfitzge: where are we using it with zero-copy requirements? All I see is the use in creating a slice.

kskalski: that's exactly that - we re-interpret memory containing a sequence of `WireInstructionHeader` as `&[WireInstructionHeader]`, that's what wincode calls zero-copy deserialization.

I wanted to actually suggest doing just that using proper APIs, but there is a caveat, we only deserialize the *body* without the len, so we can't do `let headers: &[WireInstructionHeader] = deserialize(input)`, we need to re-implement manually a part of that call.

message/src/versions/v1/mod.rs
```diff
@@ -17,8 +17,18 @@ pub use {config::*, error::*, message::*};
 ///  - data_len
 ///
 /// This is used to parse the instruction portion of a V1 message.
+#[deprecated(since = "4.3.0", note = "Use WireInstructionHeader instead")]
 pub type InstructionHeader = (u8, u8, [u8; 2]);
 
+#[cfg_attr(feature = "wincode", derive(wincode::SchemaRead, wincode::SchemaWrite))]
+#[cfg_attr(feature = "wincode", wincode(assert_zero_copy))]
+#[repr(C)]
+pub struct WireInstructionHeader {
```
apfitzge, line 26: Is there any reason this needs to be public? Seems like we don't actually need this public...

apfitzge: nvm it's used in transaction tests 🫠

kskalski (commented):
LGTM

tao-stones (approved):
lg for tx-metadata

apfitzge:
Holding off on merge - @joncinque and I are de-tangling some version bumps rn and don't want to make it worse by adding a new breaking change in the mix.

apfitzge:
realizing this is actually not breaking - since we only added a new type and deprecated the existing. removing the label.

PR #799, opened 2026-06-29
feat: non-hashing hasher for Hash

apfitzge:
### Problem
- We use `Hash` as a key in many `HashMap`s and hashing them shows ups in profiles
	- We do not need an additional proof of history in every hashmap lookup

### Solution
- Implement a non-hashing hasher for Hash
	- implementation copied from the non-hashing AddressHasher

### DIFF
--- a/hash/Cargo.toml
+++ b/hash/Cargo.toml
@@ -23,6 +23,7 @@ copy = []
 decode = ["dep:five8"]
 default = []
 frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro"]
+rand = ["dep:rand"]
 sanitize = ["dep:solana-sanitize"]
 serde = ["dep:serde", "dep:serde_derive"]
 std = []
@@ -33,6 +34,7 @@ borsh = { workspace = true, optional = true }
 bytemuck = { workspace = true, optional = true }
 bytemuck_derive = { workspace = true, optional = true }
 five8 = { workspace = true, optional = true }
+rand = { workspace = true, optional = true }
 serde = { workspace = true, optional = true }
 serde_derive = { workspace = true, optional = true }
 solana-atomic-u64 = { workspace = true, optional = true }
--- a/hash/src/hasher.rs
+++ b/hash/src/hasher.rs
@@ -1,0 +1,175 @@
+use {
+    crate::HASH_BYTES,
+    core::{
+        cell::Cell,
+        hash::{BuildHasher, Hasher},
+        mem,
+    },
+    rand::{rng, Rng},
+};
+
+/// A faster, but less collision resistant hasher for hashes.
+///
+/// Specialized hasher that uses a random 8 bytes subslice of the
+/// hash as the hash value. Should not be used when collisions
+/// might be used to mount DOS attacks.
+///
+/// Using this results in about 4x faster lookups in a typical hashmap.
+#[derive(Default)]
+pub struct HashHasher {
+    offset: usize,
+    state: u64,
+}
+
+impl Hasher for HashHasher {
+    #[inline]
+    fn finish(&self) -> u64 {
+        self.state
+    }
+    #[inline]
+    fn write(&mut self, bytes: &[u8]) {
+        debug_assert_eq!(
+            bytes.len(),
+            HASH_BYTES,
+            "This hasher is intended to be used with hashes and nothing else"
+        );
+        // This slice/unwrap can never panic since offset is < HASH_BYTES - mem::size_of::<u64>()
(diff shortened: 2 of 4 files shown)


hash/src/hasher.rs
```diff
@@ -0,0 +1,165 @@
...
+    ///
+    /// The position of the slice is determined initially
+    /// through random draw and then by incrementing a thread-local
+    /// This way each hashmap can be expected to use a slightly different
+    /// slice. This is essentially the same mechanism as what is used by
+    /// `RandomState`
+    fn default() -> Self {
+        std::thread_local!(static OFFSET: Cell<usize>  = {
+            let mut rng = rng();
+            Cell::new(rng.random_range(0..HASH_BYTES - mem::size_of::<u64>()))
```
apfitzge, line 82: @joncinque - I copied this code from the AddressHasher. Is there a reason we use the Cell stuff here? Why not just rng() on every call to default()?

apfitzge:
not sure if this approach is good or maybe we just define a generic [u8; N] random offset hasher and create aliases for Address, Hash, (Signature?).

wdyt @joncinque?

hash/src/lib.rs
```diff
@@ -132,6 +137,16 @@ impl FromStr for Hash {
     }
 }
 
+/// Custom impl of Hash for Hash.
+///
+/// This allows us to skip hashing the length of the hash
+/// which is always the same anyway.
+impl core::hash::Hash for Hash {
```
apfitzge, line 144: this is actually separate and affects all current hashing. I can pull into a separate PR if desired.

`Address` has something similar as this - so no reason to not implement it similarly for `Hash` to avoid hashing the length.

(This was caught because the non-hashing hasher expected a 32-byte slice, and was getting an 8 byte slice for the length!)

hash/src/hasher.rs
```diff
@@ -0,0 +1,175 @@
...
+        super::HashHasherBuilder,
+        crate::{Hash, HASH_BYTES},
+        core::hash::{BuildHasher, Hasher},
+    };
+
+    fn hash_with_bytes(bytes: u8) -> Hash {
+        Hash::new_from_array([bytes; HASH_BYTES])
+    }
+
+    fn hash_with_distinct_windows() -> Hash {
```
apfitzge, line 120: Hash::new_unique works a bit different than Address::new_unique. Just using this to make sure every 8-byte window is a distinct u64.

hash/src/hasher.rs
joncinque: https://doc.rust-lang.org/src/std/hash/random.rs.html#56 should give the full story, I lean towards keeping it as is unless we have a compelling reason to change it

hash/src/lib.rs
```diff
@@ -3,7 +3,7 @@
 #![cfg_attr(feature = "frozen-abi", feature(min_specialization))]
 #[cfg(feature = "borsh")]
 use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
-#[cfg(feature = "frozen-abi")]
+#[cfg(feature = "std")]
```
joncinque: Since the `std` feature on its own doesn't do anything, how about gating this with `#[cfg(any(feature = "rand", feature = "frozen-abi"))]`?

joncinque (dismissed):
Looks great overall! Feel free to land, just the bit on the `std` feature, if you want

hash/src/lib.rs
apfitzge: the std feature already exists. what's the point of it if it does nothing on its' own? Should I just get rid of it?

joncinque: It only exists because we used to need it, and removing it would be a breaking change

apfitzge: done.

joncinque (dismissed):
Looks great! We can extract the common hasher in a future PR

hash/Cargo.toml
```diff
@@ -22,7 +22,8 @@ bytemuck = ["copy", "dep:bytemuck", "dep:bytemuck_derive"]
 copy = []
 decode = ["dep:five8"]
 default = []
-frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro"]
+frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro", "std"]
+rand = ["dep:rand", "std"]
```
joncinque: We don't need these anymore
```suggestion
frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro"]
rand = ["dep:rand"]
```

apfitzge: done

joncinque (approved):
Looks great!

PR #806, opened 2026-07-02
packet: add StableAbi to public types

kskalski:
#### Problem
Another crate without `StableAbi` support

#### Summary of Changes
Derive `StableAbi/StableAbiSample` on `Meta` and `Packet`, add a manual impl for the bitflags `PacketFlags`, and freeze `Packet`'s bincode ABI digest.

`PacketFlags` truncates to defined bits: all are defined today, but this keeps sampling valid if a flag is ever removed.

### DIFF
--- a/packet/Cargo.toml
+++ b/packet/Cargo.toml
@@ -25,7 +25,12 @@ serde = [
     "dep:serde_with",
     "solana-pubkey/serde",
 ]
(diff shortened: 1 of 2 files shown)


joncinque (approved):
Simple and clean, looks great!

PR #809, opened 2026-07-05
loader-v3-interface: surface non-EOF read errors in optional trailing bool

### LINKED ISSUE #808: Optional trailing bool swallows `ReadError`
#### Problem

The SDK supports optional trailing booleans. Detecting such a `bool` is safe if and only the `bool` is present or if the reader has reached the end.

Right now, the code silently swallows all other errors such that more severe failures remain unknown, i.e. `io::ErrorKind::PermissionDenied`.

https://github.com/anza-xyz/solana-sdk/blob/ca5f28aeb8789dde888e0cf650ec94da3b2dc389/loader-v3-interface/src/instruction.rs#L236

#### Fix

```rust
let value = match reader.take_byte() {
    Ok(0) => false,
    Ok(1) => true,
    Ok(byte) => return Err(invalid_bool_encoding(byte)),
    Err(wincode::io::ReadError::ReadSizeLimit(_)) => DEFAULT,
    Err(e) => return Err(e.into())
};
```

plutohan:
The optional trailing bool decoder swallowed every read error, so a genuine failure (e.g. `PermissionDenied`) was silently treated as a missing bool and decoded to `DEFAULT`.

Now only `ReadError::ReadSizeLimit` (end of input) falls back to `DEFAULT`; any other error propagates. Added a regression test with a reader that fails after the prefix.

Fixes #808

### DIFF
--- a/loader-v3-interface/src/instruction.rs
+++ b/loader-v3-interface/src/instruction.rs
@@ -233,7 +233,11 @@ unsafe impl<'de, C: ConfigCore, const DEFAULT: bool> SchemaRead<'de, C>
             Ok(0) => false,
             Ok(1) => true,
             Ok(byte) => return Err(invalid_bool_encoding(byte)),
-            Err(_) => DEFAULT,
+            // A reader that reaches the end without any byte means the trailing
+            // `bool` was simply absent, so fall back to `DEFAULT`. Any other read
+            // error is a genuine failure and must be surfaced.
+            Err(wincode::io::ReadError::ReadSizeLimit(_)) => DEFAULT,
+            Err(err) => return Err(err.into()),
         };
         dst.write(value);
         Ok(())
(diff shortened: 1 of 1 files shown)


joncinque:
@buffalojoec can you take a look at this?

loader-v3-interface/src/instruction.rs
```diff
@@ -748,4 +752,47 @@ mod tests {
         data.push(2);
         assert_invalid_trailing_bool(&data);
     }
+
+    /// A read error other than a clean end-of-input (`ReadError::ReadSizeLimit`,
+    /// e.g. the `PermissionDenied` from #808) must be surfaced, not silently
+    /// treated as a missing trailing `bool`.
```
buffalojoec: nit: this isn't the error that you surface in the test.

buffalojoec (commented):
Good fix, makes sense! Just adjust the test comment, or remove it, and then we'll land it.

loader-v3-interface/src/instruction.rs
plutohan: Adjusted, thanks!

buffalojoec (approved):

PR #810, opened 2026-07-06
packet: zero-initialize buffer in Packet::default

### LINKED ISSUE #722: Potential soundness issue in `solana-packet` 4.1.0
Hello, and thank you for maintaining this crate.

I found a sanitizer/Miri failure reachable through public crate APIs using safe Rust code. I may be missing crate-specific preconditions, but the behavior looks worth checking because safe callers should not be able to trigger undefined behavior.

## Summary

- Crate: `solana-packet`
- Version tested: `4.1.0`
- API paths with similar failures observed:
  - `Packet::default`

## Observed diagnostics

```text
Miri undefined behavior: error: Undefined Behavior: constructing invalid value of type [u8; 1232]: at [0], encountered uninitialized memory, but expected an integer
```

## Reproduction

The snippets below are minimal readable reproducers. Each PoC also includes the source location most relevant to the reported failure.

### PoC 1: `test_solana_packet0::generated_test_0`

Relevant source location:

- `src/lib.rs:231-233`: `Packet::default` constructs `[u8; PACKET_DATA_SIZE]` with `MaybeUninit::uninit().assume_init()`.

Readable equivalent PoC:

```rust
#[test]
fn poc() {
    let _ = solana_packet::Packet::default();
}
```

Observed diagnostic:

```text
error: Undefined Behavior: constructing invalid value of type [u8; 1232]: at [0], encountered uninitialized memory, but expected an integer
```

## Source review and suggested fix

Brief reasoning:
- `Default` creates an uninitialized `[u8; PACKET_DATA_SIZE]`.
- That is not a valid initialized Rust array value, even if later code only intends to use a prefix.

Suggested fix:
- Store the payload as `MaybeUninit<u8>` storage or initialize the buffer explicitly before exposing it through safe APIs.

Thanks again for taking a look.

plutohan:
`Packet::default` built its buffer with `MaybeUninit::uninit().assume_init()`, which is undefined behavior for `[u8; PACKET_DATA_SIZE]` (uninitialized integers). Miri flags it through the existing `test_deserialize_slice`:

```
error: Undefined Behavior: constructing invalid value of type [u8; 1232]: at [0], encountered uninitialized memory, but expected an integer
```

Switched to a zeroed buffer, which drops the unsafe block and the clippy allow. Zeroing 1232 bytes is a single memset, and callers overwrite the buffer anyway.

Also added `solana-packet` to `scripts/test-miri.sh` so the existing tests guard against a regression (verified locally: miri fails on the old code, passes on this branch).

Fixes #722

### DIFF
--- a/packet/src/lib.rs
+++ b/packet/src/lib.rs
@@ -315,12 +315,10 @@ impl fmt::Debug for Packet {
     }
 }
 
-#[allow(clippy::uninit_assumed_init)]
 impl Default for Packet {
     fn default() -> Self {
-        let buffer = std::mem::MaybeUninit::<[u8; PACKET_DATA_SIZE]>::uninit();
         Self {
-            buffer: unsafe { buffer.assume_init() },
+            buffer: [0; PACKET_DATA_SIZE],
             meta: Meta::default(),
         }
     }
--- a/scripts/test-miri.sh
+++ b/scripts/test-miri.sh
@@ -5,4 +5,4 @@ here="$(dirname "$0")"
 src_root="$(readlink -f "${here}/..")"
 cd "${src_root}"
 # miri is very slow; so only run very few of selective tests!
-./cargo nightly miri test -p solana-hash -p solana-account-info -p solana-account-view -p solana-instruction-view
+./cargo nightly miri test -p solana-hash -p solana-account-info -p solana-account-view -p solana-instruction-view -p solana-packet


alexpyattaev:
This zero init will be on the hot path, which is not desirable. There are no production paths hitting this now except for banking trace, which is what https://github.com/anza-xyz/agave/pull/13700 should address.

plutohan:
With #13700 in, nothing hot constructs packets through Default anymore, so the memset only hits cold callers. Default is still a safe public API of this crate and miri flags the UB through the existing test_deserialize_slice, so it seems worth keeping it sound. If a no-init constructor is useful for hot paths, an explicit unsafe fn could cover that.

alexpyattaev (approved):
Agreed, better safe than sorry. Thank you!

alexpyattaev:
This likely needs to be rebased to clear CI.

plutohan:
Rebased onto master.

joncinque (approved):
Thanks!

PR #812, opened 2026-07-07
packet: add wincode serialization support

kskalski:
#### Problem
`packet` lacks wincode support

#### Summary of Changes
Add wincode Schema impls for `Packet, Meta`, and `PacketFlags` so packets can
be (de)serialized with wincode. The encoding matches the existing bincode
format byte-for-byte, verified by a shared frozen-abi digest across both
serializers.

#### Testing
frozen-abi abi_serializers is extended to test wincode roundtrip and abi digest

### DIFF
--- a/packet/Cargo.toml
+++ b/packet/Cargo.toml
@@ -30,7 +30,9 @@ frozen-abi = [
     "dep:solana-frozen-abi-macro",
     "serde",
     "solana-pubkey/frozen-abi",
+    "wincode",
 ]
+wincode = ["dep:wincode", "solana-pubkey/wincode"]
 
 [dependencies]
 bincode = { workspace = true, optional = true }
@@ -42,6 +44,7 @@ serde_with = { workspace = true, optional = true, features = ["macros"] }
 solana-frozen-abi = { workspace = true, optional = true, features = ["frozen-abi"] }
 solana-frozen-abi-macro = { workspace = true, optional = true, features = ["frozen-abi"] }
 solana-pubkey = { workspace = true }
+wincode = { workspace = true, optional = true, features = ["alloc"] }
 
 [dev-dependencies]
 solana-packet = { path = ".", features = ["dev-context-only-utils"] }
--- a/packet/Cargo.toml
+++ b/packet/Cargo.toml
@@ -44,7 +44,7 @@ serde_with = { workspace = true, optional = true, features = ["macros"] }
 solana-frozen-abi = { workspace = true, optional = true, features = ["frozen-abi"] }
 solana-frozen-abi-macro = { workspace = true, optional = true, features = ["frozen-abi"] }
 solana-pubkey = { workspace = true }
-wincode = { workspace = true, optional = true, features = ["alloc"] }
+wincode = { workspace = true, optional = true }
 
 [dev-dependencies]
 solana-packet = { path = ".", features = ["dev-context-only-utils"] }
--- a/packet/src/lib.rs
+++ b/packet/src/lib.rs
@@ -82,6 +82,7 @@ pub struct Meta {
     pub size: usize,
     pub addr: IpAddr,
     pub port: u16,
+    #[cfg_attr(feature = "wincode", wincode(with = "PodPacketFlags"))]
     pub flags: PacketFlags,
     remote_pubkey: Pubkey,
 }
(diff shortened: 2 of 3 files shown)


packet/src/lib.rs
```diff
@@ -102,6 +113,61 @@ impl ::solana_frozen_abi::stable_abi::StableAbi for PacketFlags {
...
+
+#[cfg(feature = "wincode")]
+unsafe impl<const N: usize, C: ConfigCore> SchemaWrite<C> for LenPrefixedByteArray<N> {
+    type Src = [u8; N];
+
+    const TYPE_META: TypeMeta = TypeMeta::join_types([
+        <u64 as SchemaWrite<C>>::TYPE_META,
+        <[u8; N] as SchemaWrite<C>>::TYPE_META,
+    ])
+    .keep_zero_copy(false);
```
joncinque, line 137: Just so I'm sure I understand, do we have to disable zero-copy because the length prefix is thrown away?

packet/src/lib.rs
```diff
@@ -102,6 +113,61 @@ impl ::solana_frozen_abi::stable_abi::StableAbi for PacketFlags {
     }
 }
 
+// Every bit is a defined flag, so any byte is a valid `PacketFlags`; it is a
+// single `u8` in memory and can be serialized as raw bytes.
+#[cfg(feature = "wincode")]
+wincode::pod_wrapper! {
+    unsafe struct PodPacketFlags(PacketFlags);
+}
```
joncinque, line 121: Maybe I'm misunderstanding, but the documentation for `pod_wrapper` states that it isn't needed for `u8`, and `PacketFlags` is a `u8`: https://github.com/anza-xyz/wincode/blob/771cde235f0d5bbd4e53f8bed2838cf4caffcca2/wincode/src/schema/containers.rs#L148

Is this needed because of whatever the `bitflags!` macro does underneath?

packet/Cargo.toml
```diff
@@ -42,6 +44,7 @@ serde_with = { workspace = true, optional = true, features = ["macros"] }
 solana-frozen-abi = { workspace = true, optional = true, features = ["frozen-abi"] }
 solana-frozen-abi-macro = { workspace = true, optional = true, features = ["frozen-abi"] }
 solana-pubkey = { workspace = true }
+wincode = { workspace = true, optional = true, features = ["alloc"] }
```
joncinque: Is the alloc feature needed? I removed it and the feature-powerset and tests worked

joncinque (commented):
It's too bad that the bincode adaptations are needed, but looks good! Just one point and some questions

packet/src/lib.rs
kskalski: It's because of how `bitflags!` handles the type - it actually wraps the u8 in a private InternalBitFlags type that isn’t a wincode schema type

packet/src/lib.rs
kskalski: right, to make schema treat a target type as zero-copy, it needs to have an exactly matching memory layout to one that is found in wire-format. Since wire-format includes the len, the wire bytes are not the same (or even the same size) as the object - array.

packet/Cargo.toml
kskalski: ah, right, works with removed alloc feature

joncinque (approved):
Looks great to me!

packet/src/lib.rs
```diff
@@ -102,6 +113,61 @@ impl ::solana_frozen_abi::stable_abi::StableAbi for PacketFlags {
...
+
+    const TYPE_META: TypeMeta = TypeMeta::join_types([
+        <u64 as SchemaRead<'de, C>>::TYPE_META,
+        <[u8; N] as SchemaRead<'de, C>>::TYPE_META,
+    ])
+    .keep_zero_copy(false);
+
+    fn read(mut reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+        let len = <u64 as SchemaRead<'de, C>>::get(reader.by_ref())?;
+        if len != N as u64 {
```
vadorovsky, line 162: This conversion from `usize` to `u64` is unfortunate. Any chance we could use `usize` everywhere? Might be a dumb question, I might be missing some context about wincode internals.

vadorovsky (approved):
One non-blocking comment.

packet/src/lib.rs
kskalski: I could deserialize as `usize` here (on the wire usize is always encoded with 8 bytes, at least in default config), this will automatically restrict the wire len to the platform-specific max value for usize... which might be a bit of a downgrade, since then behavior differs depending on platform.

I think given that 
* usize -> u64 is a widening conversion and 
* this is a private helper used only with a single constant in this crate, 
* the failure might happen only if we were on 128-bit platform and *user would pass N > u64::MAX* as generic parameter of LenPrefixedByteArray
* since the wire format will use 8 bytes other parts of impl do `N as 64` anyway in some form

it's fine to leave as is.

PR #813, opened 2026-07-07
deps: bump crossbeam-epoch to 0.9.20 (RUSTSEC-2026-0204)

kskalski:
CI fails on changes based on master

```
error: 1 vulnerability found!
warning: 9 allowed warnings found
Crate:     crossbeam-epoch
Version:   0.9.18
Title:     Invalid pointer dereference in `fmt::Pointer` impl for `Atomic` and `Shared` when the underlying pointer is invalid
Date:      2026-07-06
ID:        RUSTSEC-2026-0204
URL:       https://rustsec.org/advisories/RUSTSEC-2026-0204
Solution:  Upgrade to >=0.9.20
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -939,9 +939,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+checksum = "2d6914041f254d6e9176c01941b21115dcfb7089e55135a35411081bd106ef3f"
 dependencies = [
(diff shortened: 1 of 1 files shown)


apfitzge (approved):
LGTM - bump from 0.9.18 to 0.9.20:

```
Version 0.9.20

    Fix invalid pointer dereference in fmt::Pointer impl for Atomic/Shared when the underlying pointer is invalid. This fixes unsoundness that was not fully addressed in 0.9.19's fix. (#1276)

Version 0.9.19

    Fix null pointer dereference in fmt::Pointer impl for Atomic/Shared when it is a null pointer. (#1273)
    Fix return value of Atomic::fetch_update. (#1197)
    Improve compatibility with ThreadSanitizer. (#998)
    Allow unsized types in Guard::defer_destroy. (#1201)
    Use 64-bit integer for epoch representation on 32-bit platforms if AtomicU64 is available. (#1230)
```

PR #814, opened 2026-07-07
message: add borrowed loaded addresses view for AccountKeys

apfitzge:
### Problem
- `AccountKeys` required dynamic loaded addresses to come from an owned `LoadedAddresses`
- Callers with already-borrowed writable/readonly address slices cannot use them and have to allocate temporary `Vec`s if they want to use existing interfaces
	- Use here is in an external scheduler which has ptrs / lens to the resolved addresses - these live inside the custom shmem allocator not global allocator (thus not compatible with Vec as is).

### Summary of Changes
- Added `v0::LoadedAddressesView<'a>` for borrowewd loaded address slices.
- Updated `AccountKeys` to store the borrowed view internally
- Added new constructor for slice-backed loaded addresses
- Existing interfaces are not broken, only requires a minor version bump!

### DIFF
--- a/message/src/account_keys.rs
+++ b/message/src/account_keys.rs
@@ -1,5 +1,9 @@
 use {
-    crate::{compiled_instruction::CompiledInstruction, v0::LoadedAddresses, CompileError},
+    crate::{
+        compiled_instruction::CompiledInstruction,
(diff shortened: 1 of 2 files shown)


tao-stones (approved):
Nice!

PR #817, opened 2026-07-09
nonce-account: zero-copy get_system_account_kind

2501babe:
port `fast_get_system_account_kind()` from https://github.com/anza-xyz/agave/pull/13709, a zero-copy impl of `get_system_account_kind()`. this function is called for every incoming transaction in the agave central scheduler

### DIFF
--- a/nonce-account/src/lib.rs
+++ b/nonce-account/src/lib.rs
@@ -58,17 +58,26 @@ pub enum SystemAccountKind {
 }
 
 pub fn get_system_account_kind(account: &AccountSharedData) -> Option<SystemAccountKind> {
-    if system_program::check_id(account.owner()) {
-        if account.data().is_empty() {
-            Some(SystemAccountKind::System)
-        } else if account.data().len() == State::size() {
-            let nonce_versions: Versions = account.state().ok()?;
-            match nonce_versions.state() {
-                State::Uninitialized => None,
-                State::Initialized(_) => Some(SystemAccountKind::Nonce),
-            }
-        } else {
-            None
+    if !system_program::check_id(account.owner()) {
+        return None;
+    }
+
+    let data = account.data();
+
+    if data.is_empty() {
+        Some(SystemAccountKind::System)
+    } else if data.len() == State::size() {
+        const NONCE_VERSIONS_LEGACY: u32 = 0;
+        const NONCE_VERSIONS_CURRENT: u32 = 1;
+        const NONCE_STATE_INITIALIZED: u32 = 1;
+
+        let versions_tag = u32::from_le_bytes(data.get(..4)?.try_into().ok()?);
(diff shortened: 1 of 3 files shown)


nonce-account/src/lib.rs
```diff
@@ -58,17 +58,26 @@ pub enum SystemAccountKind {
...
+
+    if data.is_empty() {
+        Some(SystemAccountKind::System)
+    } else if data.len() == State::size() {
+        const NONCE_VERSIONS_LEGACY: u32 = 0;
+        const NONCE_VERSIONS_CURRENT: u32 = 1;
+        const NONCE_STATE_INITIALIZED: u32 = 1;
+
+        let versions_tag = u32::from_le_bytes(data.get(..4)?.try_into().ok()?);
+        let state_tag = u32::from_le_bytes(data.get(4..8)?.try_into().ok()?);
```
apfitzge, line 75: because State::size() is const this inlines correctly to just reading u32s directly w/o bounds/conversion checks 💪

nonce-account/src/lib.rs
```diff
@@ -58,17 +58,26 @@ pub enum SystemAccountKind {
...
 
 pub fn get_system_account_kind(account: &AccountSharedData) -> Option<SystemAccountKind> {
-    if system_program::check_id(account.owner()) {
-        if account.data().is_empty() {
-            Some(SystemAccountKind::System)
-        } else if account.data().len() == State::size() {
-            let nonce_versions: Versions = account.state().ok()?;
-            match nonce_versions.state() {
-                State::Uninitialized => None,
-                State::Initialized(_) => Some(SystemAccountKind::Nonce),
```
apfitzge, line 68: is it possible to have a nonce that has invalid state? previously that would have been rejected, but now it'd be accepted, so long as it still has the Initialized tag

2501babe: this is a bit quirked up because the two `state()` functions are completely different

`account.state().ok()` is a bincode parse via `StateMut` and produces `Versions(State(Data))`. `Versions` and `State` are these enum tags (can fail), `Data` is `(Pubkey, DurableNonce, u64)` (arbitrary)

`nonce_versions.state()` is just a call to this:

```rust
pub fn state(&self) -> &State {
    match self {
        Self::Legacy(state) => state,
        Self::Current(state) => state,
    }   
}
```

2501babe: incidentally the test in this pr is the same as the test in the original agave pr, except that the agave test also asserts my impl produces the same result as the sdk impl in master for all test cases

apfitzge: Got it.
We've checked size. We've checked enum tag. So everything left is POD, so cannot fail.

apfitzge (approved):

PR #824, opened 2026-07-15
message: add no-std `is_maybe_writable_with_reserved_addresses`

grod220:
### Problem

`is_maybe_writable()` is gated behind the std feature. The gate exists because is has an `Option<HashSet<Address>>` parameter. This makes it unusable for no-std environments (like usage in the new [programmatic signer program](https://github.com/solana-program/ed25519-programmatic-signer/blob/main/signer/client/src/instruction.rs#L30-L47)).

### Solution

The logic itself is just arithmetic and scans over slices. So it can work with no-std if we swap the HashSet for BTreeSet. However, to prevent a breaking change, we need to introduce a new method: `is_maybe_writable_with_reserved_addresses()` that makes use of shared internals.

### Semver change

Minor bump. The existing `is_maybe_writable()` signature is unchanged. The new method is additional. Should possibly discuss if we should add deprecation attributes to the previous method though.

### DIFF
--- a/message/src/legacy.rs
+++ b/message/src/legacy.rs
@@ -22,7 +22,7 @@ use {
         compiled_instruction::CompiledInstruction, compiled_keys::CompiledKeys,
         inline_nonce::advance_nonce_account_instruction, MessageHeader,
     },
-    alloc::vec::Vec,
+    alloc::{collections::BTreeSet, vec::Vec},
     core::convert::TryFrom,
     solana_address::Address,
     solana_hash::Hash,
@@ -579,6 +579,22 @@ impl Message {
         super::is_writable_index(i, self.header, &self.account_keys)
     }
 
+    /// Returns true if the account at the specified index was requested as writable.
+    /// This has the same semantics as `is_maybe_writable` but is no-std.
+    pub fn is_maybe_writable_v2(
+        &self,
+        i: usize,
+        reserved_account_keys: Option<&BTreeSet<Address>>,
+    ) -> bool {
+        super::is_maybe_writable_v2(
+            i,
+            self.header,
+            &self.account_keys,
+            &self.instructions,
+            reserved_account_keys,
+        )
+    }
+
     /// Returns true if the account at the specified index is writable by the
     /// instructions in this message. The `reserved_account_keys` param has been
     /// optional to allow clients to approximate writability without requiring
@@ -636,8 +652,13 @@ impl Message {
 #[cfg(test)]
 mod tests {
     use {
-        super::*, crate::MESSAGE_HEADER_LENGTH, alloc::vec, core::str::FromStr,
-        solana_instruction::AccountMeta, std::collections::HashSet,
+        super::*,
+        crate::MESSAGE_HEADER_LENGTH,
+        alloc::{collections::BTreeSet, vec},
+        core::str::FromStr,
+        solana_instruction::AccountMeta,
+        std::collections::HashSet,
+        test_case::test_case,
     };
 
     #[test]
@@ -722,8 +743,15 @@ mod tests {
         assert_eq!(message.program_position(2), Some(1));
     }
 
-    #[test]
-    fn test_is_maybe_writable() {
+    #[test_case(0, true, true; "writable signer")]
+    #[test_case(1, false, true; "first readonly signer")]
+    #[test_case(2, false, true; "second readonly signer")]
+    #[test_case(3, false, true; "reserved writable unsigned account")]
+    #[test_case(3, true, false; "writable unsigned account without reserved keys")]
+    #[test_case(4, true, true; "writable unsigned account")]
+    #[test_case(5, false, true; "readonly unsigned account")]
+    #[test_case(6, false, true; "out of bounds")]
+    fn test_is_maybe_writable(key_index: usize, expected: bool, with_reserved_account_keys: bool) {
         let key0 = Address::new_unique();
         let key1 = Address::new_unique();
         let key2 = Address::new_unique();
@@ -743,15 +771,20 @@ mod tests {
         };
 
         let reserved_account_keys = HashSet::from([key3]);
+        let reserved_account_keys_v2 = BTreeSet::from([key3]);
+        let maybe_reserved_account_keys =
+            with_reserved_account_keys.then_some(&reserved_account_keys);
+        let maybe_reserved_account_keys_v2 =
+            with_reserved_account_keys.then_some(&reserved_account_keys_v2);
 
-        assert!(message.is_maybe_writable(0, Some(&reserved_account_keys)));
-        assert!(!message.is_maybe_writable(1, Some(&reserved_account_keys)));
-        assert!(!message.is_maybe_writable(2, Some(&reserved_account_keys)));
-        assert!(!message.is_maybe_writable(3, Some(&reserved_account_keys)));
-        assert!(message.is_maybe_writable(3, None));
-        assert!(message.is_maybe_writable(4, Some(&reserved_account_keys)));
-        assert!(!message.is_maybe_writable(5, Some(&reserved_account_keys)));
-        assert!(!message.is_maybe_writable(6, Some(&reserved_account_keys)));
+        assert_eq!(
+            message.is_maybe_writable(key_index, maybe_reserved_account_keys),
+            expected,
+        );
+        assert_eq!(
+            message.is_maybe_writable_v2(key_index, maybe_reserved_account_keys_v2),
+            expected,
+        );
     }
 
     #[test]
--- a/message/src/legacy.rs
+++ b/message/src/legacy.rs
@@ -604,10 +604,10 @@ impl Message {
     /// fetching the latest set of protocol-reserved addresses. If this method
     /// is called by the runtime, the latest set of reserved addresses must be
     /// passed.
-    pub fn is_maybe_writable_with_reserved_addresses<T: AddressSet>(
+    pub fn is_maybe_writable_with_reserved_addresses(
         &self,
         i: usize,
-        reserved_addresses: Option<&T>,
+        reserved_addresses: Option<&impl AddressSet>,
     ) -> bool {
         super::is_maybe_writable(
             i,
--- a/message/src/legacy.rs
+++ b/message/src/legacy.rs
@@ -588,33 +588,33 @@ impl Message {
     #[cfg(feature = "std")]
     #[deprecated(
         since = "4.4.0",
-        note = "Use `is_maybe_writable_with_address_set` instead"
+        note = "Use `is_maybe_writable_with_reserved_addresses` instead"
     )]
     pub fn is_maybe_writable(
         &self,
         i: usize,
         reserved_account_keys: Option<&HashSet<Address>>,
     ) -> bool {
-        self.is_maybe_writable_with_address_set(i, reserved_account_keys)
+        self.is_maybe_writable_with_reserved_addresses(i, reserved_account_keys)
     }
 
     /// Returns true if the account at the specified index is writable by the
-    /// instructions in this message. The `reserved_account_keys` param is
+    /// instructions in this message. The `reserved_addresses` param is
     /// optional to allow clients to approximate writability without requiring
-    /// fetching the latest set of reserved account keys. If this method is
-    /// called by the runtime, the latest set of reserved account keys must be
+    /// fetching the latest set of protocol-reserved addresses. If this method
+    /// is called by the runtime, the latest set of reserved addresses must be
     /// passed.
-    pub fn is_maybe_writable_with_address_set<T: AddressSet>(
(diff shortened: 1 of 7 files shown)


message/src/versions/v0/mod.rs
```diff
@@ -399,38 +398,50 @@ impl Message {
...
     }
 
-    /// Returns true if the account at the specified index is in the reserved
-    /// account keys set. Before loading addresses, we can't detect reserved
-    /// account keys properly so this shouldn't be used by the runtime.
-    #[cfg(feature = "std")]
-    fn is_account_maybe_reserved(
+    /// Returns true if the account at the specified index was requested as writable.
+    /// This has the same semantics as `is_maybe_writable` but is no-std.
+    pub fn is_maybe_writable_v2(
```
febo: Could this be called `is_writable` instead? Based on the comments, that seems to be what the function is checking.

message/src/versions/v0/mod.rs
```diff
@@ -399,38 +398,50 @@ impl Message {
...
         &self,
         key_index: usize,
-        reserved_account_keys: Option<&HashSet<Address>>,
+        reserved_account_keys: Option<&BTreeSet<Address>>,
```
febo: We could introduce a trait to represent a `Set`, then the function could be generic and `is_maybe_writable` marked as deprecated.

febo (commented):
Couple of initial comments about the implementation.

message/src/versions/v0/mod.rs
apfitzge: this would also let us use faster hashsets instead of the default hasher

apfitzge:
just a note - this whole maybe writable stuff SHOULD go away. it's a ridiculous complication that should have never been introduced. but it's not been a high priority yet.

grod220:
> just a note - this whole maybe writable stuff SHOULD go away.

Are you suggesting in this PR, we should just have `fn is_writable(index)` without the `reserved_account_keys: Option<Set<Address>>` parameter? or this is just a note about the future protocol change?

apfitzge:
> > just a note - this whole maybe writable stuff SHOULD go away.
> 
> Are you suggesting in this PR, we should just have `fn is_writable(index)` without the `reserved_account_keys: Option<Set<Address>>` parameter? or this is just a note about the future protocol change?

just noting of a future protocol change

message/src/versions/v0/mod.rs
grod220: I think we'll need to wait until the protocol changes before changing the `is_maybe` prefix on this. It looks like we previously had the shorter naming, but was intentionally deprecated given we couldn't offer an exact answer anymore given the runtime demotions.

relevant: https://github.com/anza-xyz/agave/pull/541

message/src/versions/v0/mod.rs
grod220: Also makes code re-use between the different versions easier. Added 👍

message/src/lib.rs
```diff
@@ -176,14 +176,32 @@ fn is_writable_index(i: usize, header: MessageHeader, account_keys: &[Address])
...
 /// Returns true if the account at the specified index is in the optional
 /// reserved account keys set.
-#[cfg(feature = "std")]
 #[inline(always)]
-fn is_account_maybe_reserved(
+fn is_account_maybe_reserved<T: AddressSet>(
     i: usize,
     account_keys: &[Address],
-    reserved_account_keys: Option<&HashSet<Address>>,
+    reserved_account_keys: Option<&T>,
```
apfitzge: just a thought, and feel free to reject the idea!


----

for None the behavior is consistent with a blanket impl of the trait for Option<T> where T:AddressSet.

gets rid of the nest in this fn, simplifies the internal signatures a bit.

message/src/legacy.rs
```diff
@@ -586,10 +586,25 @@ impl Message {
...
+        self.is_maybe_writable_v2(i, reserved_account_keys)
+    }
+
+    /// Returns true if the account at the specified index is writable by the
+    /// instructions in this message. The `reserved_account_keys` param is
+    /// optional to allow clients to approximate writability without requiring
+    /// fetching the latest set of reserved account keys. If this method is
+    /// called by the runtime, the latest set of reserved account keys must be
+    /// passed.
+    pub fn is_maybe_writable_v2<T: AddressSet>(
```
apfitzge: `is_maybe_writable_with_address_set` seems better than v2 to me

apfitzge (commented):
changes look good, left a couple minor comments for consideration.

message/src/lib.rs
grod220: I see. Yep, can update that 👍

apfitzge (dismissed):

grod220:
@febo all good from your end?

message/src/versions/v0/mod.rs
```diff
@@ -395,36 +394,39 @@ impl Message {
...
-    /// account keys set. Before loading addresses, we can't detect reserved
-    /// account keys properly so this shouldn't be used by the runtime.
-    #[cfg(feature = "std")]
-    fn is_account_maybe_reserved(
+    /// Returns true if the account at the specified index was requested as
+    /// writable. Before loading addresses, we can't demote write locks properly
+    /// so this should not be used by the runtime. The `reserved_account_keys`
+    /// param is optional to allow clients to approximate writability without
+    /// requiring fetching the latest set of reserved account keys.
+    pub fn is_maybe_writable_with_address_set<T: AddressSet>(
```
febo: nit: I am still not sure about using `is_maybe_writable*`. For me it is not clear what "maybe writable" means in this context, since the function returns a `bool`, so either the account is writable or not.

message/src/versions/v0/mod.rs
```diff
@@ -395,36 +394,39 @@ impl Message {
...
         &self,
         key_index: usize,
-        reserved_account_keys: Option<&HashSet<Address>>,
+        reserved_account_keys: Option<&T>,
```
febo: nit: Should this be named `address_set` so it matches the name of the function?

message/src/versions/v1/message.rs
```diff
@@ -380,10 +380,36 @@ impl Message {
...
         reserved_account_keys: Option<&HashSet<Address>>,
+    ) -> bool {
+        self.is_maybe_writable_with_address_set(key_index, reserved_account_keys)
+    }
+
+    /// Returns `true` if the account at the specified index was requested as
+    /// writable.
+    ///
+    ///
+    /// # Important
```
febo, line 399: nit: Other "versions" of this function do not have an "# Important" section. Perhaps we can follow the same pattern in all of them?

message/src/versions/v1/message.rs
```diff
@@ -380,10 +380,36 @@ impl Message {
...
+    /// to allow clients to approximate writability without requiring fetching the latest
+    /// set of reserved account keys.
+    ///
+    /// Program accounts are demoted from writable to readonly, unless the upgradeable
+    /// loader is present in which case they are left as writable since upgradeable
+    /// programs need to be writable for upgrades.
+    pub fn is_maybe_writable_with_address_set<T: AddressSet>(
+        &self,
+        key_index: usize,
+        reserved_account_keys: Option<&T>,
```
febo: nit: Technically we don't need a generic type here, we could have:
```rust
pub fn is_maybe_writable_with_address_set(
        &self,
        key_index: usize,
        reserved_account_keys: Option<&impl AddressSet>,
```
Might keep the syntax a bit simpler, but this is totally cosmetic. 😊

febo (dismissed):
Left a few nits, totally optional.

message/src/versions/v0/mod.rs
grod220: The issue is that this function cannot definitively make that conclusion given the runtime's demotion behavior. So going with `is_writable()` would overpromise and is the reason it was changed away from this in the past. The doc comments (copied from the previous version) allude to this saying "Returns true if... *requested* as writable".

message/src/versions/v0/mod.rs
grod220: I think doing so we'd lose some documentation about expectations for that param as address_set is more generic.

But this leads me to think the `is_maybe_writable_with_address_set` is not really specific about this either. And also, we could swap the usage of "keys" for "addresses". 

Let me know what you think about this renaming.

message/src/versions/v1/message.rs
grod220: This code comment was copied from is_maybe_writable above

message/src/versions/v1/message.rs
grod220: Updated ✅

message/src/lib.rs
```diff
@@ -176,22 +176,49 @@ fn is_writable_index(i: usize, header: MessageHeader, account_keys: &[Address])
...
+
+/// Returns true if the account at the specified index is in the reserved
+/// account keys set.
 #[inline(always)]
-fn is_account_maybe_reserved(
+fn is_account_maybe_reserved<T: AddressSet>(
     i: usize,
     account_keys: &[Address],
-    reserved_account_keys: Option<&HashSet<Address>>,
+    reserved_account_keys: T,
```
apfitzge: oh i see this is done here,but nowhere else. Would rather we are consistent; either use Option<T> or just T everywhere?

apfitzge:
let's update the PR title since the function name changed

message/src/versions/v1/message.rs
febo: Yeah, I was commenting on the fact that this is the only one that adds that comment under an "Important" section. Wonder if we should the same to all of them?

message/src/lib.rs
```diff
@@ -228,14 +255,13 @@ fn is_upgradeable_loader_present(account_keys: &[Address]) -> bool {
...
-#[cfg(feature = "std")]
 #[inline(always)]
-fn is_maybe_writable(
+fn is_maybe_writable<T: AddressSet>(
     i: usize,
     header: MessageHeader,
     account_keys: &[Address],
     instructions: &[CompiledInstruction],
-    reserved_account_keys: Option<&HashSet<Address>>,
+    reserved_account_keys: T,
```
febo: Should we have this as `&impl AddressSet` as well?

message/src/lib.rs
febo: Also, we can replace the use of `T` by `&impl AddressSet`.

message/src/versions/v1/message.rs
grod220: Oh I misunderstood. You are suggesting adding the `# Important` header to the other message versions? Just added.

message/src/lib.rs
grod220: Done

message/src/lib.rs
```diff
@@ -228,14 +255,13 @@ fn is_upgradeable_loader_present(account_keys: &[Address]) -> bool {
...
 /// passed.
-#[cfg(feature = "std")]
 #[inline(always)]
 fn is_maybe_writable(
     i: usize,
     header: MessageHeader,
     account_keys: &[Address],
     instructions: &[CompiledInstruction],
-    reserved_account_keys: Option<&HashSet<Address>>,
+    reserved_account_keys: &impl AddressSet,
```
febo: I think @apfitzge mentioned this before, but it is not clear why `reserved_account_keys` is not an `Option` anymore. If this is on purpose, then I think the comments need updating since still mentioned the parameter being optional. Also, should it be names `reserved_addresses`?

message/src/lib.rs
```diff
@@ -176,22 +176,49 @@ fn is_writable_index(i: usize, header: MessageHeader, account_keys: &[Address])
...
+}
+
+/// Returns true if the account at the specified index is in the reserved
+/// account keys set.
 #[inline(always)]
 fn is_account_maybe_reserved(
     i: usize,
     account_keys: &[Address],
-    reserved_account_keys: Option<&HashSet<Address>>,
+    reserved_account_keys: &impl AddressSet,
```
febo: Same here, not clear why `reserved_account_keys` is not optional anymore + naming.

febo (commented):
Looks good to me. Couple of final nits.

message/src/lib.rs
grod220: Will update. Suppose we should probably keep the `*_keys` naming on the old public deprecated functions though.

message/src/lib.rs
grod220: The blanket option impl ([related](https://github.com/anza-xyz/solana-sdk/pull/824#discussion_r3599893120)) makes it... optional lol. But you're right, it's more declarative to include it in param signature and just rely on that blanket impl to make the .contains() usage simpler for the other helper.

apfitzge (approved):

febo (approved):
Looks great!

PR #825, opened 2026-07-15
rent: Export `GetSysvar` trait

febo:
### Problem

Currently to use `Rent::get`, it is necessary to enable the `"sysvar"` feature on solana-rent and also add the solana-get-sysvar crate as a dependency. This can be simplified if solana-rent re-exports the `GetSysvar` trait.

### Solution

Add an export to `GetSysvar` from solana-rent.

### DIFF
--- a/rent/src/sysvar.rs
+++ b/rent/src/sysvar.rs
@@ -1,8 +1,7 @@
-pub use solana_sdk_ids::sysvar::rent::{check_id, id, ID};
-use {
-    crate::Rent,
-    solana_get_sysvar::{impl_get_sysvar, GetSysvar},
-    solana_sysvar_id::impl_sysvar_id,
+use {crate::Rent, solana_get_sysvar::impl_get_sysvar, solana_sysvar_id::impl_sysvar_id};
+pub use {
+    solana_get_sysvar::GetSysvar,
+    solana_sdk_ids::sysvar::rent::{check_id, id, ID},
 };
 
 impl_sysvar_id!(Rent);


grod220 (approved):

grod220:
Also realizing this pattern is not on any sysvar crates actually. At the moment, folks need to import `solana-get-sysvar = "1.0.0"` directly. 

Shall we just add all the others too here while we are at it?

febo:
> Also realizing this pattern is not on any sysvar crates actually. At the moment, folks need to import `solana-get-sysvar = "1.0.0"` directly.
> 
> Shall we just add all the others too here while we are at it?

I think we should – let's see of @joncinque agrees.

joncinque (approved):
So I'm sure I understand, users still need to import `GetSysvar` somehow, correct? Meaning, with this change, they do:

```
use solana_rent::sysvar::GetSysvar;
```

Someone might end up importing the same trait from many different crates if they use multiple sysvars, but that shouldn't be too problematic. The ease of use of just needing one crate is worth it in my opinion.

febo:
> So I'm sure I understand, users still need to import `GetSysvar` somehow, correct? Meaning, with this change, they do:
> 
> ```
> use solana_rent::sysvar::GetSysvar;
> ```
> 
> Someone might end up importing the same trait from many different crates if they use multiple sysvars, but that shouldn't be too problematic. The ease of use of just needing one crate is worth it in my opinion.

Yeah, they still need to import it but it is easier to "find" the trait with the re-export.

PR #827, opened 2026-07-16
get-sysvar: Optimize trait implementation

febo:
### Problem

Currently, `get_sysvar` is not as performant as the one in pinocchio. It consumes about `48` CUs more, but it could have the same performance.

### Solution

This PR refactor the trait implementation by:

- Removing the double-error conversion
- Removing the unnecessary initialization of padding bytes (see [rustdocs](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initialization-invariant))
- Use of `ID` constant instead of `id()` function

### DIFF
--- a/get-sysvar/src/lib.rs
+++ b/get-sysvar/src/lib.rs
@@ -8,23 +8,50 @@ use {solana_address::Address, solana_program_error::ProgramError};
 // Stable `$crate` paths for `impl_get_sysvar!`, which expands downstream.
 #[doc(hidden)]
 pub mod __private {
-    pub use {crate::get_sysvar_unchecked, solana_program_error::ProgramError};
+    pub use solana_program_error::ProgramError;
+
+    /// Syscall success code.
+    //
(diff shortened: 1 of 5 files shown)


get-sysvar/src/lib.rs
```diff
@@ -73,62 +99,59 @@ pub fn get_sysvar(
...
+    match __private::sol_get_sysvar(sysvar_id, var_addr, offset, length) {
+        __private::SUCCESS => Ok(()),
+        __private::OFFSET_LENGTH_EXCEEDS_SYSVAR => Err(ProgramError::InvalidArgument),
+        __private::SYSVAR_NOT_FOUND => Err(ProgramError::UnsupportedSysvar),
         // Unexpected errors are folded into `UnsupportedSysvar`.
         _ => Err(ProgramError::UnsupportedSysvar),
     }
 }
 
-fn sol_get_sysvar(sysvar_id: *const u8, var_addr: *mut u8, offset: u64, length: u64) -> u64 {
```
febo, line 91: This was moved to the `__private` module so the macro can reused it.

joncinque (approved):
Looks great!

grod220 (approved):
Nice catching that double nested match 💪

PR #829, opened 2026-07-17
[bls12-381] Add library for bls12-381 syscalls

samkim-crypto:
#### Problem

[SIMD-388](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0388-bls12-381-syscalls.md) is already activated on mb, but we don't yet have a library in the sdk that developers can use to easily invoke the syscall.

#### Summary of Changes

I added the `solana-bls12-381` library. I organized the library via structs (g1, g2, scalar, etc), which is different to how the `solana-bn254` library is organized where it is organized by syscall operations. I think organizing things by structs is more intuitive and inline with existing cryptographic libraries like dalek, blstrs, ark, etc.

There is a bit more work to be done on the testing side. This PR is already big, so I just added the sanity check tests. Let me add more comprehensive tests incrementally on follow-ups.

### DIFF
--- a/bls12-381/Cargo.toml
+++ b/bls12-381/Cargo.toml
@@ -1,0 +1,33 @@
+[package]
+name = "solana-bls12-381"
+description = "Solana BLS12-381"
+documentation = "https://docs.rs/solana-bls12-381"
+version = "0.1.0"
+authors = { workspace = true }
+repository = { workspace = true }
+homepage = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+include = ["src/**/*"]
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+all-features = true
+rustdoc-args = ["--cfg=docsrs"]
+
+[dependencies]
+bytemuck = { workspace = true, features = ["derive", "min_const_generics"] }
+solana-define-syscall = { workspace = true }
+thiserror = { workspace = true }
+
+[target.'cfg(not(target_os = "solana"))'.dependencies]
+solana-bls12-381-syscall = { workspace = true }
+
+[target.'cfg(target_os = "solana")'.dependencies]
+solana-define-syscall = { workspace = true }
+
+[dev-dependencies]
+array-bytes = { workspace = true }
+
+[lints]
+workspace = true
--- a/bls12-381/Cargo.toml
+++ b/bls12-381/Cargo.toml
@@ -17,8 +17,6 @@ rustdoc-args = ["--cfg=docsrs"]
 
 [dependencies]
 bytemuck = { workspace = true, features = ["derive", "min_const_generics"] }
-solana-define-syscall = { workspace = true }
-thiserror = { workspace = true }
 
 [target.'cfg(not(target_os = "solana"))'.dependencies]
 solana-bls12-381-syscall = { workspace = true }
--- a/bls12-381/Cargo.toml
+++ b/bls12-381/Cargo.toml
@@ -25,6 +25,7 @@ bytemuck_derive = { workspace = true, optional = true }
 
 [target.'cfg(not(target_os = "solana"))'.dependencies]
 bytemuck = { workspace = true, features = ["min_const_generics"] }
+bytemuck_derive = { workspace = true }
 solana-bls12-381-syscall = { workspace = true }
 
 [target.'cfg(target_os = "solana")'.dependencies]
--- a/bls12-381/Cargo.toml
+++ b/bls12-381/Cargo.toml
@@ -23,13 +23,13 @@ bytemuck = ["dep:bytemuck", "dep:bytemuck_derive"]
 bytemuck = { workspace = true, features = ["min_const_generics"], optional = true }
 bytemuck_derive = { workspace = true, optional = true }
 
-[target.'cfg(not(target_os = "solana"))'.dependencies]
+[target.'cfg(any(target_os = "solana", target_arch = "bpf"))'.dependencies]
+solana-define-syscall = { workspace = true }
+
+[target.'cfg(not(any(target_os = "solana", target_arch = "bpf")))'.dependencies]
 bytemuck = { workspace = true, features = ["min_const_generics"] }
 bytemuck_derive = { workspace = true }
 solana-bls12-381-syscall = { workspace = true }
-
-[target.'cfg(target_os = "solana")'.dependencies]
-solana-define-syscall = { workspace = true }
 
 [dev-dependencies]
 array-bytes = { workspace = true }
--- a/bls12-381/Cargo.toml
+++ b/bls12-381/Cargo.toml
@@ -15,10 +15,16 @@ targets = ["x86_64-unknown-linux-gnu"]
 all-features = true
 rustdoc-args = ["--cfg=docsrs"]
 
+[features]
+default = ["bytemuck"]
+bytemuck = ["dep:bytemuck", "dep:bytemuck_derive"]
+
 [dependencies]
-bytemuck = { workspace = true, features = ["derive", "min_const_generics"] }
+bytemuck = { workspace = true, features = ["min_const_generics"], optional = true }
+bytemuck_derive = { workspace = true, optional = true }
 
 [target.'cfg(not(target_os = "solana"))'.dependencies]
+bytemuck = { workspace = true, features = ["min_const_generics"] }
 solana-bls12-381-syscall = { workspace = true }
 
 [target.'cfg(target_os = "solana")'.dependencies]
@@ -26,6 +32,7 @@ solana-define-syscall = { workspace = true }
 
 [dev-dependencies]
 array-bytes = { workspace = true }
+bytemuck = { workspace = true, features = ["min_const_generics"] }
 
 [lints]
 workspace = true
--- a/bls12-381/README.md
+++ b/bls12-381/README.md
@@ -14,10 +14,10 @@ verification and zero-knowledge proof (e.g., Groth16) validation.
   `bytemuck::Pod`. Developers can cast transaction instruction data directly
   into curve points without heap allocations.
 - **CU-Optimized Mutations:** Provides `_assign` variants for all group
-  operations (e.g., `add_assign`), allowing in-place memory mutations
-  to strictly control Compute Unit (CU) consumption.
-- **Safe & Unchecked APIs:** Exposes both fully validated group
-  operations and `_unchecked` variants that skip subgroup checks for cheaper
+  operations (e.g., `add_assign`), which write into a caller-supplied
+  `MaybeUninit` buffer to strictly control Compute Unit (CU) consumption.
+- **Validated & Unchecked APIs:** Group operations validate their operands by
+  default, with `_unchecked` variants that skip subgroup checks for cheaper
   point accumulation.
 - **Ergonomic Pairings:** Includes `pairing_check` for evaluating if the
   product of multiple pairings equals the identity element, avoiding costly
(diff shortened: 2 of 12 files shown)


samkim-crypto:
Tagging @zz-sol for review on the cryptography part and @joncinque for the repo integration part.

bls12-381/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+[dependencies]
+bytemuck = { workspace = true, features = ["derive", "min_const_generics"] }
+solana-define-syscall = { workspace = true }
+thiserror = { workspace = true }
+
+[target.'cfg(not(target_os = "solana"))'.dependencies]
+solana-bls12-381-syscall = { workspace = true }
+
+[target.'cfg(target_os = "solana")'.dependencies]
+solana-define-syscall = { workspace = true }
```
zz-sol: seems like duplicated dep as this is already on line 20?

bls12-381/src/g1.rs
```diff
@@ -0,0 +1,392 @@
...
+            }
+        }
+    }
+
+    /// Point subtraction returning a new allocated point.
+    /// Skips subgroup checks.
+    pub fn sub_unchecked(&self, other: &Self, endianness: Endianness) -> Option<Self> {
+        let mut result = MaybeUninit::<Self>::uninit();
+        let success =
+            self.sub_unchecked_assign(other, unsafe { result.assume_init_mut() }, endianness);
```
zz-sol: The [official Rust docs](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.assume_init_mut) for `MaybeUninit::assume_init_mut()` say explicitly:

> Calling this when the content is not yet fully initialized causes undefined behavior

and

> ` .assume_init_mut()` cannot be used to initialize a `MaybeUninit`.

credit: gpt-sol

how about


``` rust
let mut result = Self::zeroed();

let success =
    self.sub_unchecked_assign(other, &mut result, endianness);
```

zz-sol (approved):

zz-sol (dismissed):

bls12-381/README.md
```diff
@@ -0,0 +1,75 @@
...
+[SIMD-0388](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0388-bls12-381-syscalls.md).
+
+This crate provides the standard interface for Solana program
+developers to perform pairing-based cryptography, such as BLS signature
+verification and zero-knowledge proof (e.g., Groth16) validation.
+
+## Features
+
+- **Zero-Copy Deserialization:** Types are `#[repr(transparent)]` and implement
+  `bytemuck::Pod`. Developers can cast transaction instruction data directly
```
joncinque, line 14: We don't need to change anything in this PR, but we might want to start using wincode's zero-copy mode instead, since I believe it gets slightly better performance.

bls12-381/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+edition = { workspace = true }
+include = ["src/**/*"]
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+all-features = true
+rustdoc-args = ["--cfg=docsrs"]
+
+[dependencies]
+bytemuck = { workspace = true, features = ["derive", "min_const_generics"] }
```
joncinque: In case we decide to use wincode in the future, it would be nicer to have the bytemuck trait derivations happen with a `bytemuck` feature, similar to other crates in the sdk.

Edit: since the offchain implementation requires the bytemuck feature enabled, we might need to add the crate itself as a dependency with the bytemuck feature enabled if `target_os != "solana"`

bls12-381/src/g1.rs
```diff
@@ -0,0 +1,392 @@
...
+    /// Safely negates the point by subtracting it from infinity.
+    pub fn neg(&self, endianness: Endianness) -> Option<Self> {
+        Self::infinity(endianness).checked_sub(self, endianness)
+    }
+
+    /// In-place point addition.
+    ///
+    /// WARNING: This operation skips the prime-order subgroup check
+    /// for performance. Only use this if inputs are known to be valid.
+    pub fn add_unchecked_assign(
```
joncinque: nit: it's a bit more idiomatic to have `_unchecked` at the end of the function name
```suggestion
    pub fn add_assign_unchecked(
```

bls12-381/src/g1.rs
```diff
@@ -0,0 +1,392 @@
...
+            }
+        }
+    }
+
+    /// Point addition returning a new allocated point.
+    /// Skips subgroup checks.
+    pub fn add_unchecked(&self, other: &Self, endianness: Endianness) -> Option<Self> {
+        let mut result = MaybeUninit::<Self>::uninit();
+        let success =
+            self.add_unchecked_assign(other, unsafe { result.assume_init_mut() }, endianness);
```
joncinque: Although this might be ok since we're assigning in the function, the lines are technically undefined behavior: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initialization-invariant

Maybe `add_assign_unchecked` should just take a `MaybeUninit` directly?

bls12-381/src/g1.rs
```diff
@@ -0,0 +1,392 @@
...
+        if success {
+            Some(unsafe { result.assume_init() })
+        } else {
+            None
+        }
+    }
+
+    /// Safe in-place point addition.
+    /// Validates both operands (field, curve, subgroup) prior to addition.
+    pub fn checked_add_assign(&self, other: &Self, out: &mut Self, endianness: Endianness) -> bool {
```
joncinque: nit: we can just call this `add_assign` -- people can use the unchecked version if they want.

We might also want to have this take a `MaybeUninit` directly

bls12-381/src/g1.rs
```diff
@@ -0,0 +1,392 @@
...
+        if !self.validate(endianness) || !other.validate(endianness) {
+            return false;
+        }
+        self.add_unchecked_assign(other, out, endianness)
+    }
+
+    /// Safe point addition returning a new allocated point.
+    pub fn checked_add(&self, other: &Self, endianness: Endianness) -> Option<Self> {
```
joncinque: nit: same here, we can go with `add`

bls12-381/src/g1.rs
```diff
@@ -0,0 +1,392 @@
...
+        } else {
+            None
+        }
+    }
+
+    /// In-place point subtraction.
+    ///
+    /// WARNING: This operation skips the prime-order subgroup check
+    /// for performance. Only use this if inputs are known to be valid.
+    pub fn sub_unchecked_assign(
```
joncinque: nit: same here for consistency
```suggestion
    pub fn sub_assign_unchecked(
```

bls12-381/src/g1.rs
```diff
@@ -0,0 +1,392 @@
...
+        &self,
+        other: &Self,
+        out: &mut Self,
```
joncinque: this will probably need to be a `MaybeUninit`

bls12-381/src/g1.rs
```diff
@@ -0,0 +1,392 @@
...
+        if success {
+            Some(unsafe { result.assume_init() })
+        } else {
+            None
+        }
+    }
+
+    /// Safe in-place point subtraction.
+    /// Validates both operands prior to subtraction.
+    pub fn checked_sub_assign(&self, other: &Self, out: &mut Self, endianness: Endianness) -> bool {
```
joncinque: nit: remove the `checked` here

bls12-381/src/g1.rs
```diff
@@ -0,0 +1,392 @@
...
+        if !self.validate(endianness) || !other.validate(endianness) {
+            return false;
+        }
+        self.sub_unchecked_assign(other, out, endianness)
+    }
+
+    /// Safe point subtraction returning a new allocated point.
+    pub fn checked_sub(&self, other: &Self, endianness: Endianness) -> Option<Self> {
```
joncinque: nit: same here, remove the `sub`

bls12-381/src/g1.rs
```diff
@@ -0,0 +1,392 @@
...
+        } else {
+            None
+        }
+    }
+
+    /// In-place scalar multiplication.
+    ///
+    /// Note: The underlying syscall inherently performs full validation
+    /// (field, curve, and subgroup checks) on the input point.
+    pub fn mul_assign(&self, scalar: &Scalar, out: &mut Self, endianness: Endianness) -> bool {
```
joncinque: probably take in a `MaybeUninit` as the out param?

bls12-381/src/g1.rs
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    /// Returns the underlying compressed byte array via a memory copy.
+    pub const fn to_bytes(&self) -> [u8; G1_COMPRESSED_POINT_SIZE] {
+        self.0
+    }
+
+    /// In-place decompression into an uncompressed affine point.
+    /// Inherently performs format, field, curve, and subgroup validation.
+    pub fn decompress_assign(&self, out: &mut G1Point, endianness: Endianness) -> bool {
```
joncinque: Probably take a `MaybeUninit` here too

bls12-381/src/g2.rs
```diff
@@ -0,0 +1,392 @@
...
+    /// Safely negates the point by subtracting it from infinity.
+    pub fn neg(&self, endianness: Endianness) -> Option<Self> {
+        Self::infinity(endianness).checked_sub(self, endianness)
+    }
+
+    /// In-place point addition.
+    ///
+    /// WARNING: This operation skips the prime-order subgroup check
+    /// for performance. Only use this if inputs are known to be valid.
+    pub fn add_unchecked_assign(
```
joncinque: nit: `add_assign_unchecked`

bls12-381/src/g2.rs
```diff
@@ -0,0 +1,392 @@
...
+        &self,
+        other: &Self,
+        out: &mut Self,
```
joncinque: `MaybeUninit` here too

bls12-381/src/g2.rs
```diff
@@ -0,0 +1,392 @@
...
+        if success {
+            Some(unsafe { result.assume_init() })
+        } else {
+            None
+        }
+    }
+
+    /// Safe in-place point addition.
+    /// Validates both operands (field, curve, subgroup) prior to addition.
+    pub fn checked_add_assign(&self, other: &Self, out: &mut Self, endianness: Endianness) -> bool {
```
joncinque: Remove `checked_` here

bls12-381/src/g2.rs
```diff
@@ -0,0 +1,392 @@
...
+        if !self.validate(endianness) || !other.validate(endianness) {
+            return false;
+        }
+        self.add_unchecked_assign(other, out, endianness)
+    }
+
+    /// Safe point addition returning a new allocated point.
+    pub fn checked_add(&self, other: &Self, endianness: Endianness) -> Option<Self> {
```
joncinque: Remove `checked_` here

bls12-381/src/g2.rs
```diff
@@ -0,0 +1,392 @@
...
+        } else {
+            None
+        }
+    }
+
+    /// In-place point subtraction.
+    ///
+    /// WARNING: This operation skips the prime-order subgroup check
+    /// for performance. Only use this if inputs are known to be valid.
+    pub fn sub_unchecked_assign(
```
joncinque: `sub_assign_unchecked` here

bls12-381/src/g2.rs
```diff
@@ -0,0 +1,392 @@
...
+        &self,
+        other: &Self,
+        out: &mut Self,
```
joncinque: Take a `MaybeUninit` here

bls12-381/src/g2.rs
```diff
@@ -0,0 +1,392 @@
...
+        if success {
+            Some(unsafe { result.assume_init() })
+        } else {
+            None
+        }
+    }
+
+    /// Safe in-place point subtraction.
+    /// Validates both operands prior to subtraction.
+    pub fn checked_sub_assign(&self, other: &Self, out: &mut Self, endianness: Endianness) -> bool {
```
joncinque: Remove `checked_` here

bls12-381/src/g2.rs
```diff
@@ -0,0 +1,392 @@
...
+        if !self.validate(endianness) || !other.validate(endianness) {
+            return false;
+        }
+        self.sub_unchecked_assign(other, out, endianness)
+    }
+
+    /// Safe point subtraction returning a new allocated point.
+    pub fn checked_sub(&self, other: &Self, endianness: Endianness) -> Option<Self> {
```
joncinque: Remove `checked_` here

bls12-381/src/g2.rs
```diff
@@ -0,0 +1,392 @@
...
+        } else {
+            None
+        }
+    }
+
+    /// In-place scalar multiplication.
+    ///
+    /// Note: The underlying syscall inherently performs full validation
+    /// (field, curve, and subgroup checks) on the input point.
+    pub fn mul_assign(&self, scalar: &Scalar, out: &mut Self, endianness: Endianness) -> bool {
```
joncinque: Take a `MaybeUninit` here

bls12-381/src/g2.rs
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    /// Returns the underlying compressed byte array via a memory copy.
+    pub const fn to_bytes(&self) -> [u8; G2_COMPRESSED_POINT_SIZE] {
+        self.0
+    }
+
+    /// In-place decompression into an uncompressed affine point.
+    /// Inherently performs format, field, curve, and subgroup validation.
+    pub fn decompress_assign(&self, out: &mut G2Point, endianness: Endianness) -> bool {
```
joncinque: Take a `MaybeUninit` here

bls12-381/src/pairing.rs
```diff
@@ -0,0 +1,167 @@
...
+pub struct GtElement(pub [u8; GT_ELEMENT_SIZE]);
+
+/// In-place product of pairings for a batch of G1 and G2 points.
+///
+/// Computes `e(P_1, Q_1) * ... * e(P_n, Q_n)` and writes the resulting
+/// `GtElement` to the `out` reference.
+pub fn pairing_map_assign(
+    g1_points: &[G1Point],
+    g2_points: &[G2Point],
+    out: &mut GtElement,
```
joncinque: Take a `MaybeUninit` here?

joncinque (commented):
Looks great overall! Many comments, all small

bls12-381/Cargo.toml
samkim-crypto: Good idea. Done!

bls12-381/README.md
samkim-crypto: Okay agreed. I'll do this in a separate follow-up 🙏

bls12-381/src/g1.rs
samkim-crypto: Yeah I made the changes. Let me know what you think.

bls12-381/Cargo.toml
samkim-crypto: Nice catch. Should be addressed now.

bls12-381/src/g1.rs
samkim-crypto: Yeah this should be addressed now along with Jon's comments.

joncinque (commented):
Looks good to me! Will just need some more gating on the bytemuck feature to make it work. If it's too annoying, we can just have the bytemuck derivations always available, and add feature-gating in some future breaking change

samkim-crypto:
Okay yeah I just forgot to include the `bytemuck_derive`. Things should be fixed now, so let me know what you think 🙏

joncinque (commented):
Oh, that was easy, looks great to me!

And also, since it's no-std, can you add the crate to the non-alloc crates in `./scripts/check-no-std.sh`?

joncinque:
We'll need to modify all instances of `target_os = "solana"` to `any(target_os = "solana", target_arch = "bpf")`, otherwise the syscall implementation gets picked up

joncinque (approved):
Looks great to me!

PR #830, opened 2026-07-17
nonce: add frozen-abi traits to DurableNonce

2501babe:
to support https://github.com/anza-xyz/agave/issues/13909 id like to derive the abi traits for `DurableNonce` so that it can be (in a snapshot-format-safe manner) added to `BlockhashQueue`. my current idea is to not serialize/deserialize it, but to populate it from the deserialized `last_hash`. but to pass abi tests we need to either have these traits or hand-impl `AbiExample` for `BlockhashQueue`

### DIFF
--- a/nonce/src/state.rs
+++ b/nonce/src/state.rs
@@ -1,5 +1,7 @@
 #[cfg(feature = "serde")]
 use serde_derive::{Deserialize, Serialize};
+#[cfg(feature = "frozen-abi")]
+use solana_frozen_abi_macro::{AbiExample, StableAbi, StableAbiSample};
 #[cfg(feature = "wincode")]
 use wincode::{SchemaRead, SchemaWrite};
 use {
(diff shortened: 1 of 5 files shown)


nonce/src/state.rs
kskalski, line 82: for completeness, could we also add derives to all types that support serialization?

2501babe: sure, though i assumed the philosophy was more "abi traits imply we do or want to include this in important serialized structs", which we wouldnt want to with `Data` / `State` / `Versions`

kskalski: That might make sense for `AbiExample`, which IMHO should be avoided and I'm planning to actually start removing it from agave as much as possible (or completely).
The new thing is `StableAbi`, which should be added to basically any struct that implements `Serialize+Deserialize` / `SchemaRead+SchemaWrite` - let users decide if they want to benefit from that or not.

Also, those traits are just supporting traits, the key element is attaching `api_digest` (for `AbiExample`) or `abi_digest` (for `StableAbi`), which indeed pins the format and highlights the importance of the type as "top level" and needing to be "stable".

kskalski (approved):

PR #832, opened 2026-07-17
chore(deps)!: Bump wincode to 0.6.0

cpubot:


kskalski:
Will be a pain to roll-out, but I suppose we could follow the same pattern as before - treat it with minor releases across SDK, since wincode is not a feature enabled by default?

As an explainer - there were a couple of safety issues with a APIs that had to be resolved with breaking changes.

Cargo.toml
```diff
@@ -351,7 +351,7 @@ tiny-bip39 = "2.0.0"
 toml = "0.8.23"
 uriparse = "0.6.4"
 wasm-bindgen = "0.2.100"
-wincode = { version = "0.5.0", features = ["derive"], default-features = false }
```
robertkiel, line 355: Any reason to not add feature `bv-strict`?

This prevents [malleabilities](https://github.com/anza-xyz/wincode/pull/347) of the `BitVec` implementation.

cpubot: It rejects payloads serialized with the `serde` implementation in certain cases. It should likely be rolled out independently from the 0.6.0 bump where we can verify those rejection cases aren't problematic. Secondly, it should be introduced specifically in the sub-crates that use `bv`, not globally.

joncinque:
> Will be a pain to roll-out, but I suppose we could follow the same pattern as before - treat it with minor releases across SDK, since wincode is not a feature enabled by default?

Sorry for the lateness here, but yeah we can move forward with this. We did run into issues with that strategy last time because some SPL crates used by Agave updated to the wincode v0.5 compatible SDK dependencies, ie spl-token-interface -> solana-address with a higher version. That meant we couldn't then use Agave crates in other parts of the program repo, since they were on wincode v0.4 at the time.

I would like to disable Rust dependabot in the program repos for SDK crates, to make things simpler in the short term.

And we should also backport a version limitation in Agave, same as we did last time: https://github.com/anza-xyz/agave/pull/12227

kskalski (approved):
LGTM, I think agave now entangles most of the SDK crates through wincode trait interface, so it's more likely than before that we need to first publish all the crates with `wincode` feature and then bump them all at the same time.

PR #833, opened 2026-07-20
[define-syscall] Add `curve_constants` module for curve syscall constants

samkim-crypto:
#### Problem

The syscall functions associated with elliptic curves takes in a curve ID and perform operations based on it. Some examples of curve ID constants are [here](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0565-secp256r1-syscalls.md#new-curve-id-constants).

Currently, these constants live in various places:
- The curve25519 curve ID live in the `solana-curve25519` crate
- The BLS12-381 curve ID live in the `solana-syscall` crate.

It would be nice to have these constants in one place in the `solana-define-syscall` crate as they are syscall-related constants that are validator-client-independent.

#### Summary of Changes

I added a `curve_constants` module inside the `solana-define-syscall` crate and added the curve IDs there.

If there is a better place to host these constants, please let me know.

### DIFF
--- a/define-syscall/src/curve_constants.rs
+++ b/define-syscall/src/curve_constants.rs
@@ -1,0 +1,44 @@
+//! Constants and identifiers for Solana elliptic curve sycalls.
+//!
+//! This module defines the unique identifiers for supported elliptic curves,
+//! encoding formats, and algebraic group operations. These constants are used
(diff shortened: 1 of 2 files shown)


joncinque (approved):
This makes sense to me! I was thinking about whether an enum representation makes sense, but we're better off having the exact u64 constants here, since those are the parameters for the syscalls.

We can always build an enum representation using these constants with a higher level library

PR #839, opened 2026-07-23
[define-sycall] fix big_mod_exp syscall api

zz-sol:
fix the API for big_int_mod_exp so it returns an u64.


samkim-crypto (approved):
Thanks! Can you include this in https://github.com/solana-foundation/solana-improvement-documents/pull/585 as well?

PR #841, opened 2026-07-23
Add --levels flag to order-crates-for-publishing.py

kskalski:
#### Problem
The topological sort already removes packages in batches: each pass of the ordering loop drops every package whose dependencies have all been removed, so a batch is a set of mutually independent crates. This information isn't however exposed to the user of the script.

#### Summary of Changes
Record the batch number as a publishing "level".

Crates sharing a level are independent of each other and depend only on crates in earlier levels, so each level can be published in parallel once all earlier levels land.

The new --levels flag prints "<level>\t<path>"; without it the output is unchanged, so existing callers (including CI) are unaffected. The header comment documents the flag and a grep-based one-liner for grouping a subset of crates (e.g. those using wincode) into levels.

### DIFF
--- a/scripts/order-crates-for-publishing.py
+++ b/scripts/order-crates-for-publishing.py
@@ -6,6 +6,22 @@
 #
 # On success an ordered list of Cargo.toml files is written to stdout
 #
+# With |--levels| each line is prefixed with the publishing "level" the crate
+# belongs to (|<level>\t<path>|). Crates sharing a level are independent of each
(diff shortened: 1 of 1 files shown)


joncinque (commented):
This is very neat!

My only nit would be to use the term "level" instead of "wave", since this is closer to showing the level structure of the publish graph. How does that sound?

kskalski:
Yeah, sounds good, changed.

joncinque (approved):
Looks great!

PR #842, opened 2026-07-23
feat(account): add wincode serialization with separate StateMutWincode trait

kskalski:
#### Problem
Many `Account` functions rely on `SysvarSerialize` and bincode, we want to switch over from both to wincode / direct management of sysvar serialization and sizing.

#### Summary of Changes
Adds wincode-based (de)serialization for `Account`/`AccountSharedData` state, alongside bincode. The wire format matches bincode byte-for-byte; the only deliberate difference is raising the preallocation guard to 10 MiB (Solana's max account data length) so valid accounts are never rejected.

The whole wincode codec surface is a single `StateMutWincode<T>` trait: read (`state`), in-place write (`set_state`), and construction (`new_data` / `new_data_with_space`, as `Self`-returning associated fns). It is deliberately a
separate trait from the bincode `StateMut` so the two coexist when both features are enabled. Keeping construction on the trait lets a consumer migrating to wincode keep its `Type::new_data(..)` / `account.state()` call syntax — only the
trait import changes.

`set_state` and construction serialize into uninitialized capacity bounded to the target size, avoiding the allocate-zero-then-overwrite pattern and a redundant `serialized_size` pass.

### DIFF
--- a/account/src/account.rs
+++ b/account/src/account.rs
@@ -17,6 +17,10 @@ mod bincode;
 #[cfg(feature = "bincode")]
 pub use bincode::*;
 
+// NOTE: the wincode codec surface is not a peer of the bincode module above. Instead of
+// duplicating the inherent `new_data`/`serialize_data`/... methods, wincode is exposed
+// solely through the `state_traits::StateMutWincode` trait (read, write, and construct).
+
 // NOTE: `Account` and `AccountSharedData` are defined in the crate root (`lib.rs`)
 // rather than here. The frozen-abi digest of any downstream struct that holds an
 // `Account`/`AccountSharedData` field hashes that field's fully-qualified
@@ -229,6 +233,18 @@ impl Account {
         Account {
             lamports,
             data: vec![0; space],
+            owner: *owner,
+            executable: false,
+            rent_epoch: Epoch::default(),
+        }
+    }
+    /// Construct an `Account` from already-encoded `data`, taking ownership of the
+    /// buffer. Avoids the zero-fill allocation of [`Account::new`] followed by a
+    /// separate write when the caller already has the bytes.
+    pub fn new_with_data(lamports: u64, data: Vec<u8>, owner: &Pubkey) -> Self {
+        Account {
+            lamports,
+            data,
             owner: *owner,
             executable: false,
             rent_epoch: Epoch::default(),
--- a/account/src/lib.rs
+++ b/account/src/lib.rs
@@ -17,10 +17,16 @@ pub mod state_traits;
 
 // wincode config used by the `state_traits::wincode` trait. The wire format matches
 // bincode byte-for-byte; the only deliberate difference is raising the preallocation
-// guard to 10 MiB (Solana's max account data length) so valid accounts are never
-// rejected. This gates only the prealloc check, not the encoded bytes.
+// guard to the max account data length so valid accounts are never rejected. This gates
+// only the prealloc check, not the encoded bytes.
+//
+// Mirrors `solana_system_interface::MAX_PERMITTED_DATA_LENGTH` (kept in sync by the
+// assertion below) so `solana-system-interface` stays a dev-only dependency.
 #[cfg(feature = "wincode")]
 const WINCODE_PREALLOC_LIMIT: usize = 10 * 1024 * 1024;
+#[cfg(all(test, feature = "wincode"))]
+const _: () =
+    assert!(WINCODE_PREALLOC_LIMIT as u64 == solana_system_interface::MAX_PERMITTED_DATA_LENGTH);
 #[cfg(feature = "wincode")]
 pub(crate) type WincodeConfig = wincode::config::Configuration<true, WINCODE_PREALLOC_LIMIT>;
 #[cfg(feature = "wincode")]
--- a/account/src/lib.rs
+++ b/account/src/lib.rs
@@ -12,8 +12,21 @@ pub use account::*;
 #[cfg(feature = "serde")]
 mod serde;
 
-#[cfg(feature = "bincode")]
+#[cfg(any(feature = "bincode", feature = "wincode"))]
 pub mod state_traits;
+
+// wincode config used by the `state_traits::wincode` trait. The wire format matches
+// bincode byte-for-byte; the only deliberate difference is raising the preallocation
+// guard to 10 MiB (Solana's max account data length) so valid accounts are never
+// rejected. This gates only the prealloc check, not the encoded bytes.
+#[cfg(feature = "wincode")]
+const WINCODE_PREALLOC_LIMIT: usize = 10 * 1024 * 1024;
+#[cfg(feature = "wincode")]
+pub(crate) type WincodeConfig = wincode::config::Configuration<true, WINCODE_PREALLOC_LIMIT>;
+#[cfg(feature = "wincode")]
+pub(crate) const WINCODE_CONFIG: WincodeConfig = wincode::config::Configuration::default()
+    .with_preallocation_size_limit::<WINCODE_PREALLOC_LIMIT>(
+);
 
 // NOTE: `Account` and `AccountSharedData` are defined here, in the crate root, on
 // purpose. The frozen-abi digest of any downstream struct that has an
--- a/account/src/state_traits/wincode.rs
+++ b/account/src/state_traits/wincode.rs
@@ -1,0 +1,193 @@
+//! wincode-based state trait for the account types.
+//!
+//! wincode gets its own [`StateMutWincode`] trait rather than reusing
+//! [`StateMut`](crate::state_traits::StateMut), so the bincode and wincode
+//! implementations can coexist when both features are enabled. It is the crate's
+//! entire wincode codec surface: reading (`state`), in-place writing (`set_state`),
+//! and construction (`new_data`/`new_data_with_space`).
+
+use {
+    crate::{
+        Account, AccountSharedData, ReadableAccount, WincodeConfig, WritableAccount, WINCODE_CONFIG,
+    },
+    solana_instruction_error::InstructionError,
+    solana_pubkey::Pubkey,
+    wincode::{SchemaRead, SchemaWrite, WriteResult},
+};
+
+/// wincode counterpart of [`StateMut`](crate::state_traits::StateMut) that also
+/// constructs accounts whose data is wincode-encoded.
+pub trait StateMutWincode<T>: Sized {
+    fn state(&self) -> Result<T, InstructionError>;
+    fn set_state(&mut self, state: &T) -> Result<(), InstructionError>;
+
+    /// Create an account sized to the wincode-serialized length of `state`.
+    fn new_data(lamports: u64, state: &T, owner: &Pubkey) -> WriteResult<Self>;
+
+    /// Create an account of `space` bytes with `state` serialized into the front and
+    /// the remainder zero-padded; fails with `WriteSizeLimit` if `state` does not fit.
+    fn new_data_with_space(
+        lamports: u64,
+        state: &T,
+        space: usize,
+        owner: &Pubkey,
+    ) -> WriteResult<Self>;
+}
+
+impl<T> StateMutWincode<T> for Account
+where
+    T: SchemaWrite<WincodeConfig, Src = T> + for<'de> SchemaRead<'de, WincodeConfig, Dst = T>,
+{
+    fn state(&self) -> Result<T, InstructionError> {
+        deserialize_state(self)
+    }
+    fn set_state(&mut self, state: &T) -> Result<(), InstructionError> {
+        serialize_state(self, state)
+    }
+    fn new_data(lamports: u64, state: &T, owner: &Pubkey) -> WriteResult<Self> {
+        let data = wincode::config::serialize(state, WINCODE_CONFIG)?;
+        Ok(Account::new_with_data(lamports, data, owner))
+    }
+    fn new_data_with_space(
+        lamports: u64,
+        state: &T,
+        space: usize,
+        owner: &Pubkey,
+    ) -> WriteResult<Self> {
+        let mut data = serialize_bounded(state, space)?;
+        data.resize(space, 0);
+        Ok(Account::new_with_data(lamports, data, owner))
+    }
+}
+
+impl<T> StateMutWincode<T> for AccountSharedData
+where
+    T: SchemaWrite<WincodeConfig, Src = T> + for<'de> SchemaRead<'de, WincodeConfig, Dst = T>,
+{
+    fn state(&self) -> Result<T, InstructionError> {
+        deserialize_state(self)
+    }
+    fn set_state(&mut self, state: &T) -> Result<(), InstructionError> {
+        serialize_state(self, state)
+    }
+    fn new_data(lamports: u64, state: &T, owner: &Pubkey) -> WriteResult<Self> {
+        <Account as StateMutWincode<T>>::new_data(lamports, state, owner)
+            .map(AccountSharedData::from)
+    }
+    fn new_data_with_space(
+        lamports: u64,
+        state: &T,
+        space: usize,
+        owner: &Pubkey,
+    ) -> WriteResult<Self> {
+        <Account as StateMutWincode<T>>::new_data_with_space(lamports, state, space, owner)
+            .map(AccountSharedData::from)
+    }
+}
+
+fn deserialize_state<T, U>(account: &U) -> Result<T, InstructionError>
+where
+    T: for<'de> SchemaRead<'de, WincodeConfig, Dst = T>,
+    U: ReadableAccount,
+{
+    wincode::config::deserialize(account.data(), WINCODE_CONFIG)
+        .map_err(|_| InstructionError::InvalidAccountData)
+}
+
+fn serialize_state<T, U>(account: &mut U, state: &T) -> Result<(), InstructionError>
+where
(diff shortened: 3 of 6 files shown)


account/src/account/wincode.rs
```diff
@@ -0,0 +1,169 @@
...
+        state: &T,
+    ) -> WriteResult<()> {
+        shared_serialize_data(self, state)
+    }
+}
+
+// NOTE: the `create_account_*` sysvar helpers are omitted: they size accounts via the
+// bincode-gated `SysvarSerialize::size_of()`, and sizing to the wincode-serialized
+// length instead would undersize sysvars whose canonical size exceeds it. These helpers
+// are being deprecated/removed in favor of caller-provided sizing.
```
kskalski: possibly we can get rid of more functions?

account/src/account.rs
```diff
@@ -17,6 +17,11 @@ mod bincode;
 #[cfg(feature = "bincode")]
 pub use bincode::*;
 
+#[cfg(all(feature = "wincode", not(feature = "bincode")))]
```
kskalski: one downside of that is feature unification, when some crate forces bincode on `solana-account`, it will disable the new wincode based functions for others...

brooksprumo:
PR is currently marked as in draft. Is it ready for review? Or just an FYI?

kskalski:
I think it's ready, marked for review, though I suppose at this point let's agree on high level approach first before reviewing impl.

account/src/lib.rs
```diff
@@ -12,9 +12,23 @@ pub use account::*;
...
+#[cfg(any(feature = "bincode", feature = "wincode"))]
 pub mod state_traits;
 
+// wincode config shared by the `account::wincode` and `state_traits::wincode`
+// sub-modules. The wire format matches bincode byte-for-byte; the only deliberate
+// difference is raising the preallocation guard to 10 MiB (Solana's max account data
+// length) so valid accounts are never rejected. This gates only the prealloc check,
+// not the encoded bytes.
+#[cfg(feature = "wincode")]
+const WINCODE_PREALLOC_LIMIT: usize = 10 * 1024 * 1024;
```
brooksprumo, line 26: Should we use `MAX_PERMITTED_DATA_LENGTH` from system-interface?


```suggestion
const WINCODE_PREALLOC_LIMIT: usize = solana_system_interface::MAX_PERMITTED_DATA_LENGTH as usize;
```

I'm not sure if we're trying to limit other dependencies or not though.

account/src/account/wincode.rs
```diff
@@ -0,0 +1,169 @@
...
+        let data = wincode::config::serialize(state, WINCODE_CONFIG)?;
+        Ok(Account::new_with_data(lamports, data, owner))
+    }
+    pub fn new_ref_data<T: SchemaWrite<WincodeConfig, Src = T>>(
+        lamports: u64,
+        state: &T,
+        owner: &Pubkey,
+    ) -> WriteResult<RefCell<Self>> {
+        Account::new_data(lamports, state, owner).map(RefCell::new)
+    }
```
joncinque: I did a grep in agave, and found no usage of this function -- maybe we don't add it?

account/src/account/wincode.rs
```diff
@@ -0,0 +1,169 @@
...
+        Ok(Account::new_with_data(lamports, data, owner))
+    }
+    pub fn new_ref_data_with_space<T: SchemaWrite<WincodeConfig, Src = T>>(
+        lamports: u64,
+        state: &T,
+        space: usize,
+        owner: &Pubkey,
+    ) -> WriteResult<RefCell<Self>> {
+        Account::new_data_with_space(lamports, state, space, owner).map(RefCell::new)
+    }
```
joncinque: Same with this, found no usage of it in agave

account/src/account/wincode.rs
```diff
@@ -0,0 +1,169 @@
...
+    ) -> WriteResult<Self> {
+        Account::new_data(lamports, state, owner).map(AccountSharedData::from)
+    }
+    pub fn new_ref_data<T: SchemaWrite<WincodeConfig, Src = T>>(
+        lamports: u64,
+        state: &T,
+        owner: &Pubkey,
+    ) -> WriteResult<RefCell<Self>> {
+        AccountSharedData::new_data(lamports, state, owner).map(RefCell::new)
+    }
```
joncinque: Same with this, seems unused

account/src/account/wincode.rs
```diff
@@ -0,0 +1,169 @@
...
+        Account::new_data_with_space(lamports, state, space, owner).map(AccountSharedData::from)
+    }
+    pub fn new_ref_data_with_space<T: SchemaWrite<WincodeConfig, Src = T>>(
+        lamports: u64,
+        state: &T,
+        space: usize,
+        owner: &Pubkey,
+    ) -> WriteResult<RefCell<Self>> {
+        AccountSharedData::new_data_with_space(lamports, state, space, owner).map(RefCell::new)
+    }
```
joncinque: And this doesn't seem to be used either

account/src/account/wincode.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+/// Serialize a value into an `Account`'s data.
+pub fn to_account<S, T>(state: &S, account: &mut T) -> Option<()>
+where
+    S: SchemaWrite<WincodeConfig, Src = S>,
+    T: WritableAccount,
+{
+    wincode::config::serialize_into(account.data_as_mut_slice(), state, WINCODE_CONFIG).ok()
+}
```
joncinque: With `deserialize_data` and `serialize_data`, do we need these?

account/src/account/wincode.rs
joncinque: Yes please!

account/src/state_traits/wincode.rs
```diff
@@ -0,0 +1,89 @@
...
+    solana_instruction_error::InstructionError,
+    std::cell::Ref,
+    wincode::{SchemaRead, SchemaWrite},
+};
+
+/// wincode counterpart of [`StateMut`](crate::state_traits::StateMut).
+pub trait StateMutWincode<T> {
+    fn state(&self) -> Result<T, InstructionError>;
+    fn set_state(&mut self, state: &T) -> Result<(), InstructionError>;
+}
```
joncinque: There seems to be some overlap with the `deserialize_data` and `serialize_data` that are implemented directly on the Account types. I would vote with doing one or the other, ie *either* a trait *or* the functions on the types.

To be more Rusty, the trait is probably the better option, but I'd be curious to hear @brooksprumo 's take

account/src/account.rs
joncinque: As mentioned in my other comment, maybe we should *just* do the wincode trait in that case

account/src/account.rs
```diff
@@ -234,6 +239,18 @@ impl Account {
...
+    /// separate write when the caller already has the bytes.
+    pub fn new_with_data(lamports: u64, data: Vec<u8>, owner: &Pubkey) -> Self {
+        Account {
+            lamports,
+            data,
+            owner: *owner,
+            executable: false,
+            rent_epoch: Epoch::default(),
+        }
+    }
```
joncinque, line 252: I can't believe we didn't have this before 😅

joncinque (commented):
The code looks good to me, we just need to choose which of it we want to add in

account/src/state_traits/wincode.rs
brooksprumo: Traits are nice in that could have both wincode and bincode trait impls, whereas the methods impls can only do one.

How much work is it to remove the method impls and replace with traits?

account/src/lib.rs
kskalski: Looks like it doesn't cause a cycle, but I'm not completely sure we want this new dependency arc, so I'm putting it on a separate commit d93353aaf8dad71c6e29767b740881d912bcdf53 - will let Jon comment on whether it's fine

account/src/state_traits/wincode.rs
kskalski: Ok, I rewrote this PR - we actually can use just the trait to satisfy all needs of agave by adding constructor functions, so now the `wincode` feature changes are purely additive and can coegist with bincode

account/src/account.rs
```diff
@@ -234,6 +238,18 @@ impl Account {
             rent_epoch: Epoch::default(),
         }
     }
+    /// Construct an `Account` from already-encoded `data`, taking ownership of the
+    /// buffer. Avoids the zero-fill allocation of [`Account::new`] followed by a
+    /// separate write when the caller already has the bytes.
+    pub fn new_with_data(lamports: u64, data: Vec<u8>, owner: &Pubkey) -> Self {
```
brooksprumo, line 244: Can the `owner` param be by-value instead?

```suggestion
    pub fn new_with_data(lamports: u64, data: Vec<u8>, owner: Pubkey) -> Self {
```

For me, when I see a param as a reference, I don't expect it to be copied unconditionally. So it's somewhat surprising when only looking at the function signature that the impl always makes a copy. As a user, I'd rather see that I need to pass in a Pubkey value.

account/src/state_traits/wincode.rs
```diff
@@ -0,0 +1,193 @@
...
+fn deserialize_state<T, U>(account: &U) -> Result<T, InstructionError>
+where
+    T: for<'de> SchemaRead<'de, WincodeConfig, Dst = T>,
+    U: ReadableAccount,
+{
+    wincode::config::deserialize(account.data(), WINCODE_CONFIG)
+        .map_err(|_| InstructionError::InvalidAccountData)
+}
+
+fn serialize_state<T, U>(account: &mut U, state: &T) -> Result<(), InstructionError>
```
brooksprumo, line 97: Just noting that if `serialize_into()` fails and this fn returns an Err (with account data too small) that `account` may be in an unknown/bad/modified state.

I think that's an ok trade off though, for performance. The caller shouldn't touch `account` if the fn fails.

An alternative could be to take `U` by value, and only return it when `Ok`. I imagine that'll be a little worse for perf since Rust has a known compiler bug where it doesn't do NRVO.

brooksprumo (commented):
Here's a finding from codex:

```
• When both bincode and wincode are enabled, Account::new_data(...) resolves to bincode’s inherent method even if StateMutWincode is imported, because Rust always prioritizes inherent methods over trait methods. This means Cargo feature unification can either make a wincode-only state fail to
  compile or silently select bincode for a type implementing both codecs.

  Could we give the wincode constructors distinct names, such as new_data_wincode and new_data_with_space_wincode? That would preserve ergonomic call syntax while making codec selection unambiguous under combined features. Otherwise, callers must use explicit UFCS:

  <Account as StateMutWincode<T>>::new_data(...)
```

I'm not sure if I agree with the codex suggestions, but wanted to get your thoughts.

account/src/account.rs
kskalski: makes sense, though it brings a bit of a dilemma:
* it is inconsistent with all the other functions, which take `&Pubkey` and use it for the same purpose
* should the new trait also adopt this convention? then it's not a strightforward "switch features + add imports" change, we will need to augment all the callsites to move to the wincode

account/src/state_traits/wincode.rs
kskalski: I think this is fine as is, the account could be mutated, but on Err there is no contract / expectation that it stays consistent. 

We also do not expose any uninitialized memory in this way, the function operates on `&mut [u8]`, so slice taken from `U` must already be a valid initialized memory, which we just overwrite partially.

kskalski:
>   compile or silently select bincode for a type implementing both codecs.
> 
> I'm not sure if I agree with the codex suggestions, but wanted to get your thoughts.

Yeah, I think it isn't worth adding names that will be long / hard / impossible to migrate over from later.

The important points:
* silent move from wincode to bincode is a perf, not correctness problem, since they are expected to be binary compatible
* we are going to remove bincode feature across the dependency chains, so we will actively track any unified bincode that is still present, this situation is thus just temporary
* there is always an escape hatch of explicitly calling the new trait as suggested above

On the other hand, if we apply your suggestion to use `Pubkey` instead of `&Pubkey`, it would make sense to switch the trait functions too and then we no longer have a problem of "silent" use of bincode impl... though we do have a few other problems like needing to touch all the call sites.

account/src/account.rs
brooksprumo: Gotcha. Keeping `&Pubkey` is fine with me then.

brooksprumo:
Given this
> I think it isn't worth adding names that will be long / hard / impossible to migrate over from later.

Plus this
> * we are going to remove bincode feature across the dependency chains, so we will actively track any unified bincode that is still present, this situation is thus just temporary

I think it's fine to leave as is. Thanks!

brooksprumo (dismissed):
:shipit:

robertkiel:
Note that `bincode` first checks the serialized size and fails before mutating state

https://github.com/anza-xyz/solana-sdk/blob/e37ac49438169ca9f19e3c893fca65b6fe7d0441/account/src/account/bincode.rs#L24-L27

`wincode` only does this for statically sized types, i.e. `TypeMeta::Static`. `Account` and `AccountSharedData` seem to default to `TypeMeta::Dynamic`.

So this is a breaking change.

kskalski:
> Note that `bincode` first checks the serialized size and fails before mutating state
> 
> https://github.com/anza-xyz/solana-sdk/blob/e37ac49438169ca9f19e3c893fca65b6fe7d0441/account/src/account/bincode.rs#L24-L27
> 
> `wincode` only does this for statically sized types, i.e. `TypeMeta::Static`. `Account` and `AccountSharedData` seem to default to `TypeMeta::Dynamic`.

This doesn't matter though, since we only pass the same limited size mut slice to serialization, so even if wincode doesn't early return checking `serialized_size`, it will still fail trying to serialize any larger amount. Or do you mean that mismatch in returned value of `serialized_size` vs what size `serialize` actually produces is an issue?
 
> So this is a breaking change.

Activating new code requires switching imported trait and in many cases actually switching activated features of a crate, so I'm not sure this qualified as breaking.

robertkiel:
see also https://github.com/anza-xyz/wincode/issues/418

robertkiel:
> Activating new code requires switching imported trait and in many cases actually switching activated features of a crate, so I'm not sure this qualified as breaking.

... sure, it is behind a feature flag but sooner or later, this code will be used.

Regarding the exploitability:

An attacker may have a valid account. The attacker may now update its `data` so that `serialize_into` fails due to space boundaries. However, the property `lamports` is written before `data`, so the account state is now inconsistent.

kskalski:
Wouldn't runtime just revert the whole instruction and all mutations of the account, including `lamports` and partially updated `data` if serialize / `set_state` returned `Err`?

robertkiel:
> Wouldn't runtime just revert the whole instruction and all mutations of the account, including `lamports` and partially updated `data` if serialize / `set_state` returned `Err`?

Can't say for sure right now. This needs some further exploration.

account/src/lib.rs
```diff
@@ -12,9 +12,22 @@ pub use account::*;
...
-#[cfg(feature = "bincode")]
+#[cfg(any(feature = "bincode", feature = "wincode"))]
 pub mod state_traits;
 
+// wincode config used by the `state_traits::wincode` trait. The wire format matches
+// bincode byte-for-byte; the only deliberate difference is raising the preallocation
+// guard to the max account data length so valid accounts are never rejected. This gates
+// only the prealloc check, not the encoded bytes.
+#[cfg(feature = "wincode")]
+const WINCODE_PREALLOC_LIMIT: usize = solana_system_interface::MAX_PERMITTED_DATA_LENGTH as usize;
```
joncinque: Rather than introducing the dependency on `solana-system-interface`, let's redefine the const and add a static assertion under `#[cfg(test)]` to make sure that they match, like we do at https://github.com/anza-xyz/solana-sdk/blob/34fd0f28813c780d8d17d124de5e3dcd8f0bb3f2/cpi/src/lib.rs#L279-L280

joncinque (commented):
One last bit, the rest looks good to me

account/src/lib.rs
kskalski: done

joncinque (dismissed):
Looks great to me, thanks!

account/src/lib.rs
```diff
@@ -12,9 +12,30 @@ pub use account::*;
...
+//
+// Mirrors `solana_system_interface::MAX_PERMITTED_DATA_LENGTH` (kept in sync by the
+// assertion below) so `solana-system-interface` stays a dev-only dependency.
+#[cfg(feature = "wincode")]
+const WINCODE_PREALLOC_LIMIT: usize = 10 * 1024 * 1024;
+#[cfg(all(test, feature = "wincode"))]
+static_assertions::const_assert_eq!(
+    WINCODE_PREALLOC_LIMIT as u64,
+    solana_system_interface::MAX_PERMITTED_DATA_LENGTH
+);
```
brooksprumo: Can also do:

```rust
const _: () = assert!(WINCODE_PREALLOC_LIMIT as u64 == solana_system_interface::MAX_PERMITTED_DATA_LENGTH);
```

So then a new dep on `static_assertions` isn't required.

Edit: Not a blocker for merging. Feel free to disregard.

joncinque (dismissed):

account/src/lib.rs
kskalski: yeah, it's cleaned. changed

brooksprumo (approved):

PR #843, opened 2026-07-24
fix: SanitizedMessage::get_durable_nonce static account only

apfitzge:
### Problem
- `SanitizedMessage::get_durable_nonce` is in inconsistent with `SVMMessage::get_durable_nonce` since we have activated `7VVhpg5oAjAmnmz1zCcSHb2Z9ecZB2FQqpnEwReka9Zm`
	- specifically it still returns Some for non-static acct index
	- this is only used in rpc in agave 

### Summary of Changes
- Only return Some if the nonce acct is a static acct index

### DIFF
--- a/message/src/sanitized.rs
+++ b/message/src/sanitized.rs
@@ -347,7 +347,7 @@ impl SanitizedMessage {
                     if !self.is_writable(idx) {
                         None
                     } else {
-                        self.account_keys().get(idx)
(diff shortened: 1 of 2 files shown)


apfitzge:
the flow of ALT nonce hasn't been supported for over a year...this can be considered a patch fix imo.

tao-stones (approved):
good call, looks good to me

Issue #831, opened 2026-07-17
[message] Deserializer and sanitizer support massively oversized V1 transactions


robertkiel:
_Requires #832_

[SIMD 385](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0385-transaction-v1.md) bumps the transaction size limit to 4096 bytes. However, neither the deserializer nor the `sanitize()` function enforce this limit. This allows a semi-honest user of the platform to submit oversized transactions. See PoC for a transaction that passes `sanitize()` with a size of ~4.2 MiB.

### Details

The code specifies a maximum transaction size but apparently, this constant is only used in tests and not enforced by the deserializer or the sanitizer.

[Code](https://github.com/anza-xyz/solana-sdk/blob/33e4209a50cfadd6f7be2f5323d7380c8a537843/message/src/versions/v1/mod.rs#L35-L36)
```rust
/// Maximum transaction size for V1 format in bytes.
pub const MAX_TRANSACTION_SIZE: usize = 4096;
```

### Suggestion

By utilizing wincode's `deserialize_exact` [API](https://docs.rs/wincode/latest/wincode/config/fn.deserialize_exact.html), the deserializer automatically rejects trailing bytes and defines an upper bound for DST-allocations, such as `Vec<T>`. This does not enforce the exact upper bound of 4096 bytes per transaction but provides an easy and effective way to safe the network from handling and filtering oversized transactions.

[Code](https://github.com/anza-xyz/solana-sdk/blob/5a0be40fca67bbd5c1b2c3a3e9c825c62538bb2d/message/src/versions/v1/message.rs#L709-L715)
```rust
pub fn deserialize(input: &[u8]) -> wincode::ReadResult<Message> {
    wincode::config::deserialize_exact(input, wincode::config::DefaultConfig::default().with_preallocation_size_limit::<MAX_TRANSACTION_SIZE>())
}
```

### PoC

Add to `message/tests/v1_sanitize_bypass.rs`

```rust
//! Exercises the V1 `Message::validate` / `sanitize` gate and shows what it does
//! *not* catch.
//!
//! `validate()` (message/src/versions/v1/message.rs:434) bounds the account and
//! instruction *counts* (<= 64 each) and every account/program index, but it
//! never bounds the *total serialized size*. The documented per-transaction cap
//! `MAX_TRANSACTION_SIZE = 4096` (v1/mod.rs:36) is a dead constant — it is never
[...]

PR #846, opened 2026-07-24
message: enforce transaction size limit in V1 deserialize

valentynkit:
## Problem

`v1::Message::deserialize` deserializes with wincode's default config, which permits preallocations up to 4 MiB and ignores trailing bytes. So the documented `MAX_TRANSACTION_SIZE` (`message/src/versions/v1/mod.rs`) is never enforced and the constant is unused outside tests. A buffer that declares an oversized instruction is allocated for before any check runs, and `sanitize()` bounds counts and indices but never total size (#831 has a message that passes `sanitize()` at roughly 4 MiB).

## Change

Bound preallocation to `MAX_TRANSACTION_SIZE` and reject trailing bytes with `deserialize_exact`:

```rust
pub fn deserialize(input: &[u8]) -> wincode::ReadResult<Message> {
    wincode::config::deserialize_exact(
        input,
        wincode::config::DefaultConfig::default()
            .with_preallocation_size_limit::<MAX_TRANSACTION_SIZE>(),
    )
}
```

Same two-part pattern agave uses for repair requests (`deserialize_exact` plus a config whose preallocation limit is the size cap). It also wires the previously dead `MAX_TRANSACTION_SIZE`.

This is hardening of the public `deserialize` helper, not a fix for a live remote DoS: agave already caps before the deserializer (RPC at `MAX_TRANSACTION_SIZE`, TPU/QUIC at `PACKET_DATA_SIZE`), and V1's length prefixes are `u8`/`u16` so there is no small-input preallocation amplification. The value is a self-defending public API for callers that decode untrusted bytes without pre-capping, plus strict framing.

One open point: the generic `VersionedMessage`/`VersionedTransaction` paths still decode with the default config, and a test there intentionally accepts an over-max transaction. I left those permissive since agave caps at ingress, but can thread the limit through them too if you'd prefer.

## Testing

`cargo test -p solana-message --features "wincode,serde"` passes (125 tests), including two new ones: an oversized message is rejected while the default-config `wincode::deserialize` still accepts it, and a valid message round-trips.

Addresses #831.

### DIFF
--- a/message/src/versions/v1/message.rs
+++ b/message/src/versions/v1/message.rs
@@ -60,7 +60,7 @@ use {
         compiled_keys::CompiledKeys,
         v1::{
             MessageError, TransactionConfig, TransactionConfigMask, MAX_ADDRESSES, MAX_HEAP_SIZE,
-            MAX_INSTRUCTIONS, MAX_SIGNATURES, MIN_HEAP_SIZE,
+            MAX_INSTRUCTIONS, MAX_SIGNATURES, MAX_TRANSACTION_SIZE, MIN_HEAP_SIZE,
         },
         AccountKeys, AddressSet, CompileError, MessageHeader,
     },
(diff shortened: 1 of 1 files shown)


apfitzge:
it's checked in agave, which is consistent with the way it's done for legacy and v0.

valentynkit:
Thanks. Missed the check in agave, closing.

robertkiel:
see related issue in `wincode` itself http://github.com/anza-xyz/wincode/issues/417

PR #848, opened 2026-07-27
Bump bytemuck and bytemuck_derive versions to 1.25 and 1.11

samkim-crypto:
#### Problem

There are some minimal version issue for `bytemuck` and `bytemuck_derive` crates as I was adding the syscall library for bls12-381 (https://github.com/anza-xyz/solana-sdk/pull/829).

The [cryptography](https://github.com/anza-xyz/cryptography) repo specifies  `bytemuck` v1.25 and `bytemuck_derive` v1.10` as the minimum versions while the solana-sdk is still on v1.21 and v1.8 respectively.

#### Summary of Changes

I bumped the `bytemuck` and `bytemuck_derive` versions in the solana-sdk. 

I looked at the change logs for [bytemuck](https://github.com/Lokathor/bytemuck/blob/main/changelog.md) and [bytemuck_derive](https://github.com/Lokathor/bytemuck/blob/main/derive/changelog.md) and it seems like the version bumps are minor updates that don't directly impact crates in the sdk. But if there are reasons we don't want to make this upgrade, then let me know and I can figure another way to resolve the CI in https://github.com/anza-xyz/solana-sdk/pull/829.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -618,18 +618,18 @@ checksum = "7575182f7272186991736b70173b0ea045398f984bf5ebbb3804736ce1330c9d"
 
 [[package]]
 name = "bytemuck"
-version = "1.25.0"
+version = "1.25.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c8efb64bd706a16a1bdde310ae86b351e4d21550d98d056f22f8a7f7a2183fec"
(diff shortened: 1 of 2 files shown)


joncinque:
Since we're providing libraries from this repo, we should aim to use the lowest acceptable version of dependencies. "Acceptable" meaning it has only the features we need, without any known security vulnerabilities.

Unless we have a specific reason to be using the higher versions in the cryptography repo, we should lower the dependencies over there rather than increase the versions over here

PR #850, opened 2026-07-27
[ed25519] update keypair dependencies from dalek to cryptography

zz-sol:
update keypair dependencies from dalek to cryptography


keypair/src/lib.rs
```diff
@@ -92,29 +95,35 @@ impl TryFrom<&[u8]> for Keypair {
...
+                "candidate keypair byte array is the wrong length",
+            ))
+        })?;
+        // The first half is the secret key seed; the second half is the
+        // expected public key. Reconstruct the signing key from the seed and
+        // verify that its derived public key matches the provided one.
+        let secret_key: [u8; Keypair::SECRET_KEY_LENGTH] = keypair_bytes
+            [..Keypair::SECRET_KEY_LENGTH]
+            .try_into()
+            .expect("secret key length is checked above");
```
samkim-crypto: As good security hygiene, it might be good to zeroize this out since it contains secret key material that gets copied.

keypair/src/lib.rs
```diff
@@ -92,29 +95,35 @@ impl TryFrom<&[u8]> for Keypair {
...
+        let signing_key = SigningKey::from(secret_key);
+        if signing_key.verification_key().as_ref() != &keypair_bytes[Keypair::SECRET_KEY_LENGTH..] {
```
samkim-crypto: It might also be worth doing the comparison in constant time for good security hygiene. We probably don't want to introduce a new dependency like `subtle` here, so we can either add a manual byte-by-byte comparison loop that does not short-circuit or (probably better) add a `from_keypair_bytes` equivalent to `solana-ed25519`.

samkim-crypto (commented):
Looks good overall. Some nits below, but it might be cleaner to add something like dalek's `from_keypair_bytes` into the `solana-ed25519` crate as noted below.

keypair/src/lib.rs
samkim-crypto: @zz-sol Sorry, I meant we should minimize the number of direct dependencies, so we should not add a new direct dependency like `subtle` to this crate.

But it would be good to get @joncinque's take here though. We will have crates like `subtle` in our dependency graph anyway because dalek has it as its own dependency, so maybe it doesn't matter to add. Overall, it would be cleaner to add `from_keypair_bytes` to `solana-ed25519` to make this function simpler and hide zeroization and constant time eq behind `solana-ed25519`.

zz-sol: oh i see. somehow i read it as "We probably want to introduce a new dependency like subtle here"

keypair/src/lib.rs
```diff
@@ -34,22 +36,25 @@ impl Keypair {
...
     /// Constructs a new `Keypair` using secret key bytes
     pub fn new_from_array(secret_key: [u8; 32]) -> Self {
-        Self(ed25519_dalek::SigningKey::from(secret_key))
+        Self(SigningKey::from(secret_key))
     }
 
     /// Returns this `Keypair` as a byte array
     pub fn to_bytes(&self) -> [u8; KEYPAIR_LENGTH] {
-        self.0.to_keypair_bytes()
+        let mut bytes = [0u8; KEYPAIR_LENGTH];
```
joncinque, line 48: nit: how about using `MaybeUninit` here to avoid zero-initializing?

keypair/src/lib.rs
```diff
@@ -92,29 +97,47 @@ impl TryFrom<&[u8]> for Keypair {
...
+                "keypair bytes do not specify same pubkey as derived from their secret key",
+            )));
+        }
+        Ok(Self(signing_key))
     }
 }
 
 #[cfg(test)]
-static_assertions::const_assert_eq!(Keypair::SECRET_KEY_LENGTH, ed25519_dalek::SECRET_KEY_LENGTH);
+static_assertions::const_assert_eq!(Keypair::SECRET_KEY_LENGTH * 2, KEYPAIR_LENGTH);
```
joncinque, line 152: nit: to preserve the old behavior, this should rather check against `size_of::<SigningKey>()`
```suggestion
static_assertions::const_assert_eq!(Keypair::SECRET_KEY_LENGTH, size_of::<SigningKey>());
```

joncinque (commented):
Just a couple of tiny comments, looks great otherwise

keypair/src/lib.rs
joncinque: I'm always happy to limit dependencies, and in this case, it seems pretty easy to avoid adding them, so let's inline whatever we need here directly.

keypair/src/lib.rs
zz-sol: Just tried it and reverted.

`MaybeUninit` seems to require an **unsafe** block to get the value back out (`assume_init`), and the safe slice helper that would keep the copies length-checked (`write_copy_of_slice`) only stabilized in 1.93, (c.f. our 1.89 MSRV).

joncinque: We don't need to be afraid of unsafe code, especially when it's clear that we're using the functions correctly.

Either way, we can do the work in a follow-up PR, since the rest of the crate is zero-initializing currently.

keypair/src/lib.rs
```diff
@@ -88,33 +93,72 @@ impl Keypair {
...
+// Fails to compile if the underlying library's secret key length ever diverges
+// from `Keypair::SECRET_KEY_LENGTH`.
+#[cfg(test)]
+const _: fn(&[u8; Keypair::SECRET_KEY_LENGTH]) -> SigningKey = SigningKey::from_bytes;
```
joncinque, line 156: Heh, clever

joncinque (approved):
Looks good to me!

keypair/src/lib.rs
zz-sol: claude came up with the idea

samkim-crypto (approved):

PR #852, opened 2026-07-28
Reorder publish steps to first push tag and then publish to crates.io

kskalski:
#### Problem
`cargo release` publishes to crates.io and then pushes the version-bump commit + tag in a single non-atomic step, with no rebase/retry if that push is rejected non-fast-forward. When another release (or any commit) lands on `master` in that window, the crate ends up published on crates.io while its git commit and tag fail to push, leaving the repo inconsistent.

#### Summary of Changes
- Add `--no-push` to the `cargo release` call, so publishing + the local commit/tag still happen, but the git push is separated out.
- After release, push the branch and tag with `git push --atomic` in a retry loop (up to 5 attempts). On rejection, `git pull --rebase origin <branch>`, re-point the tag onto the rebased commit with `git tag -f`, and retry. Exit non-zero if all attempts are exhausted.

### DIFF
--- a/scripts/publish-rust.sh
+++ b/scripts/publish-rust.sh
@@ -28,15 +28,14 @@ old_version=$(readCargoVariable version "Cargo.toml")
 package_name=$(readCargoVariable name "Cargo.toml")
 tag_name="${package_name//solana-/}"
 
-# Publish the new version and commit + tag the repo change locally.
-# The push is deliberately deferred (--no-push) so it can be retried against a
-# freshly synced master below; a concurrent publish of another crate may move
-# master forward between our rebase and our push, which would otherwise reject
-# the push non-fast-forward *after* the crate is already live on crates.io.
+# Publishing to crates.io is irreversible, so it is done last. If the tag push
+# loses a race with a concurrent release, the run fails before anything is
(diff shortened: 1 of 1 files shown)


joncinque (commented):
I'm not sure if this is the right approach, since it's possible for the published crate to not correspond to any tag in the repo, correct?

I don't know if there is an ideal solution, which is why I never opted to change anything in these flows. You always have a chance to run into a bad situation unfortunately.

We could potentially do things slightly differently, by instead doing:
* bump the version
* run `cargo package` to create the crate to be published, make sure it works
* try to push the tag. if it works go to next step. otherwise, rebase and go back to the previous step
* publish the package to crates

This makes sure that every tag corresponds to published crates, but it's also much more complicated. What do you think?

kskalski:
Agreed, we should just make the step that is hardest to reverse / fix be the last one. 

I'm not sure if this is too complicated - we can still use the existing `cargo` / `git` commands and the implementation is actually pretty simple, don't even need a retry loop, which can be left to the user, since most failures in the middle of publish are leaving no persistent trace.

joncinque (approved):
Well, that wasn't so bad! Looks great to me, thanks

PR #855, opened 2026-07-30
feat(loader-v3-interface): read v7.0 wire format on serde/bincode path

kskalski:
#### Problem
solana-v3-interface should preserve backwards compatibility even when used without wincode feature.

#### Summary of Changes
Accept legacy tag-only Upgrade/Close/DeployWithMaxDataLen payloads (no trailing bool) via serde deserialize_with + default_on_eof, mirroring the wincode OptionalTrailingBool behavior.

### DIFF
--- a/loader-v3-interface/src/instruction.rs
+++ b/loader-v3-interface/src/instruction.rs
@@ -111,6 +111,10 @@ pub enum UpgradeableLoaderInstruction {
         /// Optional on the wire: when the trailing byte is absent, this
         /// decodes to `true`.
         #[cfg_attr(feature = "wincode", wincode(with = "OptionalTrailingBool<true>"))]
+        #[cfg_attr(
+            feature = "serde",
+            serde(deserialize_with = "deserialize_optional_trailing_bool::<_, true>")
+        )]
         close_buffer: bool,
     },
 
@@ -140,6 +144,10 @@ pub enum UpgradeableLoaderInstruction {
         /// Optional on the wire: when the trailing byte is absent, this
         /// decodes to `true`.
         #[cfg_attr(feature = "wincode", wincode(with = "OptionalTrailingBool<true>"))]
+        #[cfg_attr(
+            feature = "serde",
+            serde(deserialize_with = "deserialize_optional_trailing_bool::<_, true>")
+        )]
         close_buffer: bool,
     },
 
@@ -173,6 +181,10 @@ pub enum UpgradeableLoaderInstruction {
         /// Optional on the wire: when the trailing byte is absent, this
         /// decodes to `false`.
         #[cfg_attr(feature = "wincode", wincode(with = "OptionalTrailingBool<false>"))]
+        #[cfg_attr(
+            feature = "serde",
+            serde(deserialize_with = "deserialize_optional_trailing_bool::<_, false>")
+        )]
         tombstone: bool,
     },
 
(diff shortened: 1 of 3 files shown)


loader-v3-interface/src/instruction.rs
```diff
@@ -173,6 +181,10 @@ pub enum UpgradeableLoaderInstruction {
         /// Optional on the wire: when the trailing byte is absent, this
         /// decodes to `false`.
```
kskalski, line 182: Also double-checked the write path -> old code deserialization
https://github.com/anza-xyz/agave/blob/698ac85e1b98212a51747609ca6d5abf2ba4583e/programs/bpf_loader/src/lib.rs#L157
which goes to https://github.com/anza-xyz/solana-sdk/blob/dca44b61a9c36ecb130d05df3ce49abd98a2eaff/bincode/src/lib.rs#L17 allowing trailing bytes (wincode also ignores trailing bytes unless using `deserialize_exact` API)
- we should be fine with always emitting the extra bytes in new version and old validators should ignore them

kskalski:
I think getting this compatibility fix in will allow us to remove most of remaining bincode uses in agave, as loader-v3 comes up as the only blocker for most of the migrations now. :)

loader-v3-interface/src/instruction.rs
```diff
@@ -790,3 +827,76 @@ mod tests {
         );
     }
 }
+
+/// Backward-compatibility of the serde/bincode deserializer with the v7.0 wire
+/// format, in which `DeployWithMaxDataLen`, `Upgrade`, and `Close` carried no
+/// trailing `bool`. Mirrors the wincode `legacy_*` tests above.
+#[cfg(all(test, feature = "serde"))]
+mod serde_tests {
```
buffalojoec: I actually think these `cfg(feature = "wincode"/"serde")` aren't used in the tests.

Can we just merge these three cases with their existing wincode counterparts and remove the `cfg(featre = "wincode")` restriction on `mod tests`?

This way, we can see these layouts immediately side-by-side, and they can run on the same buffer.

loader-v3-interface/src/instruction.rs
buffalojoec: Yep, I also checked this out as well. One thing we could do is write a test for `with_limit` and make sure using `PACKET_DATA_SIZE` for the limit doesn't produce divergent behavior between wincode and bincode.

I think it's slightly gratuitous, though, since the optional trailing instructions are fixed size, well below `PACKET_DATA_SIZE`.

loader-v3-interface/src/instruction.rs
```diff
@@ -263,6 +275,31 @@ unsafe impl<C: ConfigCore, const DEFAULT: bool> SchemaWrite<C> for OptionalTrail
...
+/// serde/bincode counterpart of `OptionalTrailingBool<true>`: decodes a
+/// trailing `bool` that is absent in the legacy wire format as `true`. A clean
+/// end-of-input yields `true`; a present byte is decoded as its value.
+#[cfg(feature = "serde")]
+fn deserialize_optional_trailing_bool_true<'de, D>(deserializer: D) -> Result<bool, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let value: DefaultTrueBool = solana_serde::default_on_eof(deserializer)?;
+    Ok(value.0)
```
buffalojoec: I think you can just use `ignore_eof_error` directly and just pass `true` and then we don't need the struct. Does that work? Or maybe we can pass `OptionalTrailingBool<true>`? Just trying to avoid the new type if possible.

buffalojoec (commented):
Nice, this is great! Thanks for handling this. This should allow us to bump the interface on Agave.

I left a few suggestions about tightening things up, but the implementation is performing correctly.

loader-v3-interface/src/instruction.rs
kskalski: Added both features as crate's self dev-dependency, to make that work this way, I guess this is fine

loader-v3-interface/src/instruction.rs
kskalski: Cool, that left out just the adapter function, looks cleaner now.

loader-v3-interface/src/instruction.rs
buffalojoec: Actually I think we can just write a few tests with `solana_bincode::limited_deserialize` if we want to be extra safe. It's the same deserializer all the builtins use, and it sets up all the bincode options for us. That way we can also make sure the error handlings are the same for both wincode and bincode.

buffalojoec (commented):
Nice, this looks great. Just the open question about `limited_deserialize`, but I can test it in follow-up, too, if you want this unblocked.

buffalojoec (approved):

loader-v3-interface/src/instruction.rs
kskalski: I thought you had in mind a test in agave that would check the exact payload, types, options used there with a `PACKET_DATA_SIZE` limit. I guess we can settle for something in a follow up.

buffalojoec: Yeah, I'll take care of it in the Loader program, thanks!

PR #857, opened 2026-08-03
[poseidon] Use no-std when `target_os = "solana"` or `target_arch = "bpf"`

### LINKED ISSUE #858: [poseidon] Implicitly links std on the Solana target
### Problem

`solana-poseidon` uses a syscall when compiled for Solana and does not require the Rust standard library on that path.
However, the crate does not currently declare `#![no_std]`, so Rust implicitly links `std` even when `target_os = "solana"`.

This causes strict `no_std` Solana programs, such as those built with Quasar, to fail with:
```
error[E0152]: found duplicate lang item `panic_impl`
```

The program provides its own panic handler, while `std`, pulled in through `solana-poseidon`, provides another one.
This can be reproduced by importing `solana-poseidon` into a `no_std` program built with `Quasar`.

### Fix

Declare the crate as `no_std` only for the Solana target:
```
#![cfg_attr(target_os = "solana", no_std)]
```
The Solana implementation already uses only core-compatible functionality and the `sol_poseidon` syscall.
`thiserror` is already configured with default features disabled in the workspace and is compatible with `no_std`.

gwalen:
`solana-poseidon` already has a syscall-only implementation for Solana, but the crate does not declare `no_std`. 
Rust implicitly links `std`, which conflicts with panic handlers provided by strict `no-std` programs like `Quasar`
This change disables `std` only on `target_os = "solana";` host behavior is unchanged.

Fixes #858

### DIFF
--- a/poseidon/Cargo.toml
+++ b/poseidon/Cargo.toml
@@ -17,12 +17,12 @@ rustdoc-args = ["--cfg=docsrs"]
 [dependencies]
 thiserror = { workspace = true }
 
-[target.'cfg(not(target_os = "solana"))'.dependencies]
+[target.'cfg(any(target_os = "solana", target_arch = "bpf"))'.dependencies]
+solana-define-syscall = { workspace = true }
+
+[target.'cfg(not(any(target_os = "solana", target_arch = "bpf")))'.dependencies]
 ark-bn254 = { workspace = true }
 light-poseidon = { workspace = true }
-
-[target.'cfg(target_os = "solana")'.dependencies]
-solana-define-syscall = { workspace = true }
 
 [lints]
 workspace = true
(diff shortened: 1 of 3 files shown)


joncinque (commented):
Thanks for your contribution!

I'm on-board with the idea, since it'll take a long time to use no-std dependencies for the fallback implementation of poseidon hashing.

However, let's support it totally, by gating the line you added, and all other `cfg`s, by also supporting `bpf`, ie changing:

```
target_os = "solana"
```

to

```
any(target_os = "solana", target_arch = "bpf")
```

Then we can also add `solana-poseidon` as a no-std crate to check in https://github.com/anza-xyz/solana-sdk/blob/58ca02e893b7553b71ba1b702f4f8d0565da64a6/scripts/check-no-std.sh#L10

Does that make sense?

gwalen:
Definitely make sense, I was also thinking about it.
Will add those changes and push the commit soon

joncinque (approved):
Thanks for your contribution!

PR #863, opened 2026-08-06
feat(nonce-account): use wincode state trait

kskalski:
3743f36c — **read nonce state through `StateMutWincode`**
Replaces the ad-hoc `wincode::deserialize` in `verify_nonce_account` with the trait; `lamports_per_signature_of` now follows the feature too, where before it stayed on bincode even with `wincode` on.

c9a856d2 — **make bincode an optional (default) feature**
`solana-account/bincode` + `solana-nonce/serde` move behind a `bincode` feature that's in `default`, so a wincode-only consumer drops `bincode`, `serde`, `serde_bytes`, `serde_derive` and `solana-sysvar`

### DIFF
--- a/nonce-account/Cargo.toml
+++ b/nonce-account/Cargo.toml
@@ -15,15 +15,20 @@ all-features = true
 rustdoc-args = ["--cfg=docsrs"]
 
 [features]
+default = ["bincode"]
+bincode = ["solana-account/bincode", "solana-nonce/serde"]
 wincode = ["solana-account/wincode", "solana-nonce/wincode"]
 
 [dependencies]
-solana-account = { workspace = true, features = ["bincode"] }
+solana-account = { workspace = true }
 solana-hash = { workspace = true }
-solana-nonce = { workspace = true, features = ["serde"] }
+solana-nonce = { workspace = true }
 solana-sdk-ids = { workspace = true }
 
 [dev-dependencies]
 bincode = { workspace = true }
+# Tests build accounts with the bincode constructors whichever codec the lib is built with.
+solana-account = { workspace = true, features = ["bincode"] }
 solana-fee-calculator = { workspace = true }
+solana-nonce = { workspace = true, features = ["serde"] }
 solana-pubkey = { workspace = true, features = ["std"] }
--- a/nonce-account/Cargo.toml
+++ b/nonce-account/Cargo.toml
@@ -15,14 +15,13 @@ all-features = true
 rustdoc-args = ["--cfg=docsrs"]
 
 [features]
-wincode = ["dep:wincode", "solana-nonce/wincode"]
+wincode = ["solana-account/wincode", "solana-nonce/wincode"]
 
 [dependencies]
 solana-account = { workspace = true, features = ["bincode"] }
 solana-hash = { workspace = true }
 solana-nonce = { workspace = true, features = ["serde"] }
 solana-sdk-ids = { workspace = true }
-wincode = { workspace = true, optional = true }
 
 [dev-dependencies]
 bincode = { workspace = true }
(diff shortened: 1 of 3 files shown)


nonce-account/Cargo.toml
```diff
@@ -15,16 +15,20 @@ all-features = true
 rustdoc-args = ["--cfg=docsrs"]
 
 [features]
-wincode = ["dep:wincode", "solana-nonce/wincode"]
+default = ["bincode"]
```
kskalski, line 18: this attempts to make this change a minor only eligible, though strictly speaking we are now adding knob that will remove some APIs under `default-features = false`, which could be considered breaking

kskalski:
Any chance for a review here? Turns out `solana-nonce-account` is the only crate that forces `bincode` feature on `solana-account` in agave and thus blocking a large bunch of migrations.

nonce-account/src/lib.rs
```diff
@@ -1,18 +1,28 @@
...
+use {
+    solana_hash::Hash,
+    solana_nonce::{state::Data, versions::Versions},
     std::cell::RefCell,
 };
 
+#[cfg(any(feature = "bincode", feature = "wincode"))]
 pub fn create_account(lamports: u64) -> RefCell<AccountSharedData> {
+    // bincode's `new_data_with_space` is inherent while wincode's is a trait method, so this
+    // shadows to bincode whenever that codec is compiled in. The encodings agree.
```
joncinque, line 25: This is a bit strange, since every other function will prefer the wincode version if both are enabled, correct? It would be better to stay consistent and use wincode everywhere if both features are enabled.

On the other hand, people probably won't enable both features, so maybe it's ok

nonce-account/Cargo.toml
joncinque: Feel free to break and publish a major version if you prefer, this crate is a leaf in the dependency graph of the sdk repo, and has very few dependents: https://crates.io/crates/solana-nonce-account/reverse_dependencies

nonce-account/Cargo.toml
```diff
@@ -15,16 +15,20 @@ all-features = true
...
+solana-account = { workspace = true }
 solana-hash = { workspace = true }
-solana-nonce = { workspace = true, features = ["serde"] }
+solana-nonce = { workspace = true }
 solana-sdk-ids = { workspace = true }
-wincode = { workspace = true, optional = true }
 
 [dev-dependencies]
 bincode = { workspace = true }
+# Tests build accounts with the bincode constructors whichever codec the lib is built with.
```
joncinque, line 30: The way this is setup, we won't currently run the wincode tests. If we decide to make a breaking change and *not* have bincode as a default feature, we should probably enable the wincode feature on this crate

joncinque (commented):
This looks great to me! Just a few small points

2501babe:
confirmed all `Versions` variants are byte-equal in both directions just to safe

nonce-account/src/lib.rs
2501babe: fwiw this function is almost certainly only used for tests

2501babe (approved):

nonce-account/Cargo.toml
2501babe: imo hurts nothing to land this and also follow up removing the default in a major

nonce-account/Cargo.toml
kskalski: We should add nonce-account as dev dependency with wincode enabled then I guess.

nonce-account/Cargo.toml
kskalski: I think I will merge this one and publish minor, then I can follow-up with switch over for a major so things are cleaned up better by default

PR #864, opened 2026-08-06
ci: Use environment variables to bind inputs

joncinque:
#### Problem

It's possible to do script injection attacks on certain workflow inputs. In particular, the `package_path` in the publish job is just a string in this repo, and not a choice, so it's possible for someone with write permission on the repo to do something nasty.

#### Summary of changes

Bind all inputs to environment variables.


yihau (approved):
🔥

PR #867, opened 2026-08-08
svm-transaction: prep-work for ALT relaxation

2501babe:
* move as much SVMMessage functionality as possible into SVMStaticMessage
* add signed/unsigned account counts to SVMStaticMessage
* add is_requested_writable() to SVMStaticMessage with blanket impl
* move SVMTransaction functions into new SVMStaticTransaction and make the former an automatic alias
* add static traits to SanitizedVersionedTransaction

in other words this reduces `SVMMessage` to nothing but `account_keys()`, `is_writable()`, and `get_durable_nonce()`

i cant promise this will be my _only_ sdk commit, since its honestly hard to keep track of all the different transaction wrapper types and which of them i might truly desire to use where, but as best i can tell this gives me what i need to start refactoring agave. other than `is_requested_writable()` and the two account counters (which are impled directly so that we can have one impl of of the more complex function) everything is just loosening constraints. `is_invoked()` and `is_signer()` are safe to be static because program ids and signers are not allowed to come from ALTs

`is_requested_writable()` is so that i can impl a static durable nonce helper in `runtime/`, which i am _not_ adding to any SVM traits because it could affect consensus if used in the wrong place. my basic idea is either:
1. i can put a slightly _looser_ nonce check in `check_transactions.rs` for leader and replay, by ignoring the fact that a nonce account may be demoted to read-only. the real check would be deferred to svm. this is consensus-safe as long as replay single-batches
2. i can put a slightly _tighter_ nonce check in the leader-only check function which discards a nonce transaction which uses a reserved key as a nonce or uses the nonce as a program id. the latter is not consensus-safe for replay because if LoaderV3 is present (which may be via ALT), the nonce would not be demoted, and if the nonce is valid it would become fee-only. in this case i would remove replay nonce checking from runtime entirely and again rely on svm, which is consensus-safe as long as replay single-batches

in both cases the goal is to allow all the check functions to be bounded by `Static` traits. then i would aim to be `Static` upstream of that on the leader side and use a list of loaded addresses produced in svm in all post-execution runtime work. this remains fairly speculative which is why im trying to land prs like this, ie that are nice refactors on their own even if ALT relaxation hits some major speedbumps later

the idea is land this, major release, bump version and rebase in https://github.com/anza-xyz/agave-sdk/pull/42, land that, major release, bump version and rebase in https://github.com/anza-xyz/agave/pull/14446, land that, then actually get to work. i feel like it is safer to speculatively land sdk changes first rather than to continuously rebase across three repos that may change out from under me, since the sdk/agave-sdk changes are straightforward improvements even if no agave pr besides the fixup for the interface change ever landed

the agave-sdk pr fails ci because it depends on this one. however, the _agave_ pr passes, because i pointed its sdk and agave-sdk deps to the other two prs' branches on github. this illustrates that everything compiles happily across the three repos

### DIFF
--- a/svm-transaction/src/svm_message.rs
+++ b/svm-transaction/src/svm_message.rs
@@ -12,6 +12,7 @@ use {
 
 mod sanitized_message;
 mod sanitized_transaction;
+mod sanitized_versioned_transaction;
 // inlined to avoid solana-nonce dep
 #[cfg(test)]
 static_assertions::const_assert_eq!(
@@ -20,7 +21,8 @@ static_assertions::const_assert_eq!(
 );
 const NONCED_TX_MARKER_IX_INDEX: u8 = 0;
 
-pub trait SVMStaticMessage {
+// - Debug to support legacy logging
+pub trait SVMStaticMessage: Debug {
     /// Get the transaction version.
     fn version(&self) -> TransactionVersion;
 
@@ -42,6 +44,12 @@ pub trait SVMStaticMessage {
     /// Returns the number of requested write-locks in this message.
     /// This does not consider if write-locks are demoted.
     fn num_write_locks(&self) -> u64;
+
+    /// Return the number of readonly signed static account keys.
+    fn num_readonly_signed_static_accounts(&self) -> u8;
+
+    /// Return the number of readonly unsigned static account keys.
+    fn num_readonly_unsigned_static_accounts(&self) -> u8;
 
     /// Return the recent blockhash.
     fn recent_blockhash(&self) -> &Hash;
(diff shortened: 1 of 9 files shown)


apfitzge:
> this is consensus-safe as long as replay single-batches

This is the case right now, but it might change in the future. Is there no way to do this w/ batched txns?

apfitzge (approved):
LGTM

2501babe:
>This is the case right now, but it might change in the future. Is there no way to do this w/ batched txns?

the nonce checks in runtime are not necessary for consensus at all, the trick is just whether what changes are the _same_ under consensus rules across two versions without feature gate. with single batches it doesnt matter but with real batches there are weird edge cases where runtime would reject something that svm accepts, so a different consensus result

PR #870, opened 2026-08-10
feat(nonce-account)!: drop bincode from default features

kskalski:
Follow-up to #863, which put both codecs behind features but kept `bincode` in `default` to avoid changing what existing consumers get. Nothing is enabled by default now, so a consumer picks its codec explicitly and a wincode-only one stops paying for `bincode`, `serde`, `serde_bytes`, `serde_derive` and `solana-sysvar`.

With no default features the unit tests would compile down to just `test_get_system_account_kind`, the one test that needs no codec, so the crate takes a dev-dependency on itself with `wincode` enabled. Test builds therefore exercise the wincode path while still constructing their fixtures with the bincode constructors, which the existing dev-dependencies keep available.

This is a breaking change: `default-features = false` was already a no-op after #863, but a consumer on plain defaults now loses the codec-dependent half of the API until it names a feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3930,6 +3930,7 @@ dependencies = [
  "solana-fee-calculator",
  "solana-hash",
  "solana-nonce",
+ "solana-nonce-account",
  "solana-pubkey",
  "solana-sdk-ids",
 ]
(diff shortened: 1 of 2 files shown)


joncinque (approved):
Looks great!

PR #875, opened 2026-08-12
feat(account): add AccountSharedData::new_with_data

kskalski:
#### Problem
Creating `AccountSharedData` with initialized data is inconvenient
```rust
    let mut account = AccountSharedData::new(1, required_data_len, &sysvar::id());
    wincode::serialize_into(account.data_as_mut_slice(), value).unwrap();
```
becomes a pattern after deprecating / removing sysvar initialization functions

#### Summary of Changes
Mirrors `Account::new_with_data` so callers holding already-encoded bytes can build an `AccountSharedData` without the zero-fill allocation of new() plus a separate write.

### DIFF
--- a/account/src/account.rs
+++ b/account/src/account.rs
@@ -356,6 +356,21 @@ impl AccountSharedData {
             rent_epoch: Epoch::default(),
         }
     }
+    /// Construct an `AccountSharedData` from already-encoded `data`, taking ownership of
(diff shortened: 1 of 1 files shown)


joncinque (commented):
This makes sense to me! I'll let @grod220 give the final signoff if this is useful

grod220 (approved):
LGTM

PR #876, opened 2026-08-12
Adds `InstructionError::BailOut`

Lichtso:
See discussion: https://github.com/anza-xyz/agave/pull/13860#discussion_r3665634659

### DIFF
--- a/instruction-error/src/lib.rs
+++ b/instruction-error/src/lib.rs
@@ -221,7 +221,9 @@ mod instruction_error_module {
         /// Builtin programs must consume compute units
         BuiltinProgramsMustConsumeComputeUnits,
 
-        /// Block production bailed out
+        /// Block production bailed out.
+        /// This discards transactions to protect the leader and does not propagate to followers.
+        /// Meaning this explicitly excludes transactions from consensus.
         BailOut,
         // Note: For any new error added here an equivalent ProgramError and its
         // conversions must also be added
--- a/instruction-error/src/lib.rs
+++ b/instruction-error/src/lib.rs
@@ -10,7 +10,7 @@ pub use {
     instruction_error_module::*,
     solana_program_error::{
         ACCOUNT_ALREADY_INITIALIZED, ACCOUNT_BORROW_FAILED, ACCOUNT_DATA_TOO_SMALL,
-        ACCOUNT_NOT_RENT_EXEMPT, ARITHMETIC_OVERFLOW, BORSH_IO_ERROR,
+        ACCOUNT_NOT_RENT_EXEMPT, ARITHMETIC_OVERFLOW, BAIL_OUT, BORSH_IO_ERROR,
         BUILTIN_PROGRAMS_MUST_CONSUME_COMPUTE_UNITS, CUSTOM_ZERO, ILLEGAL_OWNER, IMMUTABLE,
         INCORRECT_AUTHORITY, INCORRECT_PROGRAM_ID, INSUFFICIENT_FUNDS, INVALID_ACCOUNT_DATA,
         INVALID_ACCOUNT_DATA_REALLOC, INVALID_ACCOUNT_OWNER, INVALID_ARGUMENT,
@@ -39,7 +39,7 @@ mod instruction_error_module {
         feature = "frozen-abi",
         derive(AbiExample, AbiEnumVisitor, StableAbi, StableAbiSample),
         frozen_abi(
-            abi_digest = "99sLbFrZmSAM4i28P5vPDJFtB6xDgvyT92iEJpKiMPpF",
+            abi_digest = "FeTxh6dMDyYG1EdnenTpe8vpH37xDRvfksy83XKBN671",
             abi_serializer = ["bincode", "wincode"],
             test_roundtrip = "eq_and_wire"
         )
@@ -220,6 +220,9 @@ mod instruction_error_module {
 
         /// Builtin programs must consume compute units
         BuiltinProgramsMustConsumeComputeUnits,
+
+        /// Block production bailed out
+        BailOut,
         // Note: For any new error added here an equivalent ProgramError and its
         // conversions must also be added
     }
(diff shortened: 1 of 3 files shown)


instruction-error/src/lib.rs
```diff
@@ -220,6 +220,9 @@ mod instruction_error_module {
 
         /// Builtin programs must consume compute units
         BuiltinProgramsMustConsumeComputeUnits,
+
+        /// Block production bailed out
+        BailOut,
```
buffalojoec, line 227: Maybe add some comments explaining what this is for and warning maintaners to be careful about using it?

Also, as an idea to further isolate/harden it, we could give it an inner payload, and match on the specific payload we define in Agave. Just an idea, not sure it's necessary, though.

buffalojoec (commented):
Generally works for me, as long as we're all aligned on the design in https://github.com/anza-xyz/agave/pull/13860.

Is the idea to couple it with a `TransactionError::BailOut`? Is just a `TransactionError::BailOut` enough without the instruction variant?

apfitzge (commented):
Straight-forward addition and looks good to me. Will give Joe a chance to give final review/approve, but I think this is necessary for the changes we want.

I added the `breaking` label to the PR since this is a breaking change. We should do a major version bump.

@joncinque - I know breaking changes to these low level crates are annoying, this is more necessary than the previous variant we tried adding recently.
I think last time we tried to add something here we discussed possibly adding non_exhaustive in the next major bump...is that something we'd want to consider now that we have one?

buffalojoec (commented):
LGTM, I think just document the variant so people don't grab it for unrelated purposes, then good to go.

joncinque (commented):
@Lichtso already explained the rationale, so makes sense to me.

Why is this also an `InstructionError` though and not a `TransactionError` only? If a scheduler bails out, wouldn't it only bail out on the full transaction?

With this change as is, it's possible to have `TransactionError::BailOut` and also `TransactionError::InstructionError(index, InstructionError::BailOut)`. When would one be used over the other? 

Looking at the usage in the agave PR, we should be able to get by with just the TransactionError variant, but I might be missing something.

Further, it seems strange for a program to emit `BailOut`, unless we want to allow programs to decide on inclusion of transactions in blocks.

Lichtso:
> Why is this also an InstructionError though and not a TransactionError only? If a scheduler bails out, wouldn't it only bail out on the full transaction?

We need an `InstructionError` since it gets thrown well ... in an instruction. All program runtime errors are `InstructionError`s. Only during transaction loading we can throw `TransactionError` directly, since that is before any instructions.

> get by with just the TransactionError variant, but I might be missing something.

The additional `TransactionError` was added on request of @apfitzge since he did not want to overload / reuse the existing `TransactionError::CommitCancelled`.

Alternatively, we could also just keep the `TransactionError::InstructionError(_, InstructionError::BailOut)` around. I can remove the `TransactionError` again. However, with a `TransactionError` variant, we could also apply this mechanism to transaction loading, not just the program runtime.

> Further, it seems strange for a program to emit BailOut, unless we want to allow programs to decide on inclusion of transactions in blocks.

Good point, I will remove the `ProgramError` variant.

joncinque:
> We need an InstructionError since it gets thrown well ... in an instruction. All program runtime errors are InstructionErrors. Only during transaction loading we can throw TransactionError directly, since that is before any instructions.

Gotcha, I thought the error was only supposed to be thrown at the transaction level, makes sense then.

> I think last time we tried to add something here we discussed possibly adding non_exhaustive in the next major bump...is that something we'd want to consider now that we have one?

The biggest concern for non-exhaustive enums is that they require code that matches on it to include a wildcard match arm: https://doc.rust-lang.org/nightly/std/attribute.non_exhaustive.html

This means that in the code that *does* match exhaustively, we'll need to be doubly careful. I took a quick look in Agave, and I found the following that *don't* match exhaustively:

* runtime status cache: https://github.com/anza-xyz/agave/blob/88fb4364021e180e98f8cfddcafe5a33b9c7e2a9/runtime/src/serde_snapshot/status_cache.rs#L166
* scheduling utils: https://github.com/anza-xyz/agave/blob/88fb4364021e180e98f8cfddcafe5a33b9c7e2a9/scheduling-utils/src/error.rs#L17
* storage proto: https://github.com/anza-xyz/agave/blob/88fb4364021e180e98f8cfddcafe5a33b9c7e2a9/storage-proto/src/convert.rs#L799

Outside of these three usages, we're actually safe to make these enums non-exhaustive. In those exhaustive cases, we just need to add unit tests that actually iterate through all enum variants, ie https://github.com/anza-xyz/solana-sdk/blob/5190ff456079d17b64669bcb5eeac48dd595b91e/system-interface/src/error.rs#L154-L163

All that to say, I'm down to do it! I was worried it would be a non-starter. Any other thoughts?

Lichtso:
Think that because `non_exhaustive` would affect other code sites too, it should be done in a separate PR.

@joncinque and @buffalojoec, is this PR otherwise good to go?

transaction-error/src/lib.rs
```diff
@@ -240,6 +245,8 @@ impl fmt::Display for TransactionError {
              => f.write_str("Program cache hit max limit"),
             Self::CommitCancelled
              => f.write_str("CommitCancelled"),
+            Self::BailOut
+             => f.write_str("BailOut"),
```
buffalojoec, line 249: Any reason why this shouldn't also be `"Block production bailed out"`?

buffalojoec (commented):
> Think that because `non_exhaustive` would affect other code sites too, it should be done in a separate PR.

I think the idea is that we'd only be breaking the crates once. Adding `#[non_exhaustive]` is a breaking change. It can be done in a separate PR, but we shouldn't release these libs without it if that's the direction we want to go in.

transaction-error/src/lib.rs
Lichtso: There is no real consistency here, some are the name of the variant, others their doc comment. Can change it if you want.

buffalojoec (approved):
Approving, but reiterating that if @joncinque wants these enums non-exhaustive, we could do him a favor and make it part of the same release(s). I do think a separate PR is the right way to go, though.

apfitzge:
Yeah 100% non_exhaustive should not be part of same PR, but same release iff we want to also do that.

joncinque (approved):
Yep, let's go ahead with this!

PR #877, opened 2026-08-12
fix(offchain-message): validate signing domain during deserialization

raushan728:
`OffchainMessage::deserialize` never validated that the input actually started with `SIGNING_DOMAIN`, letting bytes from any signing domain deserialize successfully and breaking domain separation.


joncinque:
Sorry, I hadn't noticed this PR, but #907 fixed it, so closing this one

Issue #819, opened 2026-07-10
[ed25519] clean up simd-0376 and disable verify_strick


zz-sol:
prior to [simd-0376](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0376-verify-strict.md), ed25519 verification uses `dalek/verify_strick`
simd-0376 allows to use `zip-215/zebra/verify_non_strick`

currently we support both methods for backward compatibility

at some point we may want to clean up the code and disable dalek verification

PR #880, opened 2026-08-13
fix(signature): use ZIP-215 semantics in `batch_verify` single-item fast path

raushan728:
- `Signature::batch_verify` documents ZIP-215-compatible semantics, matching `verify_non_strict` rather than strict `verify`.
- Its single-item fast path called strict `verify()` instead, so the same signature could pass or fail depending only on batch size.

## Context

- Introduced in 672ad60 (#724), which added `verify_non_strict` and the ZIP-215 docs but left the single-item fast path on the old strict `verify()` call.
- Related: anza-xyz/solana-sdk#819 (broader ZIP-215/verify_strict cleanup, also referenced in `verify_non_strict`'s doc comment).

### DIFF
--- a/signature/src/lib.rs
+++ b/signature/src/lib.rs
@@ -515,6 +515,25 @@ mod tests {
         )));
     }
 
+    #[test]
+    fn test_batch_verify_single_item_rejects_zip215_valid_signature() {
+        let mut signature = [0; SIGNATURE_BYTES];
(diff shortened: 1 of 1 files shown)


zz-sol:
i think we want to use dalek for now until simd-0376 is active. so we will need to wait till simd-0376 to update/merge the batch_verify function

PR #883, opened 2026-08-14
feat(feature-gate-interface): add wincode feature

kskalski:
Provides a bincode-free path for `Feature` account state. wincode shadows bincode when both features are enabled; the two encodings are byte-for-byte identical.

`solana-system-interface/bincode` moves off the dependency line into the `bincode` feature so it is no longer enabled unconditionally.

The account writers go through `StateMut`/`StateMutWincode`, which share method names, so `to_account`/`create_account` need no cfg. The readers keep a raw `deserialize` helper, as those traits are only implemented for `Account`/`AccountSharedData` while `from_account` is generic over `ReadableAccount` and `from_account_info` takes an `AccountInfo`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3557,6 +3557,7 @@ dependencies = [
  "solana-system-interface",
  "strum",
  "strum_macros",
+ "wincode",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


joncinque (approved):
Looks great, thanks!

PR #886, opened 2026-08-14
message: Deprecate `v1::Message::try_compile`

### LINKED ISSUE #871: `v1::Message::try_compile` generates a v1 transaction that will never be successfully processed
The [`v1::Message::try_compile` function](https://github.com/anza-xyz/solana-sdk/blob/9f4a3062bdd41fffcb9c220f798d762c3c1d17e4/message/src/versions/v1/message.rs#L215-L226) uses `TransactionCofig::empty()` as the default transaction config. 

This sets all values to `None` for the transaction config, most importantly `compute_unit_limits` and `loaded_accounts_data_size_limit`. I'm no expert on the runtime code, but from what I can tell [here](https://github.com/anza-xyz/agave/blob/5fb575b301dca7c15f5542805eee06aa5b23b58f/runtime-transaction/src/transaction_meta.rs#L122-L129) these values are unwrapped to `0` and `0`. 

This means that the default compute unit limits and loaded account data size are 0, so every transaction will be rejected by the network when using `try_compile` for v1 transactions.

I wrote a rust test just to confirm this behavior. I was able to point it to the community cluster running alpenglow + agave 4.2 with txv1 enabled. As it is, the transaction fails with error `MaxLoadedAccountsDataSizeExceeded`. If you uncomment the second message declaration, the transaction succeeds. If you don't also include the `.with_compute_unit_limit()` call, you get a `ComputationalBudgetExceeded` error.

Is this known behavior? I think the `try_compile` function is what most users will turn to (since this is the same pattern used for other transaction types), so it should create the transaction config with some sensible defaults. I could be missing something here, so let me know if so; otherwise I'm happy to open a PR to fix!

```rust
#[tokio::test(flavor = "multi_thread")]
async fn tmp_custom_test() {
    let rpc_url = "https://my.rpc.url";
    let keypair_path = "/path/to/.config/solana/id.json";
    let rpc_client = RpcClient::new(rpc_url.into());
    let payer = solana_keypair::read_keypair_file(&keypair_path)
        .expect("Failed to read keypair file");
    let payer_pubkey = payer.pubkey();
    let recipient = Pubkey::new_unique();
    let transfer_amount = 1_000_000;

    let sender_balance_before = rpc_client
        .get_balance(&payer_pubkey)
        .await
        .expect("Failed to get sender balance");
    println!("Sender balance before: {}", sender_balance_before);

    let recent_blockhash = rpc_client
        .get_latest_blockhash()
        .await
        .expect("Failed to get latest blockhash");

    let tx = {
        let instructions = &[transfer(&payer.pubkey(), &recipient, transfer_amount)];

        // default transaction config fails
        let message = v1::Message::try_compile(&payer.pubkey(), instructions, recent_blockhash)
            .expect("v1 message should compile");

        // uncomment and this transaction config works fine
        // let message = v1::Message::try_compile_with_config(
        //     &payer.pubkey(),
        //     instructions,
        //     recent_blockhash,
        //     TransactionConfig::empty()
        //         .with_loaded_accounts_data_size_limit(1024)
        //         .with_compute_unit_limit(200_000),
        // )
        // .expect("v1 message should compile");

        VersionedTransaction::try_new(VersionedMessage::V1(message), &[payer])
            .expect("v1 transaction should sign")
    };
    let sig = rpc_client
        .send_and_confirm_transaction(&tx)
        .await
        .expect("Failed to send transaction");

    println!("Transaction sent with signature: {}", sig);
    let sender_balance_after = rpc_client
        .get_balance(&payer_pubkey)
        .await
        .expect("Failed to get sender balance");
    println!("Sender balance after: {}", sender_balance_after);
    let recipient_balance = rpc_client
        .get_balance(&recipient)
        .await
        .expect("Failed to get recipient balance");
    println!("Recipient balance: {}", recipient_balance);
}
```

joncinque:
#### Problem

As pointed out in #871, the `try_compile` function on v1 Message creates a transaction that won't execute because its transaction configuration is unusable.

#### Summary of changes

Deprecate the function.

Fixes #871

### DIFF
--- a/message/src/versions/v1/message.rs
+++ b/message/src/versions/v1/message.rs
@@ -212,6 +212,7 @@ impl Message {
     /// # create_v1_tx(&client, instruction, &payer)?;
     /// # Ok::<(), anyhow::Error>(())
     /// ```
+    #[deprecated(since = "4.6.0", note = "Use `try_compile_with_config`")]
     pub fn try_compile(
         payer: &Address,
         instructions: &[Instruction],
--- a/transaction/src/sanitized.rs
+++ b/transaction/src/sanitized.rs
@@ -466,8 +466,13 @@ mod tests {
             &[1, 2, 3],
             vec![AccountMeta::new(payer.pubkey(), true)],
         );
-        let message =
-            v1::Message::try_compile(&payer.pubkey(), &[instruction], Hash::new_unique()).unwrap();
+        let message = v1::Message::try_compile_with_config(
+            &payer.pubkey(),
+            &[instruction],
+            Hash::new_unique(),
+            v1::TransactionConfig::empty(),
+        )
+        .unwrap();
         let versioned_tx =
             VersionedTransaction::try_new(VersionedMessage::V1(message), &[&payer]).unwrap();
 


febo (approved):
Looks great!

apfitzge (approved):

PR #887, opened 2026-08-15
fix: replace `unimplemented!()` panics with spanned `syn::Error` in proc-macro derives

raushan728:
Several `unimplemented!()` branches in `frozen-abi-macro` and `sdk-macro` panicked on unsupported derive shapes instead of emitting a `compile_error!`. Converted the reachable ones to spanned `syn::Error`, and the remaining unreachable/defensive branches for consistency.


raushan728:
Hi @buffalojoec @joncinque i'm a new contributor here and i have few small PRs open. I’d appreciate it if you could take a look at them.
#885 , #881, #880, #877

kskalski (approved):
LTGM. thanks!

PR #888, opened 2026-08-16
feat: add fallible constructors for `borsh`, `bincode`, and `wincode` instruction serialization

raushan728:
`Instruction::new_with_borsh/bincode/wincode` and `CompiledInstruction::new` (wincode) unwrapped serialization results, giving callers no way to handle failures. Added `try_new_with_*`/`try_new` siblings returning the native serializer error; existing infallible constructors now delegate to these with `.expect(...)`, preserving current behavior exactly.

`system-interface` was left untouched - its `create_instruction` helper is private, and `SystemInstruction` is a constrained first-party enum where serialization failure is practically unreachable, so propagating fallibility there would add API surface without real benefit.

### DIFF
--- a/message/src/compiled_instruction.rs
+++ b/message/src/compiled_instruction.rs
@@ -49,17 +49,34 @@ impl Sanitize for CompiledInstruction {}
 
 impl CompiledInstruction {
     #[cfg(feature = "wincode")]
+    /// Create a new instruction from a value, encoded with [`wincode`].
+    ///
+    /// This is the infallible version of [`CompiledInstruction::try_new`], and will
+    /// panic if serialization fails.
     pub fn new<T: wincode::Serialize<Src = T>>(
         program_ids_index: u8,
         data: &T,
         accounts: Vec<u8>,
     ) -> Self {
-        let data = wincode::serialize(data).unwrap();
-        Self {
+        Self::try_new(program_ids_index, data, accounts)
+            .expect("Failed to serialize instruction data")
+    }
+
+    #[cfg(feature = "wincode")]
+    /// Fallibly create a new instruction from a value, encoded with [`wincode`].
+    ///
+    /// [`wincode`]: https://docs.rs/wincode/latest/wincode/
+    pub fn try_new<T: wincode::Serialize<Src = T>>(
+        program_ids_index: u8,
+        data: &T,
+        accounts: Vec<u8>,
+    ) -> Result<Self, wincode::Error> {
+        let data = wincode::serialize(data)?;
+        Ok(Self {
             program_id_index: program_ids_index,
             accounts,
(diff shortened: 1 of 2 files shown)


apfitzge:
What's the use-case here? I could be wrong but generally these are used when a program you've written are creating txns, if you're creating the ixs incorrectly then it's unclear to me how you'd actually handle the failure beyond just failing.

Still it does seem like it's a reasonable interface change, @joncinque any thoughts?

raushan728:
In most simple cases where a program constructs an instruction directly (e.g. hardcoded data, well-known types).

The main motivation is for library/SDK authors and long running services (RPC nodes, wallet backends, indexers, bots building transactions from user-supplied or dynamic data) that want to avoid a hard panic in a library function. A panic can unwind through code that isn't expecting it
and, depending on context, take down more than just the failed transaction building call (e.g. a request handler, a batch job processing many unrelated instructions). A `Result` lets callers propagate the error through their own error types. log it skip that one instruction and continue a batch, or return a structured error to a caller instead of crashing..

message/src/compiled_instruction.rs
```diff
@@ -49,17 +49,34 @@ impl Sanitize for CompiledInstruction {}
...
+
+    #[cfg(feature = "wincode")]
+    /// Fallibly create a new instruction from a value, encoded with [`wincode`].
+    ///
+    /// [`wincode`]: https://docs.rs/wincode/latest/wincode/
+    pub fn try_new<T: wincode::Serialize<Src = T>>(
+        program_ids_index: u8,
+        data: &T,
+        accounts: Vec<u8>,
+    ) -> Result<Self, wincode::Error> {
```
buffalojoec, line 73: You can tighten this and the `Instruction` one to `wincode::WriteError`.

buffalojoec (commented):
I mean, generally the change is clean, but the internal step that can panic is a one-liner, so anyone who can't afford to panic can just call `T::serialize(data)` themselves.

I don't think the motivation is really strong here.

raushan728:
> I mean, generally the change is clean, but the internal step that can panic is a one-liner, so anyone who can't afford to panic can just call `T::serialize(data)` themselves.
> 
> I don't think the motivation is really strong here.

That's a fair point, and I think it undercuts the version of my earlier argument.

I don't have a strong case beyond ergonomics/consistency (matching the `try_new` pattern used elsewhere) and this being purely additive/non-breaking. If that's not compelling enough on its own, I'm fine narrowing this PR or closing it.

joncinque (commented):
To be honest, I'd prefer removing the existing creators rather than adding more, but that'll be a longer process. Unless there's a demonstrated need, we should avoid landing this PR.

PR #889, opened 2026-08-17
fix(transaction): require exact legacy signature count

apfitzge:
### Problem
- legacy transaction does not currently require exact equality of signature count for sanitization

### Summary of Changes
- require exact signature count for legacy Transaction

### DIFF
--- a/transaction/src/lib.rs
+++ b/transaction/src/lib.rs
@@ -250,8 +250,12 @@ impl solana_frozen_abi::rand::prelude::Distribution<Transaction>
 
 impl Sanitize for Transaction {
     fn sanitize(&self) -> Result<(), SanitizeError> {
-        if self.message.header.num_required_signatures as usize > self.signatures.len() {
+        let num_required_signatures = usize::from(self.message.header.num_required_signatures);
+        if num_required_signatures > self.signatures.len() {
             return Err(SanitizeError::IndexOutOfBounds);
+        }
+        if num_required_signatures < self.signatures.len() {
+            return Err(SanitizeError::InvalidValue);
         }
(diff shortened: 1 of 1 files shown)


transaction/src/lib.rs
```diff
@@ -250,9 +250,13 @@ impl solana_frozen_abi::rand::prelude::Distribution<Transaction>
...
 impl Sanitize for Transaction {
     fn sanitize(&self) -> Result<(), SanitizeError> {
-        if self.message.header.num_required_signatures as usize > self.signatures.len() {
+        let num_required_signatures = usize::from(self.message.header.num_required_signatures);
+        if num_required_signatures > self.signatures.len() {
             return Err(SanitizeError::IndexOutOfBounds);
         }
+        if num_required_signatures < self.signatures.len() {
+            return Err(SanitizeError::InvalidValue);
+        }
```
tao-stones: nit: why not using `match` like VersionedTransction [does](https://github.com/anza-xyz/solana-sdk/blob/master/transaction/src/versioned/mod.rs#L260-L264)?

tao-stones (commented):
Looks good to me to bring it up to VersionedTransaction's sanitization. Just want to point out it breaks existing behavior for legacy that allows extra signatures - even it is unwarranted and should not exist.

joncinque (dismissed):
Looks good to me!

transaction/src/lib.rs
apfitzge: 090bc083

tao-stones (approved):
lgtm

joncinque (approved):

PR #890, opened 2026-08-17
fix(transaction): sanitize before signature verification

apfitzge:
### Problem
- signature verification interface is dangerous because it can return true when signature/pubkey count do not match required values

### Summary of Changes
- call `sanitize` before verifying signatures

### DIFF
--- a/transaction/src/lib.rs
+++ b/transaction/src/lib.rs
@@ -1025,8 +1025,10 @@ impl Transaction {
     ///
     /// # Errors
     ///
-    /// Returns [`TransactionError::SignatureFailure`] on error.
+    /// Returns [`TransactionError::SanitizeFailure`] if the transaction is malformed, or
+    /// [`TransactionError::SignatureFailure`] if any signature is invalid.
     pub fn verify(&self) -> TransactionResult<()> {
+        self.sanitize()?;
         let message_bytes = self.message_data();
         if !self
             ._verify_with_results(&message_bytes)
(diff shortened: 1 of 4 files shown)


apfitzge:
Built on top of #889

transaction/src/lib.rs
```diff
@@ -1309,6 +1313,50 @@ mod tests {
         assert_eq!(tx.sanitize(), Err(SanitizeError::IndexOutOfBounds));
     }
 
+    #[test]
+    fn test_verify_sanitizes_transaction() {
```
DSharifi, line 1317: Minor:
We're testing multiple things here. Would be better if we make the tests more focused and split them up.

DSharifi (dismissed):

transaction/src/versioned/mod.rs
tao-stones, line 315: this function still exposes the unsafe behavior PR is fixing, does it? If we can't change this function signature to return `Vec<TransactionResult<bool>>`, perhaps should doc it to day that caller must sanitize first and should not treat all-true vec as transaction is valid?

transaction/src/lib.rs
tao-stones, line 1070: same as `versioned` comment above.

tao-stones, line 1075: think about it, one can argue that signature/static-key counts should be validated as part of `_verify_with_results()`, so maybe:
1. move `self.sanitize()?;` into this function
2. change this function signature to return `Vec<TransactinoResult<bool>>` (easier since it's private helper)
wdyt?

transaction/src/versioned/mod.rs
apfitzge: Yeah this fn is definitely bad but can't be fixed without a breaking change.

truly this function should have never existed at all.
I can add documentation.

joncinque: We'll need to do a breaking release in this crate due to #876, so how about breaking these?

apfitzge: Opting to remove this - either direction would be a breaking change, and this shouldn't have been exposed in the first place. Anyone who wants to validate signatures individually is free to do so by iterating. There's no reason we should provide a footgun for it.

Removed here: de8716a1e53c3e5e799bb3660ea75e59a5d1fa67

tao-stones (approved):
looks good to me to remove `verify_with_results()` from api; maybe wait for Jon before merge?

joncinque (approved):
Looks good to me! I've put in https://github.com/anza-xyz/agave/pull/14856 to remove usage in Agave

PR #891, opened 2026-08-17
feat(transaction): verify sanitized versioned transactions

apfitzge:
### Problem
- with #890 we'd end up sanitizing twice because there's no signature verification fn on a `SanitizedVersionedTransaction`

### Summary of Changes
- add `verify_and_hash_message` on `SanitizedVersionedTransaction`

### DIFF
--- a/transaction/src/versioned/sanitized.rs
+++ b/transaction/src/versioned/sanitized.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "verify")]
+use solana_transaction_error::{TransactionError, TransactionResult};
 use {
     crate::versioned::VersionedTransaction, alloc::vec::Vec,
     solana_message::SanitizedVersionedMessage, solana_sanitize::SanitizeError,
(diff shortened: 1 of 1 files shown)


tao-stones (dismissed):
lgtm

transaction/src/versioned/sanitized.rs
```diff
@@ -37,6 +39,42 @@ impl SanitizedVersionedTransaction {
...
+    /// Verifies that all signers have signed the provided serialized message.
+    ///
+    /// The caller must ensure that `message_bytes` is the serialized message for
+    /// this transaction.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`TransactionError::SignatureFailure`] if any signature is invalid.
+    #[cfg(feature = "verify")]
+    pub fn verify_with_serialized_message(&self, message_bytes: &[u8]) -> TransactionResult<()> {
```
joncinque: The version of this function on other transaction types, `_verify_with_results`, is private, since it can be called incorrectly, as the comment points out.

Is this meant to be used externally? If not, I would recommend keeping it private

joncinque (commented):
Looks good, just the comment on visibility of one of the new functions

transaction/src/versioned/sanitized.rs
apfitzge: actually we don't need this in agave, i thought we did but we don't

tao-stones (approved):

joncinque (approved):

PR #895, opened 2026-08-19
[bls12-381] Reject empty pairing slices, improve comments, and add additional utility functions

samkim-crypto:
#### Summary of Changes

I made some smaller improvements on the crate:

- Two behavior changes. `pairing_check` now rejects empty slices rather than reporting the identity — a zero-length batch reporting success is a verification bypass when the slices come from instruction data. `pairing_map` still returns the identity for the empty product to match the syscall. The pairing functions also moved to `Result` with a new `Bls12381Error`, so a malformed batch is distinguishable from a failed verification. Point operations keep `Option`, since the syscall only gives us one bit.
- Filled in API that every consumer would otherwise hand-roll: group generators, `is_infinity` / `is_identity`, borrowing accessors, zero-copy casting that doesn't need the `bytemuck` feature, `Scalar` constructors, and the missing `neg` variants. Also made `neg` two syscalls instead of three.
- Made the docs match what the SIMD actually guarantees. The `_assign` contract no longer promises the output buffer is left untouched on failure, and the SAFETY comments no longer cite SIMD-0388 for the full-write-on-success guarantee — they give the real reasoning instead, and there's a test for it.
- Extended the tests. The existing ones mostly used infinity points, so they would pass against an implementation that returned the identity for everything.

### DIFF
--- a/bls12-381/src/pairing.rs
+++ b/bls12-381/src/pairing.rs
@@ -57,7 +57,7 @@ impl GtElement {
             if i != one_index && self.0[i] != 0 {
                 return false;
             }
-            i += 1;
+            i = i.wrapping_add(1);
         }
         true
     }
--- a/bls12-381/src/pairing.rs
+++ b/bls12-381/src/pairing.rs
@@ -160,23 +160,52 @@ pub fn pairing(
     )
 }
 
-/// Evaluates if the product of pairings equals the identity element.
+/// Evaluates whether the product of pairings equals the identity element.
 ///
-/// Highly efficient for ZK verifiers (e.g., Groth16) as it avoids returning
-/// the raw 576-byte `GtElement` to the caller.
+/// Returns `Some(true)` if `e(P_1, Q_1) * ... * e(P_n, Q_n) == 1`,
+/// `Some(false)` if the product is some other target group element, and
+/// `None` if the check could not be run at all.
+///
+/// # Empty batches
+///
+/// Unlike [`pairing_map`], this rejects an empty batch with `None` rather
+/// than reporting the vacuously-true identity. `pairing_check` is a
+/// verification primitive, and callers typically build these slices from
+/// attacker-controlled instruction data: a zero-length batch that reports
+/// `Some(true)` is a verification bypass, not a mathematical convenience.
+/// If you genuinely want the empty product, call [`pairing_map`] and compare
+/// against [`GtElement::identity`] yourself.
+///
+/// # Correct usage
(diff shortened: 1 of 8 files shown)


bls12-381/src/pairing.rs
```diff
@@ -147,37 +209,67 @@ pub fn pairing_assign(
...
-) -> Option<bool> {
-    let mut out = MaybeUninit::uninit();
-
-    if !pairing_map_assign(g1_points, g2_points, &mut out, endianness) {
-        return None;
+) -> Result<bool, Bls12381Error> {
+    // A check that asserts nothing must not report success. Testing `g1_points`
+    // alone is enough — a mismatched (empty, non-empty) pair already trips the
+    // length check in `pairing_map_assign`.
+    if g1_points.is_empty() {
```
zz-sol, line 275: nit

seem like `pairing_check(&[g1], &[], endianness)` returns `LengthMismatch` and `pairing_check(&[], &[q], ...)` returns `EmptyBatch`

bls12-381/tests/bls12_381.rs
```diff
@@ -32,29 +72,93 @@ fn test_validated_arithmetic_routing() {
...
+fn test_failure_leaves_buffer_unwritten() {
     let valid = G1Point::infinity(Endianness::Little);
     let invalid = G1Point::from_bytes([0xFF; G1_UNCOMPRESSED_POINT_SIZE]);
 
-    let success = valid.add_assign(&invalid, &mut out, Endianness::Little);
-    assert!(!success);
+    let fill = G1Point::from_bytes([0xAB; G1_UNCOMPRESSED_POINT_SIZE]);
+    let mut out = MaybeUninit::new(fill);
+
+    assert!(!valid.add_assign(&invalid, &mut out, Endianness::Little));
```
zz-sol: should we use `add_assign_unchecked` here?

zz-sol (dismissed):

bls12-381/src/pairing.rs
samkim-crypto: Good catch! I added the length checks ahead of the mismatch. The `EmptyBatch` now means both sides are empty.

bls12-381/tests/bls12_381.rs
samkim-crypto: Yep, good call. I switched to `add_assign_unchecked`. Let me know what you think.

zz-sol (approved):

PR #897, opened 2026-08-19
loader-v3-interface: split `close_any` into dedicated builders

### LINKED ISSUE #894: `solana_loader_v3_interface::instruction::close_any()` accepts invalid parameters
Currently, calling [`close_any()`](https://github.com/anza-xyz/solana-sdk/blob/822be4736ed01c4170740c804da3b22c39a2bfe7/loader-v3-interface/src/instruction.rs#L521-L543) with the following parameter shape

```rs
close_any(close_addr, 
    recipient_addr, 
    None,                   // authority_address 
    Some(program_address),
    false)
```

returns an (invalid) `Instruction`, where the provided `program_address` is incorrectly placed in the `AccountMeta` vector at index 2 (where `authority_address` is supposed to live). Hence, this instruction will fail at runtime.

If we are OK with a breaking change, we should make the `authority_address.is_none() && program_address.is_some()` unrepresentable, such as something along the lines of

```rs
enum CloseTarget<'a> { 
    Uninitialized, 
    Buffer { authority: &'a Pubkey },
    Program { authority: &'a Pubkey, program: &'a Pubkey
}

pub fn close_any(
    close_address: &Pubkey,
    recipient_address: &Pubkey,
    target: CloseTarget,
    tombstone: bool,
) -> Instruction {
    ...
}
```

Otherwise, we could at least add a debug assertion to reject `authority_address.is_none() && program_address.is_some()`.

buffalojoec:
#### Problem
The API for Loader V3's `Close` instruction is a bit tricky to use, since we have a few different ways of overloading the instruction. This was already an issue, but it was made worse by the implementation of SIMD-0432, which adds the `tombstone: bool` parameter.

#### Summary of Changes
Simply break `close_any` up into individual instruction builders that are each responsible for one type of closure.

- `close_uninitialized`: Closes an uninitialized account. No tombstone parameter is supported.
- `close_buffer`: Closes a buffer account. No tombstone parameter is supported.
- `close_program`: Closes a program (programdata account), and allows the `tombstone: bool` parameter.
- `reclaim_tombstone`: Directly responsible for crafting the instruction to reclaim a tombstone, in which the program account is an aliased writable signer.

Closes #894

### DIFF
--- a/loader-v3-interface/src/instruction.rs
+++ b/loader-v3-interface/src/instruction.rs
@@ -500,6 +500,8 @@ pub fn set_upgrade_authority_checked(
 }
 
 #[cfg(feature = "wincode")]
+#[deprecated(since = "8.1.1", note = "Use `close_buffer` instead")]
+#[allow(deprecated)]
 /// Returns the instructions required to close a buffer account
 pub fn close(
     close_address: &Pubkey,
@@ -517,6 +519,10 @@ pub fn close(
 }
 
 #[cfg(feature = "wincode")]
+#[deprecated(
+    since = "8.1.1",
+    note = "Use `close_uninitialized`, `close_buffer`, or `close_program` instead"
+)]
 /// Returns the instructions required to close program, buffer, or uninitialized account
 pub fn close_any(
     close_address: &Pubkey,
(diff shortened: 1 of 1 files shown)


buffalojoec:
@plafer do we still want something like this, even after I removed `tombstone: bool` in https://github.com/anza-xyz/solana-sdk/pull/930?

plafer:
I think so yes, since the main problem stated in #894 remains: `close_any()` happily creates an invalid instruction in the case where `authority_address.is_none() && program_address.is_some()`.

Also, if every different type of `Close` instruction gets their own `close_*()` constructor, then making nonbreaking changes in the future is easier (such as re-adding a tombstone for example) since all you need to do is add a new function `close_xyz()` and leave the existing ones untouched. And finally IMO this enhances readability too.

PR #898, opened 2026-08-19
perf(slot-hashes, stake-history)!: replace tuple entries with repr(C) structs

kskalski:
#### Problem: 
wincode does not treat tuples as zero-copy, because Rust guarantees no tuple layout. Both sysvars stored their entries as tuples inside a `Vec`, so decoding a `SlotHashes` (20 KB) or `StakeHistory` (16 KB) account ran a per-element loop instead of a single copy. Both are decoded on every sysvar cache fill.

#### Summary of Changes: 
introduce `SlotHash` and `StakeHistoryItem` as `#[repr(C)]` padding-free structs and store those in the sysvar types. `assert_zero_copy` is gated on little-endian, since a native `u64` never qualifies as zero-copy on big-endian targets; a size assert pins the padding-free layout everywhere.

#### Breaking: 
`SlotHash` changes from a `(u64, Hash)` alias to a struct, and the `Deref` targets become `Vec<SlotHash>` / `Vec<StakeHistoryItem>`, so callers use `.slot`/`.hash` and `.epoch`/`.entry` instead of `.0`/`.1`. The serialized layout is unchanged, and is covered by a new wire-compatibility test in each crate.

This needs `slot-hashes` 4.0.0, `stake-history` 2.0.0, plus majors for `sysvar`, `program`, and `address-lookup-table-interface` since they publicly re-export the changed types.

### DIFF
--- a/address-lookup-table-interface/src/state.rs
+++ b/address-lookup-table-interface/src/state.rs
@@ -320,8 +320,8 @@ mod tests {
             slot_hashes.add(slot, Hash::new_unique());
         }
 
-        let most_recent_slot = slot_hashes.first().unwrap().0;
-        let least_recent_slot = slot_hashes.last().unwrap().0;
+        let most_recent_slot = slot_hashes.first().unwrap().slot;
+        let least_recent_slot = slot_hashes.last().unwrap().slot;
         assert!(least_recent_slot < most_recent_slot);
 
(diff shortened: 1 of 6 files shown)


grod220 (approved):
This looks good to me. But given the major bump required here, would prefer to get Jon's 👍 as well here.

joncinque (approved):
Looks great!

PR #900, opened 2026-08-20
[bls12-381] Optimize CUs

samkim-crypto:
#### Summary of Changes

I tried a bunch of attempts to optimize the CUs for the crate. The following are the changes that actually amounted to smaller CU reductions:

- Replaced the byte-loop predicates with array comparisons against constants. `GtElement::is_identity` went from 4,055 CU to 42. It sits on `pairing_check`'s success path, so a passing 2-pair check dropped from 43,527 to 38,923 CU. `G1Point::is_infinity` went 694 → 37 and `Scalar::is_zero` 79 → 17; the compressed and G2 variants are the same shape and now measure 36–37. Most of the old cost was the per-byte `i != flag_index` test, which roughly tripled the per-byte rate.
- Added `#[inline]` to the wrappers. They're thin enough that the cross-crate call was a real fraction of the work. Overhead on the allocating forms went from ~123 CU over the syscall charge to ~90, and the 8-point accumulation loop dropped 378 CU.
- Left `#[inline]` off `is_identity` and `pairing_check`. Inlining `is_identity` into `pairing_check` puts the 576-byte constant in the same frame as that function's `MaybeUninit`, costing +20 CU on every check against the ~6 it saves on a direct call. Comments on both so it doesn't get cleaned up later.
- `neg` borrows the infinity constant instead of rebuilding it. `Self::infinity(e)` was zeroing 96/192 bytes of stack per call. `neg_unchecked` now measures within 2 CU of the equivalent `sub_unchecked`.

Two things worth noting:

- Six methods lost `const fn`: `is_identity`, the four `is_infinity`, and `is_zero` since ince array equality isn't const-callable. Nothing in the crate called them in const context, but it is a public API narrowing.
- I added a compute unit section to the README.

I tested the CUs using a standard mollusk test setting. It is not hosted anywhere, but I can host it in another repo if helpful.

### DIFF
--- a/bls12-381/src/pairing.rs
+++ b/bls12-381/src/pairing.rs
@@ -14,6 +14,11 @@ pub const GT_ELEMENT_SIZE: usize = 576;
 /// Maximum number of pairs in a single batch pairing.
 pub const MAX_PAIRING_LENGTH: usize = 8;
 
+/// Canonical identity encodings, compared against by
+/// [`GtElement::is_identity`].
+const GT_IDENTITY_BE: GtElement = GtElement::identity(Endianness::Big);
+const GT_IDENTITY_LE: GtElement = GtElement::identity(Endianness::Little);
+
 /// A target group (Gt) element — a point in Fq12, 576 bytes.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 #[cfg_attr(
@@ -31,6 +36,7 @@ pub struct GtElement(
 
 impl GtElement {
     /// The multiplicative identity, in the given encoding.
+    #[inline]
     pub const fn identity(endianness: Endianness) -> Self {
         let mut bytes = [0u8; GT_ELEMENT_SIZE];
         match endianness {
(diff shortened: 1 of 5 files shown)


zz-sol (approved):
Nice! LGTM.

bls12-381/src/pairing.rs
```diff
@@ -42,34 +48,36 @@ impl GtElement {
...
+    /// replaces was const, but cost ~4,055 CU — ~7 per byte, of which the
+    /// per-byte flag-index test was two thirds — against ~45 CU for the whole
+    /// array here. This is on the hot path: [`pairing_check`] calls it on
+    /// every successful verification.
+    ///
+    // Deliberately not `#[inline]`. Inlining this into `pairing_check` puts the
+    // 576-byte identity constant in the same frame as that function's 576-byte
+    // `MaybeUninit`, costing ~20 CU on every pairing check — more than the ~6
+    // CU it saves on a direct call, which no hot path makes.
+    #[inline]
```
joncinque, line 63: This appears to still have `#[inline]` -- is that intentional?

samkim-crypto: Ah sorry, the doc is stale here. I'll address this in a follow-up when I add more tests for the crate 🙏

PR #902, opened 2026-08-23
feat(bls-signatures): introduce granular parse and point-conversion error variants

raushan728:
- `BlsError` collapsed multiple distinct parse/point-validation failures into `ParseFromString`, `ParseFromBytes`, `PointConversion` (see TODOs in `error.rs`)
- Added granular variants: `InvalidBase64`, `StringLengthExceeded`, `InvalidEncodedLength`, `InvalidLengthMultiple`, `InvalidPointEncoding`, `IdentityPointRejected`, `KeypairMismatch`
- Removed the three legacy variants, now fully superseded
- Updated `impl_from_str!` / `impl_bls_conversions!` macros to preserve exact failure cause at decode and affine-conversion boundaries
- Updated `keypair.rs` to use `InvalidEncodedLength` for length mismatches and `KeypairMismatch` for public/secret key mismatch (previously misused `ParseFromBytes`)
- Updated `secret_key.rs` to use `InvalidEncodedLength` for length mismatches

## Breaking Change
`BlsError` is not `#[non_exhaustive]`   variants added, removed. Downstream exhaustive matches on `BlsError` will need updating.

### DIFF
--- a/bls-signatures/src/error.rs
+++ b/bls-signatures/src/error.rs
@@ -14,6 +14,22 @@ pub enum BlsError {
     ParseFromString, // TODO: update after more precise error handling
     #[error("Failed to parse from bytes")]
     ParseFromBytes,
+    #[error("Failed to decode base64 string")]
+    InvalidBase64,
+    #[error("Base64 string length exceeded: max {max}, got {actual}")]
+    StringLengthExceeded { max: usize, actual: usize },
+    #[error("Invalid encoded length: expected {expected}, got {actual}")]
+    InvalidEncodedLength { expected: usize, actual: usize },
+    #[error("Invalid encoded length: expected {expected_compressed} or {expected_uncompressed}, got {actual}")]
+    InvalidLengthMultiple {
+        expected_compressed: usize,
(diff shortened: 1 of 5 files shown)


joncinque:
@samkim-crypto are these new variants useful? We can close if not

samkim-crypto:
Thanks for putting this together @raushan728 and for explicitly calling out the breaking change in the PR description.

We'd actually prefer to keep the existing error API stable. `BlsError` is a public, exhaustive enum, so removing or adding variants can break downstream code. We're not ready for a major release, and keeping the old variants alongside the new ones would still break exhaustive matches. The existing parsing and validation paths already reject these inputs, so the benefit here is more precise diagnostics. That is useful, but I don’t think it justifies the migration burden at this point.

I apologize about the TODOs in `error.rs`. They could reasonably be read as an invitation to make exactly this change, and we should have documented the compatibility constraints more clearly. Those comments should be clarified or removed.

@joncinque let's close this and I'll create a follow-up PR to remove the TODOs.

PR #905, opened 2026-08-25
feat(system-interface): add safe `try_from_u64` conversion for `SystemError`

raushan728:


joncinque (commented):
Is there any motivation for the change? If not, let's close the PR

raushan728:
Fair -  I don't have a concrete call site in this workspace where `SystemError::from(u64)` gets hit with untrusted input today. 

Motivation is preventive/consistency-based: `TryFrom<u32>` on this same type, and `InstructionError`/`ProgramError` elsewhere in the workspace, already avoid panicking on unknown codes `From<u64>` is the one inconsistent spot. Any actual `u64` decoding of `SystemError` would happen downstream (indexers, explorers, client SDKs), so I can't point to a usage here.

joncinque:
Thanks for the quick response, let's just close then

PR #906, opened 2026-08-25
{transaction,instruction}-error!: Mark enums non-exhaustive

joncinque:
#### Problem

It has been historically difficult to add new variants to `InstructionError` and `TransactionError`, since new enum variants constitute a breaking change. We avoided marking the enums as non-exhaustive in the past for fear of breaking places in Agave that do require exhaustive matching.

#### Summary of changes

In the end, there are very few call-sites in Agave that require exhaustive matching, so let's do ourselves a favor, and mark the enums non-exhaustive.

To go with this, introduce a `VARIANTS` constant, so that places that do require exhaustive matching can add a test to make sure they never forget new variants.

NOTE: `TransportError` is excluded from the change. We can mark it non-exhaustive, but it wraps `std::io::Error`, which has no const constructors without nightly Rust. That means we wouldn't be able to create a const `VARIANTS` array. There's only one call-site in Agave that converts from `TransportError` to RPC client error, so this seems acceptable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3779,6 +3779,8 @@ dependencies = [
  "solana-frozen-abi-macro",
  "solana-instruction-error",
  "solana-program-error",
+ "strum",
+ "strum_macros",
  "wincode",
 ]
 
@@ -4728,6 +4730,8 @@ dependencies = [
  "solana-instruction-error",
  "solana-sanitize",
  "solana-transaction-error",
+ "strum",
+ "strum_macros",
  "wincode",
 ]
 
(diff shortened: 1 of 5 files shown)


joncinque:
~Sorry, I have no idea why CI is failing. It runs fine on two different machines that I've tried. Gonna do a little debugging.~

Edit: I'm very silly. I hadn't rebased, so I didn't have the new `BailOut` error, so the test was working exactly as designed!

Lichtso (approved):

PR #907, opened 2026-08-25
offchain-message: validate signing domain

grod220:
This PR aligns offchain message deserialization with the header format. It ensures that `OffchainMessage::deserialize` enforces the fixed signing-domain prefix written by `OffchainMessage::serialize`.


joncinque (approved):
Looks great!

PR #909, opened 2026-08-26
transaction: guard versioned tx verify against signature/key length mismatch

### LINKED ISSUE #908: verify_and_hash_message and verify_with_results never verify signatures that map to address-lookup-table accounts
## Summary

`VersionedTransaction::verify_and_hash_message`, `VersionedTransaction::verify_with_results` (`transaction/src/versioned/mod.rs`) and the new `SanitizedVersionedTransaction::verify_and_hash_message` (`transaction/src/versioned/sanitized.rs`, #891) all pair signatures with keys using:

```rust
self.signatures.iter().zip(self.message.static_account_keys().iter())
```

For a v0 transaction where one of the required signers resolves through an address lookup table, that signer's public key sits beyond `static_account_keys()`. The `zip` then stops at the shorter side, so the lookup-table signer's signature is never checked at all, and earlier signatures are verified against the wrong positions of a truncated list.

## Reproduction shape

A sanitized v0 message with two required signers where signer 2 is carried by a lookup table:

- `signatures.len() == 2`
- `static_account_keys().len() == 1` (payer only)
- `zip` yields exactly one pair: `(signatures[0], static_keys[0])`

`signatures[1]` is silently dropped from verification, and `verify_with_results` returns a vector aligned to the truncated pairing rather than to the signature list. Every legitimate transaction of this shape gets `SignatureFailure` from `verify_and_hash_message`.

## Why this is not a one-line fix

Resolving lookup-table signers requires the actual tables, which are off-chain data the SDK transaction does not carry. The runtime can expand them; the standalone SDK cannot.

## Questions

1. Is the intended contract that these methods only verify statically-addressed signers? If so, should that limitation be documented, and should a length mismatch between `signatures` and `static_account_keys()` produce an explicit error instead of silent truncation?
2. Would an alternative API that accepts caller-resolved account keys (post table expansion) be in scope for this crate?

Happy to send a PR once the desired contract is clear. Happy to also note that #889 tightened sanitize to require an exact signature count, so inputs built through `try_new` have consistent lengths today, but nothing stops direct construction.

mayanksharma-eth:
## Summary

Fixes #908.

`VersionedTransaction::verify_and_hash_message` and `verify_with_results` (`transaction/src/versioned/mod.rs`) pair `signatures` with `static_account_keys()` via `zip` without first checking the lengths agree with `num_required_signatures`. A transaction constructed directly (bypassing `sanitize()`) with mismatched `signatures`/`static_account_keys()` lengths gets silently truncated by `zip` instead of failing verification — earlier signatures can end up verified against the wrong key positions instead of the mismatch being caught explicitly.

Note on #908's specific repro: a required signer resolved through an address lookup table can't actually occur — `Message::sanitize()` requires `num_required_signatures + num_readonly_unsigned_accounts <= static_account_keys.len()`, and `is_signer(i) = i < num_required_signatures`, so every signer is always a static key and lookup tables only ever supply non-signer accounts. `SanitizedVersionedTransaction` is unaffected for the same reason: `try_new` enforces `signatures.len() == num_required_signatures` via `sanitize_signatures()`, so the invariant always holds there. The real, narrower gap is that the *unsanitized* `VersionedTransaction` path trusts caller-supplied lengths.

## Fix

Add a length check in the shared `_verify_with_results` helper (used by both public methods) that returns explicit verification failure when `signatures.len() != num_required_signatures` or `static_account_keys().len() < num_required_signatures`, instead of relying on `zip`'s silent truncation.

## Test plan

- [x] Added `test_verify_with_results_signature_key_length_mismatch` covering a directly-constructed transaction with more signatures than static account keys.
- [x] `cargo test -p solana-transaction --lib versioned::`
- [x] `cargo build -p solana-transaction` and `--no-default-features` (no warnings)
- [x] `cargo clippy -p solana-transaction --tests`

### DIFF
--- a/transaction/src/versioned/mod.rs
+++ b/transaction/src/versioned/mod.rs
@@ -318,9 +318,20 @@ impl VersionedTransaction {
 
     #[cfg(feature = "verify")]
     fn _verify_with_results(&self, message_bytes: &[u8]) -> Vec<bool> {
+        let num_required_signatures = self.message.header().num_required_signatures as usize;
+        let static_account_keys = self.message.static_account_keys();
+        if self.signatures.len() != num_required_signatures
+            || static_account_keys.len() < num_required_signatures
+        {
+            // Mismatched lengths mean this transaction was never sanitized: zipping
+            // signatures against static keys would silently truncate to the shorter
+            // side and pair signatures with the wrong keys instead of catching this.
+            return alloc::vec![false; self.signatures.len()];
+        }
+
         self.signatures
             .iter()
-            .zip(self.message.static_account_keys().iter())
+            .zip(static_account_keys.iter())
             .map(|(signature, pubkey)| signature.verify(pubkey.as_ref(), message_bytes))
             .collect()
     }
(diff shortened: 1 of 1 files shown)


apfitzge:
This was addressed already in #890, which was in progress when you created the PR - sorry for the race!

PR #916, opened 2026-08-30
[bls12-381] LLM based cu optimizations

zz-sol:
This PR has a few minor optimizations for the CU cost for BLS12-381 operations. It also updates the documentation to reflect the latest cu cost.

### DIFF
--- a/bls12-381/README.md
+++ b/bls12-381/README.md
@@ -13,7 +13,7 @@ and zero-knowledge proof (e.g. Groth16) validation.
   `bytemuck::Pod`, so instruction data can be cast directly into curve points
   without allocation.
 - **In-place operations.** Every group operation has an `_assign` variant that
-  writes into a caller-supplied `MaybeUninit` buffer, saving ~68 CU per call
+  writes into a caller-supplied `MaybeUninit` buffer, saving ~36 CU per call
   and keeping large results off the 4KB stack.
 - **Validated and unchecked APIs.** Group operations validate their operands by
   default; `_unchecked` variants skip the subgroup check for cheaper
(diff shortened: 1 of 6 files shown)


bls12-381/README.md
```diff
@@ -78,32 +78,42 @@ network. `test_success_writes_full_buffer` pins the property.
...
-    // The early return is load-bearing: on failure `scratch` is poisoned and
-    // must not be swapped into `acc` or read afterwards.
-    if !lhs.add_assign_unchecked(p, &mut scratch, e) {
+let mut a = MaybeUninit::new(points[0]);
+let mut b = MaybeUninit::uninit();
+
+// Two additions per iteration, so the result lands back in `a` and no buffer
+// ever has to move. The early returns are load-bearing: on failure the
+// destination is poisoned and must not be read afterwards.
+for pair in points[1..].chunks_exact(2) {
```
samkim-crypto: If the number of points are even, then it seems like there is going to be a point that is not included in the sum. Should we add a logic to add the remaining point in case there is one?

bls12-381/README.md
```diff
@@ -78,32 +78,42 @@ network. `test_success_writes_full_buffer` pins the property.
...
-let mut acc = MaybeUninit::new(G1Point::infinity(e));
-let mut scratch = MaybeUninit::uninit();
-
-for p in points {
-    // SAFETY: `acc` is initialized, by `new` above and by the swap below.
-    let lhs = unsafe { acc.assume_init_ref() };
-    // The early return is load-bearing: on failure `scratch` is poisoned and
-    // must not be swapped into `acc` or read afterwards.
-    if !lhs.add_assign_unchecked(p, &mut scratch, e) {
+let mut a = MaybeUninit::new(points[0]);
```
samkim-crypto: `points[0]` panics on an empty slice, and `points[1..]` on the next line does too. The old example seeded with `G1Point::infinity(e)`, which made the loop total for any length as the empty sum is infinity, so there was nothing to special-case.

Seeding from `points[0]` saves one addition, but that's a separate optimization from the swap -> alternate change and it isn't called out anywhere. Since this examples gets pasted into programs where `points` come from instruction data, and a panic aborts the transaction, I'd rather go back to the infinity seed and keep the loop total. If we do want the saving, let's use `split_first()` and reject empty explicitly.

bls12-381/README.md
```diff
@@ -138,79 +148,125 @@ small, and depends only on how a result is returned:
...
+| `_assign` forms — write into a caller's `MaybeUninit`                                     | 14–16    |
+| `pairing_map_assign`                                                                      | 16       |
+| `pairing_check` — includes the `is_identity` comparison                                   | 69       |
+| Allocating forms — `add_unchecked`, `sub_unchecked`, `neg_unchecked`, `mul`, `decompress` | 36–38    |
 
 The allocating figure is the `Option<Self>` construction, not the byte copy: it
 is the same whether the output is a 96-byte G1 point or a 576-byte `Gt`
 element. Choosing an `_assign` form over its allocating counterpart therefore
-saves ~68 CU per operation, whatever the type.
+saves ~36 CU per operation, whatever the type.
```
samkim-crypto: Hm, if `_assign` is 14–16 and allocating is 36–38, shouldn't the difference be ~20, not 36?

The totals table below supports the ~36: `add_assign_unchecked` 144 vs `add_unchecked` 180, which puts the allocating overhead at ~52, not 36–38. Same answer from the other direction: 180 minus the 128 CU addition syscall quoted at the end of this section is 52.

I think the old table was self-consistent (90 − 22 = 68 = 218 − 150), so this looks like the delta got pasted into the absolute column. Could you double-check which number came out of the benchmark? Let me know if I am missing anything.

samkim-crypto (commented):
Looks fine to me overall. Just some comments on the docs.

samkim-crypto:
@claude review once

samkim-crypto (approved):
LGTM!

PR #918, opened 2026-08-31
sanitized: deprecate unnecessary functions

2501babe:
unless there is a reason to keep these im not aware of, i think we should remove them. while implementing my `get_durable_nonce()` replacement on `SVMStaticMessage`, i discovered there was another version of it, which behaves the same but is written somewhat differently

`get_durable_nonce()` is only used in rpc and the other two dont appear to be used at all


apfitzge (approved):
Seems reasonable to remove so there is one source of the behavior.
The only reason I can think to keep it is to avoid pulling another dep for the user.

PR #919, opened 2026-08-31
svm-transaction: add get_durable_nonce_static()

2501babe:
add a durable nonce function to SVMStaticMessage that applies the [SIMD-0602](https://github.com/solana-foundation/solana-improvement-documents/pull/602) rule to exclude program ids, to be used under feature gate in agave. we will delete the SVMMessage version and rename the new one after 602 goes live. pleasingly, since SVMMessage is a subtype of SVMStaticMessage, that change would be non-breaking

my plan in agave is:
* use `get_durable_nonce_static()` with `ban_nonce_as_program_id` as `true` without a feature gate for all the helpers that do strict nonce checks ie that are not in consensus, most notably the ingress precheck
* use `get_durable_nonce_static()` with `ban_nonce_as_program_id` as `false` (or, when SIMD-0602 moves forward, with the feature gate) in `check_transactions_before_execution()`. move the `is_writable()` check into svm

this allows relaxing `check_transactions()` and all its wrappers to `StaticTransactionWithMeta`, unblocking further development on ALT relaxation without waiting for SIMD-0602, which only matters when we actually _do_ SIMD-0192

most of the upcoming work is refactoring the banking/runtime/svm/runtime flow to tolerate unresolved transactions; we only need SIMD-0602 when it comes time to introduce fee-only ALT resolution failures

### DIFF
--- a/svm-transaction/src/svm_message.rs
+++ b/svm-transaction/src/svm_message.rs
@@ -137,6 +137,16 @@ pub trait SVMStaticMessage: Debug {
                     .filter_map(|signer_index| self.static_account_keys().get(signer_index))
             })
     }
+
+    /// If the message uses a durable nonce, return the pubkey of the nonce account.
+    /// This is identical in behavior to `SVMMessage::get_durable_nonce()`, except
+    /// we cannot fully check whether the nonce account is writable based on program ID
+    /// demotion and reserved key demotion. When `ban_nonce_as_program_id` is active,
+    /// program ID demotion no longer can apply. Reserved keys are never valid nonce
+    /// accounts, so they will always be rejected by nonce account validation.
+    fn get_durable_nonce_static(&self, ban_nonce_as_program_id: bool) -> Option<&Pubkey> {
+        get_durable_nonce_internal(self, ban_nonce_as_program_id).map(|(key, _index)| key)
+    }
 }
 
 pub trait SVMMessage: SVMStaticMessage {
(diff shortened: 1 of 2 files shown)


svm-transaction/src/svm_message.rs
```diff
@@ -148,35 +158,9 @@ pub trait SVMMessage: SVMStaticMessage {
 
     /// If the message uses a durable nonce, return the pubkey of the nonce account
     fn get_durable_nonce(&self) -> Option<&Pubkey> {
```
apfitzge, line 160: I have some concern with leaving this. To do a correct migration we need to make sure all call-sites in agave use the new fn with the bool argument depending on the feature-status.

Leaving this seems like a footgun to miss a callsite.

2501babe: i think what i will do is land the agave change adding the real writable check to svm and then do the move of `get_durable_nonce()` to `SVMStaticMessage` immediately. this pr will be blocked by that one so there is no danger of this getting released/merged and putting agave in a dangerous state

apfitzge (commented):
lgtm

@buffalojoec I think svm team needs to be added as owners of these crates. IIRC you guys wanted to sign off on changes here.

@2501babe I'm holding off on explicit approval until hearing from joe

buffalojoec:
> @buffalojoec I think svm team needs to be added as owners of these crates. IIRC you guys wanted to sign off on changes here.

I did add it to CODEOWNERS. I wonder if the repo needs a certain setting enabled, ie. "require review from code owners". cc @joncinque.

I'll review the changes, though.

joncinque:
Hm, looks like we have that enabled:
image

But I noticed we didn't have the SVM team on the repo, so I just added that

PR #920, opened 2026-09-01
account-info: Add comment to `assign`

febo:
### Problem

The `assign` method currently unsafely writes to `self.owner` through a read-only shared reference. When the method is used with other active references, the compiler might generate arbitrarily incorrect code (UB).

### Solution

Add comments to highlight the potential footgun when using this method.

### DIFF
--- a/account-info/src/lib.rs
+++ b/account-info/src/lib.rs
@@ -187,6 +187,13 @@ impl<'a> AccountInfo<'a> {
         Ok(())
     }
 
+    /// Changes the owner of the account.
+    ///
+    /// # Important
+    ///
+    /// The caller must ensure that there are no active references to the `owner` when
+    /// calling this method. This is because the `owner` is a shared read-only reference,
+    /// and changing it while there are active references can lead to undefined behavior.
     #[allow(invalid_reference_casting)]
     pub fn assign(&self, new_owner: &Address) {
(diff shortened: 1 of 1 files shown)


febo:
@joncinque Not sure whether we should take advantage of breaking changes and make the method `unsafe` or not. It is `unsafe` on the `AccountView` type.

joncinque (approved):
I'm happy with landing this as is to start, since it's true that the behavior is potentially dangerous.

Would marking it `unsafe` solve the problem though? I think we should also change the function to take in `&mut self` and allow the compiler to see if there are any outstanding borrows at the time.

What do you think?

febo:
> I'm happy with landing this as is to start, since it's true that the behavior is potentially dangerous.
> 
> Would marking it `unsafe` solve the problem though? I think we should also change the function to take in `&mut self` and allow the compiler to see if there are any outstanding borrows at the time.
> 
> What do you think?

Oh, that is a good point. `AccountView` also requires `&mut self`. Unfortunately none of these changes would really solve the issue, since duplicated accounts share the same reference. But I think it is the best we can with the current design.

joncinque:
Ah right that's a good point... in that case, let's do whatever is easiest for now, since the long-term plan is to deprecate and remove the whole crate.

PR #922, opened 2026-09-01
message/transaction!: Remove deprecated functions

joncinque:
#### Problem

There are a few deprecated functions, and we're doing a breaking release soon.

#### Summary of changes

Remove the deprecated functions.

### DIFF
--- a/svm-transaction/src/tests.rs
+++ b/svm-transaction/src/tests.rs
@@ -72,7 +72,6 @@ fn test_get_durable_nonce() {
     // No instructions - no nonce
     {
         let message = create_message_for_test(1, 1, vec![Pubkey::new_unique()], vec![], None);
-        assert!(SanitizedMessage::get_durable_nonce(&message).is_none());
         assert!(SVMMessage::get_durable_nonce(&message).is_none());
     }
 
@@ -85,7 +84,6 @@ fn test_get_durable_nonce() {
             vec![CompiledInstruction::new_from_raw_parts(1, vec![], vec![])],
             None,
         );
-        assert!(SanitizedMessage::get_durable_nonce(&message).is_none());
         assert!(SVMMessage::get_durable_nonce(&message).is_none());
     }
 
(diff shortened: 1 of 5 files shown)


joncinque:
Note that there were already breaking changes in `solana-transaction` before you landed your deprecations in #918, so we won't have a release with a deprecated `SanitizedTransaction::get_durable_nonce` function -- is that ok? I did a solana-message publish

apfitzge:
> Note that there were already breaking changes in `solana-transaction` before you landed your deprecations in #918, so we won't have a release with a deprecated `SanitizedTransaction::get_durable_nonce` function -- is that ok? I did a solana-message publish

Might be annoying for users of the library to not have a version with deprecated, which iirc has been our standard so far. That said it's totally valid wrt semver and the replacement is straight-forward with just importing a trait...and then fn names don't even change iirc.

I'm fine with just doing the break personally, but I'll let @joncinque make the final call on if that's acceptable for how we generally maintain this sdk.

2501babe:
i had been expecting a final patch release with the deprecations but i guess if there were already breaking changes in then it cant be helped

svm-transaction/src/tests.rs
```diff
@@ -16,7 +16,6 @@ use {
 };
 
 #[test]
-#[allow(deprecated)]
 fn test_get_durable_nonce() {
```
2501babe, line 19: we can delete all the `assert!(SanitizedMessage::get_durable_nonce( ... ) ... )` lines in here, they become confusing because when we delete the inherent method, this silently resolves to the trait which we already assert

joncinque: makes sense, done with [23cf6c1](https://github.com/anza-xyz/solana-sdk/pull/922/commits/23cf6c179d6c28a43775f5e3fb046910e8cc9401)

joncinque:
> I'm fine with just doing the break personally, but I'll let @joncinque make the final call on if that's acceptable for how we generally maintain this sdk.

It would be better to do the release with the deprecation, but the function is already unused by agave, so I'm not too worried. Otherwise I have to create a new branch, cherry-pick Hana's change, and release from that branch.

2501babe (approved):

PR #926, opened 2026-09-02
error types: Mark errors as non-exhaustive

joncinque:
#### Problem

We've marked error types in instruction-error and transaction-error as non-exhaustive, which will cause many breaking changes in the rest of the SDK.

We still have many error types that could be non-exhaustive, but aren't.

#### Summary of changes

Make more errors non-exhaustive. Looking through Agave, none of these are being matched on, so it should be an unnoticeable change downstream.

### DIFF
--- a/vote-interface/src/error.rs
+++ b/vote-interface/src/error.rs
@@ -7,6 +7,7 @@ use {
 
 /// Reasons the vote might have had an error
 #[derive(Debug, Clone, PartialEq, Eq, FromPrimitive, ToPrimitive)]
+#[non_exhaustive]
 pub enum VoteError {
     VoteTooOld,
     SlotsMismatch,
--- a/message/src/compiled_keys.rs
+++ b/message/src/compiled_keys.rs
@@ -19,6 +19,7 @@ pub(crate) struct CompiledKeys {
 }
 
 #[derive(PartialEq, Debug, Eq, Clone)]
+#[non_exhaustive]
 pub enum CompileError {
     AccountIndexOverflow,
     AddressTableLookupIndexOverflow,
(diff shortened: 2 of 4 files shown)


febo (approved):
Looks great!

vote-interface/src/error.rs
```diff
@@ -7,6 +7,7 @@ use {
 
 /// Reasons the vote might have had an error
 #[derive(Debug, Clone, PartialEq, Eq, FromPrimitive, ToPrimitive)]
+#[non_exhaustive]
```
apfitzge, line 10: Warnings above show details:

```
Warning:   --> vote-interface/src/error.rs:9:54
     |
   9 | #[derive(Debug, Clone, PartialEq, Eq, FromPrimitive, ToPrimitive)]
     |                                                      ^----------
     |                                                      |
     |                                                      `ToPrimitive` is not local
     |                                                      move the `impl` block outside of this constant `_IMPL_NUM_ToPrimitive_FOR_VoteError`
  10 | #[non_exhaustive]
  11 | pub enum VoteError {
     |          --------- `VoteError` is not local
     |
     = note: the derive macro `ToPrimitive` defines the non-local `impl`, and may need to be changed
     = note: the derive macro `ToPrimitive` may come from an old version of the `num_derive` crate, try updating your dependency with `cargo update -p num_derive`
     = note: an `impl` is never scoped, even when it is nested inside an item, as it may impact type checking outside of that item, which can be the case if neither the trait or the self type are at the same nesting level as the `impl`
     = note: items in an anonymous const item (`const _: () = { ... }`) are treated as in the same scope as the anonymous const's declaration for the purpose of this lint
     = note: this warning originates in the derive macro `ToPrimitive` (in Nightly builds, run with -Z macro-backtrace for more info)

```

so might need to force a higher minimum version for the num_derive crate

joncinque: #927 to take care of that

joncinque:
Rebased, this should be good for another look

apfitzge (approved):

PR #929, opened 2026-09-02
rent!: Remove deprecated fields

febo:
### Problem

The `Rent` struct includes 2 deprecated fields, which makes the memory requirement be `24` bytes (`17` bytes serialized + `7` padding bytes). Therefore it is not possible load the struct using zero-copy directly from the rent sysvar account.

### Solution

This PR removes the deprecated fields and adds a custom serializer that still requires `17` bytes. This way the struct only requires `8` bytes of memory, but still serializes into `17` bytes.

### DIFF
--- a/rent/src/lib.rs
+++ b/rent/src/lib.rs
@@ -335,12 +335,12 @@ mod tests {
         #[test]
         fn test_minimum_balance(bytes in 0usize..=MAX_PERMITTED_DATA_LENGTH as usize) {
             let default_rent = Rent::default();
-            #[allow(deprecated)]
-            let previous_rent = Rent {
-                lamports_per_byte: DEFAULT_LAMPORTS_PER_BYTE / 2,
-            };
             let default_calc = default_rent.minimum_balance(bytes);
-            assert_eq!(default_calc, previous_rent.minimum_balance(bytes));
+
+            // check that the calculation gives the same result using floats
+            let float_calc = (((ACCOUNT_STORAGE_OVERHEAD + bytes as u64) * GENESIS_LAMPORTS_PER_BYTE_YEAR) as f64
+                * f64::from_le_bytes(GENESIS_EXEMPTION_THRESHOLD)) as u64;
+            assert_eq!(default_calc, float_calc);
         }
     }
 
--- a/rent/src/lib.rs
+++ b/rent/src/lib.rs
@@ -12,6 +12,8 @@ extern crate std;
 #[cfg(feature = "sysvar")]
 pub mod sysvar;
 
+#[cfg(feature = "serde")]
+use serde::ser::SerializeStruct;
 #[cfg(feature = "frozen-abi")]
 use solana_frozen_abi_macro::{AbiExample, StableAbi, StableAbiSample};
 use solana_sdk_macro::CloneZeroed;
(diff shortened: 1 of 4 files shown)


rent/src/lib.rs
```diff
@@ -245,38 +221,74 @@ mod tests {
...
+        let rent = Rent::with_lamports_per_byte(1);
+        let serialized_len =
+            wincode::serialized_size(&rent).expect("failed to get serialized sysvar size") as usize;
+
+        let mut account = AccountSharedData::new(
+            rent.minimum_balance(serialized_len),
+            serialized_len,
+            &sysvar::id(),
+        );
+        wincode::serialize_into(account.data_as_mut_slice(), &rent).unwrap();
```
febo, line 368: This is how the `Rent` sysvar is serialized in Agave.

rent/src/lib.rs
```diff
@@ -215,22 +175,72 @@ impl Rent {
...
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for Rent {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let lamports_per_byte = u64::deserialize(deserializer)?;
+        Ok(Self { lamports_per_byte })
+    }
+}
```
joncinque, line 232: Does this need to deserialize the rest of the fields to consume the rest of whatever's in the buffer?

rent/src/lib.rs
```diff
@@ -12,34 +12,35 @@ extern crate std;
...
+/// these were deprecated and have been removed. The serialized size of the `Rent` sysvar
+/// account is still `17` bytes, which is the size of the `Rent` sysvar account.
 #[repr(C)]
 #[cfg_attr(feature = "frozen-abi", derive(AbiExample, StableAbi, StableAbiSample))]
-#[cfg_attr(
-    feature = "serde",
-    derive(serde_derive::Deserialize, serde_derive::Serialize)
-)]
-#[cfg_attr(feature = "wincode", derive(wincode::SchemaWrite, wincode::SchemaRead))]
+#[cfg_attr(feature = "wincode", derive(wincode::SchemaRead))]
```
joncinque: Same for this, shouldn't the deserialize implementation consume additional bytes?

joncinque (commented):
I think this'll work, just a couple more concerns 😅

rent/src/lib.rs
```diff
@@ -192,7 +197,19 @@ impl<'de> serde::Deserialize<'de> for Rent {
     where
         D: serde::Deserializer<'de>,
     {
-        let lamports_per_byte = u64::deserialize(deserializer)?;
+        #[derive(serde_derive::Deserialize)]
+        struct RentWithDeprecatedFields {
+            lamports_per_byte: u64,
+            _exemption_threshold: [u8; 8],
+            _burn_percent: u8,
+        }
```
joncinque: Very slick!

rent/src/lib.rs
febo: Good point, it does indeed.

rent/src/lib.rs
febo: Same here, we need to consume the additional bytes.

rent/src/lib.rs
```diff
@@ -215,22 +199,118 @@ impl Rent {
...
+        let mut state = serializer.serialize_struct("Rent", 3)?;
+        state.serialize_field("lamports_per_byte", &self.lamports_per_byte)?;
+
+        if self.lamports_per_byte == GENESIS_LAMPORTS_PER_BYTE_YEAR {
+            state.serialize_field("exemption_threshold", &GENESIS_EXEMPTION_THRESHOLD)?;
+            state.serialize_field("burn_percent", &GENESIS_BURN_PERCENT)?;
+        } else {
+            state.serialize_field("exemption_threshold", &SIMD0194_EXEMPTION_THRESHOLD)?;
+            state.serialize_field("burn_percent", &SIMD0194_BURN_PERCENT)?;
         }
```
febo, line 249: This is not super robust. If the `lamports_per_byte` gets back to `3480`, it will serialize a different `exemption_threshold` and `burn_percent` value.

febo:
Closing this in favour of creating a specific type for programs to use. The solution in this PR is "fragile" in the sense it relies on hardcoded values for the removed deprecated fields to keep the sysvar account consistent.

PR #930, opened 2026-09-03
loader-v3-interface: drop trailing bool deserializers

buffalojoec:
#### Problem
These trailing boolean deserializers needed a few more tweaks before they could go into agave, but the SIMDs they serve need a bit more research, and @kskalski is trying to get agave upgraded to wincode.

#### Summary of Changes
Just rip them out, until we're ready to re-land SIMDs 0430 & 0432.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3878,7 +3878,6 @@ dependencies = [
  "solana-loader-v3-interface",
  "solana-pubkey",
  "solana-sdk-ids",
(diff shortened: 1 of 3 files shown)


kskalski (approved):
I checked the diff against loader-v3-interface 7.0.0 tag and it now becomes almost identital to that except for switching `bincode` feature / API into `wincode`.

PR #932, opened 2026-09-04
[bls-signatures] Reduce allocation size in pairing computation

samkim-crypto:
#### Summary of Changes

I made some optimizations to minimize heap allocations. I removed temporary pairing-preparation allocations from single-signature verification and sequential/parallel same-message aggregate verification. I retained reusable message preparations and added correctness tests, allocation counters, and timing benchmarks to check each optimization.

Most of the code line changes are on the test harnesses, not core logic changes.

##### Additional test and benchmark harnesses

[a809da22]: I established the initial allocation budgets with a standalone counter, allocating and nonallocating controls, and valid/wrong-message verification fixtures.
[3ee601f3]: I added Criterion benchmarks that separated hashing and preparation costs from verification of individual signatures.
[d511b810]: I added regression tests for prepared-message reuse, cloning, signature errors, and duplicate-message grouping before changing the preparation storage.
[08162631]: I added sequential same-message aggregate allocation coverage, which exposed one remaining 19,584-byte allocation in the raw and pre-hashed helpers.
[d7fb6a6f]: I added matching sequential aggregate timing benchmarks that included both public-key and signature aggregation.
[78230e7e]: I added parallel aggregate allocation coverage and a Rayon control, using a warmed two-worker pool after tracing intermittent allocations to lazy initialization of Rayon’s sleep machinery.
[233cf7f3]: I added parallel aggregate timing benchmarks that started the clock inside a worker, measuring aggregation and verification while excluding the outer pool-submission overhead.

##### Core logic changes

[f6ae8f52]: Previously, raw and pre-hashed verification created temporary G2 preparation tables for both the hashed message and the signature, costing two heap allocations totaling 39,168 bytes per operation. I replaced those preparations with direct Miller loops and one final exponentiation, and updated the crate’s public-key wrappers to use that path while preserving validation and the public trait’s defaults. The raw, pre-hashed, and compressed-signature cases then measured zero allocations. Pre-hashed latency stayed essentially flat; the small raw and compressed-path gains coincided with a faster unchanged hashing control, so the clearest result was the allocation reduction.
[3e40b4c8]: Previously, prepared verification reused the message table but still built and discarded a 19,584-byte preparation table for every signature. Because blstrs kept its prepared coefficients private, I introduced a crate-private wrapper that owned the coefficients and called the existing blst dependency directly. This let verification reuse the message preparation and compute the signature Miller loop directly, reducing measured verification allocations from one to zero. I also adapted prepared distinct-message screening to the new storage. Prepared verification stayed around 503 µs, while constructing a prepared message continued to allocate its reusable table.
[4550c982]: After the earlier changes, sequential raw and pre-hashed aggregate verification still constructed a PreparedHashedMessage solely to call the prepared aggregate helper. This allocated a 19,584-byte message table for a single verification, even though the direct verification path was already allocation-free. I routed both paths through direct pre-hashed verification after aggregating the public keys and signatures, preserving error-propagation order. Measured allocations fell from one to zero, with no material timing change.
[e9d89d32]: The parallel raw and pre-hashed aggregate helpers also constructed a temporary prepared message before calling the prepared aggregate helper, adding one 19,584-byte allocation per operation. I retained the existing Rayon aggregation and error ordering, then passed the aggregated public key and signature directly to pre-hashed verification. In the warmed two-worker pool, measured allocations fell from one to zero, and the tightened budgets passed all three measurement passes. The timing comparison showed no material latency change.

##### Measured allocation reductions

Operation | Allocations/op: before → after | Requested bytes/op: before → after
-- | -- | --
Single verification: raw, pre-hashed, compressed signature | 2 → 0 | 39,168 → 0
Single verification: prepared message | 1 → 0 | 19,584 → 0
Sequential aggregate: raw and pre-hashed* | 1 → 0 | 19,584 → 0
Parallel aggregate: raw and pre-hashed* | 1 → 0 | 19,584 → 0
Creating a reusable message preparation | 1 → 1 | 19,584 → 19,584

##### Trying the allocation measurements

To try these measurements yourself, run the appropriate command from the workspace root.

Apple Silicon macOS:

```
cargo test \
    --locked \
    --release \
    --target aarch64-apple-darwin \
    -p solana-bls-signatures \
    --no-default-features \
    --features std,parallel \
    --test allocation_regression
```

x86-64 Linux:

```
cargo test \
    --locked \
    --release \
    --target x86_64-unknown-linux-gnu \
    -p solana-bls-signatures \
    --no-default-features \
    --features std,parallel \
    --test allocation_regression
```

### DIFF
--- a/bls-signatures/Cargo.toml
+++ b/bls-signatures/Cargo.toml
@@ -65,6 +65,11 @@ tempfile = { workspace = true }
 [lints]
 workspace = true
 
+[[test]]
+name = "allocation_regression"
+path = "tests/allocation_regression.rs"
+harness = false
+
 [[bench]]
 name = "bls_signatures"
 harness = false
--- a/bls-signatures/Cargo.toml
+++ b/bls-signatures/Cargo.toml
@@ -73,3 +73,7 @@ harness = false
 [[bench]]
 name = "bls_signatures"
 harness = false
+
+[[bench]]
+name = "verification_paths"
+harness = false
--- a/bls-signatures/src/hash.rs
+++ b/bls-signatures/src/hash.rs
@@ -1,7 +1,7 @@
 use {
-    crate::proof_of_possession::POP_DST,
+    crate::{prepared_g2::PreparedG2, proof_of_possession::POP_DST},
     alloc::vec::Vec,
-    blstrs::{G2Affine, G2Prepared, G2Projective},
+    blstrs::{G2Affine, G2Projective},
 };
 
 /// Domain separation tag used for hashing messages to curve points to prevent
@@ -41,13 +41,13 @@ impl HashedMessage {
 /// representation. It is useful when the same message is verified repeatedly
 /// against many signatures because it skips the pairing preparation step.
 ///
-/// Memory note: each `PreparedHashedMessage` includes a `G2Prepared`, which is
-/// significantly larger than a plain `HashedMessage` (roughly ~19 KiB per
-/// element in current `blstrs` implementations).
+/// Memory note: each `PreparedHashedMessage` owns a heap-allocated pairing
+/// preparation table (19,584 bytes for a nonidentity point), in addition to
+/// the hashed point. Cloning duplicates the table.
 #[derive(Clone, Debug)]
 pub struct PreparedHashedMessage {
     pub(crate) hashed_message: HashedMessage,
-    pub(crate) prepared: G2Prepared,
+    pub(crate) prepared: PreparedG2,
 }
 
 impl PreparedHashedMessage {
@@ -60,7 +60,7 @@ impl PreparedHashedMessage {
     pub fn from_hashed_message(hashed_message: &HashedMessage) -> Self {
         Self {
             hashed_message: *hashed_message,
-            prepared: G2Prepared::from(hashed_message.0),
+            prepared: PreparedG2::from(hashed_message.0),
         }
     }
 }
--- a/bls-signatures/src/prepared_g2.rs
+++ b/bls-signatures/src/prepared_g2.rs
@@ -1,0 +1,64 @@
+//! G2 pairing preparation owned by this crate.
+
+use {
+    alloc::{vec, vec::Vec},
+    blst::{blst_fp12, blst_fp6, blst_miller_loop_lines, blst_precompute_lines},
+    blstrs::{G1Affine, G2Affine},
+    group::prime::PrimeCurveAffine,
+};
+
+// The buffer length required by blst_precompute_lines and blst_miller_loop_lines.
+const MILLER_LOOP_LINES: usize = 68;
+
+#[derive(Clone, Debug)]
+pub(crate) struct PreparedG2 {
+    // Empty for the identity; otherwise exactly 68 initialized coefficients.
+    lines: Vec<blst_fp6>,
+}
+
+impl From<G2Affine> for PreparedG2 {
+    fn from(affine: G2Affine) -> Self {
+        if bool::from(affine.is_identity()) {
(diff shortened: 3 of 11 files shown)


bls-signatures/src/hash.rs
```diff
@@ -32,13 +41,13 @@ impl HashedMessage {
...
-/// significantly larger than a plain `HashedMessage` (roughly ~19 KiB per
-/// element in current `blstrs` implementations).
+/// Memory note: each `PreparedHashedMessage` owns a heap-allocated pairing
+/// preparation table (19,584 bytes for a nonidentity point), in addition to
+/// the hashed point. Cloning duplicates the table.
 #[derive(Clone, Debug)]
 pub struct PreparedHashedMessage {
     pub(crate) hashed_message: HashedMessage,
-    pub(crate) prepared: G2Prepared,
+    pub(crate) prepared: PreparedG2,
```
efagerho, line 50: Both of these store the Miller loop lines on the heap in a Vec, so what is the purpose of the PreparedG2 type that's been pulled into the crate? Is it to gain access to the computed Miller loop lines elsewhere?

samkim-crypto: Yes exactly. I used `PreparedG2` to use the cached message lines directly, since `blstrs::G2Prepared` kept them private. Both versions retained the heap-backed message table.

Previously, prepared verification also allocated a temporary table for the signature. I reused the message table and computed the signature Miller loop directly, keeping one final exponentiation. This removed one 19,584-byte allocation per verification; constructing the reusable message preparation still allocated its table.

efagerho (dismissed):

bls-signatures/src/prepared_g2.rs
```diff
@@ -0,0 +1,64 @@
...
+    lines: Vec<blst_fp6>,
+}
+
+impl From<G2Affine> for PreparedG2 {
+    fn from(affine: G2Affine) -> Self {
+        if bool::from(affine.is_identity()) {
+            return Self { lines: Vec::new() };
+        }
+
+        let mut lines = vec![blst_fp6::default(); MILLER_LOOP_LINES];
```
akhi3030: should we incorporate this change to avoid unnecessary initialization: https://github.com/filecoin-project/blstrs/pull/70/changes?

bls-signatures/Cargo.toml
```diff
@@ -65,6 +65,15 @@ tempfile = { workspace = true }
 [lints]
 workspace = true
 
+[[test]]
```
akhi3030, line 68: I can't remember how cargo works here, will this add the test to CI?  If not, can we consider adding it to CI assuming it is not very expensive to run of course.

akhi3030 (commented):
I don't think my review holds much water.  I am happy for it to land and then give it a spin to observe potential improvements.

akhi3030 (dismissed):

bls-signatures/src/prepared_g2.rs
samkim-crypto: Yes, this is a good idea. Done in [66521cc](https://github.com/anza-xyz/solana-sdk/pull/932/commits/66521cc4d3994a82ddc4f4a6514b026037c49550). This seems to have improved preparation time by ~5.4%.

bls-signatures/Cargo.toml
samkim-crypto: Yes, the `./scripts/test-stable.sh` runs `cargo test --all --tests -- --nocapture`, which should include integration-test targets such as `allocation_regression` I believe.

zz-sol (approved):
LGTM

PR #933, opened 2026-09-04
ci: use tombi for TOML formatting

DSharifi:
Replace cargo-sort with Tombi for TOML formatting and dependency ordering.

Same changes as https://github.com/anza-xyz/agave/pull/15062 and https://github.com/anza-xyz/agave-sdk/pull/62 for this repo.


joncinque (approved):
Beautiful, thanks!

PR #934, opened 2026-09-04
reward-info: add a RewardType for VAT

AshwinSekar:
We have no observability around VAT burns.

This adds a new `RewardType` for the VAT for use in https://github.com/anza-xyz/agave/pull/15095

### DIFF
--- a/reward-info/src/lib.rs
+++ b/reward-info/src/lib.rs
@@ -18,6 +18,9 @@ pub enum RewardType {
     Staking,
     Voting,
     DeactivatedStake,
+    /// The burn associated with the validator admission ticket
+    /// The accompanying RewardInfo will have a negative lamports value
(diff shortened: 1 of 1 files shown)


reward-info/src/lib.rs
```diff
@@ -18,6 +18,9 @@ pub enum RewardType {
     Staking,
     Voting,
     DeactivatedStake,
+    /// The burn associated with the validator admission ticket
+    /// The accompanying RewardInfo will have a negative lamports value
+    VAT,
```
t-nelson:
```suggestion
    VATDebit,
```

since this action is in opposition to the typename

t-nelson:
@joncinque can we get sdk signoff here?

joncinque (approved):
Works for me!

PR #936, opened 2026-09-07
[bls-signature] remove foreign types

zz-sol:
this PR removes the foreign types that were use in bls-signature crate
- e9fd489d24d6272e06b77ad82eff546c5bec87a2 wraps the `zeroized` types
- 4dacca187f73f18b67fb6933cdc2f82b080b494a wraps `blstrs::Scalar`

### DIFF
--- a/bls-signatures/src/scalar.rs
+++ b/bls-signatures/src/scalar.rs
@@ -1,0 +1,102 @@
+//! Scalar weights for linear combinations of public keys and signatures.
+
+use {blstrs::Scalar as BlstrsScalar, ff::Field, rand::rngs::OsRng};
+
+/// A scalar weight used to form linear combinations of public keys or
+/// signatures, as in [`PubkeyProjective::aggregate_with_scalars`] and
+/// [`SignatureProjective::aggregate_with_scalars`].
+///
+/// This wraps the underlying BLS12-381 scalar field element so that callers do
+/// not need a direct `blstrs` dependency (nor the `ff` and `rand` versions it
+/// happens to be pinned to) in order to build one.
+///
+/// [`PubkeyProjective::aggregate_with_scalars`]: crate::pubkey::PubkeyProjective::aggregate_with_scalars
+/// [`SignatureProjective::aggregate_with_scalars`]: crate::signature::SignatureProjective::aggregate_with_scalars
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct Scalar(pub(crate) BlstrsScalar);
+
+impl Scalar {
+    /// The additive identity.
+    pub const ZERO: Self = Self(BlstrsScalar::ZERO);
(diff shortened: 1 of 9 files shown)


zz-sol:
@claude review once

bls-signatures/src/secret_bytes.rs
```diff
@@ -0,0 +1,159 @@
...
+    },
+    zeroize::Zeroize,
+};
+
+/// A fixed-size buffer of secret-key material, zeroized on drop.
+///
+/// Dereferences to `[u8; N]`, so it can be passed anywhere the raw buffer is
+/// expected. `Debug` deliberately does not print the contents.
+#[derive(Clone, Eq, PartialEq)]
+pub struct SecretBytes<const N: usize>([u8; N]);
```
samkim-crypto: Nit: could we store `Zeroizing<[u8; N]>` inside `SecretBytes` (e.g. `SecretBytes<N>(Zeroizing<[u8; N]>`) and `Zeroizing<String>` inside `SecretString`? The current code already zeroizes on drop, but using these as private fields would let us reuse that cleanup behavior and remove the manual Drop implementations.

We'd lose the const constructors with the current `Zeroizing::new`, but this might be a little easier to maintain. We'd still want to keep the manual, redacted Debug implementations.

bls-signatures/src/scalar.rs
```diff
@@ -0,0 +1,102 @@
...
+/// [`SignatureProjective::aggregate_with_scalars`].
+///
+/// This wraps the underlying BLS12-381 scalar field element so that callers do
+/// not need a direct `blstrs` dependency (nor the `ff` and `rand` versions it
+/// happens to be pinned to) in order to build one.
+///
+/// [`PubkeyProjective::aggregate_with_scalars`]: crate::pubkey::PubkeyProjective::aggregate_with_scalars
+/// [`SignatureProjective::aggregate_with_scalars`]: crate::signature::SignatureProjective::aggregate_with_scalars
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct Scalar(pub(crate) BlstrsScalar);
```
samkim-crypto, line 31: I think a wrap here will limit the use of scalar arithmetic downstream. Maybe we should implement arithmetic traits like `Add` or `Mul`?

samkim-crypto (commented):
Looks good to me overall!

Sorry for the extra work, but these API changes would require a major version bump. I'm planning a few more optimizations in this crate as well, so I'd prefer to hold off on a major release until those are ready.

For now, could we keep the existing API, deprecate the methods we're replacing, and introduce the new types and methods alongside it? We'd also need to retain the existing `From` conversions so this can stay in v3.

I have a few additional comments below as well.

samkim-crypto (approved):
LGTM!

PR #937, opened 2026-09-08
ci(tombi): use toml version 1.0.0

DSharifi:
toml version 1.1.0 is only supported since cargo 1.94.0 which is higher than the current MSRV.

Tombi could cause formatting of toml files that would be unparseable for current msrv.


joncinque (approved):
Ah nice catch, thanks!

PR #941, opened 2026-09-09
ci: use hash-pinned versions for github actions deps

yihau:
#### Summary of Changes

let's use hash-pinned versions for github actions deps 🫡


joncinque (approved):
Okey dokey!

PR #942, opened 2026-09-10
(Backport v4) instruction: Semver-trick to v4

joncinque:
#### Problem

solana-instruction v4 was just published, but the only breaking change was not re-exporting solana-instruction-error.

#### Summary of changes

From solana-instruction v3, re-export everything from v4, and also solana-instruction-error.


instruction/Cargo.toml
```diff
@@ -15,32 +15,20 @@ all-features = true
...
+syscalls = ["solana-instruction-v4/syscalls"]
+wincode = ["solana-instruction-v4/wincode"]
 
 [dependencies]
-bincode = { workspace = true, optional = true }
-borsh = { workspace = true, optional = true }
-serde = { workspace = true, optional = true }
-serde_derive = { workspace = true, optional = true }
+solana-instruction-v4 = { package = "solana-instruction", version = "4.0.0" }
 solana-instruction-error = { workspace = true, features = ["num-traits"] }
```
febo, line 41: Would this be semver "friendly" for 3.x? The error type being re-exported is incompatible with the previous version now.

joncinque: This branch is https://github.com/anza-xyz/solana-sdk/tree/maintenance/v4.x, which I snapshotted right before the breaking change, so it should be all good

febo: Ah, right, so all good then! 😊

febo (approved):

Cargo.toml
kskalski, line 274: this should probably move to 3.0.0, at least my preliminary tests indicate that without it `zk-elgamal-proof` breaks

kskalski: though it should land in `instruction/Cargo.toml:41`

joncinque: that would be a breaking change to solana-instruction though -- if we need to bump and publish zk-elgamal-proof, that's no problem at all!

PR #943, opened 2026-09-10
scripts: Allow for self-dependency in order script

joncinque:
#### Problem

When we did the backports in the previous branch, there was an issue in the ordering script, but we never added it to master.

#### Summary of changes

Add the changes from #418 to master too.


febo (approved):
Looks great!

PR #944, opened 2026-09-10
(Backport v4) clock: Semver-trick to newest clock

joncinque:
#### Problem

There was a breaking release to the clock crate, but it was just constants.

#### Summary of changes

Use the semver trick to re-export the newest version of the struct, but keep all of the previous constants

~~NOTE: builds on https://github.com/anza-xyz/solana-sdk/pull/942~~ All ready now!

### DIFF
--- a/clock/src/lib.rs
+++ b/clock/src/lib.rs
@@ -25,10 +25,6 @@
 #[cfg(feature = "sysvar")]
 pub mod sysvar;
 
-#[cfg(feature = "serde")]
-use serde_derive::{Deserialize, Serialize};
-use solana_sdk_macro::CloneZeroed;
-
 /// The default tick rate that the cluster attempts to achieve (160 per second).
 ///
 /// Note that the actual tick rate at any given time should be expected to drift.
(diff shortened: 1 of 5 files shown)


clock/src/lib.rs
```diff
@@ -111,92 +107,8 @@ pub const MAX_TRANSACTION_FORWARDING_DELAY: usize = 6;
...
-    /// the [`timestamp_correction` and `timestamp_bounding`][tsc] features it
-    /// is calculated using a [validator timestamp oracle][oracle].
-    ///
-    /// [tsc]: https://docs.solanalabs.com/implemented-proposals/bank-timestamp-correction
-    /// [oracle]: https://docs.solanalabs.com/implemented-proposals/validator-timestamp-oracle
-    pub unix_timestamp: UnixTimestamp,
-}
-
-/// Serialized size of the `Clock` sysvar account.
-pub const SIZE: usize = size_of::<Clock>();
```
febo, line 175: Just checking that this is intentional, since the description says that it should maintain all constants.

febo (commented):
Looks good, I just have a small query.

clock/src/lib.rs
joncinque: Yeah I wasn't sure about that one -- since it's tied directly to the size of the type that we're also re-exporting, it made sense to just directly re-export the constant too

joncinque: If you prefer to define it inline, I can certainly do that

febo: I think it looks good as it is, my comment was just to double-check that it was intentional. 😀

febo (approved):

PR #946, opened 2026-09-12
[bn254] add consistency tests against eip/geth

zz-sol:


zz-sol:
@claude review once

samkim-crypto (dismissed):
Thanks for adding these! I verified the geth fixtures against upstream and ran the tests to verify that all 15 tests pass.

There is an issue open regarding how flag bits are handled on the ethereum side and our implementation of the syscall https://github.com/anza-xyz/agave/issues/3379. I think this issue is not covered by the tests in this PR, so let's add them when we address the flag issue.

zz-sol:
@claude review once

samkim-crypto (commented):
Oh sorry 🙏, I meant to suggest the flag bit as a natural follow-up work, but we don't need to address it in this PR. Let's actually push the changes on the bit flags to a follow-up PR.

The bit flag issue is related to developer facing functions (i.e. functions under `#[cfg(target_os = "solana")]`), not the syscall backend functions (i.e. functions under `#[cfg(not(target_os = "solana"))]`). A change to the syscall backend function would actually require a SIMD since that is a modification to the runtime. Let's discuss in our regular call.

I think the version that you had was fine to merge (commit [d2fe2e6](https://github.com/anza-xyz/solana-sdk/pull/946/commits/d2fe2e69dc32ee29cc8696ba2e2a301f3c58082c)). Sorry for not being clear earlier 🙏

samkim-crypto (approved):
LGTM thanks!

PR #947, opened 2026-09-12
[bls12-381] consistency tests against EIP/Geth

zz-sol:


zz-sol:
@claude review once

samkim-crypto (approved):
Looks fine to me! I checked the encoding conversions for both endianness, the non-canonical handling, and the pairing-limit cases.

PR #951, opened 2026-09-16
Bump rustls to 0.23.45

samkim-crypto:
The cargo audit CI check is failing due to rustls (RUSTSEC-2026-0285 - https://rustsec.org/advisories/RUSTSEC-2026-0285.html).

https://github.com/anza-xyz/agave/pull/15239 fixed the issue by updating the direct rustls dependency to 0.23.45 and refreshing the lockfiles. The sdk depends on rustls transitively, so I think we just need to update Cargo lock, including the required tls dependencies. I just ran

```
cargo update -p rustls@0.23.42
```


joncinque (approved):
Thanks! Merging to unblock others

PR #953, opened 2026-09-16
ci(deps): make tombi-toml point directly to commit hash

DSharifi:
### Problem
`tombi-toml` action is pointing to the git tag object hash, which points to the the commit hash of the 1.5.0 release.

Pointing to the commit hash of the tag is more explicit.

### Solution
Change the `setup-tombi` action to point to the commit hash of the 1.5.0 tag.


joncinque (approved):
Makes sense!

PR #955, opened 2026-09-16
account-view: Add next account view helper

febo:
### Problem

When migrating programs to pinocchio, it is useful to keep the pattern of using an iterator to access accounts in the processor. This avoids having to change the logic of instructions during the migration.

### Solution

Add a `next_account_view` helper as a companion to [`next_account_info`](https://github.com/anza-xyz/solana-sdk/blob/master/account-info/src/lib.rs#L387) so it can be used as a drop-in replacement.


joncinque (approved):
Looks great to me!

PR #959, opened 2026-09-17
address: Remove heap allocation from `try_find_program_address`

febo:
### Problem

Currently, `try_find_program_address` does heap allocations for every bump seed tested when running outside a program. This can be avoided by using a stack allocated array.

### Solution

Pre-allocate an array with `crate::MAX_SEEDS` length so it is reused across different bump tests.


address/src/syscalls.rs
```diff
@@ -305,19 +304,36 @@ impl Address {
...
+                while i < seeds.len() {
+                    seeds[i] = u8::MAX - i as u8;
+                    i += 1;
                 }
-                bump_seed[0] -= 1;
+                seeds
+            };
+
+            if seeds.len() >= crate::MAX_SEEDS {
+                return None;
```
joncinque, line 320: I'm surprised we never had this check before 😅 confirming that the syscall does this check, however it only does a `>` check, which is technically incorrect: https://github.com/anza-xyz/agave/blob/f14a033e30eb7f34649e9d64cd3d5e7f13b28cf1/syscalls/src/lib.rs#L810

Thankfully, it does call `Pubkey::create_program_address` which does the check correctly

joncinque (approved):
Looks great to me! Let's do it in the syscall implementation too.

While looking through the code, we could eventually inline part of the `create_program_address` code, since we don't need to do the bounds checks on every single iteration. We could also tag some of the branches as cold to speed things up

PR #960, opened 2026-09-18
 frozen-abi: add stable-abi feature for digests on stable Rust

kskalski:
##### Problem
The `frozen-abi` feature requires nightly Rust. Its `AbiExample` api digester specializes blanket impls, so it needs `specialization`. 
The `StableAbi` digester doesn't, but it rides on the same feature regardless, so consumers build on nightly even for `abi_digest` coverage alone.

## Summary of Changes
- add `stable-abi` feature to `solana-frozen-abi`, carrying the `StableAbi` digester (`abi_digest`) alone
- mirror the feature in `solana-frozen-abi-macro`
- enable `stable-abi` from `frozen-abi`, so existing consumers see no change
- no-op the `AbiExample`/`AbiEnumVisitor` derives under plain `stable-abi`, so call sites still compile
- add `stable-abi` to every crate defining `frozen-abi`, forwarding to the dependencies' `stable-abi`
- move those crates' `StableAbi` gates to `stable-abi`, leaving `min_specialization` under `frozen-abi`
- reject `api_digest` without an `abi_digest` under plain `stable-abi`, since such a type would get no test at all
- gate the four api-only digests on `frozen-abi`
- undeprecate the macro's `frozen-abi` feature, since it now selects the api digester code
- digest the whole workspace on stable Rust in `test-frozen-abi.sh`
- exclude `stable-abi` from the no-std check, since it pulls in std

## Testing
- the whole workspace digests on stable Rust under `stable-abi`, across 113 packages
- the nightly `frozen-abi` build keeps every api digest passing, 47 tests in total
- the api digest test names match master one for one, so api coverage did not shrink
- clippy covers the workspace in the `stable-abi` configuration with warnings denied
- a downstream crate on stable Rust digests its own wincode type, with no `#![feature(min_specialization)]`
- the same crate hits the new error once its type carries `api_digest` alone
- a crate requesting no macro feature still compiles an `api_digest` attribute, so no existing build breaks


DSharifi (approved):
Makes sense

fee-structure/Cargo.toml
```diff
@@ -15,13 +15,14 @@ all-features = true
 rustdoc-args = ["--cfg=docsrs"]
 
 [features]
-frozen-abi = ["dep:solana-frozen-abi"]
+frozen-abi = ["solana-frozen-abi/frozen-abi", "stable-abi"]
+stable-abi = ["solana-frozen-abi/stable-abi"]
```
joncinque: It might not matter, but this crate doesn't implement any of the stable-abi traits, only frozen-abi -- is it worth adding that?

transaction-error/Cargo.toml
```diff
@@ -17,18 +17,24 @@ rustdoc-args = ["--cfg=docsrs"]
...
-    "dep:solana-frozen-abi",
-    "dep:solana-frozen-abi-macro",
-    "serde",
+    "solana-frozen-abi-macro/frozen-abi",
+    "solana-frozen-abi/frozen-abi",
     "solana-instruction-error/frozen-abi",
+    "stable-abi",
+]
+stable-abi = [
+    "serde",
```
joncinque, line 26: Sorry if this is a silly question, but why does `stable-abi` need `serde`? Is it because we test both bincode and wincode compatibility in the abi tests?

joncinque (dismissed):
Looks great to me! It's also much clearer for the frozen-abi feature to mean something very specific now

transaction-error/Cargo.toml
kskalski: Yeah, since default `abi_serializer` is `bincode`, we need serde unconditionally. Possibly in the future we could switch the default to wincode and make bincode `abi_digest` support as an optional feature default-off

however reviewing the feature propagation turned something else - many crates only do `api_digest` without using `StableAbi` (well, possibly it's something to fix in a follow up, I didn't check exactly which would benefit from binary compatibility checks instread), so I trimmed down the feature propagation for them

kskalski:
I re-wired the feature propagation - stable-abi now means “this crate has StableAbi digests, or aggregates crates that do”: That is 20 SDK crates — `account, address, epoch-schedule, fee-calculator, hard-forks, hash, inflation, instruction-error, message, nonce, packet, program, pubkey, rent, sdk, signature, stake-history, transaction, transaction-error, vote-interface`. 15 crates that previously had `stable-abi` introduced, now got reverted to master shape.

joncinque (approved):
Looks great to me!

DSharifi (approved):

